PR #50, opened 2019-04-24
minor fixes

butonic:
see inline comments for details

- [x] ~~while up and downloads with phoenix work, uploads are not yet stripped of their multipart envelope. will fix that tomorrow.~~ was an outdated phoenix I was using

### DIFF
--- a/cmd/revad/config/config.go
+++ b/cmd/revad/config/config.go
@@ -19,6 +19,7 @@
 package config
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/spf13/viper"
@@ -52,7 +53,11 @@ func reGet(prefix string, kv *map[string]interface{}) {
 		if c, ok := val.(map[string]interface{}); ok {
 			reGet(prefix+"."+k, &c)
 		} else {
-			(*kv)[k] = v.Get(prefix + "." + k)
+			newVal := v.Get(prefix + "." + k)
+			fmt.Println("reGet", prefix, k, "val:", val, "->", newVal)
+			if newVal != nil {
+				(*kv)[k] = newVal
+			}
 		}
 	}
 
--- a/cmd/revad/svcs/httpsvcs/ocdavsvc/get.go
+++ b/cmd/revad/svcs/httpsvcs/ocdavsvc/get.go
@@ -19,12 +19,14 @@
 package ocdavsvc
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"time"
 
 	rpcpb "github.com/cernbox/go-cs3apis/cs3/rpc"
+	"github.com/cernbox/reva/pkg/token"
 	storageproviderv0alphapb "github.com/cernbox/go-cs3apis/cs3/storageprovider/v0alpha"
 	"github.com/cernbox/reva/cmd/revad/svcs/httpsvcs/utils"
 )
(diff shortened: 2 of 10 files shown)


cmd/revad/config/config.go
```diff
@@ -52,7 +53,11 @@ func reGet(prefix string, kv *map[string]interface{}) {
 		if c, ok := val.(map[string]interface{}); ok {
 			reGet(prefix+"."+k, &c)
 		} else {
-			(*kv)[k] = v.Get(prefix + "." + k)
+			newVal := v.Get(prefix + "." + k)
+			fmt.Println("reGet", prefix, k, "val:", val, "->", newVal)
+			if newVal != nil {
```
butonic: only overwrite config if newval is present, fixes using environment variables configuration

cmd/revad/svcs/httpsvcs/ocdavsvc/get.go
```diff
@@ -40,50 +42,51 @@ func (s *svc) doGet(w http.ResponseWriter, r *http.Request) {
...
+	dRes, err := client.InitiateFileDownload(ctx, dReq)
 	if err != nil {
 		logger.Error(ctx, err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if res.Status.Code != rpcpb.Code_CODE_OK {
-		logger.Println(ctx, res.Status)
+	if dRes.Status.Code != rpcpb.Code_CODE_OK {
```
butonic: this was using the `res` from the previous request, which is wrong. I prefixed the variables for clarity and to prevent these copy paste errors.

cmd/revad/svcs/httpsvcs/ocdavsvc/get.go
```diff
@@ -92,6 +95,15 @@ func (s *svc) doGet(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	//TODO: make header / auth configurable, check if token is available before doing stat requests
+	tkn, ok := token.ContextGetToken(ctx)
+	if !ok {
+		logger.Error(ctx, errors.New("could not read token from context"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	httpReq.Header.Set("X-Access-Token", tkn)
```
butonic: The up and download requests need to be authenticated, otherwise the storage driver cannot extract the userid. For now, I hardcoded the example header. Need to think if oidc or basic auth makes sense here. AFAICT only the internal token is used.
@labkode Maybe implement a dummy token writer and credential_strategy as they are not really needed? Or rather ... they should not be available ... then again maybe you want to just do basic auth for up and download ...

cmd/revad/svcs/httpsvcs/ocdavsvc/put.go
```diff
@@ -182,27 +184,27 @@ func (s *svc) doPut(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		logger.Error(ctx, err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if res.Status.Code != rpcpb.Code_CODE_OK {
-		logger.Println(ctx, res.Status)
+	if uRes.Status.Code != rpcpb.Code_CODE_OK {
+		logger.Println(ctx, uRes.Status)
```
butonic: this was using the `res` from the previous request, which is wrong. I prefixed the variables for clarity and to prevent these copy paste errors.

cmd/revad/svcs/httpsvcs/ocdavsvc/put.go
```diff
@@ -211,6 +213,15 @@ func (s *svc) doPut(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	//TODO: make header / auth configurable, check if token is available before doing stat requests
+	tkn, ok := token.ContextGetToken(ctx)
+	if !ok {
+		logger.Error(ctx, errors.New("could not read token from context"))
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	httpReq.Header.Set("X-Access-Token", tkn)
```
butonic: same as for download

pkg/storage/fs/eos/eos.go
```diff
@@ -40,12 +40,12 @@ import (
 )
 
 func init() {
-	registry.Register("storage.fs.eos", New)
+	registry.Register("eos", New)
```
butonic: AFAICT providers have a short name, eg `local` or `eos`.

pkg/storage/fs/eos/eos.go
```diff
@@ -40,12 +40,12 @@ import (
...
 }
 
 var hiddenReg = regexp.MustCompile(`\.sys\..#.`)
 
-var logger = log.New("storage.fs.eos")
+var logger = log.New("storage-provider-eos")
```
butonic: adjust to local storage logger name

cmd/revad/svcs/httpsvcs/ocdavsvc/get.go
labkode: @butonic I'm refactoring the token manager bussiness as it is used in many places, so it makes sense to inject it in the context

PR #56, opened 2019-04-26
split mtime into sec and nanos

butonic:
I was getting wrong mtimes. I assume because utils.UnixNanoToTS is not working as expected.

In any case I think it makes more sense to pass around Seconds and Nanos independently

### DIFF
--- a/pkg/eosclient/eosclient.go
+++ b/pkg/eosclient/eosclient.go
@@ -741,7 +741,12 @@ func (c *Client) mapToFileInfo(kv map[string]string) (*FileInfo, error) {
 	}
 
 	// mtime is split by a dot, we only take the first part, do we need subsec precision?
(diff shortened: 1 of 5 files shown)


pkg/eosclient/eosclient.go
```diff
@@ -741,7 +741,9 @@ func (c *Client) mapToFileInfo(kv map[string]string) (*FileInfo, error) {
 	}
 
 	// mtime is split by a dot, we only take the first part, do we need subsec precision?
-	mtime, err := strconv.ParseUint(strings.Split(kv["mtime"], ".")[0], 10, 64)
+	mtime := strings.Split(kv["mtime"], ".")
+	mtimesec, err := strconv.ParseUint(mtime[0], 10, 64)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

PR #73, opened 2019-05-07
svsc: cleanup services on server shutdown. Closes #72

### LINKED ISSUE #72: svcs:  allow cleanup logic on server shutdown

labkode:


cmd/revad/httpserver/httpserver.go
```diff
@@ -197,33 +214,32 @@ func (s *Server) registerMiddlewares() error {
...
-	svcs := map[string]http.Handler{}
 	for svcName, newFunc := range Services {
 		if s.isEnabled(svcName) {
 			svc, err := newFunc(s.conf.Services[svcName])
 			if err != nil {
 				err = errors.Wrap(err, "error registering new http service")
 				return err
 			}
-			svcs[svc.Prefix()] = prometheus.InstrumentHandler(svc.Prefix(), svc.Handler())
+			s.handlers[svc.Prefix()] = prometheus.InstrumentHandler(svc.Prefix(), svc.Handler())
```
golangcibot: SA1019: prometheus.InstrumentHandler is deprecated: InstrumentHandler has several issues. Use the tooling provided in package promhttp instead. The issues are the following: (1) It uses Summaries rather than Histograms. Summaries are not useful if aggregation across multiple instances is required. (2) It uses microseconds as unit, which is deprecated and should be replaced by seconds. (3) The size of the request is calculated in a separate goroutine. Since this calculator requires access to the request header, it creates a race with any writes to the header performed during request handling.  httputil.ReverseProxy is a prominent example for a handler performing such writes. (4) It has additional issues with HTTP/2, cf. https://github.com/prometheus/client_golang/issues/272.  (from `staticcheck`)

PR #77, opened 2019-05-08
implement Close for storageprovidersvc

butonic:


labkode:
@butonic I'm quite against optional interfaces, see this nice article: https://blog.merovius.de/2017/07/30/the-trouble-with-optional-interfaces.html

If you were to implement a storage you have no clue that you can implement Closer for adding additional logic. We suffer quite a lot of this problem in this issue: https://github.com/cernbox/reva/issues/59

So let's try to avoid it and add a new method to it :)

butonic:
@labkode changed to explicit Shutdown() method

PR #78, opened 2019-05-08
[WIP] storage/local notify

butonic:
This is a PoC to make the local storage provider
- listen for file changes
- propagate change via mtime updates up to the root (actually up to the subfolder of the root)

Listening for file changes is disabled by default and can be enabled by setting `witch = true`, eg:
  ```
  [grpc.services.storageprovidersvc.drivers.local]
  root = "/data"
  watch = true
  ```

Several things I noticed:
- we should have a way to Close() services, @labkode [already implemented the basics](https://github.com/cernbox/reva/pull/73), I just added [some code to Close storages if they implement the io.Closer interface](https://github.com/cernbox/reva/pull/77)
- in New() I see currently no way to get a proper logger. We could either
  - pass the logger in the New function, or
  - pass in a context that has the logger set.
  I prefer the latter.
- the current propagation implementation could live just fine as a standalone service. I also noticed that an embedded kv store might make sense as a dedicated service. I assume when I start integrating a kv store based filecache with file notifications I will need them in the storage implementation. So maybe not making them standalone services is the right thing to do. The current notification and propagation would work as a bash script just fine ... for the cache I need an entry point to make it update metadata when something changes. Will look into that with s3 tomorrow.
- sharing, if implemented via acls uses the `xattr` `system` namespace. I was wondering if we could store our own arbitrary attributes that the user cannot change himself (eg for system tags) It seems not only the namespace, but also the names you can assign a value are hardcoded. At least with `setfattr` I could not set a `system.mykey` as root ... that at least will prevent us from storing system tags in the filesystem. However, anything the user can change could be stored as xattr: user tags, maybe share expiry (might require some clever naming to link it to the share).


cmd/revad/httpserver/httpserver.go
```diff
@@ -197,33 +214,32 @@ func (s *Server) registerMiddlewares() error {
...
-	svcs := map[string]http.Handler{}
 	for svcName, newFunc := range Services {
 		if s.isEnabled(svcName) {
 			svc, err := newFunc(s.conf.Services[svcName])
 			if err != nil {
 				err = errors.Wrap(err, "error registering new http service")
 				return err
 			}
-			svcs[svc.Prefix()] = prometheus.InstrumentHandler(svc.Prefix(), svc.Handler())
+			s.handlers[svc.Prefix()] = prometheus.InstrumentHandler(svc.Prefix(), svc.Handler())
```
golangcibot: SA1019: prometheus.InstrumentHandler is deprecated: InstrumentHandler has several issues. Use the tooling provided in package promhttp instead. The issues are the following: (1) It uses Summaries rather than Histograms. Summaries are not useful if aggregation across multiple instances is required. (2) It uses microseconds as unit, which is deprecated and should be replaced by seconds. (3) The size of the request is calculated in a separate goroutine. Since this calculator requires access to the request header, it creates a race with any writes to the header performed during request handling.  httputil.ReverseProxy is a prominent example for a handler performing such writes. (4) It has additional issues with HTTP/2, cf. https://github.com/prometheus/client_golang/issues/272.  (from `staticcheck`)

butonic:
1. inotify is not the proper mechanism to watch for file changes because it requires 48 bytes of memory per file. This does not scale to millions of files. Furthermore, it does not guarantee that you will be notified for all changes.
2. the kernel has an audit mechanism which has a proper queue that can be used to audit changes to a file tree
3. a proper solution should use a file system to cover not only etag propagation and tree size accounting, but also versions and trash bin

PR #99, opened 2019-05-29
build local storage on windows, make grace os independent

butonic:
make it possible to build on windows

@labkode @glpatcern please test on windows and macos!

### DIFF
--- a/pkg/storage/fs/local/local_unix.go
+++ b/pkg/storage/fs/local/local_unix.go
@@ -16,7 +16,7 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-// +build linux
+// +build !windows
 
 package local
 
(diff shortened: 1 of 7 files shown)


pkg/storage/fs/local/local_unix.go
```diff
@@ -0,0 +1,64 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+// +build linux
```
tboerger: I think it would be safe to say !windows

glpatcern:
@labkode, this is good to go to master now 👍

PR #100, opened 2019-05-29
svcs: Preferences service

Mohitty:


labkode:
@Mohitty your PR contains your previous PRs, please only include the changes related to the work you are doing, you need update your master branch to the github.com/cs3org/reva master branch and rebase your work on top of that. Please read this guide https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request so I'll help you understand the steps to achieve it.

I'd also recommend you if you are not familiar with Git to read this webpage:
http://johnmathews.eu/rys-git-tutorial.html

It's the best git tutorial I've read in my career.

cmd/revad/svcs/grpcsvcs/preferencessvc/preferencessvc.go
```diff
@@ -0,0 +1,123 @@
...
+func (s *service) Close() error {
+	return nil
+}
+
+func (s *service) SetKey(ctx context.Context, req *preferencesv0alphapb.SetKeyRequest) (*preferencesv0alphapb.SetKeyResponse, error) {
+	//	log := appctx.GetLogger(ctx)
+	key := req.Key
+	value := req.Val
+
+	md, ok := metadata.FromIncomingContext(ctx)
```
labkode: You don't need to use this. You need to extract the user from the context, you can obtain it like we do it here:

https://github.com/cs3org/reva/blob/master/pkg/storage/fs/eos/eos.go#L108

That is more elegant, change your code accordingly.

Mohitty: Done!

labkode:
Last modification, in Go concurrent access to a map is not thread/goroutine safe, you can protect the access to the map with a mutex.

See this example:

https://gobyexample.com/mutexes

You need to call Lock() every time you read/write a value from the map
and call defer Unlock() just after you call Lock().

PR #119, opened 2019-06-19
implement viewer and editor roles

butonic:
together with https://github.com/owncloud/phoenix/pull/1332 this allows updating share permissions to viewer and editor roles, reflected in eos as `rx!d` and `rwx`. I also map all the permissions between the permission set and the cs2 resource permissions ... although that seems awkward ...

implementing something like a co owner needs a beter understanding of how eos acls work. we need to change the way acls are processed anyway because there might be user.acl and sys.acls as well as posix permissions. at least that needs to be written down and specified somewhere. the official docs for EOS were quite revealing: https://github.com/cern-eos/eos/blob/master/doc/configuration/permission.rst

I don't yet know what to make of acls not being available on files ... or that only the direct parent dir acls are taken into account, not all in the tree to the root ... weird ...

### DIFF
--- a/pkg/storage/fs/eos/eos.go
+++ b/pkg/storage/fs/eos/eos.go
@@ -297,7 +297,9 @@ func getEosACLPerm(set *storageproviderv0alphapb.ResourcePermissions) (string, e
 		b.WriteString("x")
 	}
 
-	if !set.Delete {
+	if set.Delete {
+		b.WriteString("+d")
+	} else {
 		b.WriteString("!d")
 	}
 
@@ -442,6 +444,8 @@ func (fs *eosStorage) getGrantPermissionSet(mode string) *storageproviderv0alpha
 	}
 	if strings.Contains(mode, "!d") {
 		p.Delete = false
+	} else if strings.Contains(mode, "+d") {
+		p.Delete = true
 	}
 	// x
 	if strings.Contains(mode, "x") {
--- a/pkg/storage/fs/eos/eos.go
+++ b/pkg/storage/fs/eos/eos.go
@@ -284,12 +284,27 @@ func getEosACLType(gt storageproviderv0alphapb.GranteeType) (string, error) {
 }
 
 // TODO(labkode): fine grained permission controls.
-func getEosACLPerm(rp *storageproviderv0alphapb.ResourcePermissions) (string, error) {
-	if rp.Delete {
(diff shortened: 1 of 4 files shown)


cmd/revad/svcs/grpcsvcs/storageprovidersvc/storageprovidersvc.go
```diff
@@ -841,3 +830,76 @@ func (s *service) toInfo(md *storage.MD) *storageproviderv0alphapb.ResourceInfo
...
+		// versions
+		ListFileVersions:   p.ListFileVersions,
+		RestoreFileVersion: p.RestoreFileVersion,
+
+		// ?
+		GetPath:  p.GetPath,
+		GetQuota: p.GetQuota,
+	}
+}
+func set2permissions(s *storage.PermissionSet) *storageproviderv0alphapb.ResourcePermissions {
```
golangcibot: `set2permissions` is unused (from `deadcode`)

butonic: @labkode shoud I move those to pkg/storage/storage.go? would introduce a dependency on the cs3 apis in that module.

butonic: Ok I commented this and added a TODO with regard to using cs3 permissions instead of an intermediary struct.

labkode:
Just one little thing and I am happy to merge, the delete permission is granted with *+d*, it is not granted with *w*.

So we need to add logic like:

```
if set.Delete == true {
  b.WriteString("+d")
} else {
   b.WriteString("!d")
}
```

pkg/storage/fs/eos/eos.go
```diff
@@ -284,12 +284,29 @@ func getEosACLType(gt storageproviderv0alphapb.GranteeType) (string, error) {
...
+	}
+	if set.CreateContainer || set.InitiateFileUpload || set.Delete || set.Move {
+		b.WriteString("w")
+	}
+	if set.ListContainer {
+		b.WriteString("x")
 	}
 
-	return "rx", nil
+	if set.Delete == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `set.Delete` (from `gosimple`)

PR #133, opened 2019-07-02
tail always starts with slash

butonic:
tail always starts with / ... this gets rid of redundant // in the url path


butonic:
already in https://github.com/cs3org/reva/commit/844b42c3786963648a2dd8230e7100c817f47013

PR #134, opened 2019-07-02
add json based user and auth manager

butonic:
can be used to hold virtual users that are not available in the system


butonic:
is already in https://github.com/cs3org/reva/commit/346900d76bb76057aa79f29c197a53eab717d670

PR #140, opened 2019-07-09
use correct grantee

butonic:
AddGrant should take the grantee from the grant ... not the user from the context

Obviously we need to add some checking if the opaque id is set ... but I wanted to point this out before someone tries to use this.


labkode:
@butonic the usernaname param is to specify under which role the eos command would run, not the recipient of the grant, the recipient is already defined in the eosAcl

PR #149, opened 2019-07-17
add initial config generation

butonic:
This adds an initial configuration command.
It can be used to let the relad binary generate a working configuration for local development. Currently only `oidc` and `basic`credentials strategies are supported. As a bonus you can also configure the data driver (`local` and owncloud for now) and directory (defaults to `./data`). It will also generate a `./user.json` with the hardcoded oidc users if that file does not exist, yet.

### DIFF
--- a/cmd/reva/gen.go
+++ b/cmd/reva/gen.go
@@ -1,0 +1,78 @@
+// Copyright 2018-2019 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


labkode:
@butonic yesterday I wonder that the cleanest approach will be to have this functionality in the client applications (reva) and leave the server as simple as possible. Can you move this code to the cmd/reva application as a new command and also remove the dependency on owncloud specific names? Can you use the existing ones (einstein and company).Thanks

butonic:
@butonic I originally had it in reva, but then you need both binaries to set up a basic server. Which is why I moved it to revad ... einstein etc make sense ...

butonic:
I added the demo users as well as the oidc users ... making oidc use demo users is something I dont have the time for currently.

cmd/reva/gen.go
```diff
@@ -0,0 +1,75 @@
...
+		// cmd.Args()[0] is the subcommand command
+		// cmd.Args()[1] will be the subcommands arguments
+		if len(cmd.Args()) < 1 {
+			fmt.Println(genUsage)
+			os.Exit(1)
+		}
+		subcommand := cmd.Args()[0]
+		for _, v := range subcmds {
+			if v.Name == subcommand {
+				v.Parse(cmd.Args()[1:])
```
golangcibot: Error return value of `v.Parse` is not checked (from `errcheck`)

Issue #163, opened 2019-07-29
Fix panic when disabling http server


labkode:
```
6:21PM WRN grace/grace.go:172 > error reading pidfile pid=24835 pkg=grace
6:21PM INF grace/grace.go:181 > pidfile written to gateway.pid pid=24835 pkg=grace
6:21PM INF main.go:88 > running on 4 cpus pid=24835
6:21PM INF grpcserver/grpcserver.go:239 > chainning grpc unary interceptor log with priority 200 pid=24835 pkg=grpcserver
6:21PM INF grpcserver/grpcserver.go:268 > chainning grpc streaming interceptor log with priority 200 pid=24835 pkg=grpcserver
panic: runtime error: index out of range

goroutine 14 [running]:
main.main.func2(0xc000244a90, 0x1, 0x1, 0xc000244ac0, 0x1, 0x1, 0xc00008cc00, 0xc000300240)
        /home/gonzalhu/Development/reva/cmd/revad/main.go:126 +0x14d
created by main.main
        /home/gonzalhu/Development/reva/cmd/revad/main.go:125 +0x3dc

```

unixlab:
I looked into this and I think I understood the problem.

At `reva/cmd/revad` from line 118 to 125 the code assume that the arrays `servers` and `listeners` have at least two entries ( index 0 as http server and index 1 as grpc server ):
```
if !coreConf.DisableGRPC {
	go func() {
		if err := servers[1].(*grpcserver.Server).Start(listeners[1]); err != nil {
			log.Error().Err(err).Msg("error starting the grpc server")
			watcher.Exit(1)
		}
	}()
}
```
But if the http server is disabled, we only have one entry ( index 0 ) and illegally access index 1.

I assumed that `servers` and `listeners` are arrays because it's possible that there will be a third, fours and maybe fives server one day.

I tried a few ways to fix this and submitted my final solution as pull request => https://github.com/cs3org/reva/pull/188
This change is bigger than I initially thought it would be, but this should be safe even with more services.

labkode:
Closed by #165

PR #165, opened 2019-07-30
WIP: tracing + refactor + gateway service. 

labkode:

PR #188, opened 2019-08-13
fix panic when disabling http server

unixlab:
fix for issue https://github.com/cs3org/reva/issues/163

Proposal: we always use the elements with index 0 and shift the arrays after each server.
Because the code runs in a go routine and therefore runs async, we need to push the variables into the function.

### DIFF
--- a/cmd/revad/main.go
+++ b/cmd/revad/main.go
@@ -22,6 +22,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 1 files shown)


cmd/revad/main.go
```diff
@@ -107,21 +108,31 @@ func main() {
...
+		go func(server grace.Server, listener net.Listener) {
+			if err := server.(*grpcserver.Server).Start(listener); err != nil {
 				log.Error().Err(err).Msg("error starting the grpc server")
 				watcher.Exit(1)
 			}
-		}()
+		}(servers[0], listeners[0])
+		// shift servers and listeners array
+		if len(servers) > 1 {
+			servers = servers[1:]
```
golangcibot: SA4006: this value of `servers` is never used (from `staticcheck`)

cmd/revad/main.go
```diff
@@ -107,21 +108,31 @@ func main() {
...
+			listeners = listeners[1:]
```
golangcibot: SA4006: this value of `listeners` is never used (from `staticcheck`)

labkode:
@unixlab this fix has been implemented in https://github.com/cs3org/reva/pull/165, sorry for this double work. Take a look at the implementation there to see if you find any obvious problem

labkode:
Closed by #165

PR #216, opened 2019-08-26
Set LDAP schema attributes

refs:
## What?

Minimal approach to set ldap schema attributes using values provided in the config file.

- Defaults to Active Directory attributes if no values are provided.
- Minimal use of reflection (attributes type tags).

## Usage

```toml
[grpc.services.authsvc.user_managers.ldap.schema]
   mail = "mail"
   uid = "objectSid"
   displayName = "displayName"
   dn = "dn"
```

### DIFF
--- a/cmd/revad/revad.toml
+++ b/cmd/revad/revad.toml
@@ -78,7 +78,13 @@ root = "/var/tmp/reva/data"
 [grpc.services.authsvc]
 auth_manager = "demo"
 token_manager = "jwt"
-user_manager = "demo"
+user_manager = "ldap"
+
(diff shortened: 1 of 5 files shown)


labkode:
Hi @refs, two commetns:

1) Can you avoid using reflection?
2) Can you use the standard test package for testing?

Thanks!

refs:
> Hi @refs, two commetns:
> 
>     1. Can you avoid using reflection?
> 
>     2. Can you use the standard test package for testing?
> 
> 
> Thanks!

Hi @labkode ! sure 👍

PR #225, opened 2019-08-29
Refactoring of the error responses and of the logging around them

glpatcern:
Once it's completed, this closes #162.

On the way, the status package is being extended.

PR #231, opened 2019-09-02
Mini-refactoring of the error functions, following PR #225

glpatcern:


labkode:
Check the CONTRIBUTORS.md files, choose one username not two :)

glpatcern:
Yeah, that is because in Github I used a different username than in my local box :-| ...fixed now

PR #239, opened 2019-09-05
store the userid in the context when authenticating credentials

butonic:
Authenticating credentials is used to establish the identity of a user. The credentials can be
* username & password - demo, ldap and json auth managers do this
* a clientid & token - the oidc auth manager does this, in the future we also need basic oauth tokens for device specific authentication

All of the above establish the identity of a user, which is why we should either return a UserID struct or put it into the context.

This PR for now implements how this works when using the json auth provider.

@labkode Do you agree that the result of the auth should be a userid or nil? Should we use a return parameter or the context? I prefer a return parameter ... which can be nil or a UserID struct.

I ran into this because the current reva login command only works with userids, not usernames. but a login is the process of identifying a user, so the login should imo work with usernames.

### DIFF
--- a/cmd/revad/svcs/grpcsvcs/authsvc/authsvc.go
+++ b/cmd/revad/svcs/grpcsvcs/authsvc/authsvc.go
@@ -130,7 +130,6 @@ func (s *service) GenerateAccessToken(ctx context.Context, req *authv0alphapb.Ge
 	log := appctx.GetLogger(ctx)
 	username := req.ClientId
 	password := req.ClientSecret
-	uid := &typespb.UserId{OpaqueId: username}
 
 	ctx, err := s.authmgr.Authenticate(ctx, username, password)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


butonic:
hm full user object or user idwith opacke id and idp?

labkode:
@butonic maybe userId may suffice you're rigth

butonic:
Using the context allows auth implementations to store arbitrary data. the oidc auth manager uses it to remember the claims the user agreed to which is needed later on when resolving user data.

I don't know what is necessary to handle kerberos authentication, but I doubt returning just a userid is good enough. Let's go with an updated context as the return value.

butonic:
- [ ] update ldap backend
- [ ] update demo backend
- [ ] update oidc backend

labkode:
@butonic, if the authenticate call would return a full userID you would not need the magic inside the context right? I am trying to come up with the simpler approach for people reading the code to get it easily.

butonic:
@labkode hmm ... will check tomorrow ...
I added a commit that makes all user and auth managers return proper userids, including a uuid generated opaqueid and a idp url

At least that is what I think we should aim for.

labkode:
@butonic I see a lack of functionality in the CS3 APIs. The user as you said will login with a username that is only unique to an IDP, therefore, internally we must use the userID as much as possible to ACLs and user identification. This means that we should have in the CS3 APIs a registry to IDPs that can authenticate their users by username.

In deployments that I've been seen,  the user will chose an institution and login there with their user names and password typically, then, we use the configured Id property to identify it correclty (opaqueid) and we know where the user comes from (idp), having a unique disitributed map of all the users in the system.

butonic:
@labkode after restarting the build travis was green ... the linter segfaulted .... wtf

PR #243, opened 2019-09-06
Create reference when accepting a share

labkode:
A,K.A mounting the share


cmd/revad/users.json
```diff
@@ -1,12 +1,22 @@
 [
 	{
 		"id": {
-			"opaque_id": "a48cb0ed-3375-4b94-bdcd-3cac6e201b7b",
-			"idp": "example.com"
+			"opaque_id": "labradorsvc",
```
butonic: we should use random strings / uuds as opaqueids so developers don't accidentially assume they should be visible in the ui or can be used in the same way as the username. And maybe you did not want to change the users json?

IMO we should default to the demo users?

labkode: My bad, I will remove it from the PR

labkode:
Overcomed by #394

PR #244, opened 2019-09-06
list correct trash properties

butonic:
@labkode can you clarify the trashbin Key and Path properties? AFAICT the eos storage returns the parent path of the file / folder that was deleted and uses the name as the key ... I did the same for the owncloud storage ... but it seems like key should be something akin to a unique id ... but then either eos does tho wrong thing by cutting off the base name (or i misread that code) ... or the CS3 api is missing a basename property for the file that was deleted.

### DIFF
--- a/cmd/revad/svcs/httpsvcs/ocdavsvc/trashbin.go
+++ b/cmd/revad/svcs/httpsvcs/ocdavsvc/trashbin.go
@@ -20,20 +20,24 @@ package ocdavsvc
 
 import (
 	"context"
+	"encoding/xml"
+	"fmt"
 	"net/http"
 	"net/url"
 	"path"
 	"strings"
+	"time"
 
 	authv0alphapb "github.com/cs3org/go-cs3apis/cs3/auth/v0alpha"
 	rpcpb "github.com/cs3org/go-cs3apis/cs3/rpc"
 	storageproviderv0alphapb "github.com/cs3org/go-cs3apis/cs3/storageprovider/v0alpha"
 	"github.com/cs3org/reva/cmd/revad/svcs/httpsvcs"
+	"github.com/cs3org/reva/cmd/revad/svcs/httpsvcs/utils"
 	"github.com/cs3org/reva/pkg/appctx"
 	ctxuser "github.com/cs3org/reva/pkg/user"
 )
 
-// TrashbinHandler handles version requests
+// TrashbinHandler handles trashbin requests
 type TrashbinHandler struct {
 }
 
@@ -81,8 +85,7 @@ func (h *TrashbinHandler) Handler(s *svc) http.Handler {
 		}
 		if key == "" && r.Method == "PROPFIND" {
 
-			// webdav should be death: baseURI is encoded as part of the
-			// response payload in href field
+			// baseURI is encoded as part of the response payload in href field
 			baseURI := path.Join("/", s.Prefix(), "remote.php/dav/trash-bin", username)
 			ctx = context.WithValue(r.Context(), ctxKeyBaseURI, baseURI)
 			r = r.WithContext(ctx)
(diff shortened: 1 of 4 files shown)


labkode:
> @labkode can you clarify the trashbin Key and Path properties? AFAICT the eos storage returns the parent path of the file / folder that was deleted and uses the name as the key ... I did the same for the owncloud storage ... but it seems like key should be something akin to a unique id ... but then either eos does tho wrong thing by cutting off the base name (or i misread that code) ... or the CS3 api is missing a basename property for the file that was deleted.

Sure!

Ref here: https://cs3org.github.io/cs3apis/#cs3.storageproviderv0alpha.RestoreRecycleItemRequest

I realized that the Ref field is useless in that rpc.

To restore recycle items EOS uses a key, called the restore key that you can get if you do: 

```
# Deletion Time            UID      GID      SIZE         TYPE          RESTORE-KEY      RESTORE-PATH                                                    
# ==============================================================================================================================
Fri Mar 15 14:46:59 2019   gonzalhu it       25422        file          000000000af0c416          
```

To restore:

```
eos recycle restore 000000000af0c416
```

will move the file from the trashbin to 

```
/eos/user/g/gonzalhu/university/msc.pdf 
```

butonic:
ok, so the restorepath is the full path, including the basename of the file or folder. eg `/projects/old/readme.md` for a deleted file `readme.md` in `/projects/old` the path in the list recycle item would be `/projects/old/readme.md`?

labkode:
> ok, so the restorepath is the full path, including the basename of the file or folder. eg `/projects/old/readme.md` for a deleted file `readme.md` in `/projects/old` the path in the list recycle item would be `/projects/old/readme.md`?

Yes, path is always the full path, in all CS3APIs

cmd/revad/svcs/httpsvcs/ocdavsvc/trashbin.go
```diff
@@ -226,6 +182,166 @@ func (h *TrashbinHandler) listTrashbin(w http.ResponseWriter, r *http.Request, s
...
+	}
+
+	// when allprops has been requested
+	if pf.Allprop != nil {
+		// return all known properties
+		response.Propstat = append(response.Propstat, propstatXML{
+			Status: "HTTP/1.1 200 OK",
+			Prop:   []*propertyXML{},
+		})
+		response.Propstat[0].Prop = append(response.Propstat[0].Prop, s.newProp("oc:trashbin-original-filename", item.Key))
```
labkode: I think this should be path also

butonic: yeah ... we neet to translate a bit what the webdav api wants here: this is from the phoenix.owncloud.com instance whan listing the trashbin
```xml
<d:response>
    <d:href>/remote.php/dav/trash-bin/demo/197</d:href>
    <d:propstat>
        <d:prop>
            <oc:trashbin-original-filename>.hidden</oc:trashbin-original-filename>
            <oc:trashbin-original-location>First/Second/Third/.hidden</oc:trashbin-original-location>
            <oc:trashbin-delete-datetime>Wed, 05 Jun 2019 10:10:14 GMT</oc:trashbin-delete-datetime>
            <d:getcontentlength>0</d:getcontentlength>
            <d:resourcetype/>
        </d:prop>
    </d:propstat>
</d:response>
```
so yeah ... `oc:trashbin-original-filename` is completely redundant ... will be gzipped away ... so we can just use the basename of `oc:trashbin-original-location` which is the CS3 Path.

I'll update the PR ...

butonic:
@labkode rebased and corrected the path properties as well as the owncloud storage.

labkode:
@butonic travis green and I merge

butonic:
oops, this accidentially also contained a fix for the capabilities, will submit a PR to update any config files...

PR #248, opened 2019-09-10
use correct id property, fileid is not used by the desktop client

butonic:
tested with owncloudcmd ... i refactored that away ... turns out the current owncloud client uses id ... not fileid .... hmmmm but phoenix might use file id ... should we make phoenix use the id?

yuck they have different meaning ... fileid is the oc10 numeric fileid ... id is a unique string.

### DIFF
--- a/cmd/revad/svcs/httpsvcs/ocdavsvc/propfind.go
+++ b/cmd/revad/svcs/httpsvcs/ocdavsvc/propfind.go
@@ -209,33 +209,53 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *storage
 			Status: "HTTP/1.1 200 OK",
 			Prop:   []*propertyXML{},
 		})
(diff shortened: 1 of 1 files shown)


labkode:
Let's stick to the protocol: https://github.com/cernbox/smashbox/blob/master/protocol/protocol.md

The field should be oc:id instead of oc:fileid

OC-FileId is only used in headers

butonic:
well ... fileid is used by phoenix ... maybe they can default to id. for now phoenix seems to handle `fileid` as string just fine.

however I don't know it it is used to tell the ocs sharing api to share files. if so, phoenix just seems to send back to us what we sent. and it is our job to work with the file id ...

We are bending the rules of the osc sharing api here ... oc 10 won't be able to handle that. and phoenix is intended  to work with oc10 as well.

I would always render the string, either as `id` or `fileid`, depending on what was requested.` Which is what this PR now does.

PR #249, opened 2019-09-10
Feature/ocisprovider use backend

butonic:
This PR makes the oidc provider use an auth and user manager.

Configure like this (only oidc relevant properties shown):
```toml
[http]
enabled_services = [
    "oidcprovider",
    "wellknown"
]

# HTTP middlewares

[http.middlewares.auth]
credential_strategy = "oidc"
skip_methods = [
    "/oauth2",
    "/oauth2/auth", 
    "/oauth2/token", 
    "/oauth2/introspect",
    "/oauth2/userinfo", 
    "/oauth2/sessions", 
    "/.well-known/openid-configuration",
]

[http.services.oidcprovider]
prefix = "oauth2"
auth_manager = "json"
user_manager = "json"

[http.services.oidcprovider.auth_managers.json]
users = "/data/users.json"

[http.services.oidcprovider.user_managers.json]
users = "/data/users.json"
```

### DIFF
--- a/cmd/revad/svcs/httpsvcs/oidcprovider/auth.go
+++ b/cmd/revad/svcs/httpsvcs/oidcprovider/auth.go
@@ -76,7 +76,6 @@ func (s *svc) doAuth(w http.ResponseWriter, r *http.Request) {
 		}
 		return
 	}
-	// TODO(jfd): use reva sepcific implementation that uses existing auth managers
 	actx, err := s.authmgr.Authenticate(ctx, username, password)
 	if err != nil {
 		s.oauth2.WriteAuthorizeError(w, ar, err)
--- a/cmd/revad/svcs/httpsvcs/oidcprovider/auth.go
+++ b/cmd/revad/svcs/httpsvcs/oidcprovider/auth.go
@@ -22,7 +22,9 @@ import (
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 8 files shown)


labkode:
@butonic for the auth manager this will skip any authentication going through the grpc authsvc rigtht? if that is the case I don't like it. For the user manager as we don't have one cs3 api yet for them no problem, for http services should rely solely on the grpc apisl

butonic:
The oidc provider is a fairly distinct thing from the core api services. No one is supposed to run it in production. It is intended as a bare minimum implementation that will allow you to run reva and have a working installation including oidc based authentication (because in the future all our clients will switch to it). The idea is that you can test and develop reva as a standalone system and swap out pieces to integrate it better with your infrastructure. If you don't want to nag your admins with the oidc integraion you can still use ldap with the build it oidc provider. If you feel comfortable you can replace the built in idp with something like keycloak, kopano connect d or whatever openid connect provider is already in place. Similar for the storage ... but that is a different topic.

I am using the built in oidc provider handler with the json user and auth manager in my gateway config. The backend config uses the oidc auth and user manager. It does not neet to connect to ldap or json. That is the idea. So that the oidcprovider does not use the authsvc is by design ... does that make sense?

refs:
@butonic could you provide your full configuration? I'm encountering a redirection loop on Phoenix that prevents me from logging in

PR #258, opened 2019-09-16
Import shares using share create instead of update 

IljaN:
I couldn't make it work with update share which makes sense because the  share doesn't exist.

### DIFF
--- a/cmd/reva/import.go
+++ b/cmd/reva/import.go
@@ -19,17 +19,41 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
-	"io"
 	"log"
 	"os"
 	"path"
 
 	rpcpb "github.com/cs3org/go-cs3apis/cs3/rpc"
+	typespb "github.com/cs3org/go-cs3apis/cs3/types"
+
 	storageproviderv0alphapb "github.com/cs3org/go-cs3apis/cs3/storageprovider/v0alpha"
 	usershareproviderv0alphapb "github.com/cs3org/go-cs3apis/cs3/usershareprovider/v0alpha"
 )
+
+// share representation in the import metadata
+type share struct {
+	Path           string `json:"path"`
+	ShareType      string `json:"shareType"`
+	Type           string `json:"type"`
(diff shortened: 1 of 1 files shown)


cmd/reva/import.go
```diff
@@ -19,18 +19,41 @@
 package main
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
-	"io"
+	rpcpb "github.com/cs3org/go-cs3apis/cs3/rpc"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

cmd/reva/import.go
```diff
@@ -40,133 +64,115 @@ func importCommand() *command {
...
 
-		sClient, err := getStorageProviderClient()
-		if err != nil {
-			return err
-		}
-		uClient, err := getUserShareProviderClient()
-		if err != nil {
-			return err
-		}
+		return importShares(exportPath)
```
refs: `importShares` returns an error. Could you handle the error here instead? not within the function

```go
err := importShares(exportPath)
if err != nil {
  return err
}

return nil

File: cmd/reva/import.go [hunk 3]
```diff
@@ -40,133 +64,115 @@ func importCommand() *command {
...
+	}
+	return cmd
+}
 
-		ctx := getAuthContext()
+//Imports shares from an shares.jsonl file in exportPath. The files must already be present on the storage
```
refs: typo: Imports shares from **a** shares.jsonl

cmd/reva/import.go
```diff
@@ -40,133 +64,115 @@ func importCommand() *command {
...
+		return err
+	}
+	sharing, err := getUserShareProviderClient()
+	if err != nil {
+		return err
+	}
+	ctx := getAuthContext()
+	sharesJSONL, err := os.Open(path.Join(exportPath, "shares.jsonl"))
+	if err != nil {
+		log.Fatal(err)
```
refs: following the previous comment, return the error here and deal with it in the caller

IljaN:
@refs fixed, thanks :)

PR #267, opened 2019-09-21
Add tests for crypto

IljaN:
More unit-tests


labkode:
@tboerger probably not but for the time being I merge @IljaN effort :)

PR #298, opened 2019-10-16
[WIP] Implement thumbnailssvc

TheOneRing:


cmd/revad/svcs/httpsvcs/thumbnailssvc/thumbnailssvc.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"bytes"
+	"fmt"
+	"image"
+	"image/jpeg"
+	"image/png"
+	"net/http"
+	"net/url"
+	"strconv"
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

labkode:
@TheOneRing Hi Hannah, thanks a lot for working on this PR. If it could help you in your implementation this is what we do in CERNBox for the ownCloud preview application to generate the thumbnails:
https://github.com/cernbox/revaold/blob/master/ocproxy/api/api.go#L5667

internal/http/services/thumbnails/thumbnails.go
```diff
@@ -0,0 +1,221 @@
...
+	Request *http.Request
+	URL     string
+
+	FileType string
+	Size     uint
+
+	TempFileName string
+}
+
+func newThumbnailHandler(s *svc, w http.ResponseWriter, r *http.Request) *thumnailhandler {
```
refs: functions are package scoped. a more appropriate name would be `newHandler`.

In general I'd move away from postfixing this with "-Handler" as it can be read as "thumbnailHandler implements the Handler interface" [source](https://golang.org/pkg/net/http/#Handler). Instead, your type `thumbnailHandler` is an abstraction that does not implement such interface.

You're implementing the Handler interface on your thumbnails>service.go on `svc` [method set](https://golang.org/ref/spec#Method_sets)

internal/http/services/thumbnails/thumbnails.go
```diff
@@ -0,0 +1,221 @@
...
+		u = &url.URL{}
+		u.Scheme = "http" // from where should that come
+		u.Host = r.Host
+	} else {
+		var err error
+		u, err = url.Parse(s.conf.WebDavHost)
+		if err != nil {
+			t.Log.Err(err).Msg("Failed to partse host: " + s.conf.WebDavHost)
+			t.Writer.WriteHeader(http.StatusInternalServerError)
+			return nil
```
refs: For the time being it does the job, but I wonder if it'd be possible not having this side effects, or decoupling the side effects from `t` and instead of mutating `t` along the body of the function, instead return a [composite struct literal](https://golang.org/ref/spec#Composite_literals) with the fields filled. It makes the logic easier to follow.

refs: One way of doing so is doing logging at the callers and return the error, this way `t` shouldn't need a logger.

butonic:
moved to https://github.com/owncloud/ocis-thumbnails

labkode:
Closed, moved to:
https://github.com/owncloud/ocis-thumbnails

PR #319, opened 2019-10-17
update standalone.oidc.toml and related config

butonic:
the implicit flow with `esponse_types = ["id_token token"]` is currently broken for phoenix: https://github.com/owncloud/phoenix/issues/2288

so we fall back to code

also added a phoenix config.json for that flow that corresponds to the standalone.oidc.toml


labkode:
Awesome, for history, phoenix in version 0.2.24 worked with the previous config

PR #328, opened 2019-10-22
implement context based path wrapper for home storage provider

butonic:
This PR
1. implements a context based path wrapper for a `/home` storage provider
  - use it to mount a storage provider at `/home` in the cs3 namespace that will prefix paths with the username from the context. requires users to have a non empty username.
- other path wrapper implementations can be added by implementing the interface

2. introduces a namespace prefix for dav/files handler
 
@labkode this allows prefixing the path before making a cs3 call. For now I hardcoded the namespace `/home` for the `remote.php/webdav` endpoint and introduced a configurable namespace for the `remote.php/dav/files` endpoint that defaults to `/` eke the root, which is what works with eos in your setup. In the examples I explained how to mount a storage provider with the owncloud driver at `/oc` in the cs3 namespace and prefix the path for the `remote.php/dav/files` endpoint so that paths are handled correctly.


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -739,14 +801,17 @@ func (s *service) trimMountPrefix(fn string) (string, error) {
 	return "", errors.New(fmt.Sprintf("path=%q does not belong to this storage provider mount path=%q"+fn, s.mountPath))
 }
 
-func getFS(c *config) (storage.FS, error) {
-	if f, ok := registry.NewFuncs[c.Driver]; ok {
-		return f(c.Drivers[c.Driver])
+func (s *service) wrap(ctx context.Context, ri *storageproviderv0alphapb.ResourceInfo) error {
+	ri.Id.StorageId = s.mountID
```
labkode: We must not change storage ids

butonic: That is the code from the old wrap(). AFAICT it is used to fill the storage id with the storage providers mountid, because the storage driver does not provide it.

PR #345, opened 2019-10-31
build: fix Makefile to use go env GOPATH

labkode:
Needed to export the GOPATH/bin for building on Mac.

Take the opportunity to also clean deps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ VERSION=`git describe --always`
 GO_VERSION=`go version | awk '{print $$3}'`
 
 off: 
+	echo ${PATH}
 	GORPOXY=off
 	echo BUILD_DATE=${BUILD_DATE}
 	echo GIT_COMMIT=${GIT_COMMIT}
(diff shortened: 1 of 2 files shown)


tboerger:
Using go run is much more accurate, than anybody is also using the same defined version of the tools as they are managed by the go modules config.

labkode:
@tboerger I tried with go run but to version I need to version the dep inside the go.mod in Reva, running go run github.com/..@v1.0.0 outside the the reva repo didn't work for me.

If you can give it a try I prefer to just run go run that this workaround!

tboerger:
What do you have against versioning it via go.mod? That ensures that this tool is really the same version for everybody. Imho that's a benefit as I gathered the knowledge that it could lead to different behavior otherwise.

labkode:
@tboerger the license of `golangci-lint` is GPL and such license cannot be part of any Reva dependencies to guarantee its permissive adoption, that's the reason why I can't put it into the dependencies of the project.

tboerger:
You could also replace golangci-lint by https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck which is MIT licensed :)

labkode:
golangci-lint has more checks than staticcheck, indeed staticcheck is one of the test we run:

```
linters:
  enable:
    - bodyclose
    - deadcode
    - errcheck
    - gosimple
    - govet
    - ineffassign
    - staticcheck
    - structcheck
    - typecheck
    - unused
    - varcheck
    - depguard
    - golint
    - goimports
    - unconvert
    - scopelint
    - maligned
    - misspell
    - gocritic

```

PR #281, opened 2019-10-01
WIP: List Outgoing Shares

refs:
see https://github.com/owncloud/ocis/issues/16. Once the logic runs, a refactor would be nice.

Features:
- [x] ~User shares~
- [x] ~Public shares~

Extras:
- [ ] Add reva command for listing public shares (hook it into `reva share-list` or create a new subcommand?)

new directive added: `publicshare_manager`

```toml
[http.services.ocssvc]
    publicshare_manager = "memory"
```

PR #347, opened 2019-11-04
List shares (Phoenix)

refs:
# What?

Rebuild of https://github.com/cs3org/reva/pull/281 with the latest changes.

Tasks:
- [x] User shares
  - [x] List received shares
- [x] Public shares

Leaving public shares for a different PR along with the dav api. As well as more CRUD operations regarding shares, such as delete.

Disclaimer:

Since getting a given user's home directory is still being discussed and the current implementation relies on iterating on a map keys, which is, by design, unreliable in go, the result of `GetHome()` is non-deterministic. For instance for config file with the following directives:

```toml
[grpc.services.storageregistry.drivers.static.rules]
"/home" = "localhost:12000"
"/oc" = "localhost:11000"
"123e4567-e89b-12d3-a456-426655440000" = "localhost:11000"
```

a query for a resource with path `/example/txt` will get appended either of `/home`, `/oc` or `123e4567-e89b-12d3-a456-426655440000"` where the resource exists only on `/home`


internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -0,0 +1,408 @@
...
+
+const (
+	RoleLegacy  string = "legacy"
+	RoleViewer  string = "viewer"
+	RoleEditor  string = "editor"
+	RoleCoowner string = "coowner"
+)
+
+const (
+	PermissionInvalid Permissions = 0
```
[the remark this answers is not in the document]
refs: looks like a bug on Hound. Closing

labkode:
@refs for the publicshare_manager yes we still need it as we want to plugin different backends, same model as the usershare_manager.

internal/http/services/owncloud/ocs/ocs.go
```diff
@@ -32,10 +32,11 @@ func init() {
...
 type Config struct {
-	Prefix       string           `mapstructure:"prefix"`
-	Config       ConfigData       `mapstructure:"config"`
-	Capabilities CapabilitiesData `mapstructure:"capabilities"`
-	GatewaySvc   string           `mapstructure:"gateway"`
+	Prefix         string           `mapstructure:"prefix"`
+	Config         ConfigData       `mapstructure:"config"`
+	Capabilities   CapabilitiesData `mapstructure:"capabilities"`
+	GatewaySvc     string           `mapstructure:"gateway"`
+	FilesNamespace string           `mapstructure:"files_namespace"`
```
refs: @butonic I haven't found another way to expose this path to the service other than this, since storage registries are not going to be part of the gateway, there is no way to use `GetHome`. Ideas / changes are welcomed

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -269,6 +269,7 @@ func (s *svc) UnsetArbitraryMetadata(ctx context.Context, req *storageproviderv0
 }
 
 func (s *svc) Stat(ctx context.Context, req *storageproviderv0alphapb.StatRequest) (*storageproviderv0alphapb.StatResponse, error) {
+	// TODO(refs) do we need to append home to every stat request?
```
labkode: Not at all, the gateway doesn't do any path manipulation, only find the available storage providers and send request there, is the storage provider that exposes /home. 

Gateway is just dummy :)

internal/http/services/owncloud/ocs/ocs.go
labkode: @refs now I understand your question for the storage registry in the gateway, we can add the GetHome calls to the gateway, indeed, it makes much sense to do it.

labkode: The GetHome call will just give you the home path for a user, but the client still doesn't get exposed to the storage registry neither storage providers directly.

Let me add that to the CS3APIS

refs:
@labkode ready to be merged :shipit:

refs:
> With the global namespace that is used in the example config, the gethome call is no longer needed, because the webdav namespace is the same as the cs3 namescpace and makes the sharing api use ''absolute" paths.

Reviewed the commit and went along with it. Regardless of that, for the case you're describing the code relies on a specific way to configure it, either we support both cases with a fallback case or go along with your config and get back to this later. WDYT? /cc @butonic

butonic:
I'd merge this, so we can continue working on sharing. We will discuss how to handle the namespace migration in https://github.com/owncloud/ocis/issues/44

Then we can think about weather or not we need to adjust the code again.

labkode:
I merge this now even though I need to dedicate more time to review it and won't be able to do so until the week of 25 when ai come back from holidays. I'm merging PRs from the phone

PR #355, opened 2019-11-07
allow eos to use keytab for authentication rather than gw

labkode:
Current EOS authentication relies on trusted gateways.

This change allows the connection from the storage provider to EOS by relying on EOS shared secrets, known as keytabs.


labkode:
@madsi1m this is for you :)

madsi1m:
wooooooo

PR #370, opened 2019-11-12
List recycle

butonic:
Make the gateway handle trash requests
- don't be smart: don't try to query all storages for a list trash request. 
  - merging multiple results is hard
  - trashes are specific to a storage provider
  - let clients do the work of merging / querying multiple storages
    - ocdav will query the storage registry and only use storage entries that don't start with a / so we do not get duplicate entries for /home and /oc
- fix listing trash
- fix undelete
- implement purge items
- also, allow PUT to overwrite files

### DIFF
--- a/examples/separate/frontend.toml
+++ b/examples/separate/frontend.toml
@@ -70,8 +70,32 @@ secret = "Pive-Fumkiu4"
 
 [http.middlewares.cors]
 allowed_origins = ["*"]
-allowed_methods = ["OPTIONS", "GET", "PUT", "POST", "DELETE", "MKCOL", "PROPFIND", "PROPPATCH", "MOVE", "COPY", "REPORT", "SEARCH"]
-allowed_headers = ["Origin", "Accept", "Depth", "Content-Type", "X-Requested-With", "Authorization", "Ocs-Apirequest", "If-None-Match"]
+allowed_methods = [
+	"OPTIONS",
+	"GET",
+	"PUT",
+	"POST",
+	"DELETE",
+	"MKCOL",
+	"PROPFIND",
+	"PROPPATCH",
+	"MOVE",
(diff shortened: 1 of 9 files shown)


labkode:
This PR is against the model of the gateway. The ocdav client as other client doesn't' have access to the storage registry. If you think you need to query recycles for storages then the ListRecycleRequest could include a path to filter out results. 

> * ocdav will query the storage registry and only use storage entries that don't start with a / so we do not get duplicate entries for /home and /oc

You don't get duplicate entries as the file ids for those storage providers are the same

butonic:
What should the gateway do if no ref is given. List all trash files of all storages? I implemented the gateway to currently require a ref, so that it can limit the results to a specific storage. The code also forwards the [FromTs ond ToTs as wall as any Opaque data](https://github.com/cs3org/reva/pull/370/files#diff-87a927da7f0ac3d45225420a90987c21R394-R398) to the storage in the Ref.

That part is ok for now and follows the logic you describe above.

About the ocdav part:
phoenix currently does not send a path. So we need to fetch the trash items from all storage the user has access to. [That is why I iterate over the storage providers.](https://github.com/cs3org/reva/pull/370/files#diff-a84d538a23e7f2291e3623e6c7976ecaR180-R181)

If we want to support fetching trash items from all storages in the gateway we would need to move that iterating and filtering into the gateway. I was under the impression that we should try to avoid that kind of magic in the cs3api and encourage clients to implement smarter trash lookups, eg only for the storage the user is currently looking at.

AFAICT the ocdav implementation in this PR is the least bad but backwards compatible one.

labkode:
@butonic sorry my bad, I was wrong. The code makes sense. You are already querying the correct storage provider for getting the recycle items. Merging :)

PR #372, opened 2019-11-13
initial favorites support

butonic:
good enough for testing
- only works for the owncloud storage driver
  - eos and actually any storage drive would have to take into account if a property is user specific or not... 
    - user individual metadata sucks
  - storing them in extended file attributes would leak favorites (and tags) but more flawed: users can make any of their owned files a favorite for another user by writing the corresponding xattr ... we need to check if we can use a xattr namespace (TRUSTED?) that requires a different privilege... see [CAP_SYS_ADMIN](http://man7.org/linux/man-pages/man7/capabilities.7.html) and [how to set it](https://unix.stackexchange.com/questions/454708/how-do-you-add-cap-sys-admin-permissions-to-user-in-centos-7)
    - CAP_SYS_ADMIN way to powerful a capability. friends don't let friends run reva as root ...
- we can learn from this and reimplement favorites on top of a proper tagging service
  - raised discussion in https://github.com/cs3org/cs3apis/issues/43

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -251,8 +251,17 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *storage
 			response.Propstat[0].Prop = append(response.Propstat[0].Prop, s.newProp("oc:checksums", value))
 		}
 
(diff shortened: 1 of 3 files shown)


labkode:
@butonic I merge now but let's follow up on the discussion on CS3, I'll involve Andreas from EOS as he is dealing with the same challenge :)

PR #384, opened 2019-11-21
return user on auth

butonic:
This PR makes the AuthManagers implementations return a full user object instead of just the id.
- oidc and ldap already provide the user metadata as part of the authentication process
  - more specifically:
    - the credentials are extracted using a credential strategy in the chain
    - they are passed to the gateway which will handle basic or bearer auth with the configured auth provider
    - the auth provider returns all user metadata
      - before this PR the auth provider would only return an id, which the gateway then used to look up a user in the user provider
    - this allows us to authenticate user with oidc bearer auth (and fetch the user metadata from the userinfo endpoint) as well as ldap basic auth (and fetching the user metadata from ldap)
    - if the metadata needs to be extended, or only users should be allowed in via oidc that are also avaliable in ldap a separate middleware / inspector should handle that. 
- the user manager is only used to search users for sharing

- [x] update json & demo auth manager to return all metadata
- [ ] update examples
- [ ] requires https://github.com/cs3org/cs3apis/pull/46 / https://github.com/cs3org/go-cs3apis/pull/20

### DIFF
--- a/cmd/reva/gen/gen.go
+++ b/cmd/reva/gen/gen.go
@@ -379,8 +379,6 @@ var usersTemplate = `[{{range  $i, $e := .}}{{if $i}},{{end}}
 			"idp": "{{$e.Iss}}",
 			"opaque_id": "{{$e.Sub}}",
 		},
-		"sub": "{{$e.Sub}}",
-		"iss": "{{$e.Iss}}",
 		"username": "{{$e.Username}}",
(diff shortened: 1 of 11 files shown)


labkode:
@butonic it makes sense, let me start reviewing the cs3 apis first.

butonic:
merged as part of https://github.com/cs3org/reva/pull/393

Issue #395, opened 2019-12-02
routing and authentication clarification


butonic:
Currently, the idea is that the gateway serves the cs3 api to the outside world and any client like the ocdav service. It is a facade that is used to authenticate requests and route the request to the correct cs3 service.
Authentication and routing go hand in hand, because the user may have an influence on the routing decision.

For larger deployments oidc or at least ldap are used. Both have a way to return user metadata as part of the user authorization process, which is why we want to be able to read them during authentication: https://github.com/cs3org/reva/pull/384

We still only use the userid from that response and send it to a user manager that will return the actual user object: https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/authprovider.go#L93

Regarding authentication
- [ ] get rid of the user metadata lookup. The auth manager drivers can return a proper user object. This is `auth`entication. oidc and ldap can return all the necessary user metadata. The user identifier might not be as stable as we want, but we need a cli tool to change it anyway (tracked in https://github.com/owncloud/ocis/issues/58).
- [ ] add an optional `identification` middleware that maps the user to a stable id: it will get a full userobject and can be used to maintain a mapping to a uuid based userid. For example a basic auth against an ldap server could use an `identitiy` middleware to lookup a stable user id in a SCIM services using his email. It would then replace the id in the user object with that id. If for whatever reason the email, the dn ort the uuid in ldap chinges reva can still identify the user using the SCIM service (this obviously only makes sense if SCIM is also configured as a user provider to handle tho userid to metadata lookup, eg. when listing file metadata)

Regarding authorization:
- [ ] we need to be able to add roles and permissions in a yet to implement `account` middleware, because oidc does not specify how they should be sent as claims. not every idp will be able to handle the permissions we want to use in reva / cs3

Regarding Credentials:
In some cases a service might want to send its own credentials to the client, eg. WOPI. Thes credentials need to be passed on to the service, even if the CS3 api is used to wrap it. Another example is a windows network drive  storage provider which needs to be able to send the WWW-Authenticate header in order to send the challenge to the client: https://tools.ietf.org/html/rfc4559#section-4.1

We currently replace the whatever credentials were given with a JWT that services behind the gateway use to authenticate requests. The idea isn to

butonic:
[...]

PR #397, opened 2019-12-02
make oidc claim to use as opaqueid configurable

butonic:
This PR allows configuring which claim to use as the reva opaqueid. See https://github.com/cs3org/reva/issues/395#issuecomment-560410958 for details


madsi1m:
does this do anything?

labkode:
@madsi1m  Yes, allow to configure the Unique ID of the user if you don't want to use the "sub" field, for example, at CERN we use cern_upn as the unique field rather than "sub"

labkode:
Merged in master

PR #414, opened 2019-12-05
[WIP] grpc: move auth config to service

labkode:


internal/grpc/interceptors/auth/auth.go
```diff
@@ -68,7 +59,7 @@ func parseConfig(m map[string]interface{}) (*config, error) {
 
 func skip(url string, skipped []string) bool {
 	for _, s := range skipped {
-		if strings.HasPrefix(s, url) {
```
refs: We could DRY this skip function and export the `skip` method from the http auth middleware

https://github.com/cs3org/reva/blob/master/internal/http/interceptors/auth/auth.go#L79-L88

or at least write a test for this one to be on the safe side

https://github.com/cs3org/reva/blob/master/internal/http/interceptors/auth/auth_test.go

I totally forgot to update the interceptors when updated the skip on the HTTP middlewares

labkode: Complete agree :)

labkode: I've moved the skip logic to a utils package (yes, I know but I could'nt find a better name) :)
@refs Let me know what you think

labkode:
@refs @butonic would be good to test if this branch doesn't break any authentication flow.

It also introduces a change in the OIDC implementation to rely solely on the library for token verification.

I've tested the OIDC flow with the provided Reva Idp and with a KeyCloack IdP we use here. I don't expect major havoc but worth checking if it doesn't break for Kopano.

internal/grpc/interceptors/auth/auth.go
refs: `utils` although against [package naming conventions](https://blog.golang.org/package-names) should be good enough for us. We can always refactor this in the future 🚀

labkode: @refs  I couldn't find a better name to throw it, any naming better than this will do :)

PR #416, opened 2019-12-06
WIP: Make revad start on a per directory basis

refs:
Added a new `--dir` flag that allows for starting `revad` on any directory with a set of config files.

Having a directory like so:

```
~/code/reva
❯ tree ~/automation/reva/separate
/Users/aunger/automation/reva/separate
├── auth.toml
├── frontend-oidc.toml
├── frontend.toml
├── gateway.toml
├── phoenix.oidc.config.json
├── storage-home.toml
├── storage.toml
└── users.demo.json
```

Would allow for `revad` to launch every service defined on each config file on the given folder like:

`> revad --dir ~/automation/reva/separate`

This is a preliminary implementation.

It all lives under the same terminal session so all logs are printed on STDOUT. This is more intended to speed up development as it does not handle graceful shutdown just yet.

### DIFF
--- a/cmd/revad/internal/grace/grace.go
+++ b/cmd/revad/internal/grace/grace.go
@@ -35,7 +35,7 @@ import (
 )
 
 // Watcher watches a process for a graceful restart
-// preserving open network sockets to avoid packets.
+// preserving open network sockets to avoid packet loss.
 type Watcher struct {
 	log       zerolog.Logger
 	graceful  bool
--- a/cmd/revad/main.go
+++ b/cmd/revad/main.go
@@ -56,7 +56,7 @@ var (
 	signalFlag  = flag.String("s", "", "send signal to a master process: stop, quit, reload")
 	configFlag  = flag.String("c", "/etc/revad/revad.toml", "set configuration file")
 	pidFlag     = flag.String("p", "", "pid file. If empty defaults to a random file in the OS temporary directory")
-	dirFlag     = flag.String("dir", "", "runs any toml file in the directory specified")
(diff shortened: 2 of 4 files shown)


madsi1m:
Good idea, but i prefer to run each bit in its own container so i can filter logs easy

refs:
it is still possible to run each bit on it's own process :D, this would just make it faster to spin something up

labkode:
Hi @refs, @madsi1m, for testing/dev purposes we can have such funcionality and even with restart logic, we just need to fork one chidren process per configuration file in the directory and the master process will be a dummy process.

Is not the responsibility of Reva to handle the orchestration and observability of the different services, that is delegated to enterprise-grade components like Istio/Envoy and K8s. 

Each tool for the right thing

refs:
> Is not the responsibility of Reva to handle the orchestration and observability of the different services, that is delegated to enterprise-grade components like Istio/Envoy and K8s.

100% agree, although that was not the intention of this PR. As more people join the project we're in the case that we need to explain how to run each individual config "strategy" (strategy as in 1 single file vs `n` files) when Reva, for some, is a mean to an end as they're working on the oC clients.

I'm more than open to suggestions on how to improve this, as I believe it will streamline development.

labkode:
@refs I agree for on boarding and developing against make sense.

Can you change the flag name to dev-dir instead of dir so is clear that is only for dev purposes?

Thanks!

madsi1m:
Great idea

labkode:
@refs merged :)

refs:
💃 thanks!

PR #421, opened 2019-12-10
Create and save reva releses on commits to master branch

zazola:
Adding the feature of creating commit based release in website.


tboerger:
You have not checked to simply use http://plugins.drone.io/vividboarder/drone-webdav/ for uploading, right?

labkode:
@tboerger ups! good catch, definitely should we use that

PR #428, opened 2019-12-12
Export Reva runtime

labkode:
Move main logic to package.


refs:
🎉

cmd/revad/runtime/runtime.go
```diff
@@ -0,0 +1,396 @@
...
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	"go.opencensus.io/plugin/ocgrpc"
+	"go.opencensus.io/plugin/ochttp"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/trace"
+)
+
+// Run runs a reva server with the given config file and pid file.
+func Run(config string, pidFile string) {
```
butonic: Can we pass in the map[string]interface{} here so we don't have to read the config from disk? An io.Reader would work as well, but passing in a map would be the least overhead.

labkode: @butonic of course :)

labkode:
@butonic @tboerger We'll add later log redirection so you can consume it natively in OCIS.

PR #436, opened 2019-12-13
auto create home folder if it is missing in eos

madsi1m:


labkode:
Uhmm this responsibility should be of the gateway on the GetHome call, @butonic what you think?

I prefer to create home directories on https://github.com/cs3org/reva/blob/master/internal/grpc/services/storageregistry/storageregistry.go#L125
maybe even having a new CS3 API method named CreateHome?

madsi1m:
Would be good to have this feature for all storage providers. I know the owncloud storage provider does it already. The trick is to only create it for home directories.

labkode:
I will add the creation of the home logic when the user does the logic, I think that is the place. We've been doing it like that and OC also in the PHP code in the setupFS method.

madsi1m:
Sounds good. Thinking a little further into the future, I foresee some storage’s that we will not want user home folders created on and some that we will. Can we include a config option in the storage provider config to allow/disallow this?

labkode:
That is the reason why we have a GetHome call, so only Storage Providers that are returned in that call will have a home created.

labkode:
Closed in #469 once merged

PR #446, opened 2019-12-16
misc: skip contrib makefile step

labkode:
Some people use different git usernames and mails and once the PR is accepted into master is impossible without rewriting history to change the author name.

This PR removes the automatic creation of the CONTRIBUTORS.md file by analyzing the git log and instead relies on every author to put their username and mail accordingly in the file.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,12 +1,11 @@
 - Alex Unger <zyxancf@gmail.com>
+- Aritz Brosa <aritz.brosa.iartza@cern.ch>
 - Diogo Castro <diogo.castro@cern.ch>
 - Felix Hillingshaeuser <felix@mxcore.de>
-- Giuseppe <giuseppe.lopresti@cern.ch>
 - Giuseppe Lo Presti <giuseppe.lopresti@cern.ch>
 - Hannah von Reth <vonreth@kde.org>
 - Hugo Gonzalez Labrador <github@hugo.labkode.com>
 - Ilja Neumann <ineumann@owncloud.com>
 - Jörn Friedrich Dreyer <jfd@butonic.de>
-- Michael D'Silva <md@aarnet.edu.au>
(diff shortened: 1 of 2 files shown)


tboerger:
That's why projects sometimes also provide a mailmap file like https://github.com/owncloud/core/blob/master/.mailmap

labkode:
@tboerger the problem is that when they are automatically generated, in the one you posted is the same problem, you have people repeated multiple times.

madsi1m:
You could also do a .raw file and a clean one where the .raw is automatically generated by the git log | sort | grep -v etc and the clean one people manually populate

tboerger:
The mentioned mailmap is doing the mapping, they are not duplicated.

labkode:
@tboerger in the file for example this person appears three times:

```
Bernhard Posselt <dev@bernhard-posselt.com> Bernhard Posselt <nukeawhale@gmail.com>
Bernhard Posselt <dev@bernhard-posselt.com> Bernhard Posselt <Raydiation@users.noreply.github.com>
Bernhard Posselt <dev@bernhard-posselt.com> Bernhard Posselt <BernhardPosselt@users.noreply.github.com>
```

tboerger:
Right, the first part if the mapping target, the last part is mapping the alternatives to the right name and email.I think https://github.com/git/git/blob/master/Documentation/mailmap.txt explains pretty well the structure of these mailmap files.

labkode:
@tboerger thanks for the pointer. How do you create those mailmaps? I wouldn't mind a PR :)

tboerger:
I would start with the current duplicates, put them into the mailmap, and if we find another duplicate within the contributors file it got to be added to the mailmap.

I will create a PR for an initial one.

tboerger:
There you go, the contributors file is updated based on the mailmap content: https://github.com/cs3org/reva/pull/452

PR #451, opened 2019-12-18
runtime: add host info on boot

labkode:


tboerger:
Why do you want to care about the hostname within the log? That should be a task for the log forwarding of the host.

labkode:
@tboerger if helps people on the setup @madsi1m had some problems figuring out the hostname in a kubernetes cluster. It only prints the host information, it doesn't add it as a field for all the logs.

PR #469, opened 2020-01-13
Create home directory on login on configured home storage providers and other stuff

labkode:
This PR allows to configure the gateway to create a user home directory on login.

It also adds a configuration option on the storage providers that provide "homes" for user to create their homes.

The PR only add the creation of homes for the eos storage.


labkode:
The PR also cleans the configuration adding a new [shared] section to avoid having to repeat common configurations like JWT tokens and gateway configurations.

Issue #473, opened 2020-01-15
Add layout configuration for path wrapper


labkode:
In pkg/storage/pw/context/context.go we can add a new config like:

```
type config struct { 
       Layout string `mapstructure:"layout"`
        Prefix string `mapstructure:"prefix"`
}

and then on the Wrap and Unwrap methods there is some string substitution.

For example:

```
[x.y.z]
layout = "{{.Provider}}/{{.Username}}"

```

And the we use Go templates to fill those variables:

```
u := context.GetUser(ctx)
t := template.Must(template.New("").Parse(conf.Layout))
```

madsi1m:
How about https://github.com/cs3org/reva/pull/476 ?

PR #476, opened 2020-01-16
Layout template 

madsi1m:
Addresses https://github.com/cs3org/reva/issues/473
Requires https://github.com/cs3org/reva/pull/469 to be merged first!


madsi1m:
Build will break because it needs https://github.com/cs3org/reva/pull/469 first

pkg/storage/helper/helper.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/cs3org/reva/pkg/errtypes"
+	"github.com/pkg/errors"
+
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+)
+
+type layoutTemplate struct {
+	Username      string //the username
+	UsernameLower string //the username in lowercase
+	FirstLetter   string //first letter of username in lowercase
```
butonic: Currently, `FirstLetter` limits that dir to a single rune. I'd like to see a variable length to use as the name for the parent dir. I lack the words for a better name of the `Parent` directory. Prefix does not cut it. Maybe `bucket`. For s3 storages we were considering the first n chars of an md5 / sha1 of the username to spread users among multiple buckets. But never followed up on it because randomly putting users into buckets seems nice, but in practice you'll end up with unevenly filled buckets when a few powerusers land in the same bucket. Which is why we stored a users home in the account table. That allows moving a single user to a different storage ...

This is better than the current code so still a +1 from me ;-)

madsi1m: Good idea, added `{{.UsernamePrefixCount.x}}``

labkode:
@madsi1m the layout configuration should not go into individual storages like EOS. The path wrapper   configuration should suffice to make that happen. Is there any reason why the layout configuration lives also on the eos storage?

madsi1m:
Yes, at first I wrote it just for the pw and the eos storage wasn’t using it, for example the createhome doesn’t get the home from the pw. So I changed it to a helper and included it in the storages

labkode:
@madsi1m I see so then there is no reason for having the layout on the pw, right?

madsi1m:
I guess but I’m not sure what else uses the pw so I put it there also

labkode:
@madsi1m if you can remove then I'm happy to merge ;)

madsi1m:
done

PR #503, opened 2020-02-12
add more owncloud capabilities

labkode:


internal/http/services/owncloud/ocs/capabilities.go
```diff
@@ -132,6 +132,10 @@ func (h *CapabilitiesHandler) init(c *Config) {
 		h.c.Capabilities.Dav.Reports = []string{"search-files"}
 	}
 
+	if h.c.Capabilities.Dav.ChunkingParallelUploadDisabled {
```
labkode: This is wrong ... serves to nothing

internal/http/services/owncloud/ocs/capabilities.go
```diff
@@ -287,9 +287,10 @@ type CapabilitiesFiles struct {
 
 // CapabilitiesDav holds dav endpoint config
 type CapabilitiesDav struct {
-	Chunking string   `json:"chunking" xml:"chunking"`
-	Trashbin string   `json:"trashbin" xml:"trashbin"`
-	Reports  []string `json:"reports" xml:"reports>element" mapstructure:"reports"`
+	Chunking                       string   `json:"chunking" xml:"chunking"`
+	Trashbin                       string   `json:"trashbin" xml:"trashbin"`
+	Reports                        []string `json:"reports" xml:"reports>element" mapstructure:"reports"`
+	ChunkingParallelUploadDisabled bool     `json:"chunkingParallelUploadDisabled" xml:"chunkingParallelUploadDisabled"`
```
butonic: ref https://github.com/owncloud/client/blob/master/src/libsync/capabilities.cpp#L153

labkode: Good pointer!

PR #505, opened 2020-02-12
File share manager

labkode:
Store shares in a JSON file, better for testing than in memory.


butonic:
We had the idea of persisting the grants in the filesystem, but as individual links, so the kernel can cache the stat requests. needs some discussion. will open an issue and try to describe how that might work.

PR #510, opened 2020-02-13
add a storage template package to define the user home directory

labkode:
This PR adds a new storage template package to manipulate storage layouts based on concrete and common types like a User.

The storage implementations (eos, owncloud, s3) will have a new configuration field: user_layout that if defined will be used for the storage path.

```
$reva ls /global/users/g/gonzalhu/
Documentation
$ reva ls /home/
Documentation
```

with config:

```
[grpc.services.storageprovider]
driver = "eos"
mount_path = "/home"
mount_id = "123e4567-e89b-12d3-a456-426655440000"
expose_data_server = true
data_server_url = "http://localhost:12001/data"
enable_home_creation = true

[grpc.services.storageprovider.drivers.eos]
namespace = "/eos/test/reva/users"
master_url = "root://xxxx.cern.ch"
enable_home = true
user_layout = "{{substr 0 1 .Username}}/{{.Username}}"
```


labkode:
@madsi1m this simplifies the logic of pw and the layout that you did in only one place.

If you miss something we can add it to the UserData struct.

madsi1m:
@labkode I like it!
Created a PR on your branch https://github.com/labkode/reva/pull/4

PR #523, opened 2020-02-26
support creating references on the storage provider

### LINKED ISSUE #572: create references on home storage provider
Implements also the EOS logic

labkode:
Fixes #572 

- [x] Implements Move operation on the gateway
- [x] on accepting shares, the gateway will create references on the home storage provider
- [x] Support a shadow namespace on EOS to allocate share references


pkg/storage/fs/eos/eos.go
```diff
@@ -216,29 +252,68 @@ func (fs *eosStorage) wrap(ctx context.Context, fn string) (internal string) {
 	} else {
 		internal = path.Join(fs.conf.Namespace, fn)
```
butonic: By returning the path as string you are leaking the storage internal namespace to a cs3 client. We should use an id based reference here.

labkode: In EOS this is not a problem because setting ACLs on the storage grants you access to the tree explicitly. 

Furthermore, navigating by ids while sounds attractive it brings an important performance penalty. Navigating with paths allows the gateway to decide if the path is inside the share folder (*/home/MyShare*) and trigger a dereference operation (stat to convert id to path) only on that tree segment. Allowing navigation by ids means triggering dereferencing for any operation.

butonic: I was more confused by GetHome exposing the internal path. Which was a misunderstanding. The storage driver can use GetHome to tell the API consumer which external path can be used to access the users home.

PR #559, opened 2020-03-13
ocis storage driver

butonic:
A traditional filesystem uses paths to organize files in a tree. But path based references are not persistent because they change when a file is renamed or moved. 
OCIS needs a persistent fileid to let other services reference files. Shares and favorites for example should not break when a file is moved or renamed.

A naive approach would use the fileid as the key and the path as the value for an efficient lookup. But this leads to n updates in the lookup table when a folder is renamed, because the path of all children changes.

Depending on the underlying storage a fast lookup by inode (btrfs, gpfs) or a truly unique fileid (eos) might be possible. For these a dedicated storage driver should be used. But keep in mind that inodes are reused, which might cause a purely inode based reference to point to a different file. Gpfs has an option to disable reusing inodes.

Ocisfs deconstructs a filesystem to provide a persistent fileid, an efficient file by ID lookup at the cost of more inodes.

```
/var/tmp/ocis/root/
├── nodes
│   ├── 1dd84abf-9466-4e14-bb86-02fc4ea3abcf
│   ├── 5425aaf2-8fa4-497e-b0e6-ed3f0809e350
│   │   └── bif -> ../d1bba60c-87a9-41a3-88a2-abd221abc100
│   ├── 8f16ec6c-6314-4f44-b553-d97abd464d06
│   │   ├── [MS-SFU].pdf -> ../b3951b72-cce9-4ab0-9992-f13b164ebb5b
│   │   └── foo -> ../9822b551-3c9a-4291-8a3a-4fccdbe8bcaa
│   ├── 9822b551-3c9a-4291-8a3a-4fccdbe8bcaa
│   │   └── bar -> ../5425aaf2-8fa4-497e-b0e6-ed3f0809e350
│   ├── b3951b72-cce9-4ab0-9992-f13b164ebb5b
│   └── d1bba60c-87a9-41a3-88a2-abd221abc100
├── trash
│   └── 1dd84abf-9466-4e14-bb86-02fc4ea3abcf -> ../nodes/1dd84abf-9466-4e14-bb86-02fc4ea3abcf
├── uploads
│   ├── 162c3dc6-b117-434d-b6fb-f1a350017d68.info
│   └── 322fa701-12ad-4a77-a165-28199a96aecc.info
├── users
│   └── 4c510ada-c86b-4815-8820-42cdf82c3d51 -> ../nodes/8f16ec6c-6314-4f44-b553-d97abd464d06
└── versions
```

Every file and folder is represented by an entry in the nodes folder. Directories have a children subfolder containing symlinks to child nodes. Files have a content file. A versions subfolder contains timestamp content entries. 

While this node based layout may look confusing at first sight not everything is lost. The symlink pattern used in the children folder allow navigating the filesystem using the traditional tree hierarchy. In fact it is used to implement the path based lookup.

*Why the content and children nodes? Why not use a uuid file or folder? Because we need a parent link? We can store that as an extended attribute instead of the parentname symlink.*
*We need the filename (changes on rename) and parentid (changes on move). This duplication might cause inconsistencies. Which could be mitigated using a journal with file operations.*

The etag is an extended attribute anyway. Additional properties could be stored either in the fs as extended attributes or as an additional file in a root/properties/uuid Json or ini file. Might be an option

Versions could be stored in additional root/versions/uuid/timestamp files. 

Basic file management is implemented, but this is missing:
- [x] Rename a node (inside the same directory)
- [x] Move (a node into another directory)
- [x] etag propagation
- [ ] Sharing, take from owncloud driver
- [x] Delete to trash
  - [ ] link to trashed node in a users trash folder
  - [x] make From* methods walk the path to determine if the node has been deleted parents)
- [ ] List Trash
- [ ] Restore trash item
- [ ] Versions
- [x] propagation for Move fails, see https://github.com/cs3org/reva/pull/559#issuecomment-684889931
- [x] upload
- [ ] extended attributes
- [x] get rid of FillParentAndName, BecomeParent and 
  - blob as content?
  - or using versions?
  - content addressed?

Feedback welcome.


butonic:
Move fails with
```
2020-09-01T16:17:44+02:00 INF unary code=OK end="01/Sep/2020:16:17:44 +0200" from=tcp://[::1]:37302 pkg=rgrpc service=reva start="01/Sep/2020:16:17:44 +0200" time_ns=4167182 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/cs3.gateway.v1beta1.GatewayAPI/Stat user-agent=grpc-go/1.26.0
2020-09-01T16:17:44+02:00 DBG skipping auth method=/cs3.gateway.v1beta1.GatewayAPI/Move pkg=rgrpc service=reva traceid=eeac0c5e1646257d1cf24f3d4aaeab55
2020-09-01T16:17:44+02:00 INF unary code=OK end="01/Sep/2020:16:17:44 +0200" from=tcp://[::1]:37260 pkg=rgrpc service=reva start="01/Sep/2020:16:17:44 +0200" time_ns=362160 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/cs3.storage.registry.v1beta1.RegistryAPI/GetStorageProvider user-agent=grpc-go/1.26.0
2020-09-01T16:17:44+02:00 INF unary code=OK end="01/Sep/2020:16:17:44 +0200" from=tcp://[::1]:37260 pkg=rgrpc service=reva start="01/Sep/2020:16:17:44 +0200" time_ns=285833 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/cs3.storage.registry.v1beta1.RegistryAPI/GetStorageProvider user-agent=grpc-go/1.26.0
2020-09-01T16:17:44+02:00 ERR home/jfd/Repositories/reva/internal/grpc/services/storageprovider/storageprovider.go:439 > error moving file error="ocisfs: Move: could not propagate old node: ocisfs: Propagate: readlink error: ocisfs: invalid node info '&{ParentID:8ce51b56-4382-4838-8c36-0e340045a7a7 ID: Name:bam Exists:false}'" pkg=rgrpc service=reva traceid=eeac0c5e1646257d1cf24f3d4aaeab55
2020-09-01T16:17:44+02:00 INF unary code=OK end="01/Sep/2020:16:17:44 +0200" from=tcp://[::1]:33198 pkg=rgrpc service=reva start="01/Sep/2020:16:17:44 +0200" time_ns=1361495 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/cs3.storage.provider.v1beta1.ProviderAPI/Move user-agent=grpc-go/1.26.0
2020-09-01T16:17:44+02:00 INF unary code=OK end="01/Sep/2020:16:17:44 +0200" from=tcp://[::1]:37302 pkg=rgrpc service=reva start="01/Sep/2020:16:17:44 +0200" time_ns=5021947 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/cs3.gateway.v1beta1.GatewayAPI/Move user-agent=grpc-go/1.26.0
2020-09-01T16:17:44+02:00 ERR error handling move request move_request=CODE_INTERNAL pkg=rhttp service=reva traceid=eeac0c5e1646257d1cf24f3d4aaeab55
2020-09-01T16:17:44+02:00 ERR http end="01/Sep/2020:16:17:44 +0200" host=::1 method=MOVE pkg=rhttp proto=HTTP/1.1 service=reva size=0 start="01/Sep/2020:16:17:44 +0200" status=500 time_ns=23075986 traceid=eeac0c5e1646257d1cf24f3d4aaeab55 uri=/remote.php/webdav/bam url=/remote.php/webdav/bam
```

pkg/storage/fs/ocis/path.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return
+}
+
+// WrapID returns the internal path for the id
+func (pw *Path) WrapID(ctx context.Context, id *provider.ResourceId) (*NodeInfo, error) {
+	if id == nil || id.OpaqueId == "" {
+		return nil, fmt.Errorf("invalid resource id %+v", id)
+	}
+	return &NodeInfo{ID: id.OpaqueId}, nil
```
PVince81, line 117: I find it a bit dangerous to pass around NodeInfo objects that are not fully populated. This will likely trigger bugs for code paths where a value is expected but it's not populated yet.

maybe add NodeInfo.getName() or something to resolve the name dynamically ?

or for cases where you really only need the id, then just return that ?

pkg/storage/fs/ocis/path.go
```diff
@@ -0,0 +1,189 @@
...
+	// The symlink contains the nodeID
+	if err != nil {
+		return
+	}
+
+	if fn != "/" {
+		// we need to walk the path
+		segments := strings.Split(strings.TrimLeft(fn, "/"), "/")
+		for i := range segments {
+			node.ParentID = node.ID
```
PVince81, line 80: maybe extract these bits into a function with a good name to make it clearer what is happening and why

pkg/storage/fs/ocis/persistence.go
```diff
@@ -0,0 +1,36 @@
...
+	CreateDir(ctx context.Context, node *NodeInfo) (err error)
+	CreateReference(ctx context.Context, path string, targetURI *url.URL) error
+	Move(ctx context.Context, oldNode *NodeInfo, newNode *NodeInfo) (err error)
+	Delete(ctx context.Context, node *NodeInfo) (err error)
+
+	Propagate(ctx context.Context, node *NodeInfo) (err error)
+}
+
+// PathWrapper is used to encapsulate path transformations
+type PathWrapper interface {
```
PVince81, line 25: streamline this nicer separation of interfaces for all storages ? (could be a separate task/ticket)

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+	// TODO inefficient because we might update several nodes twice, only propagate unchanged nodes?
+	// collect in a list, then only stat each node once
+	// also do this in a go routine ... webdav should check the etag async
+	err = fs.Propagate(ctx, oldNode)
+	if err != nil {
+		return errors.Wrap(err, "ocisfs: Move: could not propagate old node")
+	}
+	err = fs.Propagate(ctx, newNode)
+	if err != nil {
+		return errors.Wrap(err, "ocisfs: Move: could not propagate old node")
```
PVince81, line 177: s/old node/new node/

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+		return nil, errors.Wrap(err, "tree: error listing "+children)
+	}
+
+	names, err := f.Readdirnames(0)
+	nodes := []*NodeInfo{}
+	for i := range names {
+		link, err := os.Readlink(filepath.Join(children, names[i]))
+		if err != nil {
+			// TODO log
+			continue
```
PVince81, line 206: better fail right away ? a desktop client might not be happy about missing entries, those would be considered deleted

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+
+		nodes = append(nodes, n)
+	}
+	return nodes, nil
+}
+
+// Delete deletes a node in the tree
+func (fs *Tree) Delete(ctx context.Context, node *NodeInfo) (err error) {
+	err = fs.pw.FillParentAndName(node)
```
PVince81, line 222: filling the node feels overkill here as we'd get the name anyway

maybe need a GetName() and GetParent() method so the caller can decide whatever they need

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+	trashpath := filepath.Join(fs.DataDirectory, "trash/files", node.ID)
+	err = os.Rename(src, trashpath)
+	if err != nil {
+		return
+	}
+
+	// write a trash info ... slightly violating the freedesktop trash spec
+	t := time.Now()
+	// TODO store the original Path
+	info := []byte("[Trash Info]\nParentID=" + node.ParentID + "\nDeletionDate=" + t.Format(time.RFC3339))
```
PVince81, line 246: was that easier than using JSON ?

you also need to store the original file name for when restoring ?

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	return fs.Propagate(ctx, &NodeInfo{ID: node.ParentID})
+}
+
+// Propagate propagates changes to the root of the tree
+func (fs *Tree) Propagate(ctx context.Context, node *NodeInfo) (err error) {
+	// generate an etag
+	bytes := make([]byte, 16)
+	if _, err := rand.Read(bytes); err != nil {
```
PVince81, line 260: I seem to remember you saying that we generate the etag based on mtime ? or was it just an idea ?

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,282 @@
...
+			return
+		}
+		if err != nil {
+			err = errors.Wrap(err, "ocisfs: Propagate: readlink error")
+			return
+		}
+
+		node.BecomeParent()
+	}
+	return
```
PVince81, line 281: seems we don't have mtime propagation here

pkg/storage/fs/ocis/upload.go
```diff
@@ -0,0 +1,438 @@
...
+	"github.com/cs3org/reva/pkg/storage/utils/templates"
+	"github.com/cs3org/reva/pkg/user"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	tusd "github.com/tus/tusd/pkg/handler"
+)
+
+var defaultFilePerm = os.FileMode(0664)
+
+// TODO deprecated ... use tus
```
PVince81, line 43: separate ticket or checkbox ?

pkg/storage/fs/ocis/upload.go
```diff
@@ -0,0 +1,438 @@
...
+	if err != nil {
+		return 0, err
+	}
+	defer file.Close()
+
+	n, err := io.Copy(file, src)
+
+	// If the HTTP PATCH request gets interrupted in the middle (e.g. because
+	// the user wants to pause the upload), Go's net/http returns an io.ErrUnexpectedEOF.
+	// However, for OwnCloudStore it's not important whether the stream has ended
```
PVince81, line 293: what is "OwnCloudStore" ?

PVince81 (commented):
See comments, many already discussed partly

pkg/storage/fs/ocis/tree.go
butonic: parent id and name are now stored as an extended attribute in the node

pkg/storage/fs/ocis/upload.go
butonic: typo -> `OCISStore`

pkg/storage/fs/ocis/node.go
```diff
@@ -0,0 +1,58 @@
...
+// NodeInfo allows referencing a node by id and optionally a relative path
+type NodeInfo struct {
+	ParentID string
+	ID       string
+	Name     string
+	Exists   bool
+}
+
+// BecomeParent rewrites the internal state to point to the parent id
+func (n *NodeInfo) BecomeParent() {
```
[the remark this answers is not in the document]
butonic: implemented Child() and Parent() to return a new instance

pkg/storage/fs/ocis/revisions.go
```diff
@@ -0,0 +1,45 @@
...
+		return
+	}
+	if !node.Exists {
+		err = errtypes.NotFound(filepath.Join(node.ParentID, node.Name))
+		return
+	}
+
+	revisions = []*provider.FileVersion{}
+	nodePath := filepath.Join(fs.pw.Root(), "nodes", node.ID)
+	if items, err := filepath.Glob(nodePath + ".REV.*"); err == nil {
```
PVince81, line 25: what if nodePath is the string "abc.REV.xx.yy.txt" ? better move revisions into a separate directory

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,263 @@
...
+	relativeExternalPath, err = t.pw.Path(ctx, node)
+	return
+}
+
+// CreateDir creates a new directory entry in the tree
+func (t *Tree) CreateDir(ctx context.Context, node *NodeInfo) (err error) {
+
+	// TODO always try to fill node?
+	if node.Exists || node.ID != "" { // child already exists
+		return
```
PVince81, line 65: return an error ? important for webdav's "405 method not allowed" on already existing collection

pkg/storage/fs/ocis/tree.go
```diff
@@ -0,0 +1,263 @@
...
+	node.ID = uuid.New().String()
+
+	newPath := filepath.Join(t.DataDirectory, "nodes", node.ID)
+
+	err = os.MkdirAll(newPath, 0700)
+	if err != nil {
+		return errors.Wrap(err, "ocisfs: could not create node dir")
+	}
+
+	if err := xattr.Set(newPath, "user.ocis.parentid", []byte(node.ParentID)); err != nil {
```
PVince81, line 78: deja vu for setting xattrs.

move into a method `writeNodeInfoToXAttr(parentId, name, owner, ...)` ?
or pass in the NodeInfo and set based on the existing values ?

pkg/storage/fs/ocis/upload.go
```diff
@@ -0,0 +1,499 @@
...
+	nodePath := filepath.Join(fs.conf.Root, "nodes", node.ID)
+
+	tmp, err := ioutil.TempFile(nodePath, "._reva_atomic_upload")
+	if err != nil {
+		return errors.Wrap(err, "ocisfs: error creating tmp fn at "+nodePath)
+	}
+
+	_, err = io.Copy(tmp, r)
+	if err != nil {
+		return errors.Wrap(err, "ocisfs: error writing to tmp file "+tmp.Name())
```
PVince81, line 66: delete temp file on error or does this happen automatically ? how often ?

pkg/storage/fs/ocis/upload.go
```diff
@@ -0,0 +1,499 @@
...
+		n.ID = uuid.New().String()
+	}
+	targetPath := filepath.Join(upload.fs.conf.Root, "nodes", n.ID)
+
+	// if target exists create new version
+	if fi, err := os.Stat(targetPath); err == nil {
+		// versions are stored alongside the actual file, so a rename can be efficient and does not cross storage / partition boundaries
+		versionsPath := filepath.Join(upload.fs.conf.Root, "nodes", n.ID+".REV."+fi.ModTime().UTC().Format(time.RFC3339Nano))
+
+		if err := os.Rename(targetPath, versionsPath); err != nil {
```
PVince81, line 346: a copy might be better in case the below rename fails, in which case the file would disappear completely from view

pkg/storage/fs/ocis/upload.go
```diff
@@ -0,0 +1,499 @@
...
+	if err := os.Rename(upload.binPath, targetPath); err != nil {
+		log := appctx.GetLogger(upload.ctx)
+		log.Err(err).Interface("info", upload.info).
+			Str("binPath", upload.binPath).
+			Str("targetPath", targetPath).
+			Msg("ocisfs: could not rename")
+		return err
+	}
+
+	if err := xattr.Set(targetPath, "user.ocis.parentid", []byte(n.ParentID)); err != nil {
```
PVince81, line 368: wondering if we're going too low-level here and whether the upload code should talk to a higher level API, one that simply configures the node correctly, with xattrs and children.

PVince81 (commented):
See comments for latest additions

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+		case "l":
+			e.Label = kv[1]
+			// TODO default ... log unknown keys? or add as opaque? hm we need that for tagged shares ...
+		}
+	}
+	return e, nil
+}
+
+// Parse parses an acl string with the given delimiter (LongTextForm or ShortTextForm)
+func getACEs(ctx context.Context, fsfn string, attrs []string) (entries []*ace, err error) {
```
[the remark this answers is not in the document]
butonic: hm, the comment is totally off, I'll rewrite it in the owncloud driver as well

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+			} else {
+				if principal[:1] != "u" {
+					log.Error().Str("attr", attrs[i]).Interface("ace", e).Msg("inconsistent ace: expected user")
+					continue
+				}
+			}
+			entries = append(entries, e)
+		}
+	}
+	return entries, nil
```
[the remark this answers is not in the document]
butonic: true, we are logging the error ... but trying to continue to at least return normal metadata ... I don't think we should fail hard here, because the user cannot do anything about it. If a share does not show up he might retry creating the share ... hm

pkg/storage/fs/ocis/node.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+)
+
+// NodeInfo allows referencing a node by id and optionally a relative path
+type NodeInfo struct {
```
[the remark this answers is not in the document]
butonic: renamed to Node

pkg/storage/fs/ocis/grants.go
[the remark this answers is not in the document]
butonic: removing the error

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+}
+
+func (fs *ocisfs) getGranteeType(e *ace) provider.GranteeType {
+	if strings.Contains(e.Flags, "g") {
+		return provider.GranteeType_GRANTEE_TYPE_GROUP
+	}
+	return provider.GranteeType_GRANTEE_TYPE_USER
+}
+
+func (fs *ocisfs) getGrantPermissionSet(mode string) *provider.ResourcePermissions {
```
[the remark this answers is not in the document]
butonic: using `grantPermissionSet`

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+	if strings.Contains(mode, "v") {
+		p.ListFileVersions = true
+	}
+	if strings.Contains(mode, "V") {
+		p.RestoreFileVersion = true
+	}
+
+	// ?
+	// TODO GetPath
+	if strings.Contains(mode, "q") {
```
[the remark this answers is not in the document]
butonic: the permissions are a longer list and we could start introducing an order. Not of concern right now:
```

	//   - r read-data (files) / list-directory (directories)
	//   - w write-data (files) / create-file (directories)
	//   - a append-data (files) / create-subdirectory (directories)
	//   - x execute (files) / change-directory (directories)
	//   - d delete - delete the file/directory. Some servers will allow a delete to occur if either this permission is set in the file/directory or if the delete-child permission is set in its parent directory.
	//   - D delete-child - remove a file or subdirectory from within the given directory (directories only)
	//   - t read-attributes - read the attributes of the file/directory.
	//   - T write-attributes - write the attributes of the file/directory.
	//   - n read-named-attributes - read the named attributes of the file/directory.
	//   - N write-named-attributes - write the named attributes of the file/directory.
	//   - c read-ACL - read the file/directory NFSv4 ACL.
	//   - C write-ACL - write the file/directory NFSv4 ACL.
	//   - o write-owner - change ownership of the file/directory.
	//   - y synchronize - allow clients to use synchronous I/O with the server.
```

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	np := filepath.Join(fs.pw.Root(), "nodes", node.ID)
+	if err := xattr.Set(np, attr, getValue(e)); err != nil {
+		return err
+	}
+	return fs.tp.Propagate(ctx, node)
+}
+
+func getValue(e *ace) []byte {
```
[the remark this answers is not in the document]
butonic: `Marshal`

pkg/storage/fs/ocis/revisions.go
butonic: the entries inside the node folder all start with a uuid, `.REV.` indicates it is a revision. and what follows is a timestamp, so multiple versions can be kept in the same location as the current actual file content. This prevents new fileuploads to trigger cross storage moves when revisions accidentially are stored on another partition, because they admin mounted a different partition there. We can make a background process move old revisions to a slower storage and replace the revision five with a symbolic link in the future if necessary.

pkg/storage/fs/ocis/tree.go
butonic: ocs is making a stat call to return the proper codes. but we can return  `errtypes.AlreadyExists` here

pkg/storage/fs/ocis/grants.go
```diff
@@ -0,0 +1,380 @@
...
+
+	var attr string
+	if g.Grantee.Type == provider.GranteeType_GRANTEE_TYPE_GROUP {
+		attr = sharePrefix + "g:" + e.Principal
+	} else {
+		attr = sharePrefix + "u:" + e.Principal
+	}
+
+	np := filepath.Join(fs.pw.Root(), "nodes", node.ID)
+	if err := xattr.Set(np, attr, getValue(e)); err != nil {
```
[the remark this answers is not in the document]
butonic: no, because a grant for the same grantee (principal / user / group) overrides the permissions of a previously existing grant.

In theory we could allow multiple grants for the same user, eg. with different expiry dates ... in theory ...

pkg/storage/fs/ocis/node.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	nodePath := filepath.Join(n.pw.Root(), "nodes", n.ParentID)
+	// lookup parent id in extended attributes
+	var attrBytes []byte
+	// lookup name in extended attributes
+	if attrBytes, err = xattr.Get(nodePath, "user.ocis.owner.id"); err == nil {
+		n.ownerID = string(attrBytes)
+	} else {
+		// TODO log error
```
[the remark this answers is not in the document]
butonic: I now fail hard when the owner is missing.

pkg/storage/fs/ocis/tree.go
butonic: using `writeMetadata` but needs more refactoring

pkg/storage/fs/ocis/ocis.go
```diff
@@ -0,0 +1,421 @@
...
+func getUser(ctx context.Context) (*userpb.User, error) {
+	u, ok := user.ContextGetUser(ctx)
+	if !ok {
+		err := errors.Wrap(errtypes.UserRequired(""), "ocisfs: error getting user from ctx")
+		return nil, err
+	}
+	return u, nil
+}
+
+func (fs *ocisfs) normalize(ctx context.Context, node *NodeInfo) (ri *provider.ResourceInfo, err error) {
```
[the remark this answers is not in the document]
butonic: moved to node

pkg/storage/fs/ocis/path.go
butonic: creating new instances now

pkg/storage/fs/ocis/persistence.go
butonic: future work, api still needs some refactoring

pkg/storage/fs/ocis/tree.go
butonic: using Child(name string) and Parent()

pkg/storage/fs/ocis/tree.go
butonic: an idea ... for migrations / backup we need to be able to store an etag. the owncloud driver already has some logic for that.

pkg/storage/fs/ocis/upload.go
butonic: hm ... the put is deprecated ... but yeah we should delete the file in this case

PVince81:
let's get this merged, since the main file operations work

follow up for the remaining tasks:
- sharing: https://github.com/owncloud/product/issues/145
- trashbin implementation: https://github.com/owncloud/product/issues/209
- versions implementation: https://github.com/owncloud/product/issues/210
- extended attributes / arbitrary properties: https://github.com/owncloud/product/issues/211
- bring ocis storage to CI: https://github.com/owncloud/product/issues/212

PVince81 (approved):
Looks good, as discussed :+1:

PVince81:
weird results:
```
runsh: Total unexpected failed scenarios throughout the test run:
apiTrashbin/trashbinFilesFolders.feature:185
apiTrashbin/trashbinFilesFolders.feature:186
apiTrashbin/trashbinFilesFolders.feature:203
apiTrashbin/trashbinFilesFolders.feature:204
apiTrashbin/trashbinFilesFolders.feature:227
apiTrashbin/trashbinFilesFolders.feature:228
apiTrashbin/trashbinFilesFolders.feature:241
apiTrashbin/trashbinFilesFolders.feature:242
apiTrashbin/trashbinFilesFolders.feature:255
apiTrashbin/trashbinFilesFolders.feature:256
apiTrashbin/trashbinFilesFolders.feature:217
apiTrashbin/trashbinFilesFolders.feature:218
apiTrashbin/trashbinFilesFolders.feature:235
apiTrashbin/trashbinFilesFolders.feature:236
apiTrashbin/trashbinFilesFolders.feature:259
apiTrashbin/trashbinFilesFolders.feature:260
apiTrashbin/trashbinFilesFolders.feature:273
apiTrashbin/trashbinFilesFolders.feature:274
apiTrashbin/trashbinFilesFolders.feature:287
apiTrashbin/trashbinFilesFolders.feature:288
```

so this broke something related to trashbin on OC storage but also fixed other things ?!

PVince81:
@butonic can you add instructions in the top POST about how to test it ? it is not clear to me what env variables to switch to get the proper working setup

PVince81:
we might want for https://github.com/cs3org/reva/pull/1148 to be merged first

phil-davis:
> so this broke something related to trashbin on OC storage but also fixed other things ?!

I guess that it needs a rebase. The core commit id to use for tests is embedded in `.drone.star` - so it will be using whatever is the core commit id in `.drone.star` in this branch, but the expected-failures list will be as it is in master.

PVince81:
the failing tests are related to accessing another user's trashbin.
instead of getting a 401 we receive an empty XML response.
this is reproducible locally

```
  Scenario: Listing other user's trashbin is prohibited                                    # /srv/www/htdocs/owncloud/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature:173
    Given using old DAV path                                                               # FeatureContext::usingOldOrNewDavPath()
    And user "testtrashbin100" has been created with default attributes and skeleton files # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Brian" has been created with default attributes and without skeleton files   # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "testtrashbin100" has deleted file "/textfile1.txt"                           # FeatureContext::userHasDeletedFile()
    When user "Brian" tries to list the trashbin content for user "testtrashbin100"        # TrashbinContext::userTriesToListTheTrashbinContentForUser()
      Received empty response where XML was expected (Exception)
    Then the HTTP status code should be "401"                                              # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And the last webdav response should not contain the following elements                 # TrashbinContext::theLastWebdavResponseShouldNotContainFollowingElements()
      | path          | user            |
      | textfile1.txt | testtrashbin100 |
```

there is a slight chance that the permission refactoring that happened affects this

PVince81:
more bad luck striking:
```
0. init.................. FAIL (connection refused by `revad-services' port 20080: Connection refused)
```

PVince81:
okay, so now all tests passed... so the rebase worked...

phil-davis:
It would be nice to have #1148 merged - that will get the core API test suite having some more accept share scenarios, which are running in the ocis and ocis-reva repos today. I am not sure who is able to review-approve-merge that. Then a rebase here and it should be good.

butonic:
we should close this in faver of https://github.com/cs3org/reva/pull/1155

Issue #552, opened 2020-03-12
GetHome mixes internal and external namespace


butonic:
Currently, GetHome as implemented in the local and eos storage providers forces the storage provider to know where it is mounted. The Namespace config option used is part of the global namespace and actually exposes the internal path used by the storage driver (because the implementations return the result of the internal wrap functions).

The storage providers should return the relative path to the users home. For the owncloud driver that is `{{.Username}}`, for eos it might be `users/{{.Username}}`.

The gateway should prefix GetHome responses from storage providers with the mount point. This would create the correct global namespace without exposing service internal paths.

labkode:
Can you elaborate on that? is not clear to me what you are describing.

The storage provider exposes a mount point (/home) and the storage implementation is can mounted wherever it wants (in EOS, /eos/user/j/jon). 

The storage implementation, EOS in this case is not returning /home.

butonic:
In `wrap()` the local as well as the eos storage driver prefix the external path with the content of a variable called `Namespace`. For the local driver it defaults to `/var/tmp/reva`, for eos it defaults to `/`. The cs3 API lets clients make a `GetHome()` call. The gateway will check the storage registry and forward the request to tho storage provider responsible for homes. When the namespace has not changed the response for the eos driver will be the internal path because namespace is `/` or likely begin with `/eos` in production. Everything looks ok ... but for the local driver the response will have a path that starts with `/var/tmp/reva` because the storage drivers GetHome will prefix the external path with the namespace variable.

The current implementation requires the storage drivers to know where they are mounted in the namespace, so they return the correct path. Prefixing the storage relative external path with the mount point to make it an absolute external path is the responsibility of the gateway.

I'll  submit a PR.

labkode:
@butonic thanks, I got now the point! I'll merge that PR

PR #562, opened 2020-03-16
Let the gateway prefix relative home paths with the mount point

butonic:
fixes https://github.com/cs3org/reva/issues/552


labkode:
@butonic do we need to change anything on CreateHome?

PR #569, opened 2020-03-18
reva: fix cli commands

labkode:


cmd/reva/ls.go
```diff
@@ -69,7 +69,7 @@ func lsCommand() *command {
 			}
 
 			if *longFlag {
-				fmt.Printf("%+v %d %d %v %s\n", info.PermissionSet, info.Mtime, info.Size, info.Id, p)
+				fmt.Printf("%s %d %d %v %s\n", info.Type, info.Mtime, info.Size, info.Id, p)
```
ishank011: ResourceInfo does not have a Type member, does it? Or is it something implicit? https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.ResourceInfo

labkode: Yes it does, is indeed in the link you are pointing :), is lower case there but in the Go package is uppercase.
Is the second field

ishank011: Okay. Missed that somehow! Looks good then.

PR #570, opened 2020-03-18
fix json usershare manager

labkode:


pkg/share/manager/json/json.go
```diff
@@ -78,7 +78,15 @@ func New(m map[string]interface{}) (share.Manager, error) {
 	return mgr, nil
 }
 
-func load(file string) (*shareModel, error) {
+func loadOrCreate(file string) (*shareModel, error) {
+	_, err := os.Stat(file)
+	if os.IsNotExist(err) {
+		if err := ioutil.WriteFile(file, []byte("{}"), 0700); err != nil {
```
refs: does this file need execute permissions? or will `0644` be enough?

PR #571, opened 2020-03-18
Teach the owncloud storage driver home handling

butonic:
fixes path mappings for phoenix and oc10 acceptance tests and changes the `layout` config option to `user_layout` to align it with eos and local

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -152,9 +152,10 @@ func init() {
 
 type config struct {
 	DataDirectory string `mapstructure:"datadirectory"`
-	Scan          bool   `mapstructure:"scan"`
+	UserLayout    string `mapstructure:"user_layout"`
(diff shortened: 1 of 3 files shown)


pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -329,27 +340,37 @@ func (fs *ocfs) getRecyclePath(ctx context.Context) (string, error) {
...
-	case 3:
-		return path.Join("/", parts[1])
-	default:
-		return path.Join("/", parts[1], parts[3])
+		parts := strings.SplitN(np, "/", 4)
+		// parts = "", "<username>", "files", "foo/bar.txt"
+		switch len(parts) {
+		case 1:
+			external = "/"
+		case 2:
```
refs: you can use a `fallthrough` here, it makes it more readable to my eyes

refs: cc @butonic

PR #577, opened 2020-03-19
Leave home prefixing to the gateway

butonic:
Currently, the Path is prefixed twice. Once by the storage provider and once by the gateway.

Leave it to the gateway.

Followup of https://github.com/cs3org/reva/pull/562

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -6,6 +6,7 @@
 - Hannah von Reth <vonreth@kde.org>
 - Hugo Gonzalez Labrador <github@hugo.labkode.com>
 - Ilja Neumann <ineumann@owncloud.com>
+- Ishank Arora <ishank011@gmail.com>
 - Jörn Friedrich Dreyer <jfd@butonic.de>
 - LovisaLugnegard <lovisa.lugnegard@gmail.com>
 - Michael D'Silva <md@aarnet.edu.au>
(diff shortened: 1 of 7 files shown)


labkode:
@butonic I am bit lost with these changes.

The storage provider will prefix with the mount point. The storage driver will provide a home path relative to the storage driver.  Why the gateway would prefix the response? Shouldn't this be the storage provider already?

SP = Storage Provider
SD = Storage Driver

SD exposes "/users/jorn" for home of "jorn"
SP exposes "/home", so wraps inner SD, "/home/users/jorn".

butonic:
IMO
- the storage driver returns a relative path, hiding any internal storage layout
- the storage provider just returns this relative path because it should not know anything about where it is mounted
- the storage registry knows where storae providers are mounted in the global namespace
- the gateway should prefix the home path with the mount point of the storage to combine the knowledge from the storage registry and the storage provider

If the storage provider knows about the outside world it becomes harder to move storages in the global storage, because you have to configure it in two places: the registry and the storage provider. 

The gateway is responsible for routing requests to the correct storage and the storage registry stores the mount path &  storage id mapping for the storage providers.

AFAICT the storage provider should only know its ID.

labkode:
@butonic that is not how the system was designed. Every storage provider provides a unique mount path to create a distributed unique namespace, without depending on the gateway to change the mount paths.

This PR allows to:
```
red:reva gonzalhu$ ./cmd/reva/reva ls /oc/einstein/
Documents
red:reva gonzalhu$ ./cmd/reva/reva ls /home/
Documents
```

PR #593, opened 2020-03-23
Move OCS error handlers to pkg

ishank011:


pkg/errhandler/errhandler.go
```diff
@@ -0,0 +1,144 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package errhandler
```
refs: Nice abstraction! however, I think the chose of name is incorrect, and the base principles. While an `errhandler` package would allow for abstracting error handling, this package deals more with `OCS` messages.

On my mind working with interfaces is more appropriate as a generalized idea of error handling, then, every package that deals with the ocs api would implement the `errhandler` interface. Otherwise if you want to keep this generic, more functions will be added for different packages.

ishank011: Hi @refs, thanks for the review! The main idea behind moving this to pkg was to prevent implementing it again. We're currently implementing the OCM HTTP service and it would have the same error codes as these, and I imagine all HTTP services would be having these, and the fact that we could pass messages to add details makes these quite generic. But we could definitely rename the package name to signify that it handles HTTP error codes.

Please let me know if that would be fine. And any suggestions that you have for the package name :smile:

labkode: @ishank011 are really the OCS error responses the same for OCM? The ones that I've found in OCM only have a "MESSAGE" field:

```
{
  "message": "RESOURCE_NOT_FOUND"
}
```

ishank011: @labkode OCM defines it's own error codes and a few associated methods in https://github.com/cernbox/ocmd/blob/master/api/api.go#L14 which match with the OCS erros (https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/reqres.go#L54). We can add more if required. So instead of having separate implementations in both the services, this way we can just use a common package.

So what was previously done as https://github.com/cernbox/ocmd/blob/master/handlers/handlers.go#L308
```
ae := api.NewAPIError(api.APIErrorInvalidParameter).WithMessage("body is not json")
w.Write(ae.JSON())
```
can now be done in a similar way as OCS does.
```
errhandler.WriteError(w, r, MetaBadRequest.StatusCode, "body is not json"", nil)
```

butonic:
The OCS errors need a better implementation, because there is a difference between the OCS v1 Vs V2 http status code handling. We need to respond with different codes, depending on the endpoint. Tomorrow, I'll link to the to-do, which also references the ownCloud core implementation.

butonic:
The TODO is right above the mapping code: https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/reqres.go#L117-L138

It is derived from https://github.com/owncloud/core/commit/bacf1603ffd53b7a5f73854d1d0ceb4ae545ce9f#diff-c1d6f3585e52926be0ec6169c0f58c38R442-R466

However, it is only necessary for the v2 endpoint.

This ties the mapping to the endpoint.

We can pass a boolean isV2 flag (shudder) or a mapper function as a new parameter (meh) which clutters up the handler signatures. In this case the context might be the right place. Then the ocs handler can store the error mapper function in the context and the error handler can check if the context has a mapper function. If you feel this does not make functionality obvious enough we could also store the mapper function or a preconfigure error handler in the handler when setting up the v1 and v2 handlers.

I am leaning towards adding a mapper function to the handlers. But maybe the OCM implementation has requirements that clarify how we should implement it.
opinions?

refs:
> I am leaning towards adding a mapper function to the handlers. But maybe the OCM implementation has requirements that clarify how we should implement it.
> opinions?

My two cents: Agree with the implementation details, but my point on an untied, abstract, cold interface definition that's not tied to the OCS API still stands. Do the OCM messages care about [OCS meta statuses](https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/reqres.go#L64)? Because this implementation [requires just that](https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/reqres.go#L38). [Or at least is not reflecting a more general approach](https://github.com/cs3org/reva/pull/593/files#diff-007cbb4d2bc409d46554aa45082c3a68R37)

labkode:
@ishank011 for time being we leave the OCS on one side and the OCM on other. These two services will evolve independently. OCS and OCDAV at some point will be moved to the OCIS repo while OCM will continue to live in Reva.

ishank011:
@labkode Okay, will close this and add a similar implementation for OCM. Thanks all for the reviews!

PR #604, opened 2020-03-27
Public Access to Public Shares

refs:
Public access to public shares works, as in: lists resources.

public-shares

## Running it

Be aware that I run all this from `ocis-reva`, so one must adapt this configuration into a toml file.

### Configuration

1. add a new ocis-reva command: `storagepubliclink.go` with this config:

```golang
map[string]interface{}{
					"core": map[string]interface{}{
						"max_cpus":        cfg.Reva.StoragePublicLink.MaxCPUs,
						"tracing_enabled": true,
					},
					"shared": map[string]interface{}{
						"jwt_secret": cfg.Reva.JWTSecret,
					},
					"grpc": map[string]interface{}{
						"network": cfg.Reva.StoragePublicLink.Network,
						"address": cfg.Reva.StoragePublicLink.Addr,
						"interceptors": map[string]interface{}{
							"log": map[string]interface{}{},
						},
						"services": map[string]interface{}{
							"publicstorageprovider": map[string]interface{}{
								"mount_path":                 "/public/", // localhost:[ocdavport]/remote.php/dav/public-files/{token}/folderA/folderB/files.txt
								"mount_id":                   "e1a73ede-549b-4226-abdf-40e69ca8230d",
								"public_share_provider_addr": cfg.Reva.StoragePublicLink.PublicShareProviderAddr,
								"storage_provider_addr":      cfg.Reva.StoragePublicLink.StorageProviderAddr,
							},
							"authprovider": map[string]interface{}{
								"auth_manager": "publicshares",
								"auth_managers": map[string]interface{}{
									"publicshares": map[string]interface{}{
										"gateway_addr": "localhost:9142",
									},
								},
							},
						},
					},
				}
```
[_full file here_](https://gist.github.com/refs/ca31f16c92da0c8100aca571e7e6587e)


refs:
Totally agree, and I find option 2 very appealing indeed.

/cc @butonic

refs:
Alright, I'm getting on with it :)

madsi1m:
Patiently waiting for this merge

refs:
^^ soon

refs:
@labkode I did implement a public storage provider, however to do a status call on a resource we do need to access the storage where the file resides. For the `owncloud` storage, we do need authentication since the `auth` interceptor is enabled.

I was thinking on disabling the `auth` interceptor for the public storage, is that even possible with the current setup? So far I went with explicit declaring endpoints as insecure, since disabling the `auth` interceptor [appears not to be possible](https://github.com/cs3org/reva/blob/0dbce24f7e8b508c1ecab2d145979562888a7a76/pkg/rgrpc/rgrpc.go#L273).

If there is some conceptualizing work it needs to be done when it comes to openly exposed endpoints which I'm more than happy to help with.

So far I have explored the following solutions, both leading me and @butonic to forge a token:

1. Frontend (http) handling the token forging (as seen in this PR on the current state).
2. Storage handling the token forge (new `publicstorageprovider` forges a token on every single request).

Is extending `/rgrpc` to let service an option? If so, that would mean the `storageprovider` would need to bypass the auth, as per solution 1) and 2). This solution leads to a catch22:

a) We want a public share storage (storage#1, with auth) to deal with public shares.
b) File info are not on that storage provider, we need to ask storage#1, for which we need a token.

An ugly solution I envision would be to update the public storage(s) with the actual resource, but that would mean keeping it up to date? I have a couple things more in mind, but they're all "around" reva, not really inline with the current setup.

TL;DR: I need some pointers to get around the authentication interceptor :)

/cc @madsi1m

labkode:
@refs I think the option is to forge the token for the the public link, adding a new method to the authentication provider, we use that in Cernbox:
https://github.com/cernbox/revaold/blob/master/revad/svcs/authsvc/authsvc.go#L51

So for that to happen we need:

- [ ]  access publicly the information of a token by knowing the token (password is hidden) from the public share provider GRPC service
- [ ] implement a  new auth provider that forge the token when given a token + optional password if the public link is protected

So the http public link endpoint will ask the auth provider to forge the token given the token as parameter. This token then is valid to impersonate the user, we'll add later constrained delegation to only access the required resource.

refs:
💡that shed even more light into this, thanks! I'll get hands to work on the new auth provider. Crafting a quick MVP

labkode:
@refs thanks, I'll give you a hand asap on that area also :)

refs:
@labkode updated with the latest suggestions. Review? :)

labkode:
@refs taking a look tomorrow

pkg/auth/manager/publicshares/publicshares.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &manager{
+		c: conf,
+	}, nil
+}
+
+func (m *manager) Authenticate(ctx context.Context, token string, secret string) (*user.User, error) {
```
labkode, line 68: We passing a path as token, the split can be done better on the OCS layer, cleaner.

internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -282,25 +281,20 @@ func PublicShare2ShareData(share *link.PublicShare, r *http.Request) *ShareData
...
-		UIDOwner:     UserIDToString(share.Creator),
 		STime:        share.Ctime.Seconds, // TODO CS3 api birth time = btime
-		UIDFileOwner: UserIDToString(share.Owner),
 		Token:        share.Token,
 		Expiration:   expiration,
 		MimeType:     share.Mtime.String(),
 		Name:         r.FormValue("name"),
-		// URL:                  r.Host + "/#/s/" + share.Token, // this is broken. r.Host doesn't point to Phoenix
-		URL: "localhost:8300/#/s/" + share.Token,
+		URL:          r.Header.Get("Origin") + "/#/s/" + share.Token,
```
labkode, line 294: Shouldn't Phoenix use this path as relative? Having the hostname here means OCIS server needs to know where Phoenix is running

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -0,0 +1,315 @@
...
+	MountID                 string `mapstructure:"mount_id"`
+	PublicShareProviderAddr string `mapstructure:"public_share_provider_addr"`
+	StorageProviderAddr     string `mapstructure:"storage_provider_addr"`
+}
+
+type service struct {
+	conf                 *config
+	mountPath, mountID   string
+	publicShareProviderC link.LinkAPIClient
+	storageProviderC     provider.ProviderAPIClient
```
labkode, line 36: Use the gateway instead, services should not connect directly to storage provider.

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -0,0 +1,315 @@
...
+	MountPath               string `mapstructure:"mount_path"`
+	MountID                 string `mapstructure:"mount_id"`
+	PublicShareProviderAddr string `mapstructure:"public_share_provider_addr"`
+	StorageProviderAddr     string `mapstructure:"storage_provider_addr"`
+}
+
+type service struct {
+	conf                 *config
+	mountPath, mountID   string
+	publicShareProviderC link.LinkAPIClient
```
labkode, line 35: Use the gateway client instead, services should not connect directly to storage provider.

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -0,0 +1,315 @@
...
+)
+
+func init() {
+	rgrpc.Register("publicstorageprovider", New)
+}
+
+type config struct {
+	MountPath               string `mapstructure:"mount_path"`
+	MountID                 string `mapstructure:"mount_id"`
+	PublicShareProviderAddr string `mapstructure:"public_share_provider_addr"`
```
labkode, line 28: GatewayAddress is the only thing needed.

pkg/auth/manager/publicshares/publicshares.go
refs: right, I pushed this too fast, was pushing a path locally 🙈

internal/http/services/owncloud/ocs/conversions/main.go
refs: indeed it should, but as far as my knowledge goes it doesn't. Correct me if I'm wrong @butonic

refs:
@labkode honestly I didn't think using the gw client would work when doing a stat call to the underlying storage. Thanks for point it out, it streamlines the code quite a lot.

internal/http/services/owncloud/ocs/conversions/main.go
refs: cc'ing @kulmann perhaps you can also shed some light as per why phoenix is not dealing with relative paths there. I'd give it another try in my local environment but I'm pretty sure it will `burp`

butonic: someone needs to construct the url. we can do it in the clients or on the server side. We could return a relative path as the URL , but be aware that it might require special handling if the first path segment contains a `:`, see https://tools.ietf.org/html/rfc3986#section-4.2

IMO we should teach the clients to handle relative paths correctly. Including desktop and mobile.

labkode:
@refs how is it going?

refs:
@labkode this first approach can be merged imo. Still work need to be done for the remaining storage operations, but this allows for creating and accessing public links. The rest will be followed up by more pull requests.

internal/http/services/owncloud/ocs/conversions/main.go
labkode: Merging from now, but I'm creating an issue to keep track of it

PR #627, opened 2020-04-01
escape special xml characters in propfind response

C0rby:
When the propfind response contained a character that is a special xml character then if a client tried to parse the xml it would fail. Therefore we needed to escape the characters.

Fixes https://github.com/owncloud/ocis-reva/issues/122


butonic:
Unfortunately, this broke propfinds: `<d:resourcetype>&lt;d:collection/&gt;</d:resourcetype>` so all  directories are now broken ...

PR #629, opened 2020-04-01
EOS fix shadow root issue

madsi1m:


pkg/storage/fs/eos/eos.go
```diff
@@ -846,64 +846,16 @@ func (fs *eosfs) createShadowHome(ctx context.Context) error {
 		return errors.Wrap(err, "eos: error verifying if user home directory exists")
 	}
 
-	// TODO(labkode): only trigger creation on not found, copy from CERNBox logic.
-	err = fs.c.CreateDir(ctx, "root", home)
```
butonic, line 850: AFAIU using `root` as the owner is by design. We don't want users that use eos directly to mess with the shadow tree. It is managed exclusively by reva.

reva will manage the tree using root / elevated permissions

@labkode correct me if I am wrong

labkode: @butonic you're right, this directory is owned by root, not by the user.
Accessing the shadow tree is possible, your EOS instances are not correctly configured, that is why you cannot access the shadow tree. The machine/box where Reva runs needs to be configured:
- as a trusted unix gateway with sudo permissions
- using keytab with sudo permissions

labkode:
@madsi1m @butonic 

To properly configure EOS you need a sudo user:

```
$ eos vid ls | grep sudo
sudoer                 => uids(daemon,cbox,eosarchi,cboxwopi)
```
```
id cbox
uid=58679(cbox) gid=1028(c3) groups=1028(c3)
```

```
$ eos vid ls  | grep reva-01
tident:"*@reva-01":gid => c3
tident:"*@reva-01-01":uid => cbox
tident:"*@reva-01.cern.ch":gid => c3
tident:"*@reva-01.cern.ch":uid => cbox
tident:"unix@reva-01":gid => root
tident:"unix@reva-01":uid => root
tident:"unix@reva-01.cern.ch":gid => root
tident:"unix@reva-01.cern.ch":uid => root
```


To add a user as sudo:

```
eos vid set membership cbox +sudo
```

To add a trusted gateway:

```
eos vid add gateway reva-01 unix
$ eos vid add gateway reva-01.cern.ch unix
$ eos vid set map -tident "*@reva-01" vuid:58679 vgid:1028
$ eos vid set map -tident "*@reva-01.cern.ch" vuid:58679 vgid:1028
```

butonic:
my current vid mapping looks like this:
```
docker exec -it eos-mgm1 eos vid ls
publicaccesslevel: => 1024
sss:"<pwd>":gid => root
sss:"<pwd>":uid => root
sudoer                 => uids(admin1)
tident:"*@eos-cli1":gid => root
tident:"*@eos-cli1":uid => root
unix:"<pwd>":gid => root
unix:"<pwd>":uid => root
```

AFAICT this means I set up
* `eos-cli1` as a trusted gateway (this is where my storageprovider is running)
* unix auth to use the uid from ldap

running
```console
$ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-cli1 eos whoami
Virtual Identity: uid=1001 (99,1001) gid=1001 (99,1001) [authz:krb5] sudo* host=eos-cli1.eoscluster.cern.ch domain=eoscluster.cern.ch
```

I am confused why it is using krb auth ... might be because the client has some config I don't get yet. Anyway, I can create files and folders and create shares in eos-cli1 just fine ...

madsi1m:
this vid wrong then?

```
https:"<pwd>":gid => root
https:"<pwd>":uid => root
membership uid: daemon       => gids(adm)
membership uid: daemon       => uids(adm)
publicaccesslevel: => 1024
sss:"<pwd>":gid => root
sss:"<pwd>":uid => root
sudoer                 => uids(daemon)
tident:"*@reva-storage-global":gid => root
tident:"*@reva-storage-global":uid => root
tident:"*@reva-storage-global.dev1-md.svc.cluster.local":gid => root
tident:"*@reva-storage-global.dev1-md.svc.cluster.local":uid => root
tident:"*@reva-storage-home":gid => root
tident:"*@reva-storage-home":uid => root
tident:"*@reva-storage-home.dev1-md.svc.cluster.local":gid => daemon
tident:"*@reva-storage-home.dev1-md.svc.cluster.local":uid => daemon
unix:"<pwd>":gid => nobody
unix:"<pwd>":uid => nobody
```

butonic:
@madsi1m @labkode we should close this pr and continue in a new issue with the vid config

madsi1m:
This works now, please merge

butonic:
I agree:
```
➜  ocis git:(ocis-on-eos) ✗ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-mgm1 eos ls -l /eos/dockertest/reva/users/.shadow/
drwxrwsrw+   1 root     root                0 Apr  2 18:31 e
drwxrwsrw+   1 root     root                0 Apr  2 18:33 m
➜  ocis git:(ocis-on-eos) ✗ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-mgm1 eos ls -l /eos/dockertest/reva/users/.shadow/m
drwxrws--+   1 20001    30000               0 Apr  2 18:33 marie
➜  ocis git:(ocis-on-eos) ✗ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-mgm1 eos ls -l /eos/dockertest/reva/users/.shadow/m/marie
drwxrws--+   1 20001    30000               0 Apr  2 18:34 Shares
➜  ocis git:(ocis-on-eos) ✗ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-mgm1 eos ls -l /eos/dockertest/reva/users/.shadow/m/marie/Shares
drwxrws--+   1 root     root                0 Apr  2 18:34 frome
➜  ocis git:(ocis-on-eos) ✗ docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-mgm1 eos ls -l /eos/dockertest/reva/users/.shadow/m/marie/Shares/frome
```

without this the share folder (by default `MyShares` or in the above example `Shares`) is owned by root, leading to this error:
```
docker exec --env EOS_MGM_URL="root://eos-mgm1.eoscluster.cern.ch:1094" -it eos-cli1 eos -r 20001 30000 find --fileinfo --maxdepth 1 /eos/dockertest/reva/users/.shadow/m/marie/Shares
keylength.file=50 file=/eos/dockertest/reva/users/.shadow/m/marie/Shares/ treesize=0 container=1 files=0 mtime=1585824004.140137234 ctime=1585823989.957111421 btime=1585823989.826628305 stime=1585824004.140137234 clock=0 mode=42770 uid=0 gid=0 fxid=00000019 fid=25 ino=25 pid=24 pxid=00000018 etag=19:1585824004.140 detached=0 xattrn=sys.allow.oc.sync xattrv=1 xattrn=sys.eos.btime xattrv=1585823989.826628305 xattrn=sys.forced.atomic xattrv=1 xattrn=sys.forced.checksum xattrv=adler xattrn=sys.mask xattrv=700 xattrn=sys.mtime.propagation xattrv=1
error: no permissions to read directory /eos/dockertest/reva/users/.shadow/m/marie/Shares/
 (errc=7) (Argument list too long)
```

reva tries to impersonate the user which does not have enough permissions to access the shared folder.

@labkode another fix would be to use root to do the stat ... hmmm

butonic:
@labkode in line 650 of eos.go you are using the user to get the metadata of entries in the shadow tree:
```golang
643	func (fs *eosfs) getMDShareFolder(ctx context.Context, p string) (*provider.ResourceInfo, error) {
644		u, err := getUser(ctx)
645		if err != nil {
646			return nil, err
647		}
648	
649		fn := fs.wrapShadow(ctx, p)
650		eosFileInfo, err := fs.c.GetFileInfoByPath(ctx, u.Username, fn)
651		if err != nil {
652			return nil, err
653		}
654		// TODO(labkode): diff between root (dir) and children (ref)
655	
656		if fs.isShareFolderRoot(ctx, p) {
657			return fs.convertToResourceInfo(ctx, eosFileInfo), nil
658		}
659		return fs.convertToFileReference(ctx, eosFileInfo), nil
660	}
```

in line 650 we could use "root" instead of the username:
```golang
650		eosFileInfo, err := fs.c.GetFileInfoByPath(ctx, "root", fn)
```

but I don't think that is what you intended to do.

for now I'm good with this PR. Looking into the propfind result.

butonic:
ok phoenix showing files is caused by the reference type leaking through the gateway. Tracked in https://github.com/cs3org/reva/issues/635

PR #634, opened 2020-04-02
[WIP] OCM Invitation Workflow

ishank011:


pkg/ocm/invite/invite.go
```diff
@@ -18,6 +18,22 @@
...
+
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	invitepb "github.com/cs3org/go-cs3apis/cs3/invite/v1beta1"
+	ocm "github.com/cs3org/go-cs3apis/cs3/sharing/ocm/v1beta1"
+)
+
+// Manager is the interface that is used to perform operations to invites.
 type Manager interface {
+	// GenerateToken creates a new token for the user with a specified validity.
+	GenerateToken(ctx context.Context, user *userpb.UserId) (*invitepb.InviteToken, error)
```
labkode, line 32: You don't need the user *userpb.UserId argument here, the implementation will be able to obtain it from the context, this action is protected with the authentication middleware.

ishank011: Yep. Got it :+1:

pkg/ocm/invite/manager/memory/memory.go
```diff
@@ -0,0 +1,131 @@
...
+
+	m.Invites.Store(inviteToken.GetToken(), inviteToken)
+	return inviteToken, nil
+}
+
+func (m *manager) ForwardInvite(ctx context.Context, invite *invitepb.InviteToken, originProvider *ocm.ProviderInfo) error {
+	contexUser := user.ContextMustGetUser(ctx)
+	requestBody := url.Values{
+		"token":              {invite.GetToken()},
+		"userID":             {contexUser.GetId().GetOpaqueId()},
```
LovisaLugnegard, line 87: Is there a reason for having `userID` instead of `userId`?

PR #636, opened 2020-04-02
Resolve references

butonic:
- Allows listing actual folders in phoenix
- Also gets rid of unneded type checks that are done inside checkRef anyway


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1030,7 +1030,29 @@ func (s *svc) ListContainer(ctx context.Context, req *provider.ListContainerRequ
 	if s.isSharedFolder(ctx, p) {
 		// TODO(labkode): we need to generate a unique etag if any of the underlying share changes.
 		// the response will contain all the share names and we need to convert them to non resference types.
-		return s.listContainer(ctx, req)
+		lcr, err := s.listContainer(ctx, req)
```
labkode, line 1033: This code is to make sure the response doesn't contain REFERENCE types right?

butonic: yes

Issue #326, opened 2019-10-22
ldap user manager does not fill id, opaque_id and idp


butonic:
currently lookup of opaque_id and idp in ldap is not possible because it the ldap user provider does not fill them with anything. we at least need to be able to configure the idp. the opaque_id should be configurable by specifying which attrubute from ldap to use. we could split the userPrincipalName or configure two attributes for opaqe_id and idp (which map to sub and iss in openid world).

We should implement an id-strategy so that we can implement different solutions how to look up the id.

warning ... this is only a workaround: there might not be a way to retrieve the oidc `sub` from ldap. 

The correct solution is to implement a new [SCIM](http://www.simplecloud.info) user manager, because there are idps that support listing users using the SCIM api and there is already a draft [OpenID Connect Profile for SCIM Services](https://openid.net/specs/openid-connect-scim-profile-1_0.html). Tracked in https://github.com/cs3org/reva/issues/327

labkode:
@madsi1m will provide a configuration option for the LDAP plugin to setup the Idp

butonic:
We are dancing around the problem that we need a way to persist a user id but the different systems may know the user under different ids:
- The idp generates a `sub` claim that is unknown to eg ldap 
- LDAP generates a uuid like attrubute, but the organization might have introduced 
 a custom / different attribute to uniquely identify users, eg the [employeeNumber](https://ldapwiki.com/wiki/EmployeeNumber) because that does not change on reimports of a user in the ldap tree.
- The storage provider internally might use a numeric uid and gid for file ownership and permissions.

The perfect userid for reva is unique, persistent and non-reassignable. Why?
1. unique, so we can differentiate between two end users called 'Anna Meyer'
2. persistent, so we can identify the user on  subsequent logins, even if she marries as 'Anna Müller'
3. non-reassignable, so we treat a new 'Anna Meyer' joining the organization as a different person than the 'Anna Meyer' that left the organization 10 years ago.

Best we can do is keep track of the history uf userids and provide a migration script.

First step to make this technically possible: https://github.com/cs3org/cs3apis/pull/46
It allows the auth providers to return not only return a userid but also other metadata.

butonic:
Ultimately, we will always need to be able to update the owner and sharing permissions in an easy way. The idea is to have a cli tool that can replace a user or group id and transfer the ownership and share grantees. Tracked in https://github.com/owncloud/ocis/issues/58

Keeping a history of previous ids can be added later, as it only allows deferring the ownership change.

[...]

PR #641, opened 2020-04-03
ocm: Check for ID provider match as well when creating shares

ishank011:
When creating OCM shares, we need to take both the user ID as well as the provider ID as input, as the same user ID can be present across mesh providers, but the combination of the provider ID and user ID would be unique. I changed the logic for the JSON driver, but for ldap, there's a [comment](https://github.com/ishank011/reva/blob/master/pkg/user/manager/ldap/ldap.go#L131) that says ` this is screaming for errors if filter contains >1 %s`, tracked in https://github.com/cs3org/reva/issues/326. So that needs to be corrected.

### DIFF
--- a/examples/ocmd/users.demo.json
+++ b/examples/ocmd/users.demo.json
@@ -2,7 +2,7 @@
 	{
 		"id": {
 			"opaque_id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
-			"idp": "http://localhost:20080"
+			"idp": "http://cernbox.cern.ch"
 		},
 		"username": "einstein",
 		"secret": "relativity",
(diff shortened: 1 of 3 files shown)


labkode:
@ishank011 for OCM shares, you need to take as userID whatever is given by sharer OCM service when forwarding the invitation. The IdP and UserID sent by the remote OCM provider does not neccesary needs to exist on the receiving part, indeed, that will be the normal case.

ishank011:
@labkode Okay so are we sure that a user's ID, as provided by the sharer OCM service, would be unique across all the providers?

labkode:
@ishank011 I got your question. You're right, the user id provided by the original sharer is only unique inside that system, for us, to be unique we need take in consideration the provider.

PR #642, opened 2020-04-04
refactor and fix ocs

butonic:
- moves handlers into a hierarchy to prepare manageng pending shares 
- fixes the xml rendering for the ocs data tag in case it is a list and needs to wrap items in an element tag (which is the case when listing shares with me or others)

@refs This might impact your listing of public shares pr


labkode:
Adding @refs as reviewer

butonic:
do not merge yet ... will refactor further to make hound happy

butonic:
good to merge now IMO

internal/http/services/owncloud/ocs/response/response.go
```diff
@@ -34,13 +35,68 @@ type Response struct {
...
+				return
+			}
+		}
+		if err = e.EncodeToken(xml.EndElement{Name: dataName}); err != nil {
+			return
+		}
+	}
+
+	// we need to use reflection to determine if p.Data is an array or a slice
+	rt := reflect.TypeOf(p.Data)
```
refs, line 79: nothing against this, but I imagine this getting hit quite often, is there any way to avoid reflection here?

refs: type assertion perhaps?

```golang
package main

import (
	"fmt"
)

func main() {
	c := []string{"a", "b"}
	var a interface{} = c
	s, ok := a.([]string)
	fmt.Println(s, ok) // [a b] true
}
```

I don't have the entire problem scope on my head and I'm sure I have a blind spot as per why you came with this solution.

I would also add a benchmark perhaps, so we see how it performs. This can be done a posteriori perhaps?

labkode: Reflection is evil, should be avoided, I'll create a issue for that.

PR #653, opened 2020-04-08
fix ocs xml rendering

butonic:
fixes this:
```
3:54PM ERR home/jfd/Repositories/reva/internal/http/services/owncloud/ocs/response/response.go:172 > error encoding ocs response error="xml: EncodeToken of invalid token type" pid=6332 pkg=rhttp traceid=0193ecd6f38bc6665855fbb40c79d1a0
```


butonic:
found another bug in this

butonic:
ok, now it works fine

PR #658, opened 2020-04-14
Update Phoenix tutorial

LovisaLugnegard:
This is to clarify that there might be problems if Phoenix is cloned into a `src` folder.


LovisaLugnegard:
This issue is fixed on the Phoenix side

PR #664, opened 2020-04-16
[WIP] first toy for communicating to EOS through grpc

ffurano:
This PR will contribute with a GRPC connector to EOS instead of the current connector that forks a new shell process and parses stdout/stderr.


pkg/eosclientgrpc/eos_grpc/eos_grpc.pb.go
```diff
@@ -0,0 +1,3182 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// source: eos_grpc.proto
```
butonic, line 20: could you add a link to the source?

butonic: is it this one: https://gitlab.cern.ch/eos/grpc-proto/-/blob/master/protobuf/Rpc.proto ?

ffurano: Hi, the source is in the same directory by now, with a comment showing the command line used to compile it.

ffurano:
:-)

PR #671, opened 2020-04-20
Add ocm core to gateway

LovisaLugnegard:
Adds the method described in https://github.com/cs3org/cs3apis/pull/65 to the gateway. Note: This cannot be merged before https://github.com/cs3org/cs3apis/pull/65 is merged (build will fail until the mentioned PR is merged).


ishank011:
We need to add listShares and getShare methods here as well as to cs3apis.

Issue #4, opened 2019-02-06
Add If-Not-Match header on PUT


DeepDiver1975:
If-Match is implemented - https://github.com/cernbox/reva/blob/a6be57bfedc01eddf38148bb7015c2fb8a7a0058/services/httpsvc/ocdavsvc/put.go#L151

To be capable to ensure that overwrite upon creating a new file is not possible the If-Not-Match header with the special value '*' should be implemented.

How does it work:
the client sends If-Not-Match: '*' with a request which shall create a new file. The server checks if the file does not exist otherwise 412 is returned.

labkode:
@butonic is this still the case?

DeepDiver1975:
there is still no if-not-match in https://github.com/cernbox/reva/blob/master/cmd/revad/svcs/httpsvcs/ocdavsvc/put.go - so I guess this is still the case ....

butonic:
see https://tools.ietf.org/html/rfc2616#section-14.26

labkode:
@DeepDiver1975 @butonic thanks for the pointers, does ownCloud uses the ETAG as the entity or only uses the * case?

DeepDiver1975:
Both cases are used . If-None with * to make sure a new resource is created and If-Match 'etag' to make sure we do not overwrite by accident

butonic:
see https://github.com/cs3org/reva/pull/674/files/4d072d0601acc9e21d5e774f1f714cf5ac9fc6f0#r413058259

micbar:
@butonic if this fixed by #674 ?

DeepDiver1975:
> @butonic if this fixed by #674 ?

no - there is no If-Not-Match in the mentioned PR

DeepDiver1975:
After 7 years: if you don't care - I could not care less 🤷

PR #674, opened 2020-04-21
tus for ocdav

butonic:
This PR / branch will serve as the basis for further ocis-web development

includes https://github.com/cs3org/reva/pull/669
based on https://github.com/cs3org/reva/pull/661


butonic:
cc @PVince81 @@LukasHirt pleaase try phoenix on top of this

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -112,6 +112,12 @@ func (s *svc) handlePropfind(w http.ResponseWriter, r *http.Request, ns string)
 	}
 	w.Header().Set("DAV", "1, 3, extended-mkcol")
 	w.Header().Set("Content-Type", "application/xml; charset=utf-8")
+	// let clients know this collection supports tus.io POST requests to start uploads
+	if info.Type == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
+		w.Header().Add("Access-Control-Expose-Headers", "Tus-Version, Tus-Extension")
+		w.Header().Set("Tus-Version", "1.0.0")
```
PVince81, line 118: I thought we agreed to also set "Tus-Resumable" in the response ?

internal/http/services/owncloud/ocdav/options.go
```diff
@@ -27,9 +27,14 @@ func (s *svc) handleOptions(w http.ResponseWriter, r *http.Request, ns string) {
...
 	allow += " PUT" // TODO(jfd): only for files ... but we cannot create the full path without a user ... which we only have when credentials are sent
 
+	w.Header().Add("Access-Control-Allow-Headers", "Tus-Resumable")
+	w.Header().Add("Access-Control-Expose-Headers", "Tus-Version, Tus-Extension")
+
 	w.Header().Set("Content-Type", "application/xml")
 	w.Header().Set("Allow", allow)
 	w.Header().Set("DAV", "1, 2")
 	w.Header().Set("MS-Author-Via", "DAV")
+	w.Header().Set("Tus-Version", "1.0.0") // TODO(jfd): only for dirs?
```
PVince81, line 37: add Tus-Resumable there as well ?

internal/http/services/dataprovider/dataprovider.go
```diff
@@ -54,18 +53,6 @@ func New(m map[string]interface{}) (global.Service, error) {
 		return nil, err
 	}
 
-	if conf.Prefix == "" {
-		conf.Prefix = "data"
```
labkode, line 58: We should provide a good default, exposing on root (/) is not good as it shadows all services in standalone mode.

What about `tus`?

internal/http/services/dataprovider/dataprovider.go
```diff
@@ -103,24 +96,83 @@ func (s *svc) Handler() http.Handler {
...
+			log := appctx.GetLogger(r.Context())
+			log.Info().Msgf("tusd routing: path=%s", r.URL.Path)
+
+			switch r.Method {
+			// old fashioned download.
+
+			// GET is not part of the tus.io protocol
+			// currently there is no way to GET an upload that is in progress
+			// TODO allow range based get requests? that end before the current offset
+			case "GET":
```
labkode, line 138: GET needs to be there to download normal files, not necessary ongoing uploads.

internal/http/services/owncloud/ocdav/tus.go
```diff
@@ -0,0 +1,140 @@
...
+
+	info := sRes.Info
+	if info != nil && info.Type != provider.ResourceType_RESOURCE_TYPE_FILE {
+		log.Warn().Msg("resource is not a file")
+		w.WriteHeader(http.StatusConflict)
+		return
+	}
+
+	if info != nil {
+		clientETag := r.Header.Get("If-Match")
```
PVince81, line 98: add TODO or raise ticket for "If-None-Match" header support as well

for example one can send a request with `If-None-Match: *` to prevent overwriting an already existing file in case of concurrency (saves doing a HEAD beforehand to check for existtence)

pkg/storage/fs/local/upload.go
```diff
@@ -0,0 +1,290 @@
...
+
+	// TODO check etag with If-Match header
+	// if destination exists
+	//if _, err := os.Stat(np); err == nil {
+	// the local storage does not store metadata
+	// the fileid is based on the path, so no we do not need to copy it to the new file
+	// the local storage does not track revisions
+	//}
+
+	err := os.Rename(upload.binPath, np)
```
PVince81, line 266: we might also need a way to return the etag and oc-fileid for desktop client and others, like for the older PUT requests

cc @TheOneRing

internal/http/services/dataprovider/dataprovider.go
butonic: I think `data` is fine because it can do more than just tus.

internal/http/services/dataprovider/dataprovider.go
butonic: which is what the current code does.

internal/http/services/owncloud/ocdav/tus.go
butonic: tracked in https://github.com/cs3org/reva/issues/4

pkg/storage/fs/local/upload.go
butonic: This is the tus upload implementaion. If we need to return the etag that would have to be done in the tus handler. Well, or using the hook mechanism ... for ocdav, our webdav api, we can execute a stat request after the upload via tus.

But if the clients are speaking tus directly, indeed, we should be able to return arbitrary metadata ... 

Hm. I think this should not be part of tus, because the bytes have been transferred. the file metadata is now available in the cs3 api, which leads to another question: How long does it take for the file to become available for download? There might be antivirus scanning going on etc. Tracked in https://github.com/owncloud/ocis/issues/214

As long as this is hidden behind ocdav the stat fallback works ... transloadid uses websockets to send progress notifications of the 'assembly' to the clients. It would allow at least phoenix to show a progresspar that includes detailed postprocessing steps.

butonic:
I think I addressed all comments or raised specific issues to track them

PVince81:
- [ ] BUG: getting a 500 internal server error for non-TUS upload with PUT (works fine on reva master)

PVince81:
- [ ] BUG: on this branch I still get a CORS error:
image

it wasn't present on the butonic:tus-dataprovider branch and works fine there when testing with https://github.com/owncloud/phoenix/pull/3345 (had to hack it to bypass capability check)

butonic:
ok, the cors issues are fixed but require changes to the Makefile as well. but we now run into the funny namespacing again: https://github.com/owncloud/ocis/pull/219

internal/http/services/owncloud/ocdav/tus.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// read filename from metadata
+	meta := tusd.ParseMetadataHeader(r.Header.Get("Upload-Metadata"))
+	if meta["filename"] == "" {
+		w.WriteHeader(http.StatusPreconditionFailed)
+		return
+	}
+
+	// append filename to current dir
+	fn := path.Join(ns, r.URL.Path, meta["filename"])
```
butonic, line 59: @PVince81 this is where the wrong path is build when the ns is set to /home but the url path is taken from the dav/files endpoint, because it starts with the username ... but /home paths don't need the username. 💥 

If you could make Phoenix send tho POST to the old webdav endpoint this would work for now ...

butonic:
Ok, with the [last commit](https://github.com/cs3org/reva/pull/674/commits/9c4ff7cc45734f8704cc97665c10596ce35f9631) I can now configure
- `remote.php/webdav` to use `/home` and 
- `remote.php/dav/files` to use `/eos/users/{{substr 0 1 .Username}}/` (instead of just `/eos`)

Which allows the home storage to return references to the /eos dataprovider (which is configured not to use user_layouts because it is NOT a home storage)

The remote.php/dav/files endpoint will as the first argument always get a username which we must 'layout' the same way as in eos. I am currently relying on the requests to always use the user that is logged in ... but we could change that to layout the first path parameter, becaluse we know in the `dav/files/` endpoint always is the username.

Also ... it seems to become more clear that a global endpoint should get its own endpoint, eg. `/dav/global`

butonic:
@labkode please restart https://cloud.drone.io/cs3org/reva/1023/1/5

cmd/reva/upload.go
```diff
@@ -76,9 +82,17 @@ func uploadCommand() *command {
 					Path: target,
 				},
 			},
+			Opaque: &typespb.Opaque{
+				Map: map[string]*typespb.OpaqueEntry{
+					"Upload-Length": {
```
labkode, line 87: - [ ] Extend CS3APIS

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -243,14 +244,30 @@ func (s *service) InitiateFileDownload(ctx context.Context, req *provider.Initia
...
-	url := *s.dataServerURL
 	newRef, err := s.unwrap(ctx, req.Ref)
 	if err != nil {
 		return &provider.InitiateFileUploadResponse{
 			Status: status.NewInternal(ctx, err, "error unwrapping path"),
 		}, nil
 	}
-	url.Path = path.Join("/", url.Path, newRef.GetPath())
+	var uploadLength int64
+	if req.Opaque != nil && req.Opaque.Map != nil && req.Opaque.Map["Upload-Length"] != nil {
```
labkode, line 254: It should come from the CS3 Req

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -184,10 +191,24 @@ func (s *svc) handlePut(w http.ResponseWriter, r *http.Request, ns string) {
 		}
 	}
 
+	length, err := strconv.ParseInt(r.Header.Get("Content-Length"), 10, 64)
```
labkode, line 194: Check that we have valid length?

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -204,25 +225,44 @@ func (s *svc) handlePut(w http.ResponseWriter, r *http.Request, ns string) {
...
-	defer httpRes.Body.Close()
 
-	if httpRes.StatusCode != http.StatusOK {
+	metadata := map[string]string{
+		"filename": path.Base(fn),
+		"dir":      path.Dir(fn),
+		//"checksum": fmt.Sprintf("%s %s", storageprovider.GRPC2PKGXS(xsType).String(), xs),
+	}
+
+	upload := tus.NewUpload(r.Body, length, metadata, "")
```
labkode, line 257: In the request we don't have enough information to create an unique fingerprint to have resumable uploads, that's why is empty. This is all about the PUT request.

labkode: We use TUS here because the tus library will translate PUT to POST/PATHC so the data servers needs to only implement TUS and not PUTs.

internal/http/services/owncloud/ocdav/tus.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// initiateUpload
+
+	uReq := &provider.InitiateFileUploadRequest{
+		Ref: &provider.Reference{
+			Spec: &provider.Reference_Path{Path: fn},
+		},
+		Opaque: &typespb.Opaque{
+			Map: map[string]*typespb.OpaqueEntry{
+				"Upload-Length": {
```
labkode, line 123: CS3APIS extension

internal/http/services/owncloud/ocdav/tus.go
```diff
@@ -0,0 +1,178 @@
...
+	if uRes.Status.Code != rpc.Code_CODE_OK {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Location", uRes.UploadEndpoint)
+
+	// for creation-with-upload extension forward bytes to dataprovider
+	// TODO check this really streams
+	if r.Header.Get("Content-Type") == "application/offset+octet-stream" {
```
labkode, line 147: Better to proxy changing method?

pkg/storage/fs/eos/upload.go
```diff
@@ -0,0 +1,319 @@
...
+	return upload.info, nil
+}
+
+// GetReader returns an io.Readerfor the upload
+func (upload *fileUpload) GetReader(ctx context.Context) (io.Reader, error) {
+	return os.Open(upload.binPath)
+}
+
+// WriteChunk writes the stream from the reader to the given offset of the upload
+// TODO use the grpc api to directly stream to a temporary uploads location in the eos shadow tree
```
labkode, line 218: Send chunk to EOS directly

internal/grpc/services/storageprovider/storageprovider.go
butonic: tracked in https://github.com/cs3org/cs3apis/issues/71

cmd/reva/upload.go
butonic: tracked in https://github.com/cs3org/cs3apis/issues/71

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -184,10 +191,24 @@ func (s *svc) handlePut(w http.ResponseWriter, r *http.Request, ns string) {
...
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
 	uReq := &provider.InitiateFileUploadRequest{
 		Ref: &provider.Reference{
 			Spec: &provider.Reference_Path{Path: fn},
 		},
+		Opaque: &typespb.Opaque{
```
butonic, line 204: tracked in https://github.com/cs3org/cs3apis/issues/71

internal/http/services/owncloud/ocdav/tus.go
butonic: tracked in https://github.com/cs3org/cs3apis/issues/71

internal/http/services/owncloud/ocdav/put.go
butonic: tracked in https://github.com/cs3org/reva/issues/699

internal/http/services/owncloud/ocdav/tus.go
butonic: tracked in tracked in https://github.com/cs3org/reva/issues/699

pkg/storage/fs/eos/upload.go
butonic: tracked in https://github.com/cs3org/reva/issues/700

internal/http/services/owncloud/ocdav/put.go
butonic: https://github.com/cs3org/reva/issues/702

PR #696, opened 2020-04-29
implement ocs to http status code mapping

C0rby:
This change helps us getting our api tests to pass.
First step to resolve https://github.com/owncloud/ocis-reva/issues/26


PVince81:
seems this breaks login with Phoenix now: https://github.com/owncloud/ocis/issues/241

not sure why OCS would affect that, maybe it's the capabilities call

PR #701, opened 2020-04-30
Range request attempt WIP

PVince81:
Naive draft attempt at adding range request.
This would require the underlying data provider + storage to support range headers.

Strangely when I run this it doesn't return the "Accept-range" request.

For https://github.com/owncloud/ocis-reva/issues/12

@butonic FYI

### DIFF
--- a/internal/http/services/owncloud/ocdav/get.go
+++ b/internal/http/services/owncloud/ocdav/get.go
@@ -99,6 +99,10 @@ func (s *svc) handleGet(w http.ResponseWriter, r *http.Request, ns string) {
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+
+	if r.Header.Get("Range") != "" {
+		httpReq.Header.Set("Range", r.Header.Get("Range"))
+	}
 	httpReq.Header.Set("X-Reva-Transfer", dRes.Token)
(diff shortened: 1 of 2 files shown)


PVince81:
Ok, I missed changing my go.mod with the replace...

So the HEAD call returns the headers.

The GET call returns the full content, which is expected because the underlying layers likely don't support the range headers yet

PVince81:
Actually the "HEAD" request should also ping the backend to find out if it "accepts ranges"...

PVince81:
seems we might need the CS3 APIs to support range requests: https://github.com/owncloud/ocis-reva/issues/12

labkode:
@PVince81 I think we can add them to CS3APIS at some point, however what if we try the following for time being:

```
1) Get with Range header arrives to ocdav layer
2) ocdav will call CS3APIS:InitiateFileUpload and get the TUS server address
3) ocdav will send Range header to TUS data server
4) TUS data server will serve only the requested bytes
```

PVince81:
now that I'm a bit more familiar, maybe passing this through the Opaque would be enough ?

PVince81:
never mind, I just saw your proposal, doing a range request to the data provider sounds like a better approach

PR #712, opened 2020-05-05
Always return size in ocdav

PVince81:
Return the file size in ocdav regardless whether the mime type was
detected or not

See https://github.com/owncloud/ocis-reva/issues/67#issuecomment-623944982 for context.

Separately, we might also want to default to "application/octet-stream" instead of an empty string when the mime type was not detectable.

@butonic @labkode


PVince81:
Ohh, and now I realize this is wrong. This would return the file size also for folders.
I'll adjust the condition...

PVince81:
PR adjusted

labkode:
Thanks @PVince81 , makes a lot of sense to default to application/octet-stream, mind creating another PR?

PVince81:
@labkode done: https://github.com/cs3org/reva/pull/713

PR #714, opened 2020-05-05
Fix COPY with Depth 0

PVince81:
Honor the Depth header in Webdav COPY operation.
A Depth 0 means only duplicate the collection and not its contents.
A Depth infinity means also copy its contents.

@butonic

### DIFF
--- a/internal/http/services/owncloud/ocdav/copy.go
+++ b/internal/http/services/owncloud/ocdav/copy.go
@@ -42,8 +42,13 @@ func (s *svc) handleCopy(w http.ResponseWriter, r *http.Request, ns string) {
 	src := path.Join(ns, r.URL.Path)
 	dstHeader := r.Header.Get("Destination")
 	overwrite := r.Header.Get("Overwrite")
+	depth := r.Header.Get("Depth")
+	if depth == "" {
+		depth = "infinity"
+	}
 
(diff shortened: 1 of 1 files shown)


PVince81:
note: copy with depth infinity is still broken, raised here: https://github.com/owncloud/ocis-reva/issues/173

internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -153,7 +163,7 @@ func (s *svc) handleCopy(w http.ResponseWriter, r *http.Request, ns string) {
 		// TODO what if intermediate is a file?
 	}
 
-	err = descend(ctx, client, srcStatRes.Info, dst)
+	err = descend(ctx, client, srcStatRes.Info, dst, depth == "infinity")
```
labkode, line 166: Shouldn't we descend also in depth > 1?

PVince81: the spec doesn't allow any other values than "0" and "infinity": https://tools.ietf.org/html/rfc4918#section-9.8.3

labkode: @PVince81 Depth 1 is used by the sync clients

PR #717, opened 2020-05-06
Fix Webdav COPY to use TUS for internal upload

PVince81:
When copying single files or a files recursively, the files are
uploaded to the target on the storage. Since the storage is now expecting
a TUS workflow, this commit fixes it to now use the TUS uploader.

Fixes https://github.com/owncloud/ocis-reva/issues/173

@butonic


labkode:
@PVince81 please rebase

PVince81:
@labkode rebased and retested

PR #722, opened 2020-05-06
Cubbit example

labkode:
Run with:

```
revad -c cubbit.toml
```


alessandro-p:
Thanks for the example. I think we should add this line as well in order to point out the user service provider: 

`userprovidersvc = "address"`

Issue #710, opened 2020-05-05
TUS creation-with-upload returns 400 using OC storage


PVince81:
### Steps

1. Setup [ocis](https://github.com/owncloud/ocis)
1. Adjust go.mod to point at the latest ocis-reva master and reva master
1. `make clean build`
1. Start all service except ocis-phoenix
1. Start separate ocis-phoenix service and set the assets to point at a compiled version of https://github.com/owncloud/phoenix/pull/3436 (built with `yarn build:all`)
1. Login to "https://localhost:9200"
1. Open the network console
1. Upload a file

### Expected result
File is uploaded, no errors

### Actual result
Network call returns 400 Bad Request with no body.

### Logs
```
2020-05-05T09:08:52+02:00 DBG ocfs: unwrap: internal=/einstein/files/test/lmms-1.2.0-rc8-linux-x86_64.AppImage external=/einstein/test/lmms-1.2.0-rc8-linux-x86_64.AppImage pkg=rgrpc service=reva traceid=6e7ec7c05514015afd30486edbf4b355
2020-05-05T09:08:52+02:00 DBG ocfs: NewUpload info={"ID":"","IsFinal":false,"IsPartial":false,"MetaData":{"dir":"/einstein/test","filename":"lmms-1.2.0-rc8-linux-x86_64.AppImage"},"Offset":0,"PartialUploads":null,"Size":98168520,"SizeIsDeferred":false,"Storage":null} pkg=rgrpc service=reva traceid=6e7ec7c05514015afd30486edbf4b355
2020-05-05T09:08:52+02:00 DBG ocfs: resolved filename info={"ID":"","IsFinal":false,"IsPartial":false,"MetaData":{"dir":"/einstein/test","filename":"lmms-1.2.0-rc8-linux-x86_64.AppImage"},"Offset":0,"PartialUploads":null,"Size":98168520,"SizeIsDeferred":false,"Storage":null} pkg=rgrpc service=reva traceid=6e7ec7c05514015afd30486edbf4b355
2020-05-05T09:08:52+02:00 INF file upload data-server=http://localhost:9164/data/d1407412-faa5-4773-9069-c71f670728c9 fn=/oc/einstein/test/lmms-1.2.0-rc8-linux-x86_64.AppImage pkg=rgrpc service=reva traceid=6e7ec7c05514015afd30486edbf4b355 xs="map[md5:100 unset:1000]"
[...]

PR #725, opened 2020-05-07
Fix TUS upload with POST and payload

PVince81:
The JS lib is not sending a zero offset during upload, it is likely
optional. This fix makes it default to 0 when no Upload-Offset header is
set.

Also fixed returned header to be set on the correct response object, the
main one.

@butonic


PVince81:
Fixes https://github.com/cs3org/reva/issues/710

Tested using https://github.com/owncloud/phoenix/pull/3436

PR #727, opened 2020-05-08
Export site metrics in Prometheus

labkode:
Run like this:

```
red:reva gonzalhu$ cmd/revad/revad -c examples/metrics/metrics.toml
```
Test like this:
```
red:reva gonzalhu$ curl http://0.0.0.0:9998/metrics
# HELP revad_opencensus_io_http_server_request_count Count of HTTP requests started
# TYPE revad_opencensus_io_http_server_request_count counter
revad_opencensus_io_http_server_request_count 1
# HELP revad_opencensus_io_http_server_request_count_by_method Server request count by HTTP method
# TYPE revad_opencensus_io_http_server_request_count_by_method counter
revad_opencensus_io_http_server_request_count_by_method{http_method="GET"} 1
```

You may want to copy this branch into one of yours


redblom:
@labkode Is it possible to make the site metrics to somehow show up between the current prometheus service metrics (displayed through the exporter). I can only get them displayed if I let the current prometheus service New() function simply return with a promhttp.Handler() and remove/ignore the exporter code. Like so:
`return &svc{prefix: conf.Prefix, h: promhttp.Handler()}, nil`

labkode:
@ishank011  can you please take a look  a this corner?

ishank011:
@labkode Will look at it :+1:

ishank011:
@labkode @redblom We can define the metrics in the metrics package as has been done in #740, and then import those in the prometheus HTTP service and register them as separate views as done [here](https://github.com/census-ecosystem/opencensus-go-exporter-prometheus/blob/master/example/main.go#L50), and then initialize the counters. Initializing is important, otherwise the metrics don't show. In the metrics pkg, we can define the methods to update the stats.

redblom:
@ishank011 Thank you for the explanation. I can implement this fine, this works.
@labkode I pushed the new commits.

PR #728, opened 2020-05-08
Fix OC storage property removal

PVince81:
Whenever a non-existing property must be removed, we should ignore the
error.

Also note: PROPPATCH spec says that removing a non-existing property is ok: http://www.webdav.org/specs/rfc2518.html#ELEMENT_remove

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -1141,11 +1141,15 @@ func (fs *ocfs) UnsetArbitraryMetadata(ctx context.Context, ref *provider.Refere
 			}
 		default:
 			if err = xattr.Remove(np, mdPrefix+k); err != nil {
-				log.Error().Err(err).
-					Str("np", np).
-					Str("key", k).
-					Msg("could not unset metadata")
-				errs = append(errs, errors.Wrap(err, "could not unset metadata"))
+				// a non-existing attribute will return an error, which we can ignore
+				// (using string compare because the error type is syscall.Errno and not wrapped/recognizable)
+				if e, ok := err.(*xattr.Error); !ok || e.Err.Error() != "no data available" {
+					log.Error().Err(err).
+						Str("np", np).
+						Str("key", k).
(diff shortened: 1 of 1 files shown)


PVince81:
To test, run this twice after setting the properties in question:
```
curl -D - -u einstein:relativity -X PROPPATCH -H "Content-Type: text/xml" --data-binary "@proppatch-remove.xml" "http://localhost:9140/remote.php/webdav/test"
```
with the file proppatch-remove.xml:
```xml
<?xml version="1.0" encoding="utf-8"?>
<D:propertyupdate xmlns:D="DAV:">
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/alpha"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/beta"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/gamma"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/delta"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/epsilon"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/zeta"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/eta"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/theta"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/iota"/>
    </D:prop>
  </D:remove>
  <D:remove>
    <D:prop>
      <somename xmlns="http://example.com/kappa"/>
    </D:prop>
  </D:remove>
</D:propertyupdate>
```

Before the fix we got 500 Internal server error because "data is unavailble".
After the fix we get a 207 response.

PR #729, opened 2020-05-08
Fix OS storage move propagation

PVince81:
Propagating on the source path failed because the source path does not exist any more after moving. This fix makes it start the propagation on the parent of the source.

I've also fixed the error handling because while debugging this it caused memory pointer errors as the errors were not propagated to the outside.

@butonic


PVince81:
=> this makes the litmus webdav move tests pass!

labkode:
@PVince81 just lint and is good to merge

PVince81:
I've removed the stray debug code and fixed the linter errors by properly returning the error objects.

PVince81:
all green, please merge :-)

PR #730, opened 2020-05-08
Retry trashbin in case of timestamp collision

PVince81:
In OC storage the file name pattern for deleted files include the
timestamp: "deleted.d12345678". In some rare cases like when running
tests, it can happen that the deletion happens so fast that it causes
timestamp collisions for the same file names.

To fix this, if we detect the collision case we retry again with a
higher timestamp value.

After this, the Webdav Litmus test cleanup routine works correctly.

Fixes https://github.com/owncloud/ocis-reva/issues/184

Also see: https://github.com/owncloud/core/issues/23151

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -1213,9 +1213,17 @@ func (fs *ocfs) Delete(ctx context.Context, ref *provider.Reference) (err error)
 	}
 
 	// move to trash location
-	tgt := path.Join(rp, fmt.Sprintf("%s.d%d", path.Base(np), time.Now().Unix()))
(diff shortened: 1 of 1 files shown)


labkode:
@PVince81  perhaps a nano second resolution will work?

PVince81:
Yeah, but here I'm assuming that we want to keep the same old file name schema to stay compatibleAm 08.05.2020 13:36 schrieb "Hugo G. Labrador" <notifications@github.com>:
@PVince81  perhaps a nano second resolution will work?

PVince81:
looks like I have some leftover debug code here, will remove

PVince81:
I've removed the debug code.

PR #732, opened 2020-05-08
Local fs: metadata handling and share persistence

ishank011:


labkode:
@ishank011 can we have the recycle and versions on the .shadow directory also?

ishank011:
@labkode Cool. Will update the paths

ishank011:
@labkode We would have to modify the namespace resolution in that case. When unwrapping paths such as /root/.shadow/einstein/home/MyShares/a.txt, there will be longer matches with the versions and recycle bin paths. So should I create a separate directory for the references as well? Eg: /root/.shadow/.references/einstein/home/MyShares/a.txt

PR #740, opened 2020-05-13
New metrics package

redblom:
3 site metrics for now: number of users, number of groups, storage amount in use

### DIFF
--- a/internal/http/services/prometheus/prometheus.go
+++ b/internal/http/services/prometheus/prometheus.go
@@ -20,6 +20,7 @@ package prometheus
 
 import (
 	"net/http"
+	mtrcs "reva/pkg/metrics"
 
 	"github.com/cs3org/reva/pkg/rhttp/global"
 	"github.com/mitchellh/mapstructure"
@@ -28,6 +29,8 @@ import (
 
 func init() {
 	global.Register("prometheus", New)
+	m := mtrcs.New()
+	m.SayHello()
 }
 
 // New returns a new prometheus service
--- a/internal/http/services/prometheus/prometheus.go
+++ b/internal/http/services/prometheus/prometheus.go
@@ -58,11 +58,13 @@ func New(m map[string]interface{}, log *zerolog.Logger) (global.Service, error) 
 	view.RegisterExporter(pe)
 
 	// register the desired measures' views
-	view.Register(
+	if err = view.Register(
 		metrics.GetNumUsersView(),
 		metrics.GetNumGroupsView(),
 		metrics.GetAmountStorageView(),
-	)
+	); err != nil {
+		return nil, errors.Wrap(err, "prometheus: error registering exporter")
+	}
 
 	return &svc{prefix: conf.Prefix, h: pe}, nil
 }
--- a/internal/http/services/prometheus/prometheus.go
+++ b/internal/http/services/prometheus/prometheus.go
@@ -20,6 +20,7 @@ package prometheus
 
 import (
 	"net/http"
+	mtrcs "reva/pkg/metrics"
 
 	"contrib.go.opencensus.io/exporter/prometheus"
 	"github.com/cs3org/reva/pkg/rhttp/global"
@@ -31,6 +32,8 @@ import (
 
 func init() {
 	global.Register("prometheus", New)
+	m := mtrcs.New()
+	m.SayHello()
 }
 
 // New returns a new prometheus service
--- a/internal/http/services/prometheus/prometheus.go
+++ b/internal/http/services/prometheus/prometheus.go
@@ -21,7 +21,8 @@ package prometheus
 import (
 	"fmt"
 	"net/http"
-	"reva/pkg/metrics"
+
+	"github.com/cs3org/reva/pkg/metrics"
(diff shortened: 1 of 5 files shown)


labkode:
@redblom can you confirm that you can obtain the metrics by doing `GET /metrics`?

redblom:
> 
> 
> @redblom can you confirm that you can obtain the metrics by doing `GET /metrics`?

Yes, but only if I let the prometheus service New() function return a promhttp.Handler() like so: 
`return &svc{prefix: conf.Prefix, h: promhttp.Handler()}, nil`
Not with the existing code of the New() function. Then they don't show up between the displayed metrics.
See https://github.com/cs3org/reva/pull/727#issuecomment-627891729

labkode:
@redblom interesting, let me take a look

internal/http/services/prometheus/prometheus.go
```diff
@@ -31,6 +33,7 @@ import (
 
 func init() {
 	global.Register("prometheus", New)
+	fmt.Printf("metrics - %v \n", metrics.NumUsersMeasure.Description())
```
labkode, line 36: let's remove this debug line

labkode:
@redblom thanks for the update, only one comment and I can merge

internal/http/services/prometheus/prometheus.go
```diff
@@ -19,7 +19,9 @@
 package prometheus
 
 import (
+	"fmt"
 	"net/http"
+	"reva/pkg/metrics"
```
ishank011, line 24: Need to refer to the github repo path, I guess that's why the build isn't passing

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+			time.Sleep(4 * time.Second)
+		}
+	}()
+}
+
+// Create the measures
+var (
+	NumUsersMeasure      = stats.Int64("cs3_org_sciencemesh_site_total_num_users", "The total number of users within this site", stats.UnitDimensionless)
+	NumGroupsMeasure     = stats.Int64("cs3_org_sciencemesh_site_total_num_groups", "The total number of groups within this site", stats.UnitDimensionless)
+	AmountStorageMeasure = stats.Int64("cs3_org_sciencemesh_site_total_amount_storage", "The total amount of storage used within this site", stats.UnitDimensionless)
```
ishank011, line 50: Use stats.UnitBytes here

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+	amountStorageCounter = int64(0)
+)
+
+// getNumberUsers links to the underlying number of site users provider
+func getNumUsers() {
+	ctx := context.Background()
+	// here we must request the actual number of site users
+	// for now this is a mockup: a number increasing over time
+	numUsersCounter += int64(rand.Intn(100))
+	fmt.Printf("nrUsers = %v \n", numUsersCounter)
```
ishank011, line 65: Output to stdout isn't logged, so I guess we can remove these as well

internal/http/services/prometheus/prometheus.go
labkode: Missed that, good catch-up @ishank011

internal/http/services/prometheus/prometheus.go
redblom: Oops, yes :)

internal/http/services/prometheus/prometheus.go
redblom: @ishank011 How to resolve this? Exclude prometheus.go modifications from this pull request, and make that a separate pull request after this one is merged?

ishank011: No, just change it to `github.com/cs3org/reva/pkg/metrics`. Since the remote origin points to the repo itself, it'll resolve it itself.

internal/http/services/prometheus/prometheus.go
```diff
@@ -53,6 +56,13 @@ func New(m map[string]interface{}, log *zerolog.Logger) (global.Service, error)
 
 	view.RegisterExporter(pe)
 
+	// register the desired measures' views
+	view.Register(
```
ishank011, line 60: Check the result of Register as done [here](https://github.com/census-ecosystem/opencensus-go-exporter-prometheus/blob/master/example/main.go#L50) and return if there's an error.

pkg/metrics/metrics.go
redblom: Maybe only on storage amount ?

ishank011: Yeah, just on storage. Github shows the whole blob even if you comment on a single line.

redblom:
@labkode @ishank011 all comments looked after. New commits pushed.

ishank011:
@redblom looks good! Just can you remove the debug log which @labkode pointed out? Thanks!

redblom:
> 
> 
> @redblom looks good! Just can you remove the debug log which @labkode pointed out? Thanks!

Addressed with a new commit.

PR #743, opened 2020-05-14
ocdav: Fix parsing of PROPPATCH XML properties

PVince81:
Use the parent struct when parsing to make sure the decoder finds the
hint about "innerxml" parsing mode for proppatch properties,
and prevents mangling of said values when they contain an "xmlns"
attribute.

Fixes https://github.com/owncloud/ocis-reva/issues/203

@butonic please double check the fix as I didn't understand the lang-related code and seemingly could not create a test case that has any difference before and after the fix.

If I proppatch a value of `<foo xml:lang="en" xmlns="http://bar"/>` it will now also appear as `<foo xml:lang="en" xmlns="http://bar"/>` in the extended attribute and propfind result.

### DIFF
--- a/internal/http/services/owncloud/ocdav/proppatch.go
+++ b/internal/http/services/owncloud/ocdav/proppatch.go
@@ -19,7 +19,6 @@
 package ocdav
 
 import (
-	"bytes"
 	"encoding/xml"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


PVince81:
very strange, the linter says this is unused but it is used... I can't remove it without causing compilation errors:
```
internal/http/services/owncloud/ocdav/proppatch.go:216:6: type `xmlValue` is unused (unused)
```

PVince81:
it is used here: https://github.com/cs3org/reva/pull/743/files#diff-9f71a21d30f2217810e804a01ce9d969R218

PVince81:
ok, got it, it's the whole type/struct/class that is unused (me still learning Golang...)

PVince81:
> cases where the proppatch value was reusing an existing namespace from outside the innerxml

strange, I seemed to remember seeing a comment somewhere that said that it's anyway illegal to reuse namespaces from the parent but cannot find it anywhere in the spec (not present in http://www.webdav.org/specs/rfc4918.html#property_values)

in any case, I think there isn't much benefit in supporting this as the custom properties we'll be setting will have their own namespaces. And for known properties (like the DAV: ones) they have no complex XML values.

I'm not even sure whether Sabre DAV's XML supports this scenario of namespace reuse.

PR #747, opened 2020-05-15
REST driver for user pkg

ishank011:


labkode:
@ishank011 once you are done with this next week let's have a call so you can guide me though :)

ishank011:
@labkode yes, definitely. I'm still getting some weird unmarshalling errors while parsing the response. I'll just resolve that.

pkg/user/manager/rest/rest.go
```diff
@@ -0,0 +1,341 @@
...
+	if c.IDProvider == "" {
+		c.IDProvider = "http://cernbox.cern.ch"
+	}
+
+	return &manager{
+		conf: c,
+	}, nil
+}
+
+func (m *manager) renewAPIToken(ctx context.Context) error {
```
labkode, line 96: if this call can occur concurrently, we need to protect the m.apiToken and m.tokenExpirationTime with locks.

pkg/user/manager/rest/rest.go
```diff
@@ -0,0 +1,341 @@
...
+	return users, nil
+}
+
+func (m *manager) FindUsers(ctx context.Context, query string) ([]*userpb.User, error) {
+
+	filters := []string{"upn", "displayName", "primaryAccountEmail"}
+	users := []*userpb.User{}
+
+	for _, f := range filters {
+		url := fmt.Sprintf("%s/Identity/?filter=%s:contains:%s&field=id&field=upn&field=primaryAccountEmail&field=displayName", m.conf.APIBaseURL, f, query)
```
labkode, line 268: %s can contain arbitrary user input I imagine, we  need to whitelist so we only accept expected characters to avoid an injection attack.

ishank011: Added regexs for email and names.

pkg/user/manager/rest/rest.go
ishank011: Done

ishank011:
Hi Hugo. Thanks for the review. Addressed your comments.

PR #750, opened 2020-05-15
Localfs: return file owner info in stat

### LINKED ISSUE #749: `Stat()` does not return the owner's userid
Starting to use Reva with local storage as part of https://github.com/sciencemesh/sciencemesh/issues/56, I see that `Stat()` returns a structure like the following:
```
status {
  code: CODE_OK
  trace: "0516bb188edb496d149e1fe578f2c00e"
}
info {
  type: RESOURCE_TYPE_FILE
  id {
    storage_id: "123e4567-e89b-12d3-a456-426655440000"
    opaque_id: "fileid-dir/test.md"
  }
  etag: "\"bf82b20e17b478919514294a944a7ff6\""
  mime_type: "text/markdown; charset=utf-8"
  mtime {
    seconds: 1589545699
  }
  path: "/dir/test.md"
  permission_set {
    create_container: true
    list_container: true
  }
  size: 5068
}
```
where the owner's userid is missing.

ishank011:
Closes #749

### DIFF
--- a/pkg/storage/fs/local/local.go
+++ b/pkg/storage/fs/local/local.go
@@ -257,6 +257,11 @@ func (fs *localfs) isShareFolderChild(ctx context.Context, p string) bool {
 
 func (fs *localfs) normalize(ctx context.Context, fi os.FileInfo, fn string) *provider.ResourceInfo {
 	fn = fs.unwrap(ctx, path.Join("/", fn))
+	owner, err := getUser(ctx)
+	if err != nil {
+		return nil
+	}
+
 	md := &provider.ResourceInfo{
 		Id:            &provider.ResourceId{OpaqueId: "fileid-" + strings.TrimPrefix(fn, "/")},
 		Path:          fn,
@@ -268,6 +273,7 @@ func (fs *localfs) normalize(ctx context.Context, fi os.FileInfo, fn string) *pr
 		Mtime: &types.Timestamp{
 			Seconds: uint64(fi.ModTime().Unix()),
 		},
+		Owner: owner.Id,
 	}
 
 	return md


pkg/storage/fs/local/local.go
```diff
@@ -257,6 +257,11 @@ func (fs *localfs) isShareFolderChild(ctx context.Context, p string) bool {
 
 func (fs *localfs) normalize(ctx context.Context, fi os.FileInfo, fn string) *provider.ResourceInfo {
 	fn = fs.unwrap(ctx, path.Join("/", fn))
+	owner, err := getUser(ctx)
```
labkode, line 260: This will add the current logged in user, I'm not sure this will apply for shared resources

labkode: I think we should rely on what the storage gives us, for example, EOS wll always tell us the owner

labkode: So I thnk in the case of local storage is okay to use it like that.

ishank011: Yes, in that case, I thought we might as well return the "owner of the reference" since the reference doesn't have any info about the owner of the original file. Maybe we can pass this info about the original owner to the CreateReference method and then save that as metadata along with the reference (currently we store only the target).

glpatcern: IMHO we should really get the owner, also for local storage, independent from the currently logged in user. Or did I misunderstand @labkode's comment?

ishank011: @glpatcern yes, in case of share references, we don't have any info about the original owner in this context. That's why I suggested passing it along when the reference is created. We'll need to add this field to the cs3apis protobuf.
For cases other than the references, the logged-in user will always be the owner.

labkode:
On EOS is easy as the owner is given for free. For local storage the owner needs to be attached based on the path you are, assuming that /home and logged in as ishank the owner will be ishank.

PR #751, opened 2020-05-15
WIP prevent creating dir in another home

PVince81:
Context is here: https://github.com/owncloud/ocis-reva/issues/197
It's about preventing access to other user's homes.

Currently the OCFS propagator fails because it is performing said check and assuming we work only in the current user's home.

I tried to reuse that logic to enforce a "permission denied" in CreateDir (and later also in other calls). I'm unsure if this check is overkill and whether it should or should not include both cases where EnableHome is set and unset.

However what doesn't work correctly here is that even though [a "permission denied" error is created and returned](https://github.com/PVince81/butonic-reva/blob/ocfs-forbid-cross-home-access/pkg/storage/fs/owncloud/owncloud.go#L948), it still gets converted to an "internal server error" RPC code.

@butonic let me know if this POC makes sense at all and if yes, give me some advice about the RPC code issue

### DIFF
--- a/internal/http/services/owncloud/ocdav/mkcol.go
+++ b/internal/http/services/owncloud/ocdav/mkcol.go
@@ -19,6 +19,7 @@
 package ocdav
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"path"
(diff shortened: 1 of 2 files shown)


PVince81:
note to self, have a look at  2b018b13c24dab305d92164cf0ad4da43807060c which seems to do so for local FS

C0rby:
Wouldn't this break the `remote.php/webdav/` endpoint since it uses the same handler?

butonic:
won't be continued here

PR #755, opened 2020-05-18
Mentix service

Daniel-WWU-IT:
This PR adds Mentix (Mesh Entity Exporter) as a Reva service. This includes:
- Support for GOCDB as the data source
- HTTP endpoint for getting the data in a standardized format (independent of the used source)
- Exporting scrape configs for Prometheus
- Example configuration file
- Configuration documentation

For logging, I had to extend the rhttp `NewService` type to pass the global logger, as Mentix is actually a background service and needs logging outside of an HTTP request. I think that there should be services of a more general type so that services like Mentix can be added in a more elegant, streamlined way.


labkode:
@Daniel-WWU-IT first of all, congrats for this astonishing piece of work, nice job!

> For logging, I had to extend the rhttp `NewService` type to pass the global logger, as Mentix is actually a background service and needs logging outside of an HTTP request. I think that there should be services of a more general type so that services like Mentix can be added in a more elegant, streamlined way.
I couldn't agree more, let make sure is tracked in a Github issue for later.

One question, can you figure out the `pat` dependency where is used and from which pacakge? `go mod why` should help you.

I'm happy to merge after you make CI green (some lints are failling).

Daniel-WWU-IT:
@labkode I just updated everything to make lint happy and added some comments as well. The `path` package was used in network.utils, but I removed that dependency already (though it's a standard package).

Daniel-WWU-IT:
@labkode I created an issue for generic services: #757

labkode:
@Daniel-WWU-IT thanks! I was refering to this package:
https://github.com/cs3org/reva/pull/755/files#diff-f949e2d81c8076ebbf8af38fcbb72c1fR48

Daniel-WWU-IT:
I'm not using that package, neither directly nor indirectly. According to `go mod why`, it is used by the dataprovider service:
  ```github.com/bmizerany/pat
  github.com/cs3org/reva/internal/http/services/dataprovider
  github.com/tus/tusd/pkg/handler
  github.com/bmizerany/pat```

PR #759, opened 2020-05-19
Add option in config to disable tus

ishank011:


labkode:
@ishank011 merging for now, in the future both simple upload (PUT/GET) and TUS should be available.

ishank011:
@labkode GET is not supported by TUS. As for PUT, I'll try to create a wrapper which saves us from the config trouble.

PR #775, opened 2020-05-26
Add TUS global capabilities

PVince81:
Advertise global TUS capabilities so that clients can check whether they
should load their TUS functionality.

This is in addition to the existing advertisement using headers on
Webdav PROPFINDS or OPTIONS depending on the underlying storage.

Note that storages could have different TUS capabilities so in the
future we should look into ways to aggregate the capabilities from
configured storages in these global capability attributes.

For https://github.com/owncloud/ocis-reva/issues/177

### DIFF
--- a/cmd/reva/gen/gen.go
+++ b/cmd/reva/gen/gen.go
@@ -172,6 +172,12 @@ bigfilechunking = true
 blacklisted_files = ["foo"]
 undelete = true
 versioning = true
+[http.services.ocssvc.capabilities.capabilities.files.tus_support]
+version = "1.0.0"
+resumable = "1.0.0"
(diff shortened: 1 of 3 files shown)


PVince81:
fixed lint issue

when connecting through ocis-reva, the result looks like this:
```
...
        "files": {
          "privateLinks": false,
          "bigfilechunking": false,
          "undelete": true,
          "versioning": true,
          "blacklisted_files": [
            
          ],
          "tus_support": {
            "version": "1.0.0",
            "resumable": "1.0.0",
            "extension": "creation,creation-with-upload",
            "max_chunk_size": 0,
            "http_method_override": ""
          }
        },
...
```

PR #776, opened 2020-05-26
Add troubleshooting information

LovisaLugnegard:
Added some error logging and trouble shooting information, hopefully this won't be needed but if someone encounters this problem I hope this info saves some time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ See [https://hub.docker.com/r/cs3org/reva](https://hub.docker.com/r/cs3org/reva)
 
 Checkout the [Tutorials](https://reva.link/docs/tutorials/).
 
+## Troubleshooting
+If you see the error message "error querying database" from db.go try to remove the temporary db and restart reva, `rm /var/tmp/reva/localfs.db`.
+`rm /var/tmp/reva/` will delete everything related to reva if you need to do a clean restart.
+
 ## License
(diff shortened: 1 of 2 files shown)


ishank011:
Instead of logging here, can you modify normalize to return both `*provider.ResourceInfo` and an error, and if the error is not nil, return the error from where it is called? I try to prevent adding any logging in the packages and instead pass them to the GRPC service.

LovisaLugnegard:
Sure!

ishank011:
They already do. normalize is an internal method which doesn't. All the heavy operations are carried out here and logging with a context is quite expensive as well.

ishank011:
Also, the documentation doesn't belong in readme. You can add it to one of the tutorials.

labkode:
@LovisaLugnegard more specifically, this should be documentation for the driver causing this problem.

ishank011:
Okay yes, maybe we can add it under https://reva.link/docs/concepts/storages/#local-storage-driver?

LovisaLugnegard:
How about adding a "troubleshooting" folder under documentation? 

@labkode "this should be documentation for the driver causing this problem." - I'm not sure I understand what you mean with this?

butonic:
@LovisaLugnegard can you rebase please?

PR #681, opened 2020-04-23
Public Shares CRUD, File Public Shares Manager

refs:
## What?

- [x] Manager interface accepts UpdateRequestType
    - [x] since this is the only place where the update data exists
- [x] Update `displayname`
- [x] Update `permissions`
- [x] Update `expiration`
- [x] Update `password`
- [ ] Delete

### Other Tasks
- [x] Do not return expired shares to Phoenix.
- [x] Persistent shares manager.

## Description

The file manager, as per commit `8503ee3` needs an existing file to be used as storage, as the revad process doesn't have write permissions(1).

The db file is thread safe, all operations lock a mutex, so goroutines consuming the file are sure to be using an up-to-date version.

Example file contents:

```json
{
  "KTBfUQmYrEXy": "{\"id\":{\"opaqueId\":\"KTBfUQmYrEXy\"},\"token\":\"VjldgRWfpudX\",\"resourceId\":{\"storageId\":\"1284d238-aa92-42ce-bdc4-0b0000009162\",\"opaqueId\":\"19897125-ddd9-4939-be1d-12d86c503569\"},\"permissions\":{\"permissions\":{\"getPath\":true,\"getQuota\":true,\"initiateFileDownload\":true,\"listGrants\":true,\"listContainer\":true,\"listFileVersions\":true,\"listRecycle\":true,\"stat\":true}},\"owner\":{\"opaqueId\":\"einstein\"},\"creator\":{\"idp\":\"https://localhost:9200\",\"opaqueId\":\"einstein\"},\"ctime\":{\"seconds\":\"1587720640\",\"nanos\":939334000},\"mtime\":{\"seconds\":\"1587720755\",\"nanos\":187107000},\"expiration\":{\"seconds\":\"1587859200\"},\"displayName\":\"empty\"}",
[...]

PR #779, opened 2020-05-27
Separate PROPFIND logic for Depth 1 and infinity

PVince81:
Switching to path concatenation broke some cross-storage cases.
Since clients currently only use Depth 1, here we make sure that at
least this works.
For the infinity case, it is limited to single storage recursion for
now.

This is a workaround to unblock the public shares PR by removing the regression from view: https://github.com/cs3org/reva/pull/681

I've tested the propfinds locally and they work with all depths, but I only tested single-storage as I don't have the env for cross-storage. Based on the knowledge in https://github.com/cs3org/reva/pull/681#issuecomment-634535196 we know that the depth-1 logic does work there.

@refs


PVince81:
Important note: I chose to kept some Depth:infinity logic in place because we already have automated tests that rely on it to be able to verify results for unrelated scenarios. I don't want to disable those tests or rewrite them as it would be even more work.

PR #780, opened 2020-05-27
add security access headers for ocdav requests

karakayasemi:
Adds necessary security headers for ocdav requests. 

For owncloud/ocis-reva#66

### DIFF
--- a/internal/http/services/owncloud/ocdav/get.go
+++ b/internal/http/services/owncloud/ocdav/get.go
@@ -119,6 +119,8 @@ func (s *svc) handleGet(w http.ResponseWriter, r *http.Request, ns string) {
 	}
 
 	w.Header().Set("Content-Type", info.MimeType)
+	w.Header().Set("Content-Disposition", "attachment; filename*=UTF-8''"+
+		path.Base(info.Path)+"; filename=\""+path.Base(info.Path)+"\"")
 	w.Header().Set("ETag", info.Etag)
 	w.Header().Set("OC-FileId", wrapResourceID(info.Id))
 	w.Header().Set("OC-ETag", info.Etag)
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -122,8 +122,7 @@ func (s *svc) Handler() http.Handler {
 		ctx := r.Context()
 		log := appctx.GetLogger(ctx)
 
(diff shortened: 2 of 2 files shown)


internal/http/services/owncloud/ocdav/get.go
```diff
@@ -134,3 +138,8 @@ func (s *svc) handleGet(w http.ResponseWriter, r *http.Request, ns string) {
 		log.Error().Err(err).Msg("error finishing copying data to response")
 	}
 }
+
+// Rawurlencode https://www.php2golang.com/method/function.rawurlencode.html
+func Rawurlencode(str string) string {
```
labkode, line 143: Why the url.QueryEspace is not enough?

karakayasemi: oc10 encodes filename in `Content-Disposition` header with PHP's `rawurlencode` function. The difference between `rawurlencode` and `url.QueryEscape` is `url.QueryEscape` encodes spaces as `+` instead of `%20`. I added this function to ensure consistency with oc10.

I found `url.PathEscape` is also working in same way. So, now `url.PathEscape` has been used instead of this function with the last commit.

labkode:
Hi @karakayasemi ,

take a look at the following page for default security configurations:
https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.4

If we are providing some defaults, let's make sure they are aligned with the propositions above.

internal/http/services/owncloud/ocdav/get.go
labkode: @karakayasemi in that case, as youre' doing path.Base there no need to call `url.PathEscape` as there isn't a path. Base gives you the basename of the path.

karakayasemi:
> 
> 
> Hi @karakayasemi ,
> 
> take a look at the following page for default security configurations:
> https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.4
> 
> If we are providing some defaults, let's make sure they are aligned with the propositions above.

The only conflicting part that I see is the age of `Strict-Transport-Security`. I aligned it according to the example config.

internal/http/services/owncloud/ocdav/get.go
karakayasemi: Yes, you are right. Redundant `url.PathEscape` has been removed.

PR #784, opened 2020-05-28
Fix Link roles and permissions

refs:
This is an expedited PR to fix an issue on roles and permissions on public shares caused by a refactor on `shares.go`.

It also enables password update on the public link json manager.

For more context on the issue see https://github.com/owncloud/ocis-reva/pull/232

PR #788, opened 2020-05-29
Fix Link roles and permissions regression (second edition)

PVince81:
Resend of https://github.com/cs3org/reva/pull/784 + extra fix

This is an expedited PR to fix an issue on roles and permissions on public shares caused by a refactor on shares.go.

It also enables password update on the public link json manager.

This makes the CI tests in owncloud/ocis-reva#232 pass, the ones which were failing due to regression.


PVince81:
linted is happy. @labkode please review :-)

PR #791, opened 2020-06-02
Add option to disable TUS on OC layer

PVince81:
This way the TUS headers and capabilities are not returned when
disabled, to not advertise TUS support for clients.

Furthermore the TUS handlers are also disabled.

@butonic 

Context is in https://github.com/owncloud/ocis-reva/pull/228 and discussion in https://github.com/owncloud/ocis-reva/issues/177#issuecomment-635237630


PVince81:
seems the linter has timed out, weird

PR #792, opened 2020-06-02
[TUS] Dataprovider now support method override

PVince81:
The TUS protocol requires the ability to override the PATCH and DELETE
methods using the X-HTTP-Method-Override.

These were added in the data provider.


PVince81:
something is wrong with the linter, this is the second PR that it says timeout @labkode

labkode:
@PVince81 fixed the linter increasing the timeout

PR #797, opened 2020-06-03
Attempt to return metadata after a TUS PATCH

PVince81:
- [ ] PROBLEM: how to get at the actual file name, need to read the upload info file
- [ ] PROBLEM: cannot return headers as tusd is already sending the response

### DIFF
--- a/internal/http/services/dataprovider/dataprovider.go
+++ b/internal/http/services/dataprovider/dataprovider.go
@@ -52,6 +52,41 @@ type svc struct {
 	storage storage.FS
 }
 
+type WrappedTusHandler struct {
+	tusd.UnroutedHandler
+}
+
+func (handler *WrappedTusHandler) sendResp(w http.ResponseWriter, r *http.Request, status int) {
+	// TODO: inject extra response headers
+	w.Header().Add("X-Override-Hack", "WrappedTusHandler")
+	w.WriteHeader(status)
+
+	// FIXME: how to call parent log ?
+	//handler.UnroutedHandler.log("ResponseOutgoing", "status", strconv.Itoa(status), "method", r.Method, "path", r.URL.Path, "requestId", getRequestId(r))
+}
+
+type WrappedResponseWriter struct {
+	http.ResponseWriter
+	statusCode int
+}
+
+func (w *WrappedResponseWriter) WriteHeader(statusCode int) {
+	w.statusCode = statusCode
+	// delay this
+}
+
+func (w *WrappedResponseWriter) SendResponse() {
+	w.ResponseWriter.WriteHeader(w.statusCode)
+}
+
+func (w *WrappedResponseWriter) Header() http.Header {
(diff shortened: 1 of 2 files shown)


butonic:
This won't work because the writeHeader method flushes the cache. But we could wrap the `ResponseWriter w` which is passed in the `hander.PatchFile(w, r)` call with our [own implementation](https://golang.org/src/net/http/server.go?s=3003:5866#L85) that defers the WriteHeader call. Then we can add arbitrary headers when `hander.PatchFile(w, r)` returns and then call a `FlushResponse()` functon ...

PVince81:
I've pushed an attempt to wrap the response.

It works and I'm able to add headers after the TUS handler now.

The next challenge is finding the path to pass to the CS3 API's `GetMD`.
For this I capture the upload info from TUS' "CompleteUpload" event using a Go-routine, the approach looks really ugly.

And then I get a structure like this:
```
uploadStorage={"BinPath":"/var/tmp/reva/data/einstein/uploads/8f5ed05f-c94c-49b9-b14f-d9d3b70ae886","Idp":"https://localhost:9200","InternalDestination":"/var/tmp/reva/data/einstein/files/tus2/md5.txt","LogLevel":"trace","Type":"OwnCloudStore","UserId":"einstein","UserName":"einstein"}
```

but I haven't figured out how to resolve the correct path reference to pass the the CS3 API.
The call to `TrimPrefix` doesn't work here: https://github.com/cs3org/reva/pull/797/files#diff-d92e2cb18e0f9290c492defb7bb389abR143

@butonic any hints ?

labkode:
@PVince81 @butonic can we have a quick meeting to discuss this workflow?

PVince81:
@labkode sounds good.

Note that this is mostly driven by client development, as the old protocol had these fields returned at the end of the upload to avoid having to do subsequent calls (which are tricky for example on iOS)

labkode:
@PVince81 @butonic 

> The next challenge is finding the path to pass to the CS3 API's `GetMD`.
> For this I capture the upload info from TUS' "CompleteUpload" event using a Go-routine, the approach looks really ugly.


I've been through TUS and started refactoring code in https://github.com/cs3org/reva/pull/828.

Upon this task I've found a fundamental problem on the various implementations of TUS for the storages. By extending the interface with the `tusd.DataStore` you are cheating and bypassing any path based operation on this upload, reason you can't get the metadata for it because the internal path is not exposed to the storage provider. 

For example, if your filesystem exposes to the storage provider the path `/tmp/revad/data` but you're storing your uploads under `/tmp/revad/uploads`, you won't be able to call GetMD because the request will be relative to `/tmp/revad/data` and not to `/tmp/revad/uploads`.


So, the current implementations need to store the upload information under the configured root, that will allow to call GetMD and avoid that hack for example.

Passing internal implementation details to upper layers is a no-go.

butonic:
won't be continued here

PR #798, opened 2020-06-03
OCS fixes for create public link

PVince81:
- Return 404 if the target of createPublicLinkShare does not exist.
- Added default permission in case none was passed in the requested
- Fixed name field handling to get it from the share object and include it in the response even when empty (to be compliant with the old behavior)
- Fixed owner user id in response to not include IDP/domain (to be compliant with old behavior) https://github.com/owncloud/ocis-reva/issues/245

@refs FYI

### DIFF
--- a/internal/http/services/owncloud/ocs/conversions/main.go
+++ b/internal/http/services/owncloud/ocs/conversions/main.go
@@ -130,7 +130,7 @@ type ShareData struct {
 	// Whether the recipient was notified, by mail, about the share being shared with them.
 	MailSend string `json:"mail_send" xml:"mail_send"`
 	// Name of the public share
-	Name string `json:"name,omitempty" xml:"name,omitempty"`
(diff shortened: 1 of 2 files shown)


PVince81:
Discovered that "not found" issue when running the tests tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature:212 but omitting the SKELETON_DIR path, so no file was present for share creation.

For the perm issue, once the file exists but no permission field is given in tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature:212 the "Atoi" would fail due to empty string.

With these fixes the test tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature:212 goes further and bumps against further issues to be fixed separately.

PVince81:
fixed the lint issue

I've added a bunch of additional fixes, see top post

PR #805, opened 2020-06-05
OCDAV: fix returned timestamp format

PVince81:
RFC1123 actually recommend the use of "UT" as zimezone instead of "UTC".
The way to do it with go is using the special "RFC1123Z" time.

Fixes issue with timestamp warnings on the client and makes the format
compliant with the old implementation.

See https://github.com/golang/go/issues/13781#issuecomment-168235353

Fixes https://github.com/owncloud/ocis-reva/issues/116

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -320,7 +320,7 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 		}
 		// Finder needs the the getLastModified property to work.
 		t := utils.TSToTime(md.Mtime).UTC()
-		lastModifiedString := t.Format(time.RFC1123)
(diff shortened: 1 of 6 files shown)


internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -269,7 +269,8 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 	}
 
 	response := responseXML{
-		Href:     (&url.URL{Path: ref}).EscapedPath(), // url encode response.Href
+		//Href:     (&url.URL{Path: ref}).EscapedPath(), // url encode response.Href
+		Href:     strings.ReplaceAll(url.QueryEscape(ref), "%2F", "/"),
```
PVince81, line 273: uh oh, old unrelated change

PR #812, opened 2020-06-09
Fixed label naming

Daniel-WWU-IT:
Prometheus doesn't allow - in label names, so this fix replaces "service-type" by "service_type"


Daniel-WWU-IT:
There will be a few more changes, so I'm closing this one

PR #829, opened 2020-06-11
Separate local fs into home and with home disabled

ishank011:


pkg/storage/utils/localfs/localfs_windows.go
```diff
@@ -18,7 +18,7 @@
 
 // +build windows
```
refs, line 19: the build tad is redundant, as your filename contains `_windows` on it

ishank011: @refs Makes sense. I'll remove the tags then. Thanks!

PR #830, opened 2020-06-11
Refactor EOS fs

ishank011:


labkode:
@ishank011 ready for review?

ishank011:
@labkode Yep.

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -0,0 +1,1341 @@
...
+)
+
+const (
+	refTargetAttrKey = "reva.target"
+)
+
+var hiddenReg = regexp.MustCompile(`\.sys\..#.`)
+
+// Config holds the configuration details for the EOS fs.
+type Config struct {
```
labkode, line 57: Can this be somehow re-used?

ishank011: Yes, we can use only a single config in utils/eosfs/eosfs.go, but I wanted to keep it separately in eos and eoshome so that it can be picked up by cato. Since both of them have a few differences in their configs, using a common config will not be reflected in the documentation.

labkode: Okay

PR #833, opened 2020-06-12
Remove redundant build tags from local and owncloud fs

ishank011:


PVince81:
it seems that it's causing Windows builds to fail now: https://cloud.drone.io/owncloud/ocis-reva/492/8/3

not sure why... Golang is supposed to recognize those files based on the underscore but is still reading both versions of the file, causing duplicate symbol issues.

can we revert for now ?

PR #841, opened 2020-06-16
Propagate upload mtime to the storage

PVince81:
The value of X-OC-Mtime is now passed to InitiateFileUpload using an OpaqueEntry.

The storage FS interface's InitiateUpload was updated to include a new
metadata argument, and adjusted all known implementors.

This new metadata argument is used in the storage provider to pass in the
mtime to the storage implementations.

The ownCloud storage implementation was adjusted to read the mtime from the
metadata and set it using Chtimes. That logic was extracted in a new
private method to avoid gymnastics with path handling and also avoid the
implicit propagation that exists in SetArbitraryMetadata.

The other implementors like EOS and Local are currently ignoring the
mtime value and need to be adjusted later on.

@butonic FYI, I hope this approach makes sense. I had to extend the FS interface.

### DIFF
--- a/pkg/storage/fs/owncloud/upload.go
+++ b/pkg/storage/fs/owncloud/upload.go
@@ -81,7 +81,7 @@ func (fs *ocfs) Upload(ctx context.Context, ref *provider.Reference, r io.ReadCl
 
 // InitiateUpload returns an upload id that can be used for uploads with tus
 // TODO read optional content for small files in this request
-func (fs *ocfs) InitiateUpload(ctx context.Context, ref *provider.Reference, uploadLength int64) (uploadID string, err error) {
+func (fs *ocfs) InitiateUpload(ctx context.Context, ref *provider.Reference, uploadLength int64, metadata map[string]string) (uploadID string, err error) {
 	np, err := fs.resolve(ctx, ref)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


pkg/storage/fs/owncloud/upload.go
```diff
@@ -81,7 +81,7 @@ func (fs *ocfs) Upload(ctx context.Context, ref *provider.Reference, r io.ReadCl
 
 // InitiateUpload returns an upload id that can be used for uploads with tus
 // TODO read optional content for small files in this request
-func (fs *ocfs) InitiateUpload(ctx context.Context, ref *provider.Reference, uploadLength int64) (uploadID string, err error) {
+func (fs *ocfs) InitiateUpload(ctx context.Context, ref *provider.Reference, uploadLength int64, metadata map[string]string) (uploadID string, err error) {
```
refs, line 84: Is it intended a `map[string]string`? I think having `map[string]interface{}` is more flexible

PVince81: This is by intention to be aligned with the ArbitraryMetadata approach which has string/string.
At some point we can allow more metadata like checksum and also arbitrary webdav properties

PVince81:
@labkode please review

PR #826, opened 2020-06-10
deployment: simplify config

labkode:
This PR allows to run Reva with minimal configuration, easier for dev setups and IOP deployments:

```
$ cat examples/standalone/standalone.toml
# services to enable
[grpc.services.gateway]
[grpc.services.storageregistry]
[grpc.services.storageprovider]
[grpc.services.authprovider]
[grpc.services.authregistry]
[grpc.services.userprovider]
[grpc.services.usershareprovider]
[grpc.services.publicshareprovider]

[http.services.dataprovider]
[http.services.prometheus]

```

To test, will require to have pre-created the following file: `/var/tmp/reva/users.json`

PR #848, opened 2020-06-17
Fix Gateway config defaults

refs:
This code short circuits cases when the configuration is valid, and returns the Gateway address in many occasions, ending up in nasty `Unimplemented` calls to grpc services because the address is incorrect.


PVince81:
@labkode can you review ? this is a regression introduced by the config simplification https://github.com/cs3org/reva/pull/826

PR #852, opened 2020-06-18
Properly read permissions when creating public link

PVince81:
There was a bug where the permKey variable is redefined inside a block
due to the extra colon, so the value did not get set to the outer
variable which got stuck on the value 1.

@refs FYI


refs:
nice catch :D

PR #862, opened 2020-06-22
Revert "Remove redundant build tags (#833)"

PVince81:
This reverts commit 53c922ce9bc7a92a2d90c9bec8e1ed07a2fd23a6.

See https://github.com/cs3org/reva/pull/833#issuecomment-647361362

Should fix compilation issues on Windows in ocis-reva

@labkode @ishank011

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud_unix.go
+++ b/pkg/storage/fs/owncloud/owncloud_unix.go
@@ -16,6 +16,8 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
+// +build !windows
+
 package owncloud
 
 import (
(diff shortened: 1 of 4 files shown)


PVince81:
I'm not sure if this is sufficient, https://github.com/cs3org/reva/pull/829 might also contain changes that need to be looked at for a better separation of platforms

ishank011:
@PVince81 #829 just adds one more layer of abstraction at the fs level. For platform-wise functionality, no changes have been made there. It's just weird in this case that the platform-specific files were not identified.

PVince81:
Fantastic, it builds again on Windows: https://cloud.drone.io/owncloud/ocis-reva/493/8/3

Yeah, weird that it doesn't work. Will need further research to be able to remove those tags.

PR #871, opened 2020-06-23
downgrade grpc to 1.26.0

C0rby:
In ocis we currently face the problem that some of our dependencies rely on
grpc 1.26.0. With the upgraded grpc dependency in reva we were unable to
build ocis. And if we upgrade grpc in ocis the build failed because of
the other dependency.

I hoped we could downgrade grpc in reva until we have found a better
long term solution.

Here are some examples of our attempts to resolve this without this
downgrade. https://github.com/owncloud/ocis/pull/328#issuecomment-647455011


labkode:
@C0rby temporary downgrading, please tackle this issue also on your repo.

ishank011:
@C0rby any update on when we can revert this downgrade?

C0rby:
@ishank011, sadly one library we are using is still depending on that downgrade.
I'll try and see how we can maybe change that.

PR #875, opened 2020-06-24
Added support for differing metrics path for Prometheus to Mentix

Daniel-WWU-IT:
This PR adds support for differing metrics path when exporting scrape targets to Prometheus via Mentix. In GOCDB, a property called METRICS_PATH can now be added to override the default /metrics endpoint. This allows us to pull metrics from targets like /api/metrics etc.


labkode:
@Daniel-WWU-IT some hound lint comments above.

Daniel-WWU-IT:
It's just one actually, but it made it 3 times :D It's really just a super-unimportant comment missing, so I ignored that.

PR #878, opened 2020-06-24
TUS upload support through datagateway

kulmann:
In oCIS we want to be able to upload data through one single entrypoint and let the datagateway take care of dispatching the request to the respective ~~storage~~ data server. Reva already supports this by encoding the upload endpoint in an `X-Reva-Transfer` token.
This PR is about four things:
1. TUS clients don't understand the reva transfer token. Instead they expect to just work with a URL. In order to fill this gap, the tus code in ocdav now appends the token to the upload endpoint when the ~~storages~~ data servers are not exposed (i.e. as soon as a transfer token exists). Of course we want reva to still continue to work with the token. To achieve this, the datagateway now checks if a transfer token is present in the token verification step. If it's not present, we use the last segment of the request path as the transfer token.
2. Bring basic PATCH request support to the datagateway.
3. TUS code in ocdav didn't set the reva transfer token on the header in cases where it can actually be utilized. This is fixed in `put.go`.
4. Remove the hardcoded `ExposeDataServer = true`.

Also one small change: I've changed the datagateway to expose the `X-Reva-Transfer` constant and used it in all places where the string was hardcoded before. Makes it easier to find code fragments where the transfer token is used.

Happy to hear feedback on this.

### DIFF
--- a/internal/http/services/datagateway/datagateway.go
+++ b/internal/http/services/datagateway/datagateway.go
@@ -36,7 +36,8 @@ import (
 )
 
 const (
-	tokenTransportHeader = "X-Reva-Transfer"
+	// TokenTransportHeader holds the header key for the reva transfer token
+	TokenTransportHeader = "X-Reva-Transfer"
 )
 
 func init() {
@@ -145,7 +146,7 @@ func (s *svc) doGet(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	log := appctx.GetLogger(ctx)
 
-	token := r.Header.Get(tokenTransportHeader)
+	token := r.Header.Get(TokenTransportHeader)
 	claims, err := s.verify(ctx, token)
 	if err != nil {
 		err = errors.Wrap(err, "datagateway: error validating transfer token")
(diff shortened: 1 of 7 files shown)


kulmann:
cc @PVince81 @butonic

labkode:
@kulmann nice work. Happy to merge after approval also from @PVince81 or @butonic.

PVince81:
@kulmann how was this tested? please list scemarios (ex phoenix, cli, etc)

kulmann:
> @kulmann how was this tested? please list scenarios (ex phoenix, cli, etc)

Together with this [PR in ocis-reva](https://github.com/owncloud/ocis-reva/pull/261) which changes default configs and [PR in ocis-proxy](https://github.com/owncloud/ocis-proxy/pull/45) which shows the required `/data` endpoint in the example config, I made successful uploads:
1. with regular file upload through ocis-web (former phoenix)
2. with curl: `curl -X PUT https://localhost:9200/remote.php/dav/files/einstein/test/newfile1.txt -u einstein:relativity -d 'Hello World!' -k -v`

internal/http/services/datagateway/datagateway.go
```diff
@@ -238,3 +248,69 @@ func (s *svc) doPut(w http.ResponseWriter, r *http.Request) {
 		log.Err(err).Msg("error writing body after header were set")
 	}
 }
+
+// TODO: put and post code is pretty much the same. Should be solved in a nicer way in the long run.
```
butonic, line 252: @labkode I was pairing on this with @kulmann and while the patch works we should use the simplehttpproxy from the stdlib here ... @labkode we can look into that after this pr.

PVince81:
Ok, I've tested TUS upload with Phoenix and it worked fine :+1:

butonic:
rebasing fixde config issues. good to merge!

PR #880, opened 2020-06-25
make httpclient trust local certificates for now

butonic:
allows the datagateway to work with local deployments, see https://github.com/cs3org/reva/pull/878#pullrequestreview-437276316


kulmann:
LGTM 👍

PR #882, opened 2020-06-25
Fix up download for public link shares

butonic:
the publicstorageprovider uses the gateway to initiate up and downloads. it needs to append the token to the url because it cannot return it as a token in the response


kulmann:
LGTM

labkode:
@butonic why the dataprovider need to be exposed? The idea is that data operations always goes via datagateway.

kulmann:
> @butonic why the dataprovider need to be exposed? The idea is that data operations always goes via datagateway.

@labkode The public storage provider will always ask the gateway for an upload endpoint from a "real" storage provider. That will already respect it's own respective `ExposeDataServer` configuration. So in the public storage provider we can always expose what we got from the "real" storage provider.

labkode:
@kulmann I see now. Good to merge.

PR #890, opened 2020-06-26
Delete public share

C0rby:
I extracted the read and write db methods into methods to reduce duplication and implemented the RevokePublicShareMethod.
@PVince81, can you have a look. Did I miss something for deleting?

Implements delete for https://github.com/owncloud/ocis-reva/issues/293.

### DIFF
--- a/pkg/publicshare/manager/json/json.go
+++ b/pkg/publicshare/manager/json/json.go
@@ -345,7 +345,17 @@ func (m *manager) ListPublicShares(ctx context.Context, u *user.User, filters []
 
 // RevokePublicShare undocumented.
 func (m *manager) RevokePublicShare(ctx context.Context, u *user.User, id string) error {
-	return fmt.Errorf("RevokePublicShare method unimplemented")
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	db, err := m.readDb()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


PVince81:
is this working already ?
I thought we might need to make changes in shares.go as I saw separate implementation for user shares and public share.
If that works already, nice.

Can you enable the matching API tests and verify that they pass ?

C0rby:
I'll try the acceptance tests.

I didn't try my implementation yet to be honest.

refs:
> I'll try the acceptance tests.
> 
> I didn't try my implementation yet to be honest.

The implementation is 👍 , but curious to see oc10 end to end tests, since deleting a share actually works as in: the vuejs component deletes the element from the view, but the request essentially fail.

We can try this tomorrow @C0rby

C0rby:
Ok so there was some missing stuff. Now I implemented everything to make public link shares work and ran the acceptance tests.
All but one run successfully. The one that fails is about deleting a share of a file, which was renamed. I'll create a ticket for that.

C0rby:
Ok so there was some missing stuff. Now I implemented everything to make public link shares work and ran the acceptance tests.
All but one run successfully. The one that fails is about deleting a share of a file, which was renamed. I'll create a ticket for that.

C0rby:
This PR won't update for some reason.
Will close and reopen

PR #895, opened 2020-06-26
Fix public link file share

PVince81:
Provide virtual container on Webdav layer for the public link share
for a single file

- [x] propfind on container
- [x] propfind on file

- [x] options/head on container
- [x] options/head/get on file directly
- [x] return metadata like file size, etag, mime type
- [x] put on file directly replaces contents

- [x] check with API tests locally

- [x] manual test in Phoenix
    - [x] missing properties: fileid, id, mimetype
    - file entry should not have oc:size but only getcontentlength => unrelated, separate issue as it happens on regular dav

@butonic FYI

### DIFF
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -20,11 +20,15 @@ package ocdav
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"path"
 
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/pkg/rhttp/router"
 	tokenpkg "github.com/cs3org/reva/pkg/token"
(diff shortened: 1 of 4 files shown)


PVince81:
hmm Phoenix is not displaying anything in the folder despite the correct response.

PVince81:
would be good to know if the general approach is fine

I'll fix the bugs next week

PVince81:
ok, I've added the file id and now it works in Phoenix as well.

@refs @butonic please review

internal/http/services/owncloud/ocdav/publicfile.go
```diff
@@ -0,0 +1,219 @@
...
+			case http.MethodHead:
+				s.handleHead(w, r, h.namespace)
+			default:
+				w.WriteHeader(http.StatusMethodNotAllowed)
+			}
+		}
+	})
+}
+
+func (s *svc) adjustResourcePathInUrl(w http.ResponseWriter, r *http.Request) bool {
```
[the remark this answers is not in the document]
PVince81: ohhhh....

internal/http/services/owncloud/ocdav/publicfile.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// PublicFileHandler handles trashbin requests
+type PublicFileHandler struct {
+	namespace string
+}
+
+const (
+	tokenStatInfoKey int = iota
+)
```
refs, line 37: All this block can be turned into:

```go
type tokenStatInfoKey struct{}
```

and its usage:

```go
tokenStatInfo := ctx.Value(tokenStatInfoKey{}).(*provider.ResourceInfo)
```

Since structures are comparable, we move away from global, since packages can declare their own types.

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -20,25 +20,34 @@ package ocdav
...
+	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/pkg/rhttp/router"
 	tokenpkg "github.com/cs3org/reva/pkg/token"
 	"github.com/cs3org/reva/pkg/user"
 	"google.golang.org/grpc/metadata"
 )
 
+const (
+	tokenStatInfo int = iota
```
refs, line 40: You're not using `tokenStatInfo` anywhere in the package, perhaps a leftover from `tokenStatInfoKey`?

PVince81:
@refs adjusted, thanks for your hints

PR #900, opened 2020-06-29
forward HEAD requests to data provider to get TUS upload offset

butonic:
fixes uploads to the root folder of a user

### DIFF
--- a/internal/http/services/datagateway/datagateway.go
+++ b/internal/http/services/datagateway/datagateway.go
@@ -165,7 +165,7 @@ func (s *svc) doHead(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	log.Info().Str("target", claims.Target).Msg("sending request to internal data server")
+	log.Debug().Str("target", claims.Target).Msg("sending request to internal data server")
 
 	httpClient := rhttp.GetHTTPClient(ctx)
 	httpReq, err := rhttp.NewRequest(ctx, "HEAD", claims.Target, nil)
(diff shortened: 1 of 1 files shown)


internal/http/services/datagateway/datagateway.go
```diff
@@ -153,6 +153,44 @@ func (s *svc) verify(ctx context.Context, r *http.Request) (*transferClaims, err
...
+
+	claims, err := s.verify(ctx, r)
+	if err != nil {
+		err = errors.Wrap(err, "datagateway: error validating transfer token")
+		log.Err(err).Str("token", r.Header.Get(TokenTransportHeader)).Msg("invalid transfer token")
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
+
+	log.Info().Str("target", claims.Target).Msg("sending request to internal data server")
```
refs, line 168: debug? Log lines are valuable real state 🙄

internal/http/services/datagateway/datagateway.go
```diff
@@ -153,6 +153,44 @@ func (s *svc) verify(ctx context.Context, r *http.Request) (*transferClaims, err
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if httpRes.StatusCode != http.StatusOK {
+		w.WriteHeader(httpRes.StatusCode)
+		return
+	}
+
+	defer httpRes.Body.Close()
```
refs, line 190: I would add this line below line 178, otherwise this won't run on an error branch

PR #905, opened 2020-06-29
Add logic to ls and stat to process arbitrary metadata keys

### LINKED ISSUE #806: Modify stat to take metadata fields to be read as parameters

ishank011:
Closes #806


ishank011:
Thanks for the review @labkode! Made the change.

Issue #870, opened 2020-06-23
Make meshdirectory and mentix services use cs3apis


ishank011:
Hi @mirekys, @Daniel-WWU-IT 

As mentioned yesterday in the weekly meeting, we need to make calls to the meshdirectory service pass through the GRPC layer. Even though we'd ultimately be making a call to a gocdb server, but for other local-database-reliant drivers, in the current implementation, we just assume that the db would be present on the same machine on which the HTTP service is hosted, which won't always be the case.

I've modified the OCM provider CS3APIs definition to include all the fields which gocdb provides [here](https://cs3org.github.io/cs3apis/#cs3%2focm%2fprovider%2fv1beta1%2fresources.proto), and modified the reva methods to use these new definitions, the PR for which is [here](https://github.com/cs3org/reva/pull/864).

So I think we can carry out the following steps:
- Since the meshdirectory package's manager interface only has the `GetMeshProviders` method, which is also implemented by the OCM Provider package, we can make calls to the ocmprovider GRPC service as done [here](https://github.com/cs3org/reva/blob/master/internal/http/services/ocmd/invites.go#L120) from the meshdirectory HTTP service and move the mentix driver from meshdirectory package to ocm/provider package. Right now, we just have two packages which perform the same tasks.
- Gocdb has multiple definitions of the service and site structures, so we should use the CS3API definitions mentioned above.

Please let me know your thoughts about these.
CC: @labkode

mirekys:
Thanks @ishank011, I completely agree that we should reuse what is already in OCM Provider and move Mentix driver from meshdirectory there. It is a good idea to have one common provider site structure both for GOCDB and meshdirectory. I can prepare a pull request for the mesh directory's part.

ishank011:
Sounds good @mirekys. Thanks!

PR #906, opened 2020-06-29
meshdirectory: refactor to use ocm provider api

mirekys:
This moves most of the OCM provider functionality out of the `meshdirectory` package:

- Mesh Directory now leverages the `/cs3.ocm.provider.v1beta1.ProviderAPI/ListAllProviders` GRPC endpoint, that is already implemented by the `ocmproviderauthorizer` package
- Mentix driver, used for fetching OCM providers data from a Mentix exporter, has been moved to the `ocmproviderauthorizer` package
- Mesh Directory frontend now expects OCM provider data in the [ProviderInfo](https://github.com/cs3org/cs3apis/blob/master/cs3/ocm/provider/v1beta1/resources.proto#L77) struct format
- Mesh Directory package has been removed, only http service remains

Related to #870


labkode:
@mirekys any update on this work?

mirekys:
@labkode this is waiting for the Mentix exporter/GOCDB api to also use CS3 APIs structs (see here #870). The following code could then be removed from this pull, before it could be merged: https://github.com/cs3org/reva/pull/906/files#diff-ffed34085b7ccf28226b3573342026c2R115

ishank011:
@Daniel-WWU-IT can you take a look? It'll be good to have all the services use a common mesh provider definition.

Daniel-WWU-IT:
Just for my understanding: Mentix should export its provider data structured like the CS3API defines them? This surely can be done,  and makes a lot of sense. What you currently get from the /mentix endpoint is the "plain" data as it is stored internally. It's easy to extend that functionality and have it export the data in the CS3API structure format. From looking at their definitions, they should fit quite nicely, no need to extend them (at least for now).

Please confirm that I got this right and I'll work on it right away :)

ishank011:
@Daniel-WWU-IT I think that should be it!

mirekys:
@Daniel-WWU-IT yep, it should return a [ListAllProvidersResponse](https://github.com/cs3org/cs3apis/blob/master/cs3/ocm/provider/v1beta1/provider_api.proto#L106) formatted response, to be consistent with the OCM Provider api. You can have a look at the diff i referenced few comments ago to see, what the differences between that and current Mentix response are (not a ton of differences there :)).

pkg/ocm/provider/authorizer/mentix/mentix.go
```diff
@@ -0,0 +1,229 @@
...
+			Description:  si.Description,
+			Organization: si.Organization,
+			Domain:       si.Domain,
+			Homepage:     si.Homepage,
+			Email:        si.Email,
+			Services:     services,
+		})
+	}
+
+	status := &rpcv1beta1.Status{
```
ishank011, line 146: You can just return providers here. We're not using status anywhere.

mirekys: True indeed, I will get rid of that

Daniel-WWU-IT:
Created PR #955 to include a new CS3API exporter for Mentix. The data it exposes is just an array of `ProviderInfo` so you should be able to drop the conversion from the meshdirectory service.

mirekys:
> @mirekys @ishank011 Created PR #955 to include a new CS3API exporter for Mentix. The data it exposes is just an array of `ProviderInfo` so you should be able to drop the conversion from the meshdirectory service.

thanks a lot @Daniel-WWU-IT, I'll update this PR right away

PR #914, opened 2020-06-30
Introduce Options for the httpclient

butonic:
Currently, slow up and downloads fail because the timeout is currently hardcoded to 10 sec.

This PR allows configuring the HTTPClient as needed. It also introduces `timeout` and `insecure` config options for the oc services. I tried not change the behavior or introduce config options when a config struct was not available, eg for the CLI up and download commands.

For mentix I updated the example config.

I also made the oidc client use the rhttp httpclient, so it should also send traces.


micbar:
@butonic Can you rebase?

PR #915, opened 2020-06-30
use mutex instead of sync.Map

refs:
As it turns out, `sync.Map` has a nasty behavior I haven't digged deep enough, and "stored" connections are closed:

```
2020-06-30T16:16:57+02:00 ERR Users/aunger/code/refs/reva/internal/grpc/services/gateway/storageprovider.go:839 > error finding storage provider error="gateway: error calling GetStorageProvider: rpc error: code = Unavailable desc = write tcp [::1]:52307->[::1]:9142: use of closed network connection" pkg=rgrpc service=reva traceid=8907d5dd4f27ec946b25ffdaaf8130ee
```

This implementation, however verbose, should address the previous concern and this new one.

### DIFF
--- a/pkg/rgrpc/todo/pool/pool.go
+++ b/pkg/rgrpc/todo/pool/pool.go
@@ -41,26 +41,40 @@ import (
 	"google.golang.org/grpc"
 )
 
-var (
-	publicShareProviders = sync.Map{}
-	preferencesProviders = sync.Map{}
-	userShareProviders   = sync.Map{}
-	storageRegistries    = sync.Map{}
-	gatewayProviders     = sync.Map{}
-	storageProviders     = sync.Map{}
-	authRegistries       = sync.Map{}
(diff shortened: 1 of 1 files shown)


pkg/rgrpc/todo/pool/pool.go
```diff
@@ -83,14 +100,18 @@ func GetGatewayServiceClient(endpoint string) (gateway.GatewayAPIClient, error)
 	}
 
 	v := gateway.NewGatewayAPIClient(conn)
-	gatewayProviders.Store(endpoint, v)
+	storageProviders.conn[endpoint] = v
```
ishank011, line 103: gatewayProviders

refs: oooooh, nice catch @ishank011 ! 🎉 good eye, thanks!

PR #909, opened 2020-06-30
Thread Safe Connection Pool

refs:
## What?

We encountered `concurrent map writes` runtime errors on the pool package. This solution should address this error and provide with a safe way to access these data structures. Since they are not exported and the `sync.Map` works with `interface{}`, a type assertion is needed, making it hard to refactor this code since generics are non-existant in go (at the moment of this writing), and reflection will definitely under perform on this package, which is sort of a bottleneck.

closes #908 

ping @butonic @individual-it

PR #916, opened 2020-06-30
Use gatewayProviders instead of storageProviders on conn pool

refs:


refs:
Tested this one manually. Works just fine. It also finds out I'm not such a good copy-paster, as in the previous PR using `sync.Map` I made 2 mistakes:

https://github.com/cs3org/reva/pull/909/files#diff-7bdd3c7d88443ed0c9f3b9440e15a33bR301-R310

and most critical:

https://github.com/cs3org/reva/pull/909/files#diff-7bdd3c7d88443ed0c9f3b9440e15a33bR268-R278

Now the error I was getting was not expressive at all:

```
2020-06-30T16:16:57+02:00 ERR Users/aunger/code/refs/reva/internal/grpc/services/gateway/storageprovider.go:839 > error finding storage provider error="gateway: error calling GetStorageProvider: rpc error: code = Unavailable desc = write tcp [::1]:52307->[::1]:9142: use of closed network connection" pkg=rgrpc service=reva traceid=8907d5dd4f27ec946b25ffdaaf8130ee
```

It proved pretty hard to debug, made me think it was coming from go `sync.Map` internals; so after some digging I decided to ditch the map implementation and use a mutex instead. We could keep this one, with the memory overhead but performance benefits instead of the `sync.Map` implementation.

Since this is a bottleneck package I would go for a more performant implementation, slightly 
 sacrificing readability.

PR #919, opened 2020-07-01
Remove array from OCS Share update response

PVince81:
The OCS share response after updating a public share must not contain an
XML "element" tag. To remove it, the fix removes the array around the
response.


PVince81:
note: I discovered this by comparing the response with the one from OC 10.
then verified using curl that the "element" tag is indeed gone after this fix.

API tests don't work yet as they have other issues, will be enabled later: https://github.com/owncloud/ocis-reva/issues/252#issuecomment-652302425

PR #920, opened 2020-07-01
WIP Add wopi open

LovisaLugnegard:
Adds a call to reva for opening a file in an app provider. The call returns the provider's url.
The cs3api needs to be updated, to override and use a local version, add `replace github.com/cs3org/go-cs3apis => ../cs3apis/build/go-cs3apis` in the bottom of `go.mod`.

### DIFF
--- a/cmd/reva/open-file-in-app-provider.go
+++ b/cmd/reva/open-file-in-app-provider.go
@@ -35,34 +35,24 @@ func openFileInAppProviderCommand() *command {
 	cmd.Usage = func() string {
 		return "Usage: open-file-in-app-provider [-flags] <path> <viewMode (view, read, write)>"
 	}
+	viewMode := cmd.String("viewMode", "view", "the view permissions, defaults to view")
 
 	cmd.Action = func() error {
 		ctx := getAuthContext()
-		if cmd.NArg() < 2 {
+		if cmd.NArg() < 1 {
 			fmt.Println(cmd.Usage())
 			os.Exit(1)
 		}
 		path := cmd.Args()[0]
 
-		viewMode := getViewMode(cmd.Args()[1])
+		viewMode := getViewMode(*viewMode)
 
 		client, err := getClient()
 		if err != nil {
 			return err
 		}
-		ref := &provider.Reference{
-			Spec: &provider.Reference_Path{Path: path},
-		}
-		reqA := &provider.StatRequest{Ref: ref}
-		resA, err := client.Stat(ctx, reqA)
-		if err != nil {
-			return err
-		}
-		if resA.Status.Code != rpc.Code_CODE_OK {
-			return formatError(resA.Status)
-		}
 
-		ref1 := &provider.Reference{
+		ref := &provider.Reference{
 			Spec: &provider.Reference_Path{Path: path},
 		}
 		accessToken, ok := tokenpkg.ContextGetToken(ctx)
@@ -71,7 +61,7 @@ func openFileInAppProviderCommand() *command {
 			return err
 		}
 
-		openRequest := &providerpb.OpenFileInAppProviderRequest{Ref: ref1, AccessToken: accessToken, ViewMode: viewMode}
+		openRequest := &providerpb.OpenFileInAppProviderRequest{Ref: ref, AccessToken: accessToken, ViewMode: viewMode}
 
 		openRes, err := client.OpenFileInAppProvider(ctx, openRequest)
 		if err != nil {
--- a/cmd/reva/open-file-in-app-provider.go
+++ b/cmd/reva/open-file-in-app-provider.go
@@ -1,0 +1,103 @@
+// Copyright 2018-2020 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	providerpb "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	tokenpkg "github.com/cs3org/reva/pkg/token"
+	"github.com/pkg/errors"
+)
+
+func openFileInAppProviderCommand() *command {
+	cmd := newCommand("open-file-in-app-provider")
+	cmd.Description = func() string { return "Open a file in an external app provider" }
+	cmd.Usage = func() string {
+		return "Usage: open-file-in-app-provider [-flags] <path> <viewMode (view, read, write)>"
+	}
+
+	cmd.Action = func() error {
+		ctx := getAuthContext()
+		if cmd.NArg() < 2 {
+			fmt.Println(cmd.Usage())
+			os.Exit(1)
+		}
+		path := cmd.Args()[0]
+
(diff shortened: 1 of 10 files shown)


pkg/rgrpc/todo/pool/pool.go
```diff
@@ -370,3 +370,22 @@ func GetOCMCoreClient(endpoint string) (ocmcore.OcmCoreAPIClient, error) {
 	ocmCores.conn[endpoint] = v
 	return v, nil
 }
+
+// GetOpenFileInAppProviderClient returns a new OpenFileInAppProviderClient.
```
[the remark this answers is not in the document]
LovisaLugnegard: It's just a leftover from some testing, thanks for finding this!

internal/grpc/services/appprovider/appprovider.go
```diff
@@ -93,18 +107,75 @@ func getProvider(c *config) (app.Provider, error) {
...
+		res := &providerpb.OpenFileInAppProviderResponse{
+			Status: status.NewInternal(ctx, err, "error reading reponse body"),
+		}
+		return res, nil
 	}
-	return res, nil
+
+	return &providerpb.OpenFileInAppProviderResponse{
+		Status:         status.NewOK(ctx),
+		AppProviderUrl: string(resBody),
```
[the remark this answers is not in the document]
LovisaLugnegard: ✅ Will look into this today.

cmd/reva/open-file-in-app-provider.go
```diff
@@ -0,0 +1,103 @@
...
+
+		client, err := getClient()
+		if err != nil {
+			return err
+		}
+		ref := &provider.Reference{
+			Spec: &provider.Reference_Path{Path: path},
+		}
+		reqA := &provider.StatRequest{Ref: ref}
+		resA, err := client.Stat(ctx, reqA)
```
ishank011, line 57: No need to stat here as we're doing the same again in the gateway

internal/grpc/services/gateway/appprovider.go
```diff
@@ -24,15 +24,55 @@ import (
...
 	"github.com/pkg/errors"
 )
 
-func (s *svc) Open(ctx context.Context, req *providerpb.OpenRequest) (*providerpb.OpenResponse, error) {
-	provider, err := s.findAppProvider(ctx, req.ResourceInfo)
+func (s *svc) OpenFileInAppProvider(ctx context.Context, req *providerpb.OpenFileInAppProviderRequest) (*providerpb.OpenFileInAppProviderResponse, error) {
+
+	log := appctx.GetLogger(ctx)
+
+	c, err := s.findByPath(ctx, req.Ref.GetPath())
```
ishank011, line 40: You can directly call s.find since you have the reference object

internal/grpc/services/appprovider/appprovider.go
```diff
@@ -93,18 +108,80 @@ func getProvider(c *config) (app.Provider, error) {
...
+	folderURL := s.conf.UIURL + filepath.Dir(req.Ref.GetPath())
+
+	httpClient := rhttp.GetHTTPClient(
+		rhttp.Context(ctx),
+		// TODO make insecure configurable
+		rhttp.Insecure(true),
+		// TODO make timeout configurable
+		rhttp.Timeout(time.Duration(24*int64(time.Hour))),
+	)
+	httpReq, err := rhttp.NewRequest(ctx, "GET", wopiurl+"wopi/iop/open", nil)
```
ishank011, line 127: Shouldn't this be carried out at the driver level, where this is a driver for wopi? Other apps would not have the exact same endpoint

internal/grpc/services/appprovider/appprovider.go
```diff
@@ -38,14 +46,20 @@ func init() {
...
 	provider app.Provider
+	conf     *config
 }
 
 type config struct {
-	Driver string                 `mapstructure:"driver"`
-	Demo   map[string]interface{} `mapstructure:"demo"`
+	Driver    string                 `mapstructure:"driver"`
+	Demo      map[string]interface{} `mapstructure:"demo"`
+	IopSecret string                 `mapstructure:"iopsecret" docs:"The iopsecret used to connect to the wopiserver."`
```
ishank011, line 55: If you're passing only one value, the docs tag expects it to be the default https://github.com/cs3org/cato#extraction. If you want to add the description, it should be the second argument, or should be added as a comment

ishank011: Also, please run make after adding the tags, otherwise documentation won't be generated.

glpatcern:
> Regarding calling the wopi endpoint in the storageprovider layer [here](https://github.com/cs3org/reva/pull/920/files#diff-39c26ea9746980037b602718434dcacaR127), this should be carried out at the driver level, with a separate driver for wopi. This code disables support for any other apps which would not have the exact same endpoint as wopi does, or might just expect a simple iframe like [this](https://github.com/cs3org/reva/blob/master/pkg/app/provider/demo/demo.go#L36).

A comment concerning this: my understanding was that this endpoint is really dedicated (and "hardwired") to WOPI, which in turn is the generic way to interact with different application providers - either directly or via a WOPI "bridge" microservice (to be developed, a PoC exists at https://github.com/cs3org/wopibridge). Of course we can do everything with software, but I'm not sure there's much gain in adding another abstraction layer here. What do you think? @labkode?

Also, granted that [GetIFrame](https://github.com/cs3org/reva/blob/master/pkg/app/provider/demo/demo.go#L36) is just a demo, but I don't know any application provider that would respect that URL format - whereas a WOPI-enabled application does respect the `WOPISrc` and `access_token` query parameters.

ishank011:
@glpatcern The GetIFrame method was just an example. Do we expect all apps to expose `/wopi/iop/open`?

Even if we use WOPI bridge, that is itself another abstraction layer and it kind of defeats the purpose behind CS3APIs, which was to make sure that all applications use these. Instead of communicating with the GRPC service, this would go through one extra layer before that. So we're removing the layer from reva (which is what reva exactly does) and adding an external component to do that.

glpatcern:
(Sure, GetIFrame is just an example, just pointing out that there's no app I know about that follows that pattern...)

The point is not that all apps expose `/wopi/iop/open`, rather the reverse: all (Office-like) apps may follow the WOPI protocol or some sort of it, and therefore Reva supporting `/wopi/iop/open` of our WOPI server means Reva can interact with all those apps in one go.

At the same time, I agree that a new random app should get integrated via CS3APIs as first class choice. The WOPI protocol is an options for apps that already support it or that are close to it. To be noted that the WOPI semantic only intersects the CS3APIs semantic, it's not a subset: in particular, the logic behind locking and collaborative editing, as provided by the WOPI protocol, is richer than what we have in the CS3APIs. We may want to extend the CS3APIs in that direction, sure - this was even discussed at some stage, and it's a possibility. It remains the fact that the WOPI protocol is a pre-existing and de facto standard.

ishank011:
So for the apps that adhere to the wopi protocol (which I agree most of them should), we use the wopi driver and for others (which might not need all the functionalities that wopi provides), we have custom drivers. That is my opinion.

glpatcern:
OK, now I better understand the spirit of your comment, thanks.

Yet, apps that do not use the WOPI workflow may just not need an "open" endpoint at all, and go for a pure CS3APIs-based integration. Or at least, until we have a concrete case of an app that does require some sort of file-based open but at the same time we do not want it integrated via WOPI, I don't see the need for this abstraction.

A couple of examples of apps that would not integrate nicely via WOPI:
1) Overleaf, the cloud-based LaTeX editor: it is based on "projects" that map to folders, and it uses git on the backend. Any integration we may (and should) foresee would go far beyond a file-based "open" and would require a different kind of workflow.
2) (This is maybe a dream) VS Code web edition (I learned from Samuel about this): it would be nice to have it integrated. Again this is based on workspaces == folders rather than single files, so the file-oriented "open" does not fit.

OK, I stop here for now (and I'm off for the next 3 weeks ;-) ). Please discuss this but IMHO we could at least postpone the abstraction to a later stage if deemed necessary.

ishank011:
Sure. Enjoy your vacations!

cmd/reva/open-file-in-app-provider.go
LovisaLugnegard: Thanks for finding this!

labkode:
@LovisaLugnegard I've merged the cs3apis proposal for wopi. Can you address these comments as we need to merge this asap to align with the new APIS. And make CI green.

internal/grpc/services/appprovider/appprovider.go
LovisaLugnegard: After the discussion with @glpatcern I think the driver is for a later stage.

PR #922, opened 2020-07-01
Add tutorial on how to deploy reva together with the wopiserver locally

LovisaLugnegard:
Local setup for wopiserver and reva.

### DIFF
--- a/changelog/unreleased/add-tutorial.md
+++ b/changelog/unreleased/add-tutorial.md
@@ -1,0 +1,5 @@
+Enhancement: Add tutorial for deploying WOPI and Reva locally
+
+Add a new tutorial on how to run Reva and Wopiserver together locally
+
+https://github.com/cs3org/reva/pull/922
--- a/docs/content/en/docs/tutorials/accept-invite-tutorial.md
+++ b/docs/content/en/docs/tutorials/accept-invite-tutorial.md
@@ -36,21 +36,9 @@ git clone git@github.com:sciencemesh/accept-invite.git
 ```
 
 ## 2. Build Reva
-Go to your Reva folder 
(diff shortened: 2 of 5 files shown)


labkode:
@LovisaLugnegard for installing reva, please point to: https://reva.link/docs/getting-started/install-reva/  

There are others sections of the documentation that also define build steps,  can you also update them to point to that link?

LovisaLugnegard:
@labkode Since it's only two lines `make deps` and `make build`/`make`, I don't think it's worth the pain for the user reading the doc to open up a new tab.

labkode:
@LovisaLugnegard isn't a matter of two lines, is a matter on how to best deploy Reva, and  that is not by compiling it but to install it following the official installation guide.

LovisaLugnegard:
@labkode I think I'm missing some pieces, to me it looks like the instructions in "install reva" is not much more, but please tell me where I can find more information about this so I can point to a more specific url.   
Screenshot 2020-07-02 at 14 26 38
Screenshot 2020-07-02 at 14 27 37

labkode:
@LovisaLugnegard from the documentation you're preparing you need to point to: ```https://reva.link/docs/getting-started/install-reva/```, is the first screenshot you have posted. The preferred way to install Reva is to download the binary directly and not to compile from sources. 

Also, copy pasting  build instructions in many places is not desirable, as you need to keep track of all of them.

LovisaLugnegard:
@labkode Firstly, I agree the doc shouldn't be replicated. I still don't fully agree with you though, mainly because one reason to have a local setup of reva and wopi is local development and testing, and then compiling is needed anyway - which is more or less two lines. 
But it seems like you've stronger feelings about this than I've :)

PR #929, opened 2020-07-02
implement share handling for accepting and listing folder shares

C0rby:
I implemented share handling functionality in the owncloud storage driver. Most of the code is similar to how eosfs handles shares. There are just some small storage specific differences.

@labkode, I'm not sure if my changes in `internal/grpc/services/gateway/storageprovider.go` will break other storage drivers.
I tried to test it with eos but was not able to create share references with it, even without my changes but with the master branch it didn't work.

The reason why I'm using `path.Split` instead of `isSharedFolder` or `isShareName` is that the latter assume the path is like `/home/MyShare/somefolder/file.txt`. In my understanding that would mean you couldn't share a file from the root folder because such file would have a path like `/home/MyShare/file.txt`.

Furthermore in line 138 we only care about if the path is a file oder a folder and we can easily distinguish that with `path.Split`.
What do you think?

Closes https://github.com/owncloud/ocis-reva/issues/11

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -152,12 +152,13 @@ func init() {
 }
 
 type config struct {
-	DataDirectory string `mapstructure:"datadirectory"`
-	UploadInfoDir string `mapstructure:"upload_info_dir"`
-	UserLayout    string `mapstructure:"user_layout"`
-	Redis         string `mapstructure:"redis"`
-	EnableHome    bool   `mapstructure:"enable_home"`
-	Scan          bool   `mapstructure:"scan"`
+	DataDirectory  string `mapstructure:"datadirectory"`
+	UploadInfoDir  string `mapstructure:"upload_info_dir"`
+	ShareDirectory string `mapstructure:"sharedirectory"`
+	UserLayout     string `mapstructure:"user_layout"`
+	Redis          string `mapstructure:"redis"`
+	EnableHome     bool   `mapstructure:"enable_home"`
+	Scan           bool   `mapstructure:"scan"`
 }
 
 func parseConfig(m map[string]interface{}) (*config, error) {
@@ -178,6 +179,9 @@ func (c *config) init(m map[string]interface{}) {
 	}
 	if c.UploadInfoDir == "" {
 		c.UploadInfoDir = "/var/tmp/reva/uploadinfo"
+	}
+	if c.ShareDirectory == "" {
+		c.ShareDirectory = "/Shares"
 	}
 	// default to scanning if not configured
 	if _, ok := m["scan"]; !ok {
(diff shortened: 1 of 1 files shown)


labkode:
@C0rby the logic around the base path is wrong. When base is empty means the path is either `/` or empty string.
Access paths outside the share folder is already covered by:

```
if !s.inSharedFolder(ctx, p) {
		return s.initiateFileUpload(ctx, req)
}
```

so this condition is not triggered.

The method `isShareChild` is more semantic that the path split.   

Also, with your changes you will stat resources inside a share that don't exist in the home provider, they exist in original storage provider. This assumes that the storage drivers they need to handle resources outside its own storage provider and it not the case. Furthermore, this also assumes that you can create a reference with storage id and paths mixed, also not possible.


Please revert the changes in `internal/grpc/services/gateway/storageprovider.go`.
The filesystem implementation should work without those changes.

C0rby:
> @C0rby the logic around the base path is wrong. When base is empty means the path is either `/` or empty string.
No, when the base is empty the path can be a path to a folder.
```golang
path.Split("/home/MyShare/somefolder/") => dir == "/home/MyShare/somefolder/", base == ""
path.Split("/home/MyShare/somefile") => dir == "/home/MyShare/", base == "somefile"
```
But I'll try to make the owncloud storage driver work without the changes in `storageprovider.go`.

labkode:
@C0rby, I see, however we should not assume folder paths always end in a slash.

C0rby:
Can you clarify something for me?

```golang
// /home/MyShares/photos/
func (s *svc) isShareName(ctx context.Context, p string) bool {
	return s.split(ctx, p, 3)
}

// /home/MyShares/photos/Ibiza/beach.png
func (s *svc) isShareChild(ctx context.Context, p string) bool {
	return s.split(ctx, p, 4)
}
```

Both of these methods assume that the path contains 3 folders before the actual shared file but in the `usershareprovider.createReference` the reference path is build to `/home/MyShare/x` where `x` is the file.
Here a comment from the createReference method:
```golang
// reference path is the home path + some name
// CreateReferene(cs3://home/MyShares/x)
// that can end up in the storage provider like:
// /eos/user/.shadow/g/gonzalhu/MyShares/x
```

In my understanding this doesn't fit, either `isShareName` and `isShareChild` must be changed or `createReference`.
Or am I missing something here?

labkode:
@C0rby both of those methods assume the path contains at least **two** folder, `/home/MyShares` before the shared reference. Those methods and the create reference are correct.

In the current design we lock down to the `/home/MyShares` folder where shares can be mounted.

C0rby:
@labkode, I know about `/home/MyShares`, that part is alright.

What I mean is:
The path created by `storageprovider.CreateReference` is always 3 parts long i.e. `/home/beach.png` results in the path `/home/MyShares/beach.png` same with `/home/photos/beach.png`, this also results in `/home/MyShares/beach.png`.

So now when we call `isShareName(ctx, "/home/MyShares/beach.png")` this will return true since it consists of 3 parts even though the last part is a file not a folder.

The call to `isShareChild(ctx, "/home/MyShares/beach.png")` would then fail since it consists of 3 not 4 parts.

My problem is line 138-148 in `internal/grpc/services/gateway/storageprovider.go`
```golang
log := appctx.GetLogger(ctx)
if s.isSharedFolder(ctx, p) || s.isShareName(ctx, p) {
	log.Debug().Msgf("path:%s points to shared folder or share name", p)
	err := errtypes.PermissionDenied("gateway: cannot upload to share folder or share name: path=" + p)
	log.Err(err).Msg("gateway: error downloading")
	return &gateway.InitiateFileDownloadResponse{
		Status: status.NewInvalidArg(ctx, "path points to share folder or share name"),
	}, nil

}

if s.isShareChild(ctx, p) {
```

`isSharedFolder` returns false that is ok but `isShareName` returns true.
But even if `isShareName` would return false, `isShareChild` would return false, preventing to download the file.

labkode:
@C0rby I understand the problem now. You're trying to mount single file shares and the current checks enforce that only folders are allowed. Can you update your code to make file sharing optional rather than enabled by default?

C0rby:
Ok got it. My PR works with folder shares now only. 
I reverted the changes in storageprovider.go.

C0rby:
Accepting and listing the shares work with this PR. I opened a new issue for downloading the shared files. https://github.com/owncloud/ocis-reva/issues/359 That is not implemented here.

PR #930, opened 2020-07-02
Fix OCS public link share update values logic

PVince81:
Only update a value if it was actually set in the form.

Added support for clearing expiration date and password by
distinguishing between an empty string and whether the value is actually
set in the request.

Refactored public permission parsing to use permissionFromRequest for
both creation and update of public shares.

For https://github.com/owncloud/ocis-reva/issues/252

Makes almost all tests from tests/acceptance/features/apiSharePublicLink2/multilinkSharing.feature pass now.

- ~~run more API tests to confirm that there are no regressions~~ => CI will help with that!

### DIFF
--- a/internal/http/services/owncloud/ocs/conversions/main.go
+++ b/internal/http/services/owncloud/ocs/conversions/main.go
@@ -386,7 +386,7 @@ func Permissions2OCSPermissions(p *provider.ResourcePermissions) Permissions {
 // timestamp is assumed to be UTC ... just human readable ...
 // FIXME and ambiguous / error prone because there is no time zone ...
 func timestampToExpiration(t *types.Timestamp) string {
-	return time.Unix(int64(t.Seconds), int64(t.Nanos)).Format("2006-01-02 15:05:05")
+	return time.Unix(int64(t.Seconds), int64(t.Nanos)).UTC().Format("2006-01-02 15:05:05")
 }
 
 const (
--- a/internal/http/services/owncloud/ocs/conversions/main.go
+++ b/internal/http/services/owncloud/ocs/conversions/main.go
@@ -361,6 +361,7 @@ func publicSharePermissions2OCSPermissions(sp *link.PublicSharePermissions) Perm
 }
 
 // TODO sort out mapping, this is just a first guess
+// public link permissions to OCS permissions
 func Permissions2OCSPermissions(p *provider.ResourcePermissions) Permissions {
 	permissions := PermissionInvalid
 	if p != nil {
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -351,65 +351,50 @@ func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	// TODO: the default might change depending on allowed permissions and configs
-	permKey := 1
-
-	// handle legacy "publicUpload" arg that overrides permissions differently depending on the scenario
-	// https://github.com/owncloud/core/blob/v10.4.0/apps/files_sharing/lib/Controller/Share20OcsController.php#L447
-	if r.FormValue("publicUpload") != "" {
-		publicUploadFlag, err := strconv.ParseBool(r.FormValue("publicUpload"))
-		if err != nil {
-			log.Error().Err(err).Str("createShare", "shares").Str("publicUpload", r.FormValue("publicUpload")).Msg("could not parse publicUpload argument")
-			response.WriteOCSError(w, r, response.MetaBadRequest.StatusCode, "invalid argument value for \"publicUpload\"", err)
-			return
-		}
-
-		if publicUploadFlag {
-			// all perms except reshare
-			permKey = 15
-		}
-	}
-
-	// note: "permissions" value has higher priority than "publicUpload"
-	if r.FormValue("permissions") != "" {
-		// phoenix sends: {"permissions": 15}. See ocPermToRole struct for mapping
-		permKey, err = strconv.Atoi(r.FormValue("permissions"))
-		if err != nil {
-			log.Error().Err(err).Str("createShare", "shares").Str("permissions", r.FormValue("permissions")).Msgf("invalid type: %T", permKey)
-		}
-	}
-
-	role, ok := ocPermToRole[permKey]
-	if !ok {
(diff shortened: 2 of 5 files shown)


PVince81:
and yes, OCS is not REST... one does not pass the whole object to a PUT request...

PVince81:
/drone/src/tmp/testrunner/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature:566 and others

I'll have a look

PVince81:
seems the refactoring of permission logic broke the permission conversion from OC to CS3.
but even after verifying almost all locations I can't see where it breaks.

PVince81:
problem solved. it was a nasty colon again...

I've also added two extra fixes for expiration date and token length as they were failing in tests I'm trying to enable...

PVince81:
hmm some other failures in the same file: /drone/src/tmp/testrunner/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature:603

PVince81:
right, the old code ignored the error and I made it more strict...

I've adjusted the roles just a bit to make the tests pass. I added some comments as there will be more to adjust/align later in the roles and perms (see https://github.com/owncloud/ocis-reva/issues/292)

PVince81:
All green now!

More tests will be enabled later in https://github.com/owncloud/core/pull/37633

PVince81:
I've added yet another fix: setting any attribute for a public link share would clear its password. This was making tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature:71 in https://github.com/owncloud/core/pull/37633 fail

internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -361,6 +361,7 @@ func publicSharePermissions2OCSPermissions(sp *link.PublicSharePermissions) Perm
 }
 
 // TODO sort out mapping, this is just a first guess
+// public link permissions to OCS permissions
 func Permissions2OCSPermissions(p *provider.ResourcePermissions) Permissions {
```
refs, line 365: On an organization basis we should get rid of this conversions package in favor of letting the package do the conversions. What I mean is moving `Permissions2OCSPermissions` to the `ocs` package itself. And so on with the respective conversions.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -1549,84 +1535,128 @@ func (h *Handler) updatePublicShare(w http.ResponseWriter, r *http.Request, shar
...
+				Update: updates[k],
+			})
+			if err != nil {
+				log.Err(err).Str("shareID", shareID).Msg("sending update request to public link provider")
+				response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "Error sending update request to public link provider", err)
+				return
+			}
 		}
+		publicShare = uRes.Share
+	} else if !updatesFound {
```
refs, line 1651: if I follow the logic correctly, isn't `updatesFound` unnecessary? the reverse of `len(updates) > 0` is essentially `len(updates) == 0`

pkg/publicshare/manager/json/json.go
```diff
@@ -107,10 +107,10 @@ type manager struct {
 // CreatePublicShare adds a new entry to manager.shares
 func (m *manager) CreatePublicShare(ctx context.Context, u *user.User, rInfo *provider.ResourceInfo, g *link.Grant) (*link.PublicShare, error) {
 	id := &link.PublicShareId{
-		OpaqueId: randString(12),
+		OpaqueId: randString(15),
```
refs, line 110: is this by spec? Just curious

pkg/publicshare/manager/json/json.go
```diff
@@ -193,25 +193,32 @@ func (m *manager) UpdatePublicShare(ctx context.Context, u *user.User, req *link
...
-		p = base64.StdEncoding.EncodeToString([]byte(req.Update.GetGrant().Password))
-		share.PasswordProtected = true
+		passwordChanged = true
+		if req.Update.GetGrant().Password == "" {
+			share.PasswordProtected = false
+			newPasswordEncoded = ""
+		} else {
+			newPasswordEncoded = base64.StdEncoding.EncodeToString([]byte(req.Update.GetGrant().Password))
+			share.PasswordProtected = true
+		}
```
refs, line 221: good catch :)

labkode:
@PVince81 once you're happy with @refs comments I'll merge.

pkg/publicshare/manager/json/json.go
PVince81: I'm not sure if the spec tells about a minimum length. However that's the length we used in OC 10 and that the tests are enforcing. The longer the token, the safer it is as it will reduce collisions.

We might also need to look into "secure random" later on, as this is how tokens are generated in OC 10.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
PVince81: updatesFound is only there to trigger the 400 Bad Request in case none of the expected attributes has been passed in, basically "no update has been requested".

if you do specify a value, even if it matches already what's in the DB, then it's accepted as valid. before this fix it was saying that "setting permissions to 15 when it's already 15" is a bad request as it triggers no updates. faulty logic

PVince81: because when you set permissions to 15 when they're already 15, the updates array will be empty as we have difference checks above.

internal/http/services/owncloud/ocs/conversions/main.go
PVince81: sounds good, please post your ideas in https://github.com/owncloud/ocis-reva/issues/292 or create another tech debt ticket

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
refs: I mean the existence of the `updatesFound` variable, not the execution path :p

PVince81: the semantics of `updatesFound` is not the same as `len(updates)`. you can have a case where zero updates exist in the array but still some updates have been provided through attributes

refs: 👍 works as expected. Clarified the misunderstanding.

PVince81:
@labkode solved now, thanks

PR #935, opened 2020-07-03
Update core master tests commit for OCS share public link updates

PVince81:
Brings in new enabled tests for OCS share public link edit API: https://github.com/owncloud/core/pull/37633 that could be enabled thanks to the fix https://github.com/cs3org/reva/pull/930

@individual-it


phil-davis:
Note: I made PR #939 to confirm that the very latest core API tests are passing. Yes, they do pass. That PR got some other "No space left on device" problems in other drone pipelines.

This PR #935 can be merged safely, if someone with the necessary privs sees it.

PR #951, opened 2020-07-08
OCS Share wrap resource id correctly

PVince81:
Wrap the resource id and opaque id in OCS Share responses in the same
way like the resource id wrapping done on Webdav level.
This is important to be able to correlate the resources without
resorting to parsing the base64 bits.

Fixes https://github.com/owncloud/ocis-reva/issues/344


PVince81:
tested by creating a user share and checking the network console. here I saw that now file_source and item_source use the wrapped format

PR #952, opened 2020-07-08
Recreate shares.json in case of deletion

PVince81:
To be able to clear shares when testing, we need to be able to reload
the json files after they have been deleted.

Later on, in a real-world scenario, we need to be able to rebuild the
json file based on information stored in the storage. (see inline TODO)

@butonic

### Test steps:

1. Create a folder "test"
1. Share "test" with "Marie"
1. Refresh the page, see that the share exists
1. Delete "shares.json"
1. Refresh the page, see that the shares are gone

Before this fix: the shares would still be there until the service is reloaded
After this fix: the shares are gone directly and the file is recreated empty.

### TODOs:

- [ ] get preliminary review for approach
- [ ] remove print statements
- [ ] apply the same approach for publicshares.json

### DIFF
--- a/pkg/share/manager/json/json.go
+++ b/pkg/share/manager/json/json.go
@@ -21,6 +21,7 @@ package json
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"reflect"
@@ -32,6 +33,7 @@ import (
 	typespb "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/pkg/errtypes"
 	"github.com/cs3org/reva/pkg/share"
+	"github.com/fsnotify/fsnotify"
 	"github.com/google/uuid"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


PVince81:
maybe instead of overriding Lock/Unlock, could simply set the flag in the Save() method before and after writing to catch the events that were triggered by us

pkg/share/manager/json/json.go
```diff
@@ -154,6 +191,31 @@ func genID() string {
...
+	model, err := loadOrCreate(m.model.file)
+	if err != nil {
+		return errors.Wrap(err, "error loading the file containing the shares")
+	}
+	m.model = model
+	return nil
+}
+
+func (m *mgr) Lock() {
+	m.locked = true
```
PVince81, line 206: maybe instead of overriding Lock/Unlock, could simply set the flag in the Save() method before and after writing to catch the events that were triggered by us

labkode:
@PVince81 as this is mainly for testing: why don't you recreate the daemon again?

PVince81:
@individual-it it might be possible to restart the "reva-sharing" service alone, might be faster to boot up then restarting everything ?

PVince81:
hmm, the problem might be that the container where the services run is not the same.
they might have shared volumes but not shared processes.

restarting the service would require having some kind of API in the container that runs reva to be able to tell it to restart services.

@individual-it correct ?

individual-it:
yes we would have to talk somehow across the container to restart the service.
Anyway how would an admin do it? If a user is deleted it would be impractical to restart the complete service.
Would it make sense to recreate the shares if a change in the user is detected, e.g. the id. I guess even if a user is recreated in LDAP with the same name, email etc. the ID should be different, or?

butonic:
won't be continued here

PR #954, opened 2020-07-08
Fix file-descriptor leak

IljaN:
Clone client transport from http.DefaultTransport instead of creating a zeroed transport struct.
With an empty timeout-struct all timeout-values are zeroed disabling some timeouts and TTLs.

See https://golang.org/src/net/http/transport.go#L194 and following.

Closes https://github.com/owncloud/product/issues/100

### DIFF
--- a/pkg/rhttp/client.go
+++ b/pkg/rhttp/client.go
@@ -24,9 +24,10 @@ import (
 	"io"
 	"net/http"
 
+	"go.opencensus.io/plugin/ochttp"
+
 	"github.com/cs3org/reva/pkg/token"
 	"github.com/pkg/errors"
-	"go.opencensus.io/plugin/ochttp"
 )
 
 // GetHTTPClient returns an http client with open census tracing support.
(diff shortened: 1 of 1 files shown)


individual-it:
@jasson99 please test this fix

labkode:
@IljaN nice catch, for long term I think we should avoid to rely on the default transport and always create it explicitly without relying on package variables.

labkode:
Merge after @jasson99 validates.

IljaN:
nofail

Hatch 0.25, 150 Concurrent

pkg/rhttp/client.go
```diff
@@ -24,9 +24,10 @@ import (
...
 	"net/http"
 
+	"go.opencensus.io/plugin/ochttp"
+
 	"github.com/cs3org/reva/pkg/token"
 	"github.com/pkg/errors"
-	"go.opencensus.io/plugin/ochttp"
 )
 
 // GetHTTPClient returns an http client with open census tracing support.
```
refs, line 33: is the TODO still necessary?

IljaN:
> ...for long term I think we should avoid to rely on the default transport and always create it explicitly without relying on package variables.

Yes agree! We just need to find or tune the right values for our use-case. For now I assumed that the go-lang selected defaults should suite us. Any proposals?

jasson99:
@IljaN I tested with this fix and it works now.

PR #955, opened 2020-07-08
Added a CS3API compliant data exporter to Mentix

Daniel-WWU-IT:
A new HTTP endpoint has been added that exports data in a format compliant to the CS3API `ProviderInfo` structure.


labkode:
@Daniel-WWU-IT some hound linting

ishank011:
@Daniel-WWU-IT you can just use small-case for the first letter for internal methods. Saves you from adding comments.

Daniel-WWU-IT:
> @Daniel-WWU-IT you can just use small-case for the first letter for internal methods. Saves you from adding comments.
They aren't internal, otherwise I would have done so. I add comments to types and functions where it makes sense, not just to make the hound happy ;) And these are not the only exported funcs/types that lack comments.

ishank011:
Ideally, all the exported methods should have comments, even though they are self-explanatory ;)
Could you point me to those which aren't commented? I might as well fix those.

Daniel-WWU-IT:
> Ideally, all the exported methods should have comments, even though they are self-explanatory ;)
> Could you point me to those which aren't commented? I might as well fix those.

I think I stumbled upon a few in some other services, not sure if this has changed yet. But I'll go through my code and comment all the obvious ;)

Daniel-WWU-IT:
Comments added... hope I got all places

ishank011:
Awesome! Thanks!

PR #959, opened 2020-07-09
Remove Expired Link on Access

refs:
Since there is no background jobs scheduled to wipe out expired resources, for the time being public links are going to be removed on a "on demand" basis, meaning whenever there is an API call that access the list of shares for a given resource, we will check whether the share is expired and delete it if so.

This code will not live in the managers (drivers), since we want the behavior to be consistent across implementations.

Closes https://github.com/owncloud/ocis/issues/372

### DIFF
--- a/changelog/unreleased/remove-public-link-on-access.md
+++ b/changelog/unreleased/remove-public-link-on-access.md
@@ -1,0 +1,5 @@
+Enhancement: Remove expired Link on Access
+
+Since there is no background jobs scheduled to wipe out expired resources, for the time being public links are going to be removed on a "on demand" basis, meaning whenever there is an API call that access the list of shares for a given resource, we will check whether the share is expired and delete it if so. This code will not live in the managers (drivers), since we want the behavior to be consistent across implementations.
+
+https://github.com/cs3org/reva/pull/959
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -23,10 +23,12 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"time"
 
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	linkv1beta1 "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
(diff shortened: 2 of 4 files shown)


internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -156,6 +162,51 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
+			},
+		}
+
+		res, err := c.RemovePublicShare(ctx, req)
+		if err != nil {
+			return err
+		}
+		if res.Status.Code != rpc.Code_CODE_OK {
+			if res.Status.Code == rpc.Code_CODE_NOT_FOUND {
+				return fmt.Errorf("not found")
```
kulmann, line 201: Might be a naive question, but isn't a `not found` on a delete request something that should not break? The outcome of the request would still be the same: the share is deleted (just not by this call, but probably from another request on the same share, running in parallel). It might be worth a log statement, but I see no reason to throw an error. 🤔

refs: Good point. Logging should suffice for this case

changelog/unreleased/remove-public-link-on-access.md
```diff
@@ -0,0 +1,5 @@
+Enhancement: Remove Link on Access
```
PVince81, line 1: Remove **expired** link on access

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -142,6 +144,10 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				return
 			}
 			log.Debug().Interface("statInfo", statInfo).Msg("Stat info from public link token path")
+			err = deletePublicShareIfExpired(ctx, s, token)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
PVince81, line 149: only log here, don't write a different header.

now if an expiration is not possible, the question is whether we should even allow accessing that share.
depending on the reason for the failed deletion, it is likely that the reason is either "shares data structure corrupt" or "service not available", so maybe it's ok to return an error, but then also call `return` ?

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -156,6 +162,51 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
+		}
+		if res.Status.Code != rpc.Code_CODE_OK {
+			if res.Status.Code == rpc.Code_CODE_NOT_FOUND {
+				return fmt.Errorf("not found")
+			}
+			return fmt.Errorf("grpc delete share request failed")
+		}
+	}
+
+	return nil
```
PVince81, line 207: you should also return a bool that tells whether a deletion happened, so that the caller knows whether to continue proceeding or just exit with a 404. because once the share is deleted we should not continue with whatever operation was originally intended

internal/http/services/owncloud/ocdav/dav.go
refs: delete this way should be a noop. I'll update it. I struggled with error handling with this one.

labkode:
@PVince81 any update on this?

refs:
@labkode setting this one as WIP, as I hacked it around a little bit, and it's lacking some process. We moved it's priority to a future sprint and will retake it soon.

refs:
closing in favor of #1361

PR #960, opened 2020-07-10
Bump core commit for tests

phil-davis:
Some test-related things that have happened in https://github.com/owncloud/core

- https://github.com/owncloud/core/pull/37656 test the API response of a share-update response
- https://github.com/owncloud/core/pull/37672 retag reva issue 249 with the correct issue tags
- https://github.com/owncloud/core/pull/37662 Remove test tags for 336
- https://github.com/owncloud/core/pull/37675 Add env var TEST_REVA

Set the env var TEST_REVA so that the test runner knows that we are testing against a `cs3org/reva` system (rather than `owncloud/ocis-reva` or `owncloud/ocis`. This allows the test runner to make slightly different expectations where behavior is different. See PR https://github.com/owncloud/core/pull/37675 for the difference in public link URL reports, the first example.


phil-davis:
442 scenarios (442 passed) - good
https://cloud.drone.io/cs3org/reva/1849/4/6

PR #970, opened 2020-07-13
Add headers to SMTP client to prevent being tagged as spam

ishank011:


pkg/smtpclient/smtpclient.go
```diff
@@ -67,14 +84,22 @@ func (creds *SMTPCredentials) sendMailAuthSMTP(recipient, subject, body string)
...
-func (creds *SMTPCredentials) sendMailSMTP(recipient, subject, body string) error {
+func (creds *SMTPCredentials) sendMailSMTP(recipient, subject, message string) error {
 
 	c, err := smtp.Dial(fmt.Sprintf("%s:%d", creds.SMTPServer, creds.SMTPPort))
 	if err != nil {
 		return err
 	}
 	defer c.Close()
 
+	host, err := os.Hostname()
```
mirekys, line 95: `os.Hostname()` may come to problems when running inside k8s containers, we would be probably better off with a configurable sender hostname (like it is done e.g. in [gomail](https://github.com/go-gomail/gomail/blob/master/smtp.go#L33-L35))

mirekys: Or maybe as a fallback, we could use domain part of the `SenderMail` as hostname?

ishank011: @mirekys makes a lot of sense. I'll make the change

pkg/smtpclient/smtpclient.go
```diff
@@ -67,14 +103,22 @@ func (creds *SMTPCredentials) sendMailAuthSMTP(recipient, subject, body string)
...
+	if err != nil {
+		return err
+	}
+	if err = c.Hello(host); err != nil {
```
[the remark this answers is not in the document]
mirekys: looks good otherwise, thanks! 👍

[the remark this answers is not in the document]
ishank011: @mirekys Done, thanks!

PR #974, opened 2020-07-15
Add functionality to create webdav references for OCM shares

ishank011:


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1028,6 +1030,12 @@ func (s *svc) handleCS3Ref(ctx context.Context, opaque string) (*provider.Resour
 	return res.Info, nil
 }
 
+func (s *svc) handleWebdavRef(ctx context.Context, ri *provider.ResourceInfo) (*provider.ResourceInfo, error) {
+	// A webdav ref has the following layout: <storage_id>/<opaque_id>@webdav_endpoint
+	// TODO: Once file transfer functionalities have been added.
```
labkode, line 1035: We still need to access the remote data by issues a GET/PROPFIND response to obtain the necessary information dynamically.  

We can merge this PR and add that later. @ishank011

ishank011: Yep, I added the TODO there. Once we have that functionality, I'll update this.

labkode:
@ishank011 we can add `build(deps)` here:
https://github.com/cs3org/reva/blob/master/.github/config.yml#L10

ishank011:
@labkode Aren't the tags only those specified within square brackets? I'll anyway try and see if this works

labkode:
@ishank011 not sure either, test empirically :)

labkode:
@ishank011 the creation of the webdav reference has to be optional, for example for Cubbit, they only need to implement the share creation but not propagating it to a storage reference. Can be add a configuration flag that this needs to be explicit?

ishank011:
@labkode yes it's already there. It's created only if `commit_share_to_storage_grant`  is set to true [here](https://github.com/cs3org/reva/blob/83c8cab4c5643b90380190448b73e82b4a56bb2b/internal/grpc/services/gateway/ocmshareprovider.go#L58).

PR #983, opened 2020-07-20
[WIP] Add UID and GID to the user object from user package

### LINKED ISSUE #964: eosfs: don't rely on systemcall to obtain user unix IDs/GroupIds
In the EOS filesystem implementation we rely on using systemcalls to obtain the user uid/gid. This implies having local users where reva runs  and that sucks:
https://github.com/cs3org/reva/blob/master/pkg/storage/utils/eosfs/eosfs.go#L1219

We should obtain this information directly by calling the UserProviderAPI (through the Gateway).

ishank011:
Currently, the UID and GID for users need to be read from the local system which requires local users to be present. This change retrieves that information from the user and auth packages and adds methods to retrieve it.

Closes #964

### DIFF
--- a/changelog/unreleased/uid-gid-user-object.md
+++ b/changelog/unreleased/uid-gid-user-object.md
@@ -4,4 +4,4 @@ Currently, the UID and GID for users need to be read from the local system which
 requires local users to be present. This change retrieves that information from
 the user and auth packages and adds methods to retrieve it.
 
-https://github.com/cs3org/reva/pull/982
+https://github.com/cs3org/reva/pull/983
--- a/changelog/unreleased/uid-gid-user-object.md
+++ b/changelog/unreleased/uid-gid-user-object.md
@@ -1,0 +1,7 @@
+Enhancement: Add UID and GID to the user object from user package
+
+Currently, the UID and GID for users need to be read from the local system which
+requires local users to be present. This change retrieves that information from
+the user and auth packages and adds methods to retrieve it.
+
+https://github.com/cs3org/reva/pull/982
--- a/docs/content/en/docs/config/packages/auth/_index.md
+++ b/docs/content/en/docs/config/packages/auth/_index.md
@@ -1,0 +1,7 @@
+---
+title: "auth"
+linkTitle: "auth"
+weight: 10
+description: >
+  Configuration for the auth service
+---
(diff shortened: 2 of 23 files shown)


ishank011:
@labkode the PR is ready for review. Some initial testing is done, working on deploying it for further tests.

One thing which I'm not very sure about is [this](https://github.com/cs3org/reva/pull/983/files#diff-584bc1e66a7e7fcb9373eae4e41cd6b0R1212) part. Whenever we stat for resources, EOS returns the UID, which we encode as "uid:xyz" since we have no way of resolving it at the fs level. This is done at the [gateway](https://github.com/cs3org/reva/pull/983/files#diff-f765b4904b8979fac7d598e59211d0ffR97) layer to get the user ID back from the UID. I modified the GetUser method to return the user based on UID passed as opaque.

Please let me know if this is fine or should we switch to some other approach.

labkode:
> @labkode the PR is ready for review. Some initial testing is done, working on deploying it for further tests.
> 
> One thing which I'm not very sure about is [this](https://github.com/cs3org/reva/blob/44a380ba7464fe4693131cb407334fca40c95d83/pkg/storage/utils/eosfs/eosfs.go#L1212) part. Whenever we stat for resources, EOS returns the UID, which we encode as "uid:xyz" since we have no way of resolving it at the fs level. This is done at the [gateway](https://github.com/cs3org/reva/blob/44a380ba7464fe4693131cb407334fca40c95d83/internal/grpc/services/gateway/storageprovider.go#L851) layer to get the user ID back from the UID using [this method](https://github.com/cs3org/reva/blob/44a380ba7464fe4693131cb407334fca40c95d83/internal/grpc/services/gateway/userprovider.go#L97). I modified the GetUser method to return the user based on UID passed as opaque.
> 
> Please let me know if this is fine or if we should switch to some other approach.

What about having the EOS driver calling the userprovider grpc endpoints to resolve the uid to username and then pass it to the upper layers? This approach will avoid to have the `uid:` hack but will add a dependency to another grpc service. What do you think? Also @butonic ?

ishank011:
@labkode yes that could be an approach as well. Would save a lot of the changes I had to do at the upper levels.

labkode:
@ishank011 let's try then that approach, the drivers that need extra user information for implementing their logic they can connect to the gateway and request for specific user metadata.

Can you adapt that part to contact the Gateway API instead?

ishank011:
@labkode Cool, I'll make the change. Thanks!

ishank011:
Alternate approach #995 merged.

PR #990, opened 2020-07-22
use _ for mapstructure

butonic:
to align it with the other config options. no changelog needed as the original PR has a changelog and is unreleased.


kulmann:
post mortem: you could always add the url of a new PR to an existing (unreleased) changelog item, to make it more complete.

PR #995, opened 2020-07-23
Add UID and GID to the user object from EOS fs

ishank011:
This is an alternative approach to #983 where the gateway is available to EOS fs, from where it can resolve UIDs back to users, instead of passing it to the gateway layers.

### DIFF
--- a/internal/grpc/services/userprovider/userprovider.go
+++ b/internal/grpc/services/userprovider/userprovider.go
@@ -127,7 +127,7 @@ func (s *service) GetUserByClaim(ctx context.Context, req *userpb.GetUserByClaim
 		return res, nil
 	}
 
-	res := &userpb.GetUserByClaim{
+	res := &userpb.GetUserByClaimResponse{
 		Status: status.NewOK(ctx),
 		User:   user,
 	}
--- a/internal/grpc/services/userprovider/userprovider.go
+++ b/internal/grpc/services/userprovider/userprovider.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
-	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/pkg/rgrpc"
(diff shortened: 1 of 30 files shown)


internal/grpc/services/userprovider/userprovider.go
```diff
@@ -99,19 +100,36 @@ func (s *service) Register(ss *grpc.Server) {
...
 func (s *service) GetUser(ctx context.Context, req *userpb.GetUserRequest) (*userpb.GetUserResponse, error) {
-	user, err := s.usermgr.GetUser(ctx, req.UserId)
-	if err != nil {
-		// TODO(labkode): check for not found.
-		err = errors.Wrap(err, "userprovidersvc: error getting user")
-		res := &userpb.GetUserResponse{
-			Status: status.NewInternal(ctx, err, "error authenticating user"),
+
+	// Check if we need to retrieve user from UID or not
+	uid, err := extractUID(req.Opaque)
```
labkode, line 105: To avoid this, we should create a new method on the CS3APIS to return a User by using other claims, for example:

```
GetUserByClaim
```

labkode:
@butonic we'll introduce a new call in the CS3APIS to getUsers by arbitrary claims, like uid or usernames (for those who use them).

ishank011:
@labkode made the change. Please take a look.

pkg/eosclient/eosclient.go
```diff
@@ -269,76 +265,99 @@ func (c *Client) executeEOS(ctx context.Context, cmd *exec.Cmd) (string, string,
...
 	if err != nil {
 		return err
 	}
 
-	// since EOS Citrine ACLs are stored with uid, we need to convert username to uid
-	if aclType == acl.TypeUser {
-		recipient, err = getUID(recipient)
+	var cmd *exec.Cmd
+	if version == versionCitrine {
+		sysACL := a.CitrineSerialize()
```
labkode, line 335: This variable should be of the form: `%s:%s=`, to remove the entry:

```
 acl --user u:1010= /eos/dev
```

ishank011: As discussed, permissions would be an empty string when called from RemoveACL so it would be empty

PR #996, opened 2020-07-23
Split LDAP user filters

butonic:
The current LDAP user filters only allow a single `%s` to be replaced with the relevant string. I moved to filter templates that can use the CS3 user id properties `{{.OpaqueId}}` and `{{.Idp}}`. Furthermore,
I introduced a new find filter that is used when searching for users. An example config would be
```
userfilter = "(&(objectclass=posixAccount)(|(ownclouduuid={{.OpaqueId}})(cn={{.OpaqueId}})))"
findfilter = "(&(objectclass=posixAccount)(|(cn={{query}}*)(displayname={{query}}*)(mail={{query}}*)))"
```
As you can see the userfilter is used to lookup a single user, whereas the findfilter is for a broader search, e.g. used when searching for share recipients.

Related: https://github.com/cs3org/reva/issues/326

### DIFF
--- a/changelog/unreleased/split-ldap-filters.md
+++ b/changelog/unreleased/split-ldap-filters.md
@@ -1,0 +1,11 @@
+Enhancement: Split LDAP user filters
+
+The current LDAP user filters only allow a single `%s` to be replaced with the relevant string. We moved to filter templates that can use the CS3 user id properties `{{.OpaqueId}}` and `{{.Idp}}`. Furthermore,
+we introduced a new find filter that is used when searching for users. An example config would be
+```
+userfilter = "(&(objectclass=posixAccount)(|(ownclouduuid={{.OpaqueId}})(cn={{.OpaqueId}})))"
+findfilter = "(&(objectclass=posixAccount)(|(cn={{query}}*)(displayname={{query}}*)(mail={{query}}*)))"
+```
+As you can see the userfilter is used to lookup a single user, whereas the findfilter is for a broader search, e.g. used when searching for share recipients.
+
+https://github.com/cs3org/reva/pull/996
--- a/changelog/unreleased/split-ldap-filters.md
+++ b/changelog/unreleased/split-ldap-filters.md
@@ -1,11 +1,25 @@
 Enhancement: Split LDAP user filters
 
-The current LDAP user filters only allow a single `%s` to be replaced with the relevant string. We moved to filter templates that can use the CS3 user id properties `{{.OpaqueId}}` and `{{.Idp}}`. Furthermore,
-we introduced a new find filter that is used when searching for users. An example config would be
+The current LDAP user and auth filters only allow a single `%s` to be replaced with the relevant string.
+The current `userfilter` is used to lookup a single user, search for share recipients and for login. To make each use case more flexible we split this in three and introduced templates.
+
+For the `userfilter` we moved to filter templates that can use the CS3 user id properties `{{.OpaqueId}}` and `{{.Idp}}`:
 ```
 userfilter = "(&(objectclass=posixAccount)(|(ownclouduuid={{.OpaqueId}})(cn={{.OpaqueId}})))"
+```
+
+We introduced a new `findfilter` that is used when searching for users. Use it like this:
+```
 findfilter = "(&(objectclass=posixAccount)(|(cn={{query}}*)(displayname={{query}}*)(mail={{query}}*)))"
 ```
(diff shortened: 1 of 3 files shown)


pkg/user/manager/ldap/ldap.go
```diff
@@ -128,8 +154,8 @@ func (m *manager) GetUser(ctx context.Context, uid *userpb.UserId) (*userpb.User
 	searchRequest := ldap.NewSearchRequest(
 		m.baseDN,
 		ldap.ScopeWholeSubtree, ldap.NeverDerefAliases, 0, 0, false,
-		fmt.Sprintf(m.userfilter, uid.OpaqueId), // TODO this is screaming for errors if filter contains >1 %s
-		[]string{m.schema.DN, m.schema.UID, m.schema.Mail, m.schema.DisplayName},
+		m.getUserFilter(uid),
+		[]string{m.schema.DN, m.schema.CN, m.schema.UID, m.schema.Mail, m.schema.DisplayName},
```
butonic, line 158: we add the cn attribute so we can return a user with an id that is different from the username

pkg/user/manager/ldap/ldap.go
```diff
@@ -154,7 +180,7 @@ func (m *manager) GetUser(ctx context.Context, uid *userpb.UserId) (*userpb.User
 	}
 	u := &userpb.User{
 		Id:          id,
-		Username:    sr.Entries[0].GetAttributeValue(m.schema.UID),
+		Username:    sr.Entries[0].GetAttributeValue(m.schema.CN),
```
butonic, line 183: here we use the previously fetched cn as the username. they can be configured to the same attribute and the code will work as before

pkg/user/manager/ldap/ldap.go
```diff
@@ -244,13 +270,17 @@ func (m *manager) GetUserGroups(ctx context.Context, uid *userpb.UserId) ([]stri
 	groups := []string{}
 
 	for _, entry := range sr.Entries {
+		// FIXME this makes the users groups use the cn, not an immutable id
+		// FIXME 1. use the memberof or members attribute of a user to get the groups
+		// FIXME 2. ook up the id for each group
```
butonic, line 275: this needs to be cleaned up to actually fetch groups. AFAIU it will require the memberof overlay. In any case the member attribute needs to be made configurable as well. As part of a subsequent PR.

butonic:
@ishank011 good catch!
Updated the LDAP auth manager, and inlined the config.

The `m.c.Schema.UID` corresponds to the `user.Id.OpaqueId`, yes. `m.c.Schema.UID` corresponds to the `user.Username`. This PR allows configuring the userfilter in a way that the storage drivers can look up the user in LDAP ether with the username or with the userid: `(&(objectclass=posixAccount)(|(ownclouduuid={{.OpaqueId}})(cn={{.OpaqueId}}))`

I think we need to introduce a new api call to properly handle that though, tracked in https://github.com/cs3org/reva/pull/998

changelog/unreleased/split-ldap-filters.md
```diff
@@ -0,0 +1,25 @@
+Enhancement: Split LDAP user filters
+
+The current LDAP user and auth filters only allow a single `%s` to be replaced with the relevant string.
+The current `userfilter` the userfilter` is used to lookup a single user, search for share recipients and for login. To make each use case more flexible we split this in three and introduced templates.
```
ishank011, line 4: I think there's a typo here

pkg/user/manager/ldap/ldap.go
```diff
@@ -97,39 +101,51 @@ func New(m map[string]interface{}) (user.Manager, error) {
...
-		bindPassword: c.BindPassword,
-		idp:          c.Idp,
-		schema:       c.Schema,
-	}, nil
+	// backwards compatibility
+	c.UserFilter = strings.ReplaceAll(c.UserFilter, "%s", "{{.OpaqueId}}")
+	if c.FindFilter == "" {
+		c.FindFilter = c.UserFilter
+	}
+	c.GroupFilter = strings.ReplaceAll(c.GroupFilter, "%s", "{{.OpaqueId}}")
```
ishank011, line 109: Fallback to UserFilter if it's empty?

butonic: The group filter is completely broken imo. This pr just replaces the `%s` in case the config wes not updated. I will look at group related stuff next week.

ishank011: Sure!

PR #1000, opened 2020-07-24
meshdirectory: Add invite forward API to provider links

mirekys:
Enables forwarding of share invite to the provider selected in meshdirectory

Fixes [sciencemesh#139](https://github.com/sciencemesh/sciencemesh/issues/139)


examples/meshdirectory/static/index.html
```diff
@@ -76,12 +76,13 @@ <h5>Service description:</h5>
         },
         methods: {
             providerLink(provider) {
-                const query = this.queryParams.toString() ? `?${this.queryParams}` : ''
-                return `${this.ocmService(provider).endpoint.path}${query}`
+                const query = this.queryParams.toString() ? `?${this.queryParams}` : '';
+                const inviteApi = `${this.ocmService(provider).endpoint.path}/invites/forward`;
```
ishank011, line 80: Can we use ``const inviteApi = new URL('/invites/forward', `${this.ocmService(provider).endpoint.path}`);``? That'll take care of double slashes.

mirekys: @ishank011, I tried that initially, but that seems to always take away any path that may be already present in the provider endpoint url (`/iop/ocm`) and replace it with `/invites/forward`. I'll try to take care of that :)

ishank011: Oh okay. We can skip it then. No need to put too much effort into this

PR #1008, opened 2020-07-27
Use proper logging for ldap auth requests

butonic:
Instead of logging to stdout we now log using debug level logging or an error.


changelog/unreleased/sane-ldap-auth-logging.md
```diff
@@ -0,0 +1,5 @@
+Enhancement: Use proper logging for ldap auth requests
+
+Instead of logging to stdout we now log using debug level logging or an error.
```
refs, line 3: "error level logging"

PR #1009, opened 2020-07-27
 Remove protocol from partner domains to match gocdb config

ishank011:


examples/ocm-partners/users-ailleron.json
```diff
@@ -13,7 +13,7 @@
 	{
 		"id": {
 			"opaque_id": "mateusz5678",
-			"idp": "https://ailleron.com"
+			"idp": "ailleron.com"
```
mirekys, line 16: Hi @ishank011 thanks for updating the configs! There is just one small issue, Ailleron uses `softwaremind.com` domain in GOCDB now.

ishank011: @mirekys Will fix it. Thanks!

PR #1015, opened 2020-07-28
[Tests-Only] Bump commit id for oC10APIAcceptanceTests

phil-davis:
It gets us test scenarios that have been added/changed over the past week:
https://github.com/owncloud/core/pull/37717 Add EXPECTED_FAILURES_FILE capability to acceptance tests run.sh (I  will make a proposed PR to use that capability later this week)
https://github.com/owncloud/core/pull/37721 Modify functions to check the presence of files and folders - does more strict checking when a resource should be a file and not a folder or vice-versa
https://github.com/owncloud/core/pull/37740 Remove skipOnOcis tag for scenarios not to be skipped on ocis
https://github.com/owncloud/core/pull/37742 Fix tests for ocis with proxy


phil-davis:
The changelog step passed for thei `Tests-Only` PR - great stuff.

PR #1035, opened 2020-07-30
Refactor AppProvider workflow and protocol

### LINKED ISSUE #991: UIURL not parsed in parseConfig of appprovider
found the issue while doing some simple unit tests, see https://github.com/cs3org/reva/pull/992

glpatcern:
This PR uses https://github.com/cs3org/cs3apis/pull/85 to implement a simpler protocol from the client to the Reva gateway, and a more enriched one from the gateway to the appprovider. For now, WOPI is still the only "driver" implementing the appprovider logic.

Also, `storageID` and `UIURL` are removed from the app-provider's config: `StorageID` is auto-resolved in the gateway service, whereas the `UIURL`'s doc was misleading: we need to decide whether to drop it entirely (and surely it won't stay in the config).

En passant, this fixes #991 and includes a few other changes and fixes concerning error handling, logging and documentation.


labkode:
@glpatcern cs3apis merged, remember changelog

labkode:
@glpatcern looks already pretty good

internal/grpc/interceptors/recovery/recovery.go
```diff
@@ -45,8 +45,8 @@ func NewStream() grpc.StreamServerInterceptor {
 }
 
 func recoveryFunc(ctx context.Context, p interface{}) (err error) {
-	stack := debug.Stack()
+	debug.PrintStack()
```
ishank011, line 48: Nice!

internal/grpc/services/appprovider/appprovider.go
```diff
@@ -154,50 +146,41 @@ func (s *service) OpenFileInAppProvider(ctx context.Context, req *providerpb.Ope
 		return nil, err
 	}
 
-	httpReq, err := rhttp.NewRequest(ctx, "GET", wopiurl+"wopi/iop/open", nil)
+	httpReq, err := rhttp.NewRequest(ctx, "GET", s.conf.WopiURL+"wopi/iop/open", nil)
```
ishank011, line 149: Can we take care of missing trailing slashes here? Something like [this](https://github.com/cs3org/reva/blob/master/pkg/ocm/invite/manager/json/json.go#L210)

internal/grpc/interceptors/recovery/recovery.go
glpatcern: ;-) this was needed as I got a full stack printed with literal `\n`'s all over!

internal/grpc/services/appprovider/appprovider.go
glpatcern: Sure, makes sense

glpatcern: How does it look like? I note that the block with the HTTP request to `/wopi/cbox/endpoints` should be moved elsewhere, in some kind of init of the service. Maybe I could move it to the `init()` function? or where else?

glpatcern: Answering to myself, in fact I found a way to do that with a separate function. This appprovider will eventually be the WOPI driver, so it's OK to have it like this - commit to follow.

internal/grpc/services/appprovider/appprovider.go
```diff
@@ -48,8 +49,9 @@ func init() {
 }
 
 type service struct {
-	provider app.Provider
-	conf     *config
+	provider   app.Provider
+	conf       *config
+	appsURLMap map[string]interface{}
```
ishank011, line 54: This needs to be protected by a lock. Multiple threads can try to fill this at the same time, and even though they would probably be writing the same data, there's no ordering in a map, so we might end up with garbage values.

glpatcern: Well spotted, too much Python coding ;-) I'll fix that.

ishank011: You can consider using a sync.Map like done [here](https://github.com/cs3org/reva/blob/master/pkg/ocm/provider/authorizer/json/json.go#L83).

glpatcern: In fact, given the usage (write once, read many), for now I went for a standard `mutex.Lock()` approach. Just committed.

glpatcern: (true that once a regular refresh is implemented, the best option is a sync.Map)

ishank011: This won't work. The lock is local to each goroutine, so it won't know what it actually has to take a lock on. Ref: https://play.golang.org/p/RZfwVALEvRn. You'd have to add this lock to the service struct itself.

And why not use sync.Map in the first place since we agree it would be the better option in the long run?

glpatcern: (Thanks for your patience, I genuinely assumed Go had some higher level constructs, like mutexes being stored in some common memory. Instead, it's pretty much like C/C++, good to know)

OK, I gave it a try with sync.Map but it's not a good fit as it's not a real map; and for the long run, we want to invalidate its content and call that function again to fill it. So at that point I'll implement a proper multi-read/single-write lock for it, but for this PR, I came back to all local variables. I guess it's better to move on for now and leave that as a separate task.

internal/grpc/interceptors/recovery/recovery.go
labkode, line 48: @glpatcern can you revert this change? Even though is useful it will break log parsing. Let's create a github issue to proper address it.

glpatcern: Interestingly, Ishank marked this "nice", and was motivated by my comment above: "I got a full stack printed with literal \n's all over!". Why would this break log parsing? Because it's multiline?

glpatcern: Opened #1054 to track this.

PR #1037, opened 2020-07-30
Added a system information HTTP service

Daniel-WWU-IT:
The new service exposes system information via an HTTP endpoint. This currently only includes Reva version information (which isn't working atm in Reva's build system as it seems). The information are exposed in the form of Prometheus metrics so that we will be able to gather these in a streamlined way with some Mentix extensions yet to come.


ishank011:
Can we use the same exporter as we do for the [prometheus HTTP service](https://github.com/cs3org/reva/blob/master/internal/http/services/prometheus/prometheus.go)? That is a lot cleaner and since both of them are running similar services, it would be better to have them use the same packages.

@labkode @Daniel-WWU-IT

Daniel-WWU-IT:
Are you referring to using the `contrib.go.opencensus.io/exporter/prometheus` package? Not sure, from the name I'd say no. However, they are using the official Go client library just as I do, so indirectly they are using the same packages.

labkode:
@ishank011 that is a very good point. I keep track of it here: https://github.com/cs3org/reva/issues/1070
I'll add it to the new sprint.

PR #1051, opened 2020-08-04
Disallow sharing the shares directory

ishank011:


labkode:
@ishank011 please keep track of your commented section in a gihub issue if not already the case. That code is legimitate and there must be a way to fit to work in the oc tests.

ishank011:
@labkode Yes, I described the issue here https://github.com/cs3org/reva/issues/1053#issuecomment-669205784, it's related to the problem of looking up the user from username/UID, which @butonic and @PVince81 are working to solve. We get around the problem by checking for both the owner and creator of the share (which ideally should be the case anyway).

PR #1033, opened 2020-07-30
ocfs: lookup user to render template properly

butonic:
Currently, the username is used to construct paths, which breaks when mounting the owncloud storage driver at /oc and then expecting paths that use the username like `/oc/einstein/foo` to work, because they will mismatch the path that is used from propagation which uses `/oc/u-u-i-d` as the root, giving a `internal path outside root` error like this:

```
11:25AM ERR home/jfd/go/pkg/mod/github.com/cs3org/reva@v0.1.1-0.20200728071211-c948977dd3a0/pkg/storage/fs/owncloud/owncloud.go:1885 > could not propagate change error="internal path outside root" leafPath=/var/tmp/reva/data/4c510ada-c86b-4815-8820-42cdf82c3d51/files/TUS-upload.png pid=32062 root=/var/tmp/reva/data/einstein/files
[tusd] 2020/07/30 11:25:42 event="ResponseOutgoing" status="500" method="PATCH" path="/5137f8b9-cbad-40a2-a20b-e06a8bfd5a5a" error="internal path outside root" requestId=""
2020-07-30T11:25:42+02:00 ERR http end="30/Jul/2020:11:25:42 +0200" host=::1 method=PATCH pkg=rhttp proto=HTTP/1.1 service=reva size=27 start="30/Jul/2020:11:25:42 +0200" status=500 time_ns=912434 traceid=3c125790948fbeb2499020acafb12108 uri=/data/5137f8b9-cbad-40a2-a20b-e06a8bfd5a5a url=/data/5137f8b9-cbad-40a2-a20b-e06a8bfd5a5a
[...]

PR #1052, opened 2020-08-04
Ocfs lookup userid (update)

PVince81:
Resend of https://github.com/cs3org/reva/pull/1033

also to be able to continue working to fix the issues, if they still exist.

### DIFF
--- a/drone/oc-integration-tests/storage-oc.toml
+++ b/drone/oc-integration-tests/storage-oc.toml
@@ -23,6 +23,7 @@ data_server_url = "http://localhost:11001/data"
 [grpc.services.storageprovider.drivers.owncloud]
 datadirectory = "/drone/src/tmp/reva/data"
 redis = "redis:6379"
+userprovidersvc = "localhost:18000"
 
 [http]
 address = "0.0.0.0:11001"
--- a/examples/oc-phoenix/storage-oc.toml
+++ b/examples/oc-phoenix/storage-oc.toml
@@ -22,7 +22,8 @@ data_server_url = "http://localhost:11001/data"
 
 [grpc.services.storageprovider.drivers.owncloud]
 datadirectory = "/var/tmp/reva/data"
-
+redis = "redis:6379"
+userprovidersvc = "localhost:18000"
 
 [http]
 address = "0.0.0.0:11001"
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -179,7 +179,7 @@ func (c *config) init(m map[string]interface{}) {
 		c.Redis = ":6379"
 	}
 	if c.UserLayout == "" {
-		c.UserLayout = "{{.Username}}"
+		c.UserLayout = "{{.Id.OpaqueId}}"
 	}
 	if c.UploadInfoDir == "" {
 		c.UploadInfoDir = "/var/tmp/reva/uploadinfo"
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -41,6 +41,7 @@ import (
 	"github.com/cs3org/reva/pkg/logger"
 	"github.com/cs3org/reva/pkg/mime"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
+	"github.com/cs3org/reva/pkg/sharedconf"
 	"github.com/cs3org/reva/pkg/storage"
 	"github.com/cs3org/reva/pkg/storage/fs/registry"
 	"github.com/cs3org/reva/pkg/storage/utils/templates"
@@ -154,13 +155,14 @@ func init() {
 }
 
 type config struct {
-	DataDirectory  string `mapstructure:"datadirectory"`
-	UploadInfoDir  string `mapstructure:"upload_info_dir"`
-	ShareDirectory string `mapstructure:"sharedirectory"`
-	UserLayout     string `mapstructure:"user_layout"`
-	Redis          string `mapstructure:"redis"`
-	EnableHome     bool   `mapstructure:"enable_home"`
-	Scan           bool   `mapstructure:"scan"`
+	DataDirectory             string `mapstructure:"datadirectory"`
+	UploadInfoDir             string `mapstructure:"upload_info_dir"`
+	ShareDirectory            string `mapstructure:"sharedirectory"`
+	UserLayout                string `mapstructure:"user_layout"`
+	Redis                     string `mapstructure:"redis"`
+	EnableHome                bool   `mapstructure:"enable_home"`
+	Scan                      bool   `mapstructure:"scan"`
+	UserShareProviderEndpoint string `mapstructure:"usershareprovidersvc"`
 }
 
 func parseConfig(m map[string]interface{}) (*config, error) {
@@ -189,6 +191,7 @@ func (c *config) init(m map[string]interface{}) {
 	if _, ok := m["scan"]; !ok {
 		c.Scan = true
 	}
+	c.UserShareProviderEndpoint = sharedconf.GetGatewaySVC(c.UserShareProviderEndpoint)
 }
 
 // New returns an implementation to of the storage.FS interface that talk to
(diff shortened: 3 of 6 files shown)


PVince81:
the same tests fail on the reva repo, but they pass when I check out the whole of https://github.com/owncloud/ocis/pull/409 and the connected deps.

to align with ocis-reva defaults, I've pushed a commit that changes the default user layout for OCFS to use the opaque id

@butonic FYI

if tests pass then it will confirm that it's the ocis-reva change that makes it work.

if changing the default in Reva is fine, we can keep it here

PVince81:
litmus tests on new dav failing now:
```
4. propfind_d0........... WARNING: response href for wrong resource
```

it is likely that this will also fail in https://github.com/owncloud/ocis/pull/409 then, but there is no litmus run there apparently, not sure if we should add it on that level @individual-it

PVince81:
same failures as before. so there is more discrepancies in the setup

PVince81:
Litmus test runs fine locally with this branch and https://github.com/owncloud/ocis/pull/409. Strange...

PVince81:
Pushed another missing setting

PVince81:
- [ ] investigate whether we could go to the gateway instead

PVince81:
nice, the api tests pass.
next up: check that litmus failure

PVince81:
Right, the Litmus failure is because the test run is using "einstein" as user instead of the uuid.

If I run the test locally with "einstein" in the DAV URL I get the same failure.
But the following makes everything pass:
```
docker run -e LITMUS_URL="http://172.17.0.1:20080/remote.php/dav/files/4c510ada-c86b-4815-8820-42cdf82c3d51" -e LITMUS_USERNAME="einstein" -e LITMUS_PASSWORD="relativity" -e TESTS="props" owncloud/litmus:latest
```

now how to get this URL in Drone ?

it seems that UUID is hard-coded here https://github.com/owncloud/ocis-accounts/blob/8de0530f31ed5ffb0bbb7f7f3471f87f429cb2ea/pkg/service/v0/service.go#L45

so would be safe to use for now... I'll adjust Drone...

refs:
Rebasing https://github.com/cs3org/reva/pull/1064 onto this fixes the createShare test scenarios through the provisioning API /cc @individual-it. To test it:

```
# on this branch
git rebase fix-error-propagation
```

Replace on OCIS
run: 

```
make test-acceptance-api TEST_SERVER_URL=https://localhost:9200 TEST_OCIS=true OCIS_REVA_DATA_ROOT=/var/tmp/reva/ SKELETON_DIR=apps/testing/data/apiSkeleton BEHAT_FILTER_TAGS='~@skipOnOcis&&~@skipOnOcis-OC-Storage' TESTING_REMOTE_SYSTEM=true BEHAT_FEATURE=tests/acceptance/features/apiShareManagementBasic/createShare.feature
```

PVince81:
This PR here is ready for review @labkode @butonic @refs

drone/oc-integration-tests/storage-oc.toml
```diff
@@ -23,6 +23,7 @@ data_server_url = "http://localhost:11001/data"
 [grpc.services.storageprovider.drivers.owncloud]
 datadirectory = "/drone/src/tmp/reva/data"
 redis = "redis:6379"
+userprovidersvc = "localhost:18000"
```
butonic, line 26: Setting this will save a trip to the gateway. When left out it should fall back to the `gatewaysvc` in the `[shared]` section and work just as well.

examples/oc-phoenix/storage-oc.toml
```diff
@@ -22,7 +22,8 @@ data_server_url = "http://localhost:11001/data"
 
 [grpc.services.storageprovider.drivers.owncloud]
 datadirectory = "/var/tmp/reva/data"
-
+redis = "redis:6379"
+userprovidersvc = "localhost:18000"
```
butonic, line 26: Setting this will save a trip to the gateway. When left out it should fall back to the `gatewaysvc` in the `[shared]` section and work just as well.

examples/oc-phoenix/storage-oc.toml
```diff
@@ -22,7 +22,8 @@ data_server_url = "http://localhost:11001/data"
 
 [grpc.services.storageprovider.drivers.owncloud]
 datadirectory = "/var/tmp/reva/data"
-
+redis = "redis:6379"
```
butonic, line 25: setting `redis` is optional. it defaults to `localhost:6379`

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -175,7 +179,7 @@ func (c *config) init(m map[string]interface{}) {
 		c.Redis = ":6379"
 	}
 	if c.UserLayout == "" {
-		c.UserLayout = "{{.Username}}"
+		c.UserLayout = "{{.Id.OpaqueId}}"
```
butonic, line 182: This changes the default, but will prevent username changes from breaking the storage lookup.

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -152,13 +155,14 @@ func init() {
...
-	EnableHome     bool   `mapstructure:"enable_home"`
-	Scan           bool   `mapstructure:"scan"`
+	DataDirectory        string `mapstructure:"datadirectory"`
+	UploadInfoDir        string `mapstructure:"upload_info_dir"`
+	ShareDirectory       string `mapstructure:"sharedirectory"`
+	UserLayout           string `mapstructure:"user_layout"`
+	Redis                string `mapstructure:"redis"`
+	EnableHome           bool   `mapstructure:"enable_home"`
+	Scan                 bool   `mapstructure:"scan"`
+	UserProviderEndpoint string `mapstructure:"userprovidersvc"`
```
butonic, line 165: We need the `UserProviderEndpoint` to look up a user by its username.

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -473,6 +509,42 @@ func (fs *ocfs) getOwner(internal string) string {
...
+	}
+	// look up at the userprovider
+
+	// parts[0] contains the username or userid. use  user service to look up id
+	c, err := pool.GetUserProviderServiceClient(fs.c.UserProviderEndpoint)
+	if err != nil {
+		return nil, err
+	}
+	res, err := c.GetUser(ctx, &userpb.GetUserRequest{
+		UserId: &userpb.UserId{OpaqueId: usernameOrID},
```
butonic, line 530: When configuring the userprovider with ldap you need to configure a filter that checks both: the username attribute as well as the userid attribute.

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -881,6 +960,7 @@ func (fs *ocfs) ListGrants(ctx context.Context, ref *provider.Reference) (grants
 	grants = make([]*provider.Grant, 0, len(aces))
 	for i := range aces {
 		grantee := &provider.Grantee{
+			// TODO lookup uid from principal
```
butonic, line 963: - [ ] The user lookup here needs to be implemented before the sharing tests can pass. at least I think they should fail somewhere without it.

butonic:
@refs @PVince81 the TODO in the code for principal lookup is the only thing I think that needs to be implemented. I would expect the sharing tests to fail somewhere because of https://github.com/cs3org/reva/pull/1052#discussion_r467785480

drone/oc-integration-tests/storage-oc.toml
PVince81: the fallback did not seem to work, I had to add this

there was a strange thing where it was resolving some service, but that service did not implement the UserAPI, see https://github.com/cs3org/reva/pull/1033#issuecomment-669823201

I didn't manage to find out what service it was resolving exactly

pkg/storage/fs/owncloud/owncloud.go
PVince81: I put this there to align to the change in ocis-reva which also changes the default. I hope it's ok.
Otherwise we'll have further discrepancies.

PVince81:
@butonic all changes that were made here contributed to make CI pass. Maybe we are missing tests for triggering https://github.com/cs3org/reva/pull/1052#discussion_r467785480 ?

PR #1059, opened 2020-08-05
Improve timestamp precision while logging

ishank011:
@labkode this probably won't break log-parsing in ocis as it overrides the format, but it'd good to confirm once.


labkode:
@butonic @refs please check if in OCIS the timestamp is similar. As far as I understand you pass your own logger to Reva.

PR #1036, opened 2020-07-30
[Tests-Only] Implement expected failures file

phil-davis:
Implements `EXPECTED_FAILURES_FILE` the same as https://github.com/owncloud/ocis-reva/pull/402

1) Adjust `.drone.star` so that it runs all the core API acceptance tests except `~@notToImplementOnOCIS&&~@toImplementOnOCIS` - i.e. except the tests specifically tagged in core PR https://github.com/owncloud/core/pull/37736

2) Specify an `EXPECTED_FAILURES_FILE` and put it in a list of all the acceptance test scenarios that are expected to fail. These are things that pass on oC10 but currently have a problem on cs3org/reva

3) Add the infrastructure to be able to run local API acceptance test scenarios (`Makefile` `test-acceptance-api` target, `composer.json`, `behat.yml`,...)

4) Add local scenarios that demonstrate existing bad behavior in OCIS/reva. These are copied from the scenarios that are currently in core and tagged `skipOnOcV10`. I have moved just a few of them to demonstrate this. I will move the rest of them in a later PR.

This PR helps with issue https://github.com/owncloud/ocis-reva/issues/282 - when some examples in the Example Table of a Scenario Outline pass and some fail, we will run all of them and the ones that fail can be listed in `EXPECTED_FAILURES_FILE`

It also removes the need to change core API acceptance test code or feature files as we fix OCIS/reva issues. We can manage the local acceptance tests here in the local repo, and also can manage the list of expected failures here in the local repo (hopefully the list gets shorter as we implement and fix things!)

PR #1063, opened 2020-08-06
[Tests-Only] Bump core commit for tests and avoid preview-extension-required scenarios

phil-davis:
core PR https://github.com/owncloud/core/pull/37777 was merged. It adds a tag `preview-extension-required` to tag test scenarios for previews.

1) Bump the core commit id to the latest commit.
2) Adjust the oC10APIAcceptanceTests step so that it does not run scenarios with the `preview-extension-required` tag

Previews are not implemented yet, and the implementation detail remains to be sorted out, so do not accidentally run any of those test scenarios.

Note: similar in ocis-reva is https://github.com/owncloud/ocis-reva/pull/426 https://github.com/owncloud/ocis-reva/pull/428

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -343,7 +343,7 @@ steps:
       - git clone -b master --depth=1 https://github.com/owncloud/testing.git /drone/src/tmp/testing
       - git clone -b master --single-branch --no-tags https://github.com/owncloud/core.git /drone/src/tmp/testrunner
       - cd /drone/src/tmp/testrunner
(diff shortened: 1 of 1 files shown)


phil-davis:
@labkode at the moment this seems a daily event, updating the core commit id!
As we change things in the `owncloud/core` API tests, it is good to know that it all still passes in the various `reva` repos. And once CI passes, it may as well be merged.

After merging this, I can rebase #1036 again - it will get a conflict for sure.

PR #1073, opened 2020-08-11
[tests-only] provide and document an easy way to run all tests locally

individual-it:
This should allow developers to run all kind of tests on their local developing machines and by that improve productivity


labkode (approved):

ishank011:
@individual-it this is really nice! Thanks a lot! It'd be great if we can add make rules for these as well.

individual-it:
I've opened two new issues to track that https://github.com/cs3org/reva/issues/1106 & https://github.com/cs3org/reva/issues/1107

PR #1079, opened 2020-08-12
eosclient: get file info by fxid

labkode:


labkode:
@ishank011 track here: https://github.com/cs3org/reva/issues/1080

PR #1082, opened 2020-08-12
Minor fixes in OCM shares, gateway uploads and smtpclient

ishank011:


ishank011:
@Daniel-WWU-IT I've also included the smtp login change in this PR.

labkode:
@PVince81 can you take a look at the TUS changes?

internal/http/services/dataprovider/put.go
```diff
@@ -59,12 +59,14 @@ func (s *svc) doTusPut(w http.ResponseWriter, r *http.Request) {
 
 	fp := r.Header.Get("File-Path")
 	if fp == "" {
+		log.Error().Msg("File-Path header not present")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
 
-	length, err := strconv.ParseInt(r.Header.Get("Content-Length"), 10, 64)
```
ishank011, line 66: @labkode @PVince81 the issue here is that `Content-Length` is set by the Go HTTP client itself. So setting it manually doesn't have any effect. I noticed this when trying to enable wopi uploads to go through the datagateway. The HTTP request sent by the python has this header, but when this request is forwarded by datagateway to dataprovider, despite explicitly copying the header, it is removed.

The same is the case with the [ocdav service](https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocdav/put.go#L205). cURL requests work but requests through the go client library, e.g., https://play.golang.org/p/AMHW1ekOlya, fail.

So I wanted to change the header name from which we read the size of the request body. Perhaps, `Upload-Length` might be a better alternative?

labkode: Makes sense

labkode (approved):

Issue #1081, opened 2020-08-12
Split local and oC10 API test pipelines


phil-davis:
Currently the API tests are run in a single pipeline - first the local API tests, then the oC10 API tests.

If the local API tests fail, then the pipeline exits with failure (that is correct). In that case you do not get to see what the result of the oC10 API tests would have been - which can be annoying when trying to get everything passing.

Separate the local and oC10 API test runs into separate pipelines.

If we do that in `.drone.yml` there will be copy-paste duplication of the system setup and services sections in each pipeline.

We could switch to having `.drone.star` in similar style to https://github.com/owncloud/ocis-reva/blob/master/.drone.star - that reduces duplication of common pipeline steps and services, and allows commonly-modified parameters to be put in a `config` section at the top where they can be easily found and edited.

Opinions about making `.drone.star` ?

phil-davis:
PR #1084 implements the easy cut-paste edit to get separate pipelines.

Moving to `.drone.star` needs to be discussed.

PR #1084, opened 2020-08-13
[Tests-Only] Run local integration tests in a separate pipeline

phil-davis:
Part of issue #1081 

This is the simple copy-paste edit of `.drone.yml` to run both a local and owncloud integration tests pipeline.

Benefits:
1) get a separate pass/fail result and log for each set of tests
2) a bit less elapsed time 

Note: when the oC10 core tests are run on their own (in a separate pipeline, so the local integration tests have not been run first) then these tests are passing:
```
apiShareOperations/accessToShare.feature:20
apiShareOperations/accessToShare.feature:21
apiShareOperations/accessToShare.feature:34
apiShareOperations/accessToShare.feature:35
```

So there is some side-effect left behind by the local integration tests that was causing those `accessToShare` tests to fail. That needs investigation - issue #1085 and https://github.com/owncloud/ocis-reva/issues/438

But that is an existing issue that happens the same in `owncloud/ocis-reva` so IMO this PR can still move forward.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -308,7 +308,7 @@ steps:
 ---
 kind: pipeline
 type: docker
-name: owncloud-integration-tests
+name: local-integration-tests
 
 platform:
   os: linux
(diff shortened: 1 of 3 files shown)


labkode:
@phil-davis rebase

phil-davis:
Rebased, ready for review.

labkode (approved):

PR #1088, opened 2020-08-13
Implement GetUserByClaim for LDAP user driver

butonic:
The LDAP user driver can now fetch users by a single claim / attribute. Use an `attributefilter` like `(&(objectclass=posixAccount)({{attr}}={{value}}))` in the driver section.

partly fixes https://github.com/owncloud/ocis/issues/444


butonic:
I need to add the uid and gid to the opaque user properties ...

butonic:
With this PR and https://github.com/owncloud/ocis-proxy/pull/89 I can share files with other users when running ocis on top of eos.

PR #1064, opened 2020-08-06
Ensure stray shares get ignored

refs:
## Scope

JSON Shares Manager.

When using the json shares manager, it can be the case we found a share with a `resource_id` that no longer exists. This PR aims to address his case by changing the contract of `getPath` and return the result of the `STAT` call instead of a generic error, so we can check the cause.

This is by no means a perfect solution, but starts addressing the inconsistencies of the error handling across the code base.

PR #1090, opened 2020-08-13
Ignore Stray Public Shares

### LINKED ISSUE #1087: Fix error handling on the PublicShare manager
context: https://github.com/cs3org/reva/pull/1064

> When using the json shares manager, it can be the case we found a share with a resource_id that no longer exists. This PR aims to address his case by changing the contract of getPath and return the result of the STAT call instead of a generic error, so we can check the cause.

The same has to be essentially done for the public shares manager.

refs:
context: https://github.com/cs3org/reva/issues/1087. Closes #1087 

Separated from #1064 to keep changes atomic.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -1162,11 +1162,15 @@ func (h *Handler) listPublicShares(r *http.Request, filters []*link.ListPublicSh
 			}
 
 			statResponse, err := c.Stat(ctx, statRequest)
+			if err != nil {
+				return nil, err
+			}
 			if statResponse.Status.Code != rpc.Code_CODE_OK {
 				if statResponse.Status.Code == rpc.Code_CODE_NOT_FOUND {
-					// TODO share target was not found, we should not error here.
 					continue
 				}
+
+				return nil, errors.New(fmt.Sprintf("could not stat share target: %v, code: %v", share.ResourceId, statResponse.Status))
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -1162,14 +1162,14 @@ func (h *Handler) listPublicShares(r *http.Request, filters []*link.ListPublicSh
 			}
 
 			statResponse, err := c.Stat(ctx, statRequest)
-			if err != nil {
-				return nil, err
+			if statResponse.Status.Code != rpc.Code_CODE_OK {
+				if statResponse.Status.Code == rpc.Code_CODE_NOT_FOUND {
```
PVince81, line 1166: is status code is neither or nor not found, should still return `nil, err`, might need to format that error

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -1162,14 +1162,14 @@ func (h *Handler) listPublicShares(r *http.Request, filters []*link.ListPublicSh
 			}
 
 			statResponse, err := c.Stat(ctx, statRequest)
-			if err != nil {
```
PVince81, line 1165: we should keep detecting grpc call errors and returning `nil, err`

refs: agree 👍

PR #1094, opened 2020-08-14
OCM: Add logic for resolving storage references over webdav

ishank011:


examples/storage-references/gateway.toml
```diff
@@ -3,6 +3,9 @@
 commit_share_to_storage_grant = true
 commit_share_to_storage_ref = true
 
+[grpc.services.gateway.token_managers.jwt]
+expires = 86400
```
labkode, line 7: @ishank011 Add this as default

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -179,14 +179,30 @@ func (s *svc) InitiateFileDownload(ctx context.Context, req *provider.InitiateFi
...
 		}
 
+		if protocol == "webdav" {
+			// TODO(ishank011): pass this through the datagateway service
+			// for now, we just expose the file server to the user
+			ep, opaque, err := s.webdavRefTransferEndpoint(ctx, statRes.Info.Target, shareChild)
+			if err != nil {
+				return &gateway.InitiateFileDownloadResponse{
+					Status: status.NewInternal(ctx, err, "gateway: error creating container on webdav host: "+p),
+				}, nil
```
labkode, line 197: error message should be: error downloading ....

pkg/storage/utils/localfs/localfs.go
```diff
@@ -1147,14 +1147,14 @@ func (fs *localfs) RestoreRecycleItem(ctx context.Context, restoreKey string) er
 func (fs *localfs) propagate(ctx context.Context, leafPath string) error {
 
 	var root string
-	if fs.isShareFolderChild(ctx, leafPath) {
+	if fs.isShareFolderChild(ctx, leafPath) || strings.HasSuffix(path.Clean(leafPath), fs.conf.ShareFolder) {
 		root = fs.wrapReferences(ctx, "/")
```
labkode, line 1151: @ishank011 what is the meaning of this?

examples/storage-references/gateway.toml
ishank011: Will do

internal/grpc/services/gateway/storageprovider.go
ishank011: My bad

pkg/storage/utils/localfs/localfs.go
ishank011: This takes care of the corner case when the leaf path is the shares directory itself. Since it isn't a share child, the root we assign to it is the data directory which fails since that is not a prefix of the references directory.

labkode (approved):

Issue #1070, opened 2020-08-10
Expose all prometheus metrics under /metrics 


labkode:
@Daniel-WWU-IT @ishank011 I agree that we should expose all the necessary info under one endpoint only.

Daniel-WWU-IT:
Makes sense, and also ensures that every site exposes the necessary system info. I will merge the sysinfo service into the Prometheus service right away.

Daniel-WWU-IT:
@labkode @ishank011 See PR #1071 - the system information is now included in the main metrics.

I tried to use the OpenCensus Prometheus API, though I cannot add custom labels to a single metric; I could only add them to _all_ exported metrics, which would result in a mess. There are other ways to export metrics which include labels, though these can't be combined with what is already there. So I chose to stick with using the Prometheus API directly (this is what OpenCensus does internally just as well); I only had to create a Prometheus registry shared by OpenCensus and my SysInfo system, which seems acceptable to me.

ishank011:
@Daniel-WWU-IT after discussions, we decided that we do not want to have info about or carry out operations for any stats that some service or package is exposing in the prometheus HTTP service. The current approach forces us to pass the prometheus client registry to the sysinfo package, and as we continue adding stats from different services, we'd have to do this for all of those and this wouldn't scale well.

So please take a look at this PR https://github.com/cs3org/reva/pull/1105 which uses the opencensus stats package to add labels. We can modify the sysinfo package to work this way and this will take away all dependencies from the HTTP service.

@redblom on similar lines, this is the approach we want to follow for the metrics package as well. The prometheus client would not impose any restrictions (regarding the driver) on any services. Each of those can just register themselves. Let's follow up on your PR regarding this.

Daniel-WWU-IT:
@ishank011 I took a look at your PR and this seems like a good solution. I will incorporate the necessary changes and create a new PR.

redblom:
@Daniel-WWU-IT @ishank011  : I was having a discussion on Gitter with @ishank011 regarding the accounting metrics part and I would like to proceed with that here (@Daniel-WWU-IT  I'm aware that you may already started with additional coding based on PR #1105 on your part). 
I noticed that code is being moved around a bit due to changing insight. Anyways both accounting and sysinfo metrics should follow the same approach and implementation so I thought it would be wise to have a common agreement on that.

redblom:
[...]

PR #1114, opened 2020-08-21
The system info metrics now use OpenCensus

Daniel-WWU-IT:
As discussed in https://github.com/cs3org/reva/pull/1105 and https://github.com/cs3org/reva/issues/1070, the system info metrics are now exposed via OpenCensus instead of the Prometheus client library.


pkg/sysinfo/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+
+	"github.com/cs3org/reva/pkg/utils"
+)
+
+type sysInfoMetricsLabels = map[tag.Key]string
+
+type sysInfoMetrics struct {
```
ishank011, line 37: Do we need the struct and the global variable? Let's just get the labels and register them.

ishank011:
Also, please fix the changelog

Daniel-WWU-IT:
Done and done.

ishank011 (approved):

PR #1118, opened 2020-08-25
Read metrics from file (v2)

redblom:
@ishank011 : Metrics initializes itself, performs periodic data recording, uses sharedconfig only.
Example json metrics data file for the currently expected metrics:
`{
    "cs3_org_sciencemesh_site_total_num_users": 7000,
    "cs3_org_sciencemesh_site_total_num_groups": 550,
    "cs3_org_sciencemesh_site_total_amount_storage": 1080001003
}
`


ishank011 (commented):
Hi @redblom. Thanks for the change. A few comments:
- The initDataDriver method is called for each metrics refresh. Let's refactor it so that it's called only when initializing.
- For logging please initialize the logger like [this](https://github.com/cs3org/reva/blob/master/pkg/metrics/reader/dummy/dummy.go#L35).
- The if else construct for choosing the driver won't scale. Can you follow the registry system we follow for other packages?
- The file needs to be re-read on each refresh.

pkg/sharedconf/sharedconf.go
```diff
@@ -60,6 +62,14 @@ func Decode(v interface{}) error {
 		sharedConf.JWTSecret = "changemeplease"
 	}
 
+	if sharedConf.MetricsDataDriverType == "" {
```
ishank011, line 65: Also, remove setting the defaults here. If it's not set, then don't serve the metrics. This provides a way to disable these.

redblom:
`. The initDataDriver method is called for each metrics refresh. Let's refactor it so that it's called only when initializing.`
Actually I wanted to do that but the shared config is not there yet when no specific config has been initialized. Only then the shared config gets decoded. I don't know how to solve this in another way.

examples/metrics/metrics.toml
```diff
@@ -1,7 +1,12 @@
 [shared]
-jwt_secret = "Pive-Fumkiu4"
+# one of dummy, json.
```
ishank011, line 2: These need to be defined for the service, not shared conf.

internal/http/services/metrics/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Prefix returns the main endpoint of this service.
+func (s *svc) Prefix() string {
+	return s.prefix
+}
+
+// Unprotected returns all endpoints that can be queried without prior authorization.
+func (s *svc) Unprotected() []string {
+	return []string{"/"}
```
ishank011, line 57: Remove the "/"

pkg/metrics/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+	Prefix                string `mapstructure:"prefix"`
+	MetricsDataDriverType string `mapstructure:"metrics_data_driver_type"`
+	MetricsDataLocation   string `mapstructure:"metrics_data_location"`
+	MetricsRecordInterval int    `mapstructure:"metrics_record_interval"`
+}
+
+// Init sets sane defaults
+func (c *Config) Init() {
+	if c.Prefix == "" {
+		c.Prefix = "metrics"
```
ishank011, line 32: metrics is supposed to be the prefix for the prometheus service. This would cause conflicts.

internal/http/services/metrics/metrics.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// Close is called when this service is being stopped.
+func (s *svc) Close() error {
+	return nil
+}
+
+// Prefix returns the main endpoint of this service.
+func (s *svc) Prefix() string {
+	return s.prefix
```
ishank011, line 52: Return an empty string

pkg/metrics/driver/registry/registry.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/cs3org/reva/pkg/logger"
+	"github.com/cs3org/reva/pkg/metrics/reader"
+	"github.com/rs/zerolog"
+)
+
+var log zerolog.Logger
+
+var drivers map[string]reader.Reader // map key is driver type name
+
+func init() {
```
ishank011, line 33: Remove the logs. Let's keep it simple https://github.com/cs3org/reva/blob/master/pkg/ocm/invite/manager/registry/registry.go

pkg/metrics/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package config
+
+// Config holds the config options that need to be passed down to the metrics reader(driver)
+type Config struct {
+	Prefix                string `mapstructure:"prefix"`
```
ishank011, line 23: Remove Prefix

pkg/metrics/driver/json/json.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	d.metricsDataLocation = c.MetricsDataLocation
+
+	return nil
+}
+
+// GetNumUsers returns the number of site users
+func (d *MetricsJSONDriver) GetNumUsers() int64 {
+	return readJSON(d).NumUsers
```
ishank011, line 87: We're reading the file each time for all the three metrics, but they would be accessed simultaneously. Can you see if we can prevent reading the file three times?

pkg/metrics/driver/registry/registry.go
redblom: Ok, I do the loading like in that registry as well. That is better.
Do you agree with the GetDriver() method or do you want me to 'copy' https://github.com/cs3org/reva/blob/master/pkg/ocm/invite/manager/registry/registry.go ?

internal/http/services/metrics/metrics.go
redblom: @ishank011 loose the prefix altogether, so also remove from the config ?

pkg/metrics/driver/json/json.go
redblom: @ishank011  Hmm, that's like the initial implementation where the whole pkg got initialized on each metrics server call and all methods where read in one go. It is my recollection that it was decided to changed it to be like this. 
I have to think about this.

ishank011 (approved):

PR #1123, opened 2020-08-26
[Tests-Only] Add api trashbin tests

phil-davis:
### Description
This PR adds `apiTrashbin` bug verification scenarios.

The same as PR https://github.com/owncloud/ocis/pull/489 and PR https://github.com/owncloud/ocis-reva/pull/453

This enables future developers to be able to see the behavior in tests while they are understanding and fixing.

PR #1126, opened 2020-08-26
[Tests-Only] Also use new dav-path for empty-trashbin local scenarios 

phil-davis:
WIP


phil-davis:
works - this has been cherry-picked into #1123

PR #1131, opened 2020-08-28
Use updated cato to display nested package config in parent docs

ishank011:
Now nested configs look like this https://5f49071b89d86b0007317df7--cs3org-reva.netlify.app/docs/config/grpc/services/storageprovider/#drivers


labkode (approved):
Really cool!

PR #1140, opened 2020-09-03
Forward OpenFileInAppProvider calls for OCM shares to remote reva

### LINKED ISSUE #1032: Riva overwrite content directory
If I'm upload file to root path "/" Reva overwrite user content directory with single file. User lost all data.

Steps to reproduce:

```
cd reva/cmd/reva
echo "Test" >test.txt
./reva --insecure login basic
./reva --insecure upload test.txt /
```
after this operation directory "/var/tmp/reva/data" has only file "einstein", not directory "einstein"

ishank011:
Also closes #1032

### DIFF
--- a/changelog/unreleased/appprovider-stat.md
+++ b/changelog/unreleased/appprovider-stat.md
@@ -1,0 +1,7 @@
+Bugfix: Call the gateway stat method from appprovider
+
+The appprovider service used to directly pass the stat request to the storage
+provider bypassing the gateway, which resulted in errors while handling share
+children as they are resolved in the gateway path.
+
+https://github.com/cs3org/reva/pull/1140
--- a/cmd/reva/configure.go
+++ b/cmd/reva/configure.go
@@ -36,8 +36,8 @@ var configureCommand = func() *command {
 			return err
 		}
 
-		c := &config{Host: text}
-		if err := writeConfig(c); err != nil {
+		conf = &config{Host: text}
+		if err := writeConfig(conf); err != nil {
 			return err
 		}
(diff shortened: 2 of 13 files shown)


glpatcern (dismissed):
Looks good, and I guess the gateway's `Stat` will take care of all that is needed (including forwarding remote errors that will be embedded in the `OpenFileInAppProviderResponse`).

ishank011:
@glpatcern nope, we append `info` to the list of shares. We'll update info only if there's no error returned from the webdav stat call, otherwise we'll just let it continue to have the stat response of the reference, which it previously did.

glpatcern (commented):
Now this one looks good

SamuAlfageme (changes requested):
While testing this, we've seen that the 'username' encoded on the JWT token that gets forwarded to the WOPI Server (via Collabora) is the same for both the issuer and the receiver of the share:

```json
{
   "typ":"JWT",
   "alg":"HS256"
}{
   "userid":"...",
   "filename":"/reva/daniel/test4/test.odt",
   "username":"daniel",
   "viewmode":"VIEW_MODE_READ_WRITE",
   "folderurl":"undefined",
   "exp":1599906112,
 }
```

Therefore, it renders the same username for both parties during a collaborative editing session. cc/ @glpatcern

ishank011:
Can't do that with the current implementation. We're passing the same token around as the original user and can't make changes to it as we don't know the jwt secret of the other site. This won't be an issue in the public shares implementation so I think we can leave it be.

labkode (approved):

PR #1154, opened 2020-09-09
Progress with the eosgrpc client

ffurano:


labkode (approved):

labkode:
@ffurano @ishank011 we need to check why the go.mod are changed and why we need a different version of protobuf.

ffurano:
Good point indeed.

The eos grpc interface works only with the "standard" version of protoc. For some
historical reason instead reva was forcing an older version. I had to remove
this (old?) constraint.

Fabrizio

ishank011:
@labkode that's because apparently ocis doesn't work with the latest one https://github.com/cs3org/reva/pull/871

ishank011:
@ffurano why was the update to eos_grpc.pb.go needed though? I don't think there's any change there except those introduced by the newer version. Doesn't it work with the previous one? We won't need to uncomment the downgrade if it works anyway.

ishank011:
If there are no breaking changes, GRPC supports backward compatibility. So even if EOS runs a newer version, requests from a client running an older version should be supported. We need to followup with people from owncloud about when we can revert the downgrade but it should be fine for now.

We also need to use an older version in cs3APIs https://github.com/cs3org/cs3apis-build/commit/f1ef147b9882958e293ec063e4733604bb8ed694 because of this downgrade, but it's compatible with the newer versions.

ffurano:
Hi,

 this is not correct unfortunately, and if you uncomment that line the grpc client will not work anymore.

 If reva *really* needs to use a very old version of protoc then it's impossible to keep
the grpc protobuf file in a separate github project. It has to be embedded in reva (which
is what I had done at the beginning)

 My personal preference is for the latter (i.e. keeping the .proto file in the reva tree
and compiling it with the same protoc version as other things in reva). This is cheaper
to maintain and quite pragmatic, but does not follow the latest fancy fashions.

 So, I'd kindly prompt for a shared decision, as it's not good if one has to comment that line
every time just for running the grpc interface.


Cheers
Fabrizio

labkode:
Hi @ffurano, let's unblock this with a common ground. Let's git submodule the the EOS protobuf and we compile inside Reva.

ffurano:
Hi Hugo,

 uhm... I don't feel we are blocked on this. I am waiting for
the etag in the stat information from eos.

 Or are you referring to other things? Like the protoc version?

f



Il 21/10/20 09:27, Hugo G. Labrador ha scritto:

ishank011:
And rebase

ffurano:
done

Il 02/12/20 15:11, Ishank Arora ha scritto:

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -849,76 +761,78 @@ func (c *Client) rm(ctx context.Context, username, path string) error {
...
 		return err
 	}
 
-	msg := new(erpc.NSRequest_RmdirRequest)
+	msg := new(erpc.NSRequest_RmRequest)
 
 	msg.Id = new(erpc.MDId)
 	msg.Id.Path = []byte(path)
+	msg.Recursive = true
+	msg.Norecycle = true
```
ishank011, line 792: We would want the deleted directories to be present in the recycle bin. Same for deleting files.

ffurano: Ok

ishank011 (approved):

PR #1155, opened 2020-09-09
ocis driver: access shares, trash, revisions, metadata

butonic:
continuation of https://github.com/cs3org/reva/pull/559


pkg/storage/fs/ocis/ocis.go
```diff
@@ -207,8 +195,83 @@ func (fs *ocisfs) CreateDir(ctx context.Context, fn string) (err error) {
...
+	return relativeHome, nil
+}
+
+// TODO there really is no difference between the /Shares folder and normal nodes
+// we can store it is a node
+// but when home is enabled the "/Shares" folder should not be listed ...
+// so when listing home we need to check that flag and add the shared node
+
+// /Shares/mounted1
+// /Shares/mounted1/foo/bar
```
PVince81, line 233: invalid old example, we can only create references from a path directly under "Shares"

pkg/storage/fs/ocis/ocis.go
```diff
@@ -207,8 +195,83 @@ func (fs *ocisfs) CreateDir(ctx context.Context, fn string) (err error) {
...
+			return
+		}
+	}
+
+	if n, err = n.Child(parts[1]); err != nil {
+		return errtypes.InternalError(err.Error())
+	}
+
+	if n.Exists {
+		return errtypes.AlreadyExists(p)
```
PVince81, line 263: wondering when this scenario would happen.

when is CreateReference usually called ? at creation time or login time ?

also curious how it will work for groups as it means creating references for all group members

pkg/storage/fs/ocis/ocis.go
```diff
@@ -207,8 +195,83 @@ func (fs *ocisfs) CreateDir(ctx context.Context, fn string) (err error) {
...
+	}
+
+	if err = fs.tp.CreateDir(ctx, n); err != nil {
+		return
+	}
+
+	internal := filepath.Join(fs.pw.Root, "nodes", n.ID)
+	if err = xattr.Set(internal, referenceAttr, []byte(targetURI.String())); err != nil {
+		return errors.Wrapf(err, "ocfs: error setting the target %s on the reference file %s", targetURI.String(), internal)
```
PVince81, line 272: ocisfs

pkg/storage/fs/ocis/path.go
```diff
@@ -90,12 +94,21 @@ func (pw *Path) NodeFromID(ctx context.Context, id *provider.ResourceId) (n *Nod
...
-	return ReadNode(pw, id.OpaqueId)
+	return ReadNode(ctx, pw, id.OpaqueId)
 }
 
 // Path returns the path for node
 func (pw *Path) Path(ctx context.Context, n *Node) (p string, err error) {
-	for !n.IsRoot() {
+	var root *Node
+	if pw.EnableHome {
+		root, err = pw.HomeNode(ctx)
```
PVince81, line 104: third deja vu. move this condition into utility method ?

pkg/storage/fs/ocis/path.go
```diff
@@ -132,30 +145,34 @@ func (pw *Path) readRootLink(root string) (node *Node, err error) {
...
-	return
-}
+	u := user.ContextMustGetUser(ctx)
+	layout := templates.WithUser(u, pw.UserLayout)
 
-// Root returns the root of the storags
-func (pw *Path) Root() string {
-	return pw.root
+	segments := strings.Split(layout, "/")
+	for i := range segments {
```
PVince81, line 172: didn't we have a similar loop/logic in NodeInfo that we can reuse ?

pkg/storage/fs/ocis/persistence.go
```diff
@@ -32,9 +31,9 @@ type TreePersistence interface {
 	GetPathByID(ctx context.Context, id *provider.ResourceId) (string, error)
 	GetMD(ctx context.Context, node *Node) (os.FileInfo, error)
 	ListFolder(ctx context.Context, node *Node) ([]*Node, error)
-	CreateRoot(id string, owner *userpb.UserId) (n *Node, err error)
+	//CreateHome(owner *userpb.UserId) (n *Node, err error)
```
PVince81, line 34: make it fail as unsupported?

pkg/storage/fs/ocis/tree.go
```diff
@@ -76,24 +74,32 @@ func (t *Tree) GetPathByID(ctx context.Context, id *provider.ResourceId) (relati
...
 	n = &Node{
-		ParentID: "root",
 		pw:       t.pw,
-		ID:       id,
+		ID:       "root",
+		Name:     "", // the root node has an empty name, or use `.` ?
+		ParentID: "", // the root node has no parent, or use `root` ?
 	}
 
+	if err = createNode(n, nil); err == nil {
```
PVince81, line 86: make createNode return the node object already populated with "Exists = true" ?

pkg/storage/fs/ocis/tree.go
```diff
@@ -107,42 +113,28 @@ func (t *Tree) CreateDir(ctx context.Context, node *Node) (err error) {
...
 	if err != nil {
 		return
 	}
 	return t.Propagate(ctx, node)
 }
 
 // CreateReference creates a new reference entry in the tree
-func (t *Tree) CreateReference(ctx context.Context, path string, targetURI *url.URL) error {
-	// TODO use symlink, but set extended attribute on link (not on target)
+// TODO the
```
PVince81, line 136: incomplete todo

pkg/storage/fs/ocis/tree.go
```diff
@@ -290,8 +282,17 @@ func (t *Tree) Propagate(ctx context.Context, node *Node) (err error) {
...
+	if t.pw.EnableHome {
+		root, err = t.pw.HomeNode(ctx)
+	} else {
+		root, err = t.pw.RootNode(ctx)
+	}
+	if err != nil {
+		return
+	}
+	for err == nil && node.ID != root.ID { // TODO propagate up to where?
+		if err := xattr.Set(filepath.Join(t.pw.Root, "nodes", node.ID), "user.ocis.etag", []byte(etag)); err != nil {
```
PVince81, line 295: using the same etag value for all parent members ?

pkg/storage/fs/ocis/ocis.go
```diff
@@ -207,8 +195,83 @@ func (fs *ocisfs) CreateDir(ctx context.Context, fn string) (err error) {
...
+	if n, err = n.Child(parts[1]); err != nil {
+		return errtypes.InternalError(err.Error())
+	}
+
+	if n.Exists {
+		return errtypes.AlreadyExists(p)
+	}
+
+	if err = fs.tp.CreateDir(ctx, n); err != nil {
+		return
```
PVince81, line 267: no error/log ?

pkg/storage/fs/ocis/ocis.go
```diff
@@ -207,8 +195,83 @@ func (fs *ocisfs) CreateDir(ctx context.Context, fn string) (err error) {
...
+	}
+
+	internal := filepath.Join(fs.pw.Root, "nodes", n.ID)
+	if err = xattr.Set(internal, referenceAttr, []byte(targetURI.String())); err != nil {
```
PVince81, line 271: I guess we have no way of validating if the reference is valid or at least has the correct format (if it's a URL)

pkg/storage/fs/ocis/path.go
```diff
@@ -35,15 +35,16 @@ import (
 // Path implements transformations from filepath to node and back
 type Path struct {
 	// ocis fs works on top of a dir of uuid nodes
-	root string `mapstructure:"root"`
+	Root string `mapstructure:"root"`
 
-	// UserLayout wraps the internal path in the users folder with user information.
+	// UserLayout describles the relative path from the storages root node to the users home node.
```
PVince81, line 40: - "describes"
- "storage's root"

PVince81 (commented):
sharing code commit looks fine so far, see a few comments

pkg/storage/fs/ocis/ocis.go
butonic: we have, we could and we should so that we can use a symbolic link if the cs3 reference uses a node in the same storage provider. but that would only allow the admin to navigate the reference in that case ... not important right now.

pkg/storage/fs/ocis/ocis.go
butonic: `err` is a named return variable. if `err != nil` we return and err is returned ... go magic

pkg/storage/fs/ocis/path.go
butonic: decided on `pw.HomeOrRootNode(ctx)`

pkg/storage/fs/ocis/ocis.go
butonic: CreateReference is called when a user mounts a share. What happens when he mounts a folder with the same name

pkg/storage/fs/ocis/path.go
butonic: introduced `func (pw *Path) WalkPath(ctx context.Context, r *Node, p string, f func(ctx context.Context, n *Node) error) (*Node, error)` 
using it in three other places

PVince81:
:+1: for the latest refactorings based on my comments.

please check the linter

PVince81 (approved):
Looks fine to me except for the attr name, feel free to change

pkg/storage/fs/ocis/recycle.go
```diff
@@ -114,24 +115,95 @@ func (fs *ocisfs) ListRecycle(ctx context.Context) (items []*provider.RecycleIte
...
+		return
+	}
+
+	if n.Exists {
+		return errtypes.AlreadyExists("origin already exists")
+	}
+
+	// rename to node only name, so it is picked up by id
+	nodePath := filepath.Join(fs.pw.Root, "nodes", parts[0])
+	err = os.Rename(deletedNodePath, nodePath)
```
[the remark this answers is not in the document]
butonic: Rename replaces target. that is why I check n.Exists and return AlreadyExists. It is the responsibility of the webdav handler to translate that into 409 Conflict.

butonic:
```
-> running `props':
 0. init.................. pass
 1. begin................. pass
 2. propfind_invalid...... pass
 3. propfind_invalid2..... pass
 4. propfind_d0........... pass
 5. propinit.............. pass
 6. propset............... pass
 7. propget............... pass
 8. propextended.......... pass
 9. propmove.............. pass
10. propget............... pass
11. propdeletes........... pass
12. propget............... pass
13. propreplace........... FAIL (PROPPATCH on `/remote.php/dav/files/4c510ada-c86b-4815-8820-42cdf82c3d51/litmus/prop2': 500 Internal Server Error)
14. propget............... FAIL (Property {http://example.com/neon/litmus/}prop9 had value value9, expected newvalue9)
15. propnullns............ pass
16. propget............... pass
17. propremoveset......... FAIL (PROPPATCH remove then set)
18. propget............... FAIL (No value given for property {http://example.com/neon/litmus/}removeset)
19. propsetremove......... pass
20. propget............... pass
21. propvalnspace......... pass
22. propwformed........... pass
23. propinit.............. pass
24. propmanyns............ pass
25. propget............... pass
26. propcleanup........... pass
27. finish................ pass
<- summary for `props': of 28 tests run: 24 passed, 4 failed. 85.7%
```

PVince81 (dismissed):
Latest commits for trash, restore version and download versions look fine

PR #1158, opened 2020-09-10
[Tests-Only] Bump core commit 20200910

phil-davis:
Gets us new core tests:
https://github.com/owncloud/core/pull/37881 [Tests-Only] Refactor createShareToShareFolder.feature to use Shares folder

This provides core tests for basic sharing that use the default settings of reva:

- `share_folder` is `Shares` - so share receivers get their shares arrive in a folder called `Shares`
- auto-accept shares is off - in the test scenarios, for example, when Alice shares with Brian, then there is a step for Brian to accept the share.

These new sharing scenarios pass in oC10, and they should be able to be made to pass in reva.

This is a first installment for issue https://github.com/owncloud/ocis/issues/518

And also:
- the test refactoring in https://github.com/owncloud/core/pull/37902 [Tests-Only] Make PROPFIND response href test checks more flexible
- https://github.com/owncloud/core/pull/37882 [Tests-Only] Add some more flexible provisioning API tests for getUsers (does not really matter here, because the provisioning API does not exist and the tests are skipped anyway)


phil-davis:
Rebased - this should pass CI and is ready for review.

ishank011 (approved):

PR #1166, opened 2020-09-14
[Tests-Only] Run litmus CI for both owncloud and ocis storage

phil-davis:
Test both `ocis` and `owncloud` storage with litmus.

(after this is merged, I will then work on running the API tests with `ocis` storage as well as `owncloud` storage)

PR #1177, opened 2020-09-17
fix ocis move

butonic:
When renaming a file we updating the name attribute on the wrong node, causing the path construction to use the wrong name. This fixes the litmus move_coll test.


PVince81 (approved):
:+1:

ishank011 (approved):

phil-davis:
@butonic https://github.com/cs3org/reva/pull/1166#issuecomment-694082597

After this change, there are various `prop*` litmus tests failing on `ocis` storage.

PR #1186, opened 2020-09-20
[Tests-Only] Run local integration tests with OCIS storage

phil-davis:
The local integration tests are actually bug-demonstration scenarios of bugs that are known in `owncloud` storage. Some of those "fail" on `ocis` storage - either because the bug is fixed on `ocis` storage, or because the failing behavior is different on `ocis` storage. Either way, we skip those scenarios when testing with `ocis` storage.

I am working on issue https://github.com/owncloud/product/issues/225 to sort out which bugs are fixed on `ocis` storage and which are failing in a different way.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -409,7 +409,7 @@ steps:
 ---
 kind: pipeline
 type: docker
-name: local-integration-tests
+name: local-integration-tests-owncloud
 
 platform:
   os: linux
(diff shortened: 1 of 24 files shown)


phil-davis:
Rebased. @labkode @ishank011 this can be reviewed now. IMO it will be useful to have these local tests running on OCIS - then at least we will know if/when a PR changes some behavior of the tests that are running.

ishank011 (approved):

PR #1205, opened 2020-09-30
Allow using the username when accessing the users home

butonic:
We now allow using the userid and the username when accessing the
users home on the `/dev/files` endpoint.


kulmann (approved):
👍

labkode (approved):

PR #1208, opened 2020-09-30
Calculate etags on-the-fly for shares directory and home folder

ishank011:


butonic:
:thinking:

labkode (approved):

labkode:
@ishank011 nice! This is the last piece to make shares discoverable for the sync client. This sets the baseline, from now on, optimisations can be made, like caching this virtual etags for 30 seconds (sync discovery), etc ..

PR #1213, opened 2020-10-01
ocis driver: enforce permissions

butonic:
We now check the owner and grants on nodes in the tree to determine the actual permissions.
- [x] check up and downloads
- [x] revisions can use the original node
- [x] only allow owner of a storage to manage trash? or determine the permissions based on the original parent?
- [x] changelog
- [ ] some trash permission checks are unclear to me ... for now I only allow the owner to do some trash operations, as it is hard to determine if the current user is allowed to restore a certain trash item ... or if he can list all items ... because that is a permission not on the file but the storage afaict. @labkode how does eos handle the permission? it seems the cs3 grant permissions regarding trash need some thought ... for versions we can check the actual  file ...
- [ ] make AsResourceInfo() contain all permissions
  - which means that for requests that list or stat a node we need to read all grants to calculate the effective permissions


butonic:
can be merged for now

labkode (approved):

PR #1216, opened 2020-10-02
Add error handling for invalid references

ishank011:


ishank011:
@phil-davis this change seems to have fixed a test which was expected to fail https://cloud.drone.io/cs3org/reva/2694/8/7. It just handles the case of listing the shares folder when share references might no longer exist, maybe because they were deleted.

labkode (approved):

phil-davis:
@ishank011 @labkode when the code i improved and more test scenarios pass, that is "a good thing" (tm).
Just remove the now-passing scenarios from the expected-failures file(s).
That will get the CI green, and we have a record of exactly which scenarios are failing at each PR, and we easily if a code change has a regression (more failing scenarios) or improves things (previously-failing scenarios now pass).

See PR #1217 - that will make CI green again.

PR #1219, opened 2020-10-02
do not duplicate base

butonic:
without this I am seeing `/remote.php/dav/files/einstein/Shares/from marie/from marie/` in the webdav api at the ocdav endpoint. It breaks listing the shares folder.

@ishank011 @labkode ... is this because you are working on ocm?


ishank011 (approved):
Sorry, this must have creeped in during my last change.

PR #1233, opened 2020-10-08
Add support for a custom CodiMD mimetype

glpatcern:
This is for now hardcoded, still have to parse the appropriate config section.

### DIFF
--- a/changelog/unreleased/zmd-mimetype.md
+++ b/changelog/unreleased/zmd-mimetype.md
@@ -1,0 +1,5 @@
+Enhancement: add support for custom CodiMD mimetype
+
+The new mimetype is associated with the `.zmd` file extension.
+
+https://github.com/cs3org/reva/pull/1233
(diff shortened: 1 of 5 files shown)


ishank011 (commented):
@glpatcern can we instead call RegisterMime from the init method of appprovider? Something like adding `mime.RegisterMime(".zmd", "application/compressed-markdown")` in

https://github.com/cs3org/reva/blob/6b143c297df7ccb57ca2df822a835a6e65a74a4f/internal/grpc/services/appprovider/appprovider.go#L46-L48

glpatcern:
Makes sense - and surely we can read the config there in appprovider once it's ready.

ishank011:
Well if we can add it through appprovider, the whole process becomes much simpler 😁 

I'd still like to implement the functionality to pass configs to packages, but that can be put on the backburner for now. I'll create a PR to this branch to enable reading this config.

glpatcern:
Yes, I agree that passing configs to packages should be done as a back-burner task.

At the same time, with this PR we're saying "new mime types can be defined in the context of the appprovider only", which is fair enough for now but in general that statement could be proven wrong.

ishank011 (approved):

ishank011:
@glpatcern agreed!

PR #1236, opened 2020-10-09
[Tests-only] Bump core commit id 20201009

phil-davis:


phil-davis:
See #1245 for the next attempt at this.

PR #1242, opened 2020-10-12
[WIP] Datatx driver 

redblom:
See sciencemesh/sciencemesh#140

Datatx driver pkg setup. 
note: Datatx interface methods' signatures still indefinite.


labkode (approved):

labkode:
@redblom good progress!

redblom:
This now implements rclone driver DoTransfer, GetTransferStatus.
There are also 2 cli commands available for testing this: rclone-do-transfer, rclone-get-transfer-status
For tests you'll need a running rclone server and an rconfig file with the remotes/users specified.
Transfer eg. a cli triggered transfer between two running reva instances:

```
$ ./cmd/reva/reva rclone-do-transfer -senderEndpoint http://rclone-reva:5572/ -srcEndpoint reva-1 -srcPath /test/1gb.txt -srcToken marie -destEndpoint reva-2 -destPath /myfiles/1gb1.txt -destToken marie
```
This will echo the rclone job id.
Notes: 
- src/destEndpoint - the remotes' names as specified in the config
- the tokens are not used but required by the cli

Get status eg.:
```
$ ./cmd/reva/reva rclone-get-transfer-status -senderEndpoint http://rclone-reva:5572/ -transferID 68
```
This will echo the transfer status.

redblom:
All 3 methods (DoTransfer, GetTransferStatus, CancelTransfer) implemented.

ishank011 (changes requested):
The driver is supposed to be included in the GRPC service, not invoked directly by clients. Please take a look at the other GRPC services and make sure that this also follows that approach https://github.com/cs3org/reva/blob/master/internal/grpc/services/userprovider/userprovider.go

redblom:
@ishank011 yeah I know, I'm working on those right now. These rclone clients are for driver testing purposes only, I don't expect them to be part of the final PR.

cmd/reva/main.go
```diff
@@ -77,6 +77,9 @@ var (
 		transferGetStatusCommand(),
 		transferCancelCommand(),
 		helpCommand(),
+		rcloneDoTransfer(),
```
ishank011, line 80: We can remove these three

cmd/reva/transfer-create.go
```diff
@@ -20,32 +20,69 @@ package main
...
 	"io"
 
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	tx "github.com/cs3org/go-cs3apis/cs3/tx/v1beta1"
 )
 
 func transferCreateCommand() *command {
+	// eg. transfer file from marie @ surfsara.nl to einstein @ cern.ch
```
ishank011, line 33: Remove the comments. These are pretty self-explanatory

cmd/reva/transfer-create.go
```diff
@@ -20,32 +20,69 @@ package main
...
 
 		ctx := getAuthContext()
 		client, err := getClient()
 		if err != nil {
 			return err
 		}
 
-		transferRequest := &tx.CreateTransferRequest{}
+		// transfers are between users
+		gt := provider.GranteeType_GRANTEE_TYPE_USER
```
ishank011, line 67: Add it as a flag (default to user)

internal/grpc/services/datatx/datatx.go
```diff
@@ -81,19 +101,121 @@ func (s *service) UnprotectedEndpoints() []string {
...
+	// Notes:
+	// . rclone does NOT fail/error when the src can not be found: TODO ? check if scr exists ?
+	//   however: a get status call should give an error anyway
+	// ----------------------------------------------------------------------------------------
+
+	// TODO create a new transfer id
+	var txID int64
+
+	// the local idp; TODO must retrieve the idp through owner of the resource ? Or via config ?
+	srcRemote := "reva"
```
ishank011, line 123: This needs to be read from the user present in context

pkg/datatx/driver/rclone/rclone.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return -1, errors.Wrap(err, "json: error framing post request")
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	// TODO if we want to use basic auth, should be configurable.
+	req.SetBasicAuth("rclone", "secret")
+
+	// TODO insecure should be configurable
+	client := rhttp.GetHTTPClient(rhttp.Insecure(true))
```
ishank011, line 108: Make the client a member of the Rclone struct and reuse it

pkg/datatx/driver/rclone/rclone.go
```diff
@@ -0,0 +1,311 @@
...
+		JobID: jobID,
+	}
+
+	data, err := json.Marshal(rcloneStatusReq)
+	if err != nil {
+		return datatx.TxInfo_STATUS_INVALID, errors.Wrap(err, "error marshalling rclone req data")
+	}
+
+	transferFileMethod := "/job/status"
+	url := strings.Trim(driver.Endpoint, "/") + transferFileMethod
```
ishank011, line 167: Add proper URL handling

pkg/datatx/driver/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package config
+
+// Config holds the config options that need to be passed down to the datatx driver
+type Config struct {
+	Driver   string `mapstructure:"driver"`
```
ishank011, line 23: The driver parameter should be part of the service which initializes this package, in this case, datatx

pkg/datatx/driver/txdriver.go
```diff
@@ -0,0 +1,36 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package driver
```
ishank011, line 19: Move this file to `pkg/datatx/datatx.go` and rename package as datatx. This is the convention we follow

pkg/datatx/driver/rclone/rclone.go
```diff
@@ -0,0 +1,311 @@
...
+	transferFileMethod := "/operations/copyfile"
+	url := strings.Trim(driver.Endpoint, "/") + transferFileMethod
+	req, err := http.NewRequest("POST", url, bytes.NewReader(data))
+	if err != nil {
+		return -1, errors.Wrap(err, "json: error framing post request")
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	// TODO if we want to use basic auth, should be configurable.
+	req.SetBasicAuth("rclone", "secret")
```
ishank011, line 105: Remove this if we're not using anything which is actually needed

ishank011 (changes requested):

pkg/datatx/rclone/rclone.go
```diff
@@ -0,0 +1,354 @@
...
+		return datatx.TxInfo_STATUS_TRANSFER_IN_PROGRESS, nil
+	}
+	if !resData.Success {
+		return datatx.TxInfo_STATUS_TRANSFER_FAILED, nil
+	}
+	return datatx.TxInfo_STATUS_INVALID, nil
+}
+
+// rcloneCancelTransferReqJSON the rclone job/stop method json request
+type rcloneCancelTransferReqJSON struct {
```
ishank011, line 260: Can we use a generic struct for rcloneCancelTransferReqJSON and rcloneStatusReqJSON, since they have the same fields? (same for rcloneCancelTransferResJSON)

ishank011 (commented):
Added a few comments. I'm not sure I fully understand how the token situation works. In OCM we transfer a token when sharing a file which is used to access resources on the file owner's provider, not sure how that translates here and how it connects to that, since these will be enabled only once an OCM share has been created, right? (https://wiki.cs3mesh4eosc.eu/wps/4/task44stuff/apis)

redblom:
> 
> 
> Added a few comments. I'm not sure I fully understand how the token situation works. In OCM we transfer a token when sharing a file which is used to access resources on the file owner's provider, not sure how that translates here and how it connects to that, since these will be enabled only once an OCM share has been created, right? (https://wiki.cs3mesh4eosc.eu/wps/4/task44stuff/apis)

With data transfer a special transfer OCM share request (which will be a to be defined special datatx OCM protocol type) is done towards the anticipated destination and at destination a token is created and returned so the token is available at source site for use when doing the actual transfers (transfers are instigated at the source site so it must have a valid token).
Actually this is not described in the notes in CreateTransfer method of grpc service datatx.go. I will do that; should make it clear.

internal/grpc/services/datatx/datatx.go
```diff
@@ -81,19 +103,130 @@ func (s *service) UnprotectedEndpoints() []string {
...
+	}
+
+	srcRemote := u.GetId().GetIdp()
+	srcPath := req.GetRef().GetPath()
+
+	// Temp implementation is depending on current rclone version.
+	// (token is username for now, it is used like that in the current rclone driver implementation)
+	// TODO: re-implement when the connection-string-remote syntax is available in rclone
+	// TODO: implement proper token use
+	srcToken := u.GetUsername()
```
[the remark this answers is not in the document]
redblom: Yeah, the destination token should be available here, returned by OCM share request. But this is not implemented yet, see below: https://github.com/cs3org/reva/pull/1242#issuecomment-755258461.
Apparently a source token for the 'local' source file to be transferred should (at this point) be created somehow (see https://wiki.cs3mesh4eosc.eu/wps/4/task44stuff/apis#data-transfer 'generate token to read Hugo's file'). Maybe you could shed some light on (how to do) that.

So yes, the driver.DoTransfer token parameters are being 'abused' here. We need actual tokens.

pkg/datatx/rclone/rclone.go
redblom: I will declare them within the function there used in, that's actually what I wanted but thought it was invalid to declare a struct inside a method (my lack of golang knowledge).

ishank011:
@redblom please take a look at https://codimd.web.cern.ch/CsSOm-lxS52nV3hWjbglaw. I'd recommend merging this PR as is and starting with the OCM integration after that. I'd like to help with the coding part in that.

ishank011:
This is not ideal though. IMO, the transfer should be initiated when the recipient accepts the share through the `ocm-update-received-share` method, since we create references then only as well. So the problem of transferring the token from the recipient to the original user won't arise in the first place. That would be a much better design in my opinion.

labkode:
@ishank011 I support Ishank's comment, that is the goal of having end to end transfers driven by OCM.

redblom:
@labkode @ishank011 : To me this looks like how it has been discussed before; and as drawn in the datatx diagram. 

@labkode was however correct when showing the video, CreateTransfer returns the wrong state. It must return one of `transfer_awaiting_acceptance` or `transfer_accepted`). In this case `transfer_accepted` should be returned since in the current implementation the transfer is always accepted (there's no OCM accept share flow implemented yet). 
Once the OCM share accept flow is implemented `transfer_awaiting_acceptance` may be returned (same as `pending` in file sharing).

ishank011 (changes requested):
Persistency should be implemented in the same interface, there shouldn't be the need for a new one. For example, refer how we do it for OCM invites https://github.com/cs3org/reva/blob/master/pkg/ocm/invite/manager/json/json.go

redblom:
When the (datatx) share is received in the ocmd we know it's a datatx share through the protocol options parameter. However we currently cannot persist it as such. Share objects have no tx specifics. So the receiver does not know it's a datatx share either. How do we go about that?

ishank011:
Why can't we persist it? On the recipient side, the request would look like as described in https://wiki.cs3mesh4eosc.eu/wps/4/task44stuff/apis, so it would know it's a datatx share.

redblom:
The datatx transfer ends up as a 'marshalled' Share object in the received_shares section of the shares file at the receiver's end. What options do I have to define there that it is a datatx share; so that the receiver (and the system) when accepting this share know that after reading this share from the shares file that this is a datatx share and the transfer can commence?

In other words, in the shares file, how does a datatx share differ from a regular share?

ishank011:
Add one more parameter to the schema?

redblom:
That is one solution I was also thinking of.

So simplest:
```
cs3.sharing.ocm.v1beta1.Share { 
   ... ,
   bool is_data_transfer_share
}
```
Or maybe `share_type` in case there there could be other future types, don't know, maybe `stream` type ...

ishank011:
Sure, `share_type` would be better. Please create a PR and in the description part in the proto file, mention this use case

redblom:
https://github.com/cs3org/cs3apis/pull/109

redblom:
How about using the share ID instead of transfer ID in cs3 datatx api. The share ID is already there. That would simplify things.

redblom:
@ishank011 : I believe ocm.share manager interface method Save() needs an additional share type arg.

ishank011:
@redblom sure, feel free to make changes as you need.

redblom:
@ishank011 please review https://github.com/cs3org/reva/pull/1578

redblom:
@ishank011 : Implementing this (#1578) I think we should use `ocm-share-create` for a transfer also. An OCM share will always be the starting point anyway and the `Share.ShareType` is already introduced. So the call would be something like:
`cmd/reva/reva -insecure ocm-share-create -shareType transfer -grantee jarek1234 -granteeType user -idp softwaremind.com /home/test`
Then we change the CS3 datatx CreateTransferRequest message to contain Share (or at least the share id) (and optional opaque) fields only. *Note that the actual share id will be the transfer identifier (previously transfer id).

Now when UpdateReceivedOCMShare is called it can delegate to grpc.gateway.datatx.CreateTransfer with the share specified and the transfer can be initiated from thereon:
`pkg.datatx.TxDriver.CreateTransfer(shareID, ...)`

I think it makes more sense.

ishank011:
When `UpdateReceivedOCMShare` is called, the reference is created in the shares directory https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/ocmshareprovider.go#L260 -> https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/ocmshareprovider.go#L291-L349, where the default path is `/home/Shares/name-of-shared-resource`. We can use the same path as the destination.

For WebDAV share references, we have the target `webdav://token@idp?name=share-name`. For transfers, we don't really need a target as the whole resource will be located here and we can just pass this path to rclone.

These references are statted in the gateway storageprovider service https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/storageprovider.go#L1369-L1395. We'll need to add a case for transfer refs to just be returned as it is. We have all the existing functionalities that will allow users to make changes to any files/folders located inside this reference.

redblom:
@ishank011 I'd like to see this merged first (create transfer type OCM share) https://github.com/cs3org/reva/pull/1578

redblom:
> 
> 
> When `UpdateReceivedOCMShare` is called, the reference is created in the shares directory https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/ocmshareprovider.go#L260 -> https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/ocmshareprovider.go#L291-L349, where the default path is `/home/Shares/name-of-shared-resource`. We can use the same path as the destination.
> 
@ishank011 What would be the resulting location here, eg. `marie/name-of-shared-resource` ? And then `./reva stat /home/Shares/name-of-shared-resource` would work?

redblom:
> 
> 
> You're passing the `/remote.php/webdav` endpoints for both PUT and GET to rclone, right? Then you just need to use the path `/home/Shares/name-of-shared-resource` everywhere. Reva will resolve that based on the user in context.

I use single endpoints for each request. For transfer the rclone service is called through 2 possible endpoints, `operations/copyfile` to transfer a file and `sync/copy` to transfer a folder structure. These calls will asynchronous and immediately return a jobID (to be used for status/cancel requests).
The requests' data is in the form of a single string with all necessary parameters included (https://rclone.org/docs/#connection-strings). A folder transfer example of such a req data string would be: 
`'{"_async":true,"srcFs":":webdav,bearer_token=\"...token...\",url=\"http://reva-datatx:20080/webdav\":/big-data","dstFs":":webdav,bearer_token=\"...token...\",url=\"http://reva-2:20080/webdav\":/big-data"}'`
This would be a transfer to a 'local' path, eg `marie/big-data/...`

I experimented with accepted shares where I gave rclone the share path like `url=\"http://reva-2:20080/webdav\":/MyShares/big-data` but that did not quite work. I was expecting these files to exist somewhere in (my case) `marie/big-data/...` but for the user at `/MyShares/big-data/...`. So first I want to know if this is indeed what's expected.

ishank011:
Which service handles the requests at the `/webdav` endpoint?

redblom:
> 
> Which service handles the requests at the `/webdav` endpoint?

`../../internal/http/services/owncloud/ocdav/ocdav.go`

@ishank011 I'm also not sure what should be returned here for the transfer case (https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/storageprovider.go#L1369-L1395) exactly? Simply the provided resourceInfo? How should we establish here that this is a transfer ref?

ishank011:
Use the path `/remote.php/webdav/home/MyShares/xyz`. It should work

> Simply the provided resourceInfo?

Yes

> How should we establish here that this is a transfer ref?

Check the scheme

redblom:
> 
> 
> Use the path `/remote.php/webdav/home/MyShares/xyz`. It should work
> 
Yes, that seems to work!
>
> Simply the provided resourceInfo?
> 
> Yes
> 
> > How should we establish here that this is a transfer ref?
> 
> Check the scheme

And that should read `datatx` I suppose, because now it's `webdav`. That makes sense.

redblom:
> > Simply the provided resourceInfo?
> > Yes
> > > How should we establish here that this is a transfer ref?
> > 
> > 
> > Check the scheme
> 
> And that should read `datatx` I suppose, because now it's `webdav`. That makes sense.

@ishank011 : I can set the scheme to `datatx` here: https://github.com/cs3org/reva/blob/1765ae8db4d0d5e615fc3b55275e563dd5a1c687/internal/grpc/services/gateway/ocmshareprovider.go#L324

Then I can determine it at https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/storageprovider.go#L1369-L1395 for the new case. Is that correct?

redblom:
This PR is superseded by https://github.com/cs3org/reva/pull/2052

PR #1245, opened 2020-10-14
[tests-only] Bump core commit id for API tests

phil-davis:
This gets various refactored sharing tests running, that expect shares to be received in a folder called `Shares` and for auto-accept shares to be off - the core test scenarios have steps for the share receiver to accept the share. This means that the scenarios should be expected to pass when the various sharing features are working correctly.

The test infrastructure allows more control of how "skeleton" files are simulated. The env var `OCIS_SKELETON_STRATEGY` is set to `copy` for the local API tests on OCIS storage - the test infrastructure correctly knows how to copy skeleton files into place in that case.

The list of expected-failures is adjusted. Mostly there are test suite name and line number changes of the scenarios due to the refactoring of scenarios in core.

Some local API test scenarios are adjusted to not use skeleton files - that saves a bit of time and means that they do not rely on the skeleton "simulation" test code.


phil-davis:
@labkode @ishank011 this is ready for review.

It is a couple of weeks since I did this updating of the test infrastructure - I should do it more often!

ishank011 (approved):

PR #1246, opened 2020-10-14
Add test user to all sites

Daniel-WWU-IT:
For health monitoring of all mesh sites, we need a special user account that is present on every site. This PR adds such a user to each `users-*.json` file so that every site will have the same test user credentials (which are, of course, **test/testpass**).

This omnipresent test user allows us to perform remote checks that require authentication using the same credentials for every site, making automated check configuration much easier.

_Note that this is just a temporary solution, but for this initial phase, this is the easiest solution._


labkode (dismissed):

labkode:
@Daniel-WWU-IT please rebase

Daniel-WWU-IT:
@labcode Should be mergeable now.

Daniel-WWU-IT:
I'll close this PR and create a new one... for some reason, there was a huge mess when trying to rebase

PR #1249, opened 2020-10-16
[Tests-Only] Adjust local api tests to match owncloud ocis

phil-davis:
[WIP] put various tests back the way they are in `owncloud/ocis` and check how they pass or fail.


phil-davis:
This is out-of-date. I will do a more up-to-date PR!

Issue #860, opened 2020-06-19
Reva/ocdav seems to have memory leak when doing webdav file transfers.


redblom:
In experimenting with file transfers initiated by rclone/webdav I noticed that Reva runs out of memory very quickly with subsequent file transfers. The memory used for each transfer keeps adding up to the previously used memory until Reva runs out of memory, and then Reva is killed.
Here's the final part of the log until Reva is killed: [mem-issue-reva-webdav-file-transfer.txt](https://github.com/cs3org/reva/files/4804034/mem-issue-reva-webdav-file-transfer.txt)

It's the receiving Reva instance that takes up all the memory. In the experiments both source and destination Reva instances are running in docker containers as well as rclone.
Here's the result (docker stats) from a number of subsequent 600MB file transfers (4GB memory available for docker). Each subsequent transfer adds to the already used memory and in between transfers the memory does not gets freed up:
```
CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
2fd9dee452e9        reva-2              0.08%               1.777GiB / 3.848GiB   46.17%              615MB / 617kB       0B / 0B             10
2fd9dee452e9        reva-2              0.00%               2.668GiB / 3.848GiB   69.33%              1.23GB / 1MB        0B / 0B             10
2fd9dee452e9        reva-2              0.00%               3.225GiB / 3.848GiB   83.81%              1.84GB / 1.49MB     0B / 0B             11
2fd9dee452e9        reva-2              0.00%               3.434GiB / 3.848GiB   89.23%              2.46GB / 2.02MB     0B / 0B             12
2fd9dee452e9        reva-2              0.25%               3.547GiB / 3.848GiB   92.18%              2.56GB / 2.05MB     0B / 0B             12
 ... Reva KILLED
```
I am not experienced in the underlying implementation (TUS ?). Anyone?

PR #1253, opened 2020-10-16
ocdav: upload file to storage provider after assembling chunks

ishank011:
Closes https://github.com/cs3org/reva/issues/860


labkode:
@ishank011 let me know when is non wip

refs:
@ishank011 @labkode could you check if you have a file descriptor leaking in the chunked endpoint? We're seeing a lot of open connections to our proxy :) If you need instructions on how to see this with the ocis integration tests reach out to me at any time.

ishank011:
@labkode @butonic the reason for the memory leaks in tus uploads was that the tus client expects an io.ReadSeeker object to be passed but accepts io.Reader as well, which it [tries to typecast](https://github.com/eventials/go-tus/blob/master/upload.go#L87-L89) by copying the whole thing to an in-memory buffer declared on the heap. This stays in memory until the next invocation of the garbage collector, which causes reva to crash due to lack of free memory. A few solutions I tried are:
- Modify the tus client to initialize the buffer of a size equal to the content length, so that it is declared on the stack and that memory is freed when the function execution ends. This doesn't work because bytes internally declares slices.
 ```
(pprof) top
Showing nodes accounting for 1GB, 99.85% of 1GB total
Dropped 51 nodes (cum <= 0.01GB)
      flat  flat%   sum%        cum   cum%
       1GB 99.85% 99.85%        1GB 99.85%  bytes.makeSlice
         0     0% 99.85%        1GB 99.85%  bytes.(*Buffer).ReadFrom
         0     0% 99.85%        1GB 99.85%  bytes.(*Buffer).grow
         0     0% 99.85%        1GB 99.80%  github.com/eventials/go-tus.NewUpload
```
- Use a sync.pool of buffers but that won't scale as well, even if you have tens of concurrent uploads each worth let's say 100 MB each.
- Invoke the garbage collector manually using FreeOSMemory. That cleans a portion of the memory but not the whole thing, given that buffer tries to have a bigger capacity than its length.

The solution which does work is writing the content to a temporary file and passing the file descriptor to that. It introduces a bit of added latency but almost doesn't add at all to the memory overhead, so I think this would be the best solution. Please let me know your thoughts.

**Comment by ishank011** [comment]
I’ll make the change to other places where we initiate the tus client as well.

File: internal/http/services/owncloud/ocdav/put.go [hunk 1]
```diff
@@ -156,18 +154,45 @@ func (s *svc) handlePut(w http.ResponseWriter, r *http.Request, ns string) {
...
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	defer os.RemoveAll(fileName)
+	defer fd.Close()
+	if _, err := io.Copy(fd, r.Body); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
butonic, line 176: with this you are writing the file twice.
once on the machine that is runnung the dataprovider and once when writing it at the destination. 

The whole point of tus and resumable uploads is to not touch the disk of any intermediate services. It is too slow.

The storage provider is responsible for putting any new bytes as close as possible to the target destination, so that the final assembly does not have to copy the bytes across storage boundaries.

A simple PUT request is the worst case because ill the bytes need to make it through ... and it cannot be resumed. Which makes it problematic for large files. That might be the reason why you increased the timeout in the gateway by 10 min.

Let me assure you that there are end users trying to upload 80GB ... in the browser ... which in the past used PUT ... to a machine with a 100mbit connection ... to the nas ...

TUS goes a long way to deal with this. it disconnects from the request context so it can write as many bytes as possible. That is the reason why we are having a hard time getting debug level logging from actual uploads.

Internally, handlePut should just read the body directly and us TUS to send it to the datagateway/dataprovider.

now ... chunking v1 is a different thing. Did you check if you can translate chinking v1 to TUS? because AFAICT we should be able to internally append the incoming PUT requests for every chunk to a single TUS upload to the datagateway/provider. 

But to get things going I would be ok to just collect the chunks in ocdav and then do the upload using TUS internally.

internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -293,16 +294,26 @@ func (s *svc) descend(ctx context.Context, client gateway.GatewayAPIClient, src
 			return fmt.Errorf("status code %d", httpDownloadRes.StatusCode)
 		}
 
+		fileName, fd, err := s.createChunkTempFile()
```
butonic, line 297: see below ... this copies the file to disk before sending it out again ... just read from the body

butonic (changes requested):
We should not write temporary files for every upload and always stream it to the datagateway. See the comment above.

Also you are mixing changes to reuse the httpclient with fixing the chunking ... please send two PRs ... makes it easier to review and get smaller things in asap.

internal/http/services/owncloud/ocdav/put.go
ishank011: Sorry about the multiple changes. Let's focus only on the tus part.

With the current `/remote.php/webdav` PUT endpoint, we're not supporting resumable uploads anyway. We get the whole body at once (80 GB), store that in memory (the tus client does that for the reader to readseeker conversion), and once the transfer is complete, persist that 80 GB allocation until the next run of the garbage collector.

So with the current implementation, PUTting 80 GB won't be successful unless you have 80 GB of memory. If you're able to provide that 80 GB, for the next PUT call, you'll have to allocate another 80 GB. This is the part where the problem creeps in https://github.com/eventials/go-tus/blob/45c7ec8f5d595db5dd50a926fc8420a2c70404fb/upload.go#L87-L89
```
buf := new(bytes.Buffer)
buf.ReadFrom(reader)
stream = bytes.NewReader(buf.Bytes())
```

On my 4GB VM, I'm not able to write even 3 GB files without reva crashing.

ishank011: Sure, the POST handler works, but that assumes that the client itself knows tus and is sending binary chunks by itself, which won't be the case in a browser.

butonic: ouch ...

hm, ok so we consider PUT requests legacy ... because they cannot really be implemented without this hack ... 

The owncloud clients will either learn TUS or already have, so for legacy clients this might be ok.

butonic (dismissed):

butonic:
Ok, we should merge this and optimize handling of many small files in a subsequent PR. For files under 2MB we don't need to create temp files ... or maybe 10MB

butonic:
also, with temp files, uploading large files requires more free disk space ... which is another reason for using tus in the first place.

so if anyone tries to upload large files .... tell him to use tus ... ;-)

ishank011:
Sure, we can optimize it for smaller uploads. Will add that part as well. Thanks @butonic!

butonic:
@ishank011 any update? I'm ok to merge this as is ... we can improve performance for small files in a subsequent PR

ishank011:
@butonic running a few more tests. Will merge it then

ishank011:
@phil-davis this PR fixes the old chunking, so I removed the corresponding tests from the expected failures here https://github.com/cs3org/reva/pull/1253/commits/fef229a3d14a8c4433faafcdcd8657a75f0f475f. Please update these in ocis-reva as well.

There are some tests which still fail due to not being able to clear the log so maybe we can fix those as well https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdavUpload2/uploadFileUsingOldChunking.feature#L45

labkode (approved):

PR #1255, opened 2020-10-19
Check current node when iterating over path segments

butonic:
When checking permissions we were always checking the leaf instead of using the current node while iterating over path segments.


C0rby (approved):
LGTM :+1:

C0rby (approved):
:+1:

ishank011 (approved):

PR #1258, opened 2020-10-20
update grants in the storage provider on share update

C0rby:
The grants set by the ocis storage were not updated when updating a share.
This PR reimplements the code, that was there before but commented out, in a way that is similar to the `addGrant` logic in `CreateShare`

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -186,46 +186,31 @@ func (s *svc) UpdateShare(ctx context.Context, req *collaboration.UpdateShareReq
 	}
 
 	// TODO(labkode): if both commits are enabled they could be done concurrently.
-	/*
-		if s.c.CommitShareToStorageGrant {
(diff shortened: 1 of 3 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -186,46 +186,31 @@ func (s *svc) UpdateShare(ctx context.Context, req *collaboration.UpdateShareReq
...
+		updateGrantStatus, err := s.updateGrant(ctx, getShareRes.GetShare().GetResourceId(),
+			getShareRes.GetShare().GetGrantee(),
+			getShareRes.GetShare().GetPermissions().GetPermissions())
+		if updateGrantStatus.Code != rpc.Code_CODE_OK {
+			return &collaboration.UpdateShareResponse{
+				Status: updateGrantStatus,
+			}, err
 		}
-	*/
+	}
```
butonic, line 213: @labkode any reason why this was commented?

butonic (dismissed):

ishank011 (approved):

PR #1259, opened 2020-10-20
change response status to maintain backwards compatibility to ownClou…

C0rby:
…d 10

When executing a propfind to a resource owned by another user the service would respond with a HTTP 403.
In ownCloud 10 the response was HTTP 207. This change sets the response code to HTTP 207 to stay backwards compatible.


micbar:
@ishank011 @labkode Can you review?

ishank011 (approved):

PR #1260, opened 2020-10-20
Reuse HTTP Clients

refs:
File descriptors on the ocdav service, especially on the put handler was leaking http connections. This PR addresses this.

### DIFF
--- a/changelog/unreleased/ocdav-fix-file-descriptor-leak.md
+++ b/changelog/unreleased/ocdav-fix-file-descriptor-leak.md
@@ -2,4 +2,4 @@ Bugfix: Fix file descriptor leak on ocdav put handler
 
 File descriptors on the ocdav service, especially on the put handler was leaking http connections. This PR addresses this.
 
-https://github.com/cs3org/reva/pull/1259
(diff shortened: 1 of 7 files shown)


ishank011:
Thanks a lot for this @refs! I was looking at the storage provider tus methods and was pulling my hair. This should be good!

Can you fix the lint? We can merge it then

refs:
> Thanks a lot for this @refs! I was looking at the storage provider tus methods for the leaks and was pulling my hair. This should be good!
> 
> Can you fix the lint? We can merge it then

haha, goland, too lazy to configure watchers :D on it

refs:
@ishank011 imports should be good now. waiting for checks

butonic:
datagateway is also leaking ... @refs is adding it to this pr

butonic:
@ishank011 we should double check all places where the rhttp.GetHttpClient() is called. The context that is passed is not used anyway. So we can initialize the client when starting the services. That will allow reusing connections :rocket:

butonic (approved):

ishank011 (approved):
@butonic definitely. I'll get to fixing these next. Thanks!

PR #1264, opened 2020-10-21
fix etag propagation in ocis driver

butonic:
We now use a new synctime timestamp instead of trying to read the mtime to avoid race conditions when the stat request happens too quickly.

fixes https://github.com/owncloud/product/issues/249

### DIFF
--- a/changelog/unreleased/ocis-fix-etag-propagation.md
+++ b/changelog/unreleased/ocis-fix-etag-propagation.md
@@ -1,0 +1,6 @@
+Bugfix: Fix etag propagation in ocis driver
+
+We now use a new synctime timestamp instead of trying to read the mtime to avoid race conditions when the stat request happens too quickly.
+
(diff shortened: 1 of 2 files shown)


butonic:
@phil-davis @individual-it with this we should be able to runhttps://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdavEtagPropagation1/moveFileFolder.feature

C0rby (approved):
:+1:

phil-davis:
Note: we do not see test scenarios starting to pass here, because the core ownCloud10 API tests are only running with `OWNCLOUD` storage.
They also run with `OCIS` storage in `owncloud/ocis` repo - so tests should start passing there, after this code is applied.

I do plan to run the core ownCloud10 API tests with `OCIS` storage here in `cs3org/reva` - I just haven't sorted it out yet.

ishank011 (approved):

PR #1271, opened 2020-10-23
[tests-only] Add oC10 api tests with ocis storage

phil-davis:
OCIS storage was implemented some time ago, but we have not been running the oC10 API tests with OCIS storage. So when making changes we have not been able to have confidence that the changes are good.

This PR runs the oC10 API tests with OCIS storage. By doing this we will know when changes cause regressions (more tests will fail) and when changes improve things (more tests will pass).

1) add pipelines to `.drone.yml`  to run the oC10 API tests with the OCIS storage driver. Copy `expected-failures-on-OCIS-storage.txt` from `owncloud/ocis` (we hope that the list of failures is similar or less). Adjust the various `toml` files used in drone so they specify the data location using the setting `root` (rather than `datadirectory` that was used by the OWNCLOUD storage driver)

2) Adjust the `toml` files so that etag propagation is enabled. `treetime_accounting` and `treesize_accounting` need to be specified `true` 

3) remove apiProvisioning test scenarios from expected-failures - we do not run those in `cs3org/reva`, so they do not fail.

4) Adjust expected-failures-on-OCIS-storage for things that are also the same in cs3org/reva owncloud storage - to align the passing/failing test scenarios

5) Remove trashbin scenarios that pass on cs3org/reva with OCIS storage - these fail in `owncloud/ocis` because usernames there are not yet correctly case-insensitive, but `cs3org/reva` uses LDAP user provisioning and it "just works"

6) Remove previews scenarios - we do not run those yet in cs3org/reva

Having the separate commits above should make it clear to understand how and why the expected failures are different in `owncloud/ocis` and here in `cs3org/reva`


tests/oc-integration-tests/drone/storage-home-ocis.toml
```diff
@@ -30,7 +30,7 @@ data_server_url = "http://localhost:12001/data"
 enable_home_creation = true
 
 [grpc.services.storageprovider.drivers.ocis]
-datadirectory = "/drone/src/tmp/reva/data"
+root = "/drone/src/tmp/reva/data"
```
phil-davis, line 33: That broke lots of ocis storage tests

individual-it (approved):

phil-davis:
@ishank011 this is ready for review...

ishank011 (approved):

PR #1277, opened 2020-10-26
[tests-only] bug demonstration scenario for owncloud/ocis/issues/751

individual-it:
this demonstrates the bug in owncloud/ocis/issues/751

Its useful to know when the bug gets fixed, see https://dev.to/jankaritech/should-you-write-acceptance-tests-for-bugs-2bfm


phil-davis (commented):
Also delete the scenario the same as was deleted in https://github.com/owncloud/ocis/pull/752 ?

phil-davis:
Note: the similar PR to `owncloud/ocis` is https://github.com/owncloud/ocis/pull/752

individual-it:
> Also delete the scenario the same as was deleted in [owncloud/ocis#752](https://github.com/owncloud/ocis/pull/752) ?

it never got ported here

phil-davis:
> it never got ported here

https://github.com/cs3org/reva/blob/master/tests/acceptance/features/apiOcisSpecific/apiShareManagementBasic-deleteShare.feature exists - but actually over time it got changed into a good scenario, that does effectively the same test as what is now in core https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementBasicToShares/deleteShareFromShares.feature#L32

The one here in `cs3org/reva` is not demonstrating any different behaviour to the oC10 API test suite. We can delete it.

Do that in this PR? Or in a separate PR?

phil-davis (approved):

individual-it:
The bug mentioned in the file is not the same as talked about here, but still delete the file because it actually does not demonstrate the bug, nether the one that is mentioned here, nor the one that it claims to demonstrate

ishank011 (approved):

PR #1279, opened 2020-10-27
[tests-only] No need for a redis process for OCIS storage

phil-davis:
The OCIS storage implementation does not require `redis` to be running.
Remove it from the drone services. That will use a little bit less system resources, and more importantly, reduce future confusion.


individual-it (changes requested):
also remove that setting from the toml files

phil-davis:
> also remove that setting from the toml files

redis removed from toml files.

individual-it (approved):

ishank011 (approved):

PR #1285, opened 2020-10-28
Refactor the uploading files workflow from various clients

ishank011:


labkode:
@individual-it for some reason the oc integration test are failing (almost all), can you take a look?
Thanks

phil-davis:
Passing after a rebase.
PRs will need a rebase because the dependencies in the test infrastructure got broken due to PHP composer 2.0 update that happened in owncloud-ci/php docker today (somewhat accidentally, it was intended to just update the PHP patch release versions, but composer 2.0 came "for free" automagically). See issue https://github.com/owncloud/core/issues/38067

PR https://github.com/cs3org/reva/pull/1290 fixed that here in `cs3org/reva`

How to communicate that to everybody that codes in this repo?

labkode (approved):

pkg/storage/fs/ocis/upload.go
```diff
@@ -32,120 +32,55 @@ import (
...
-// TODO deprecated ... use tus
-
 func (fs *ocisfs) Upload(ctx context.Context, ref *provider.Reference, r io.ReadCloser) (err error) {
-
-	var n *Node
-	if n, err = fs.lu.NodeFromResource(ctx, ref); err != nil {
-		return
-	}
-
-	// check permissions
```
butonic, line 53: you removed the permissions checks ...

ishank011: I refactored the Upload method to work with the logic implemented in InitiateFileUpload, so that is taken care of there itself. I'm pushing a fix so that it works without calling that first as well.

PR #1290, opened 2020-11-01
[tests-only] bump core commit id

phil-davis:
1) similar to https://github.com/owncloud/ocis/pull/784

Includes test changes from:
https://github.com/owncloud/core/pull/38050 [tests-only] correct reshare test to use the correct path
https://github.com/owncloud/core/pull/38051 [tests-only] Improve skeleton upload error message
https://github.com/owncloud/core/pull/38052 [tests-only] test webdav locks when shares are received in a sub-folder
https://github.com/owncloud/core/pull/38056 [tests-only] split public link tests into separate files for new/old dav
https://github.com/owncloud/core/pull/38058 [tests-only] Adjust tests to better run on PGSQL

2) issue https://github.com/owncloud/core/issues/38067 - due to composer 2.0 there are issues with dependencies for the `behat` acceptance test runner dependencies when using PHP 7.2 or 7.3. Use PHP 7.4 for the `behat` test runner. (There is no need to look into what is happening with PHP  7.2 or 7.3 - we only use PHP here for the test runner anyway, so may as well use the latest)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -441,7 +441,7 @@ steps:
       - /drone/src/cmd/revad/revad -c ldap-users.toml
 
   - name: clone-oC10-test-repos
-    image: owncloudci/php:7.2
+    image: owncloudci/php:7.4
     commands:
       - git clone -b master --depth=1 https://github.com/owncloud/testing.git /drone/src/tmp/testing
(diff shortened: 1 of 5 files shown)


phil-davis:
@ishank011 ready for approve-merge

ishank011 (approved):

PR #1295, opened 2020-11-04
Test with core branch adjust-testing-on-ocis-reva-logic

phil-davis:
to demonstrate that we do not need to define `TEST_OCIS`

This PR checks that core PR https://github.com/owncloud/core/pull/38077 will be OK


phil-davis:
Acceptance tests pass. It just failed the changelog checks, because I forgot to put [tests-only]

phil-davis:
2nd commit also works.

PR #1297, opened 2020-11-04
check status codes

butonic:
checking err is not enough. the response status code needs to be checked as well.


labkode (dismissed):

labkode:
@butonic please fix the changelog

labkode (approved):

PR #1300, opened 2020-11-05
[Tests-Only] Removed scenarios that were demonstrating Oc10 issue and already exists in core

swoichha:
### Description
There are some good scenarios that were moved in the past to the OCIS repos. So, moving those scenarios back to core

### Moved scenarios
- `Get the mtime of a file inside a folder shared by public link using new webDAV version (run on OCIS)` respective scenario in Oc10 is https://github.com/owncloud/core/blob/6e5e252dcd18448709006cb2de6fe86cb6d0ecec/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature#L793
- `overwriting a file changes its mtime (new public webDAV API) (run on OCIS)` respective scenario in Oc10 is https://github.com/owncloud/core/blob/6e5e252dcd18448709006cb2de6fe86cb6d0ecec/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature#L804
- `API responds with a full set of parameters when owner changes the expireDate of a public share` respective scenario in Oc10 is https://github.com/owncloud/core/blob/6e5e252dcd18448709006cb2de6fe86cb6d0ecec/tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature#L9
- `API responds with a full set of parameters when owner changes the permission of a share` respective scenario in Oc10 is https://github.com/owncloud/core/blob/6e5e252dcd18448709006cb2de6fe86cb6d0ecec/tests/acceptance/features/apiShareUpdateToShares/updateShare.feature#L188

### Related Issue
- Part of https://github.com/owncloud/core/issues/37891


phil-davis (approved):
These scenarios got left-behind and are either already also in `owncloud/core` or are being added to `owncloud/core` now. So they can be cleaned up here. This gets the local API tests here a bit closer to what is currently in `owncloud/ocis` - it is a bit of a never-ending challenge to keep all this stuff in sync!

ishank011 (approved):

PR #1301, opened 2020-11-05
[OCS] filter share with me requests

butonic:
The OCS API now properly filters share with me requests by path and by share status.


butonic:
closing in favor of 1302

PR #1313, opened 2020-11-11
[tests-only] Bump to latest core commit for API tests

phil-davis:
Includes newly-enabled test scenarios from:
https://github.com/owncloud/core/pull/38076 Remove toImplementOnOcis tags from the scenarios

Now we can run a lot more API test scenarios. Most fail because they are still to be implemented (in reva and/or OCIS). But when the features get implemented (correctly!) we will know, because the test scenarios will start passing. `cs3org/reva` is fast at running scenarios so CI does not take much longer than before this change.


phil-davis:
Pass - good stuff.

ishank011 (approved):

PR #1314, opened 2020-11-11
Handle non-TUS uploads called directly with upload path

ishank011:


butonic:
cool, will test right away!

pkg/storage/fs/owncloud/upload.go
```diff
@@ -423,12 +426,12 @@ func (fs *ocfs) AsTerminatableUpload(upload tusd.Upload) tusd.TerminatableUpload
 
 // Terminate terminates the upload
 func (upload *fileUpload) Terminate(ctx context.Context) error {
-	if err := os.Remove(upload.infoPath); err != nil {
+	if err := os.RemoveAll(upload.infoPath); err != nil {
 		if !os.IsNotExist(err) {
```
butonic, line 430: then you can remove the IsNotExist check as well, because RemoleAll does not return an error in that case

butonic (changes requested):

pkg/storage/fs/owncloud/upload.go
ishank011: I can just revert this. The IsNotExist check is not present in localfs and this reported errors for zero-size files in that case. I'll just add it there instead.

pkg/storage/fs/ocis/upload.go
```diff
@@ -44,15 +45,28 @@ var defaultFilePerm = os.FileMode(0664)
 func (fs *ocisfs) Upload(ctx context.Context, ref *provider.Reference, r io.ReadCloser) (err error) {
 	upload, err := fs.GetUpload(ctx, ref.GetPath())
 	if err != nil {
-		return errors.Wrap(err, "ocisfs: error retrieving upload")
+		// Upload corresponding to this ID was not found.
+		// Assume that this corresponds to the resource path to which the file has to be uploaded.
+		buf := &bytes.Buffer{}
+		length, err := io.Copy(buf, r)
```
butonic, line 51: this buffers the whole file in memory before writing it to disk.

butonic: I think it makes more sense to revert the changes and use the lookup of the node before ... I can look into making the chunking work.

butonic: and this reads consumes the body, which makes the subsequent writeChunk not read anything ...

ishank011: @butonic the refactoring concerns less the chunking part and more that we can support both TUS and non-TUS uploads out of the box without ugly config changes.

This step was just to get the length of the upload since that is not explicitly provided. One way to get around this is to set `SizeIsDeferred` to true

butonic: Yes ... and `Upload()` was implementing the non tus based upload. For tus you need `InitiateUpload` and `UseIn`, which you correctly check in the dataprovider ... why did you need to touch the `Upload` implementation?

ishank011: The latest commit gets rid of the buffer issue.

The aim was to have uniform implementation for all types of uploads. Calling InitiateUpload for TUS uploads and not doing so for others didn't seem ideal. Also, since all the file systems did in fact have the `UseIn` method, the PUT calls were handled by a wrapper around the tus client which was also a hack. https://github.com/cs3org/reva/blob/ad43bae0ce75a093eb3915c5414a99f7bac7cc87/internal/http/services/dataprovider/put.go#L59

We wanted to add support for multiple protocols as started in https://github.com/cs3org/reva/pull/828, so all of these should ideally have the same workflow: Call InitiateUpload -> Send PUT/POST call to the returned URL

butonic: testing the PR

butonic (approved):

butonic:
Ok, this now works properly with ocis and the accounts service. @ishank011 Great job on the refactoring!

Let's see what the reva testsuite says.

butonic:
all green!

Issue #1275, opened 2020-10-23
delete reference when unsharing a resource


butonic:
how are we going to clean up a reference when the owner unshares a folder? 

AFAICT the gateway usershareprolider will get a RemoveShare request.
1. we delete the share in the usershareprovider,
2. and we can delete the grant in the storageprovider of the owner.

But we also need to delete the reference of the recipient. how do we do that?

butonic:
we could make the gateway delete references if it encounters invalid ones?

ishank011:
Yes, that could be one option. Or we could run a background cron. We need a periodic garbage collection functionality for a lot of cases - public shares, invalid references, expired tokens in OCM. So it'd be good to have a package for this.

Issue #543, opened 2020-03-06
eos: store share id in inherited xattr


labkode:
As a user I want to know what folder I've shared

PR #1329, opened 2020-11-23
return share-types in ocs propfind responses

C0rby:
Implements https://github.com/owncloud/ocis/issues/929

### DIFF
--- a/pkg/storage/fs/ocis/node.go
+++ b/pkg/storage/fs/ocis/node.go
@@ -34,10 +34,16 @@ import (
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/errtypes"
 	"github.com/cs3org/reva/pkg/mime"
+	"github.com/cs3org/reva/pkg/sdk/common"
 	"github.com/cs3org/reva/pkg/storage/utils/ace"
 	"github.com/pkg/errors"
 	"github.com/pkg/xattr"
 	"github.com/rs/zerolog/log"
+)
+
+const (
+	_ShareTypesKey = "http://owncloud.org/ns/share-types"
+	_UserShareType = "0"
 )
 
 // Node represents a node in the tree and provides methods to get a Parent or Child instance
@@ -291,7 +297,7 @@ func (n *Node) Owner() (id string, idp string, err error) {
 }
 
 // AsResourceInfo return the node as CS3 ResourceInfo
-func (n *Node) AsResourceInfo(ctx context.Context) (ri *provider.ResourceInfo, err error) {
+func (n *Node) AsResourceInfo(ctx context.Context, mdKeys []string) (ri *provider.ResourceInfo, err error) {
 	log := appctx.GetLogger(ctx)
 
 	var fn string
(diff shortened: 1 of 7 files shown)


labkode (changes requested):
This change is very heave on the server side.
Please add  a config option to the dav service: `include-shares-in-propfind` and set it as false by default.

butonic:
@corby the ListShares rpc can take a path argument. Use it to get a list of all grants in a folder instead of every file individually. You can send that request in parralel to the ListContainer.

When the ListContainer request has returned you can prepare a map[fileid]metadata instead of the current list.
When both coroutines have returned you can iterate over the grants and apply them by looking up the metadata using the fileid.

@labkode that should not be heavy.

labkode:
@butonic that is a good improvement, however that config option is needed to enable or disable it on demand. We don't use it for example.

C0rby:
So after talking with @butonic, I rewrote it and we read the shares directly during the ListFolder call from the extended attributes.
This should be more performant but only works for user shares right now. Public link shares will be tackled some other time.

butonic:
related:
- [eos: store share id in inherited xattr #543](https://github.com/cs3org/reva/issues/543)
- [delete reference when unsharing a resource #1275](https://github.com/cs3org/reva/issues/1275)
- [storageprovider should persist share creator #93](https://github.com/cs3org/cs3apis/issues/93)

pkg/storage/fs/ocis/node.go
```diff
@@ -387,6 +392,12 @@ func (n *Node) AsResourceInfo(ctx context.Context) (ri *provider.ResourceInfo, e
 		log.Error().Err(err).Interface("node", n).Msg("could not list attributes")
 	}
 
+	if common.FindString(mdKeys, _ShareTypesKey) != -1 {
+		if n.hasUserShares(ctx) {
+			ri.ArbitraryMetadata.Metadata[_ShareTypesKey] = "0"
```
kulmann, line 397: I was confused by the `"0"` and needed to ask. For the record, it would be really cool to have constants for the share types.

kulmann (approved):
Code looks good

pkg/storage/fs/ocis/node.go
C0rby: Yes, I agree. :+1:

pkg/storage/fs/ocis/node.go
```diff
@@ -469,3 +481,18 @@ func (n *Node) ReadGrant(ctx context.Context, grantee string) (g *provider.Grant
...
+
+func (n *Node) hasUserShares(ctx context.Context) bool {
+	g, err := n.ListGrantees(ctx)
+	if err != nil {
+		appctx.GetLogger(ctx).Error().Err(err).Msg("hasUserShares: listGrantees")
+		return false
+	}
+
+	for i := range g {
+		if strings.Contains(g[i], grantPrefix+"u") {
```
refs, line 493: `"u"` seems like a magic string. It might be because I'm disconnected from this part of the code but it tells me nothing about why `g[i]` has to contain a `prefix + u`.

I think readability could benefit of abstracting the semantics of `"u"` to a package level var with a more descriptive name. WDYT?

C0rby: Yeah, you are right. Will change it. :+1:

refs: thank you! it'd also give me some more context :)

refs (approved):

kulmann (approved):

labkode (approved):

PR #1331, opened 2020-11-23
Fix capabilities response for multiple client versions

ishank011:


labkode:
@ishank011 is not so simple.

We need the following:

```
2.7+ will use Chunking V1 if "capabilities > files > bigfilechunking" is "true" AND "capabilities > dav > chunking" is not there
2.7+ will use Chunking NG if "capabilities > files > bigfilechunking" is "true" AND "capabilities > dav > chunking" = 1.0
2.7+ will use TUS if "capabilities > files > bigfilechunking" is "false" AND "capabilities > dav > chunking" = "" AND "capabilities > files > tus_support" has proper entries.
```

labkode (changes requested):
See comment

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/uploads.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return h.c
+}
+
+func setCapabilitiesForChunkProtocol(cp chunkProtocol, c *data.CapabilitiesData) {
+	switch cp {
+	case chunkV1:
+		// 2.7+ will use Chunking V1 if "capabilities > files > bigfilechunking" is "true" AND "capabilities > dav > chunking" is not there
+		c.Capabilities.Files.BigFileChunking = true
+		c.Capabilities.Dav.Chunking = ""
```
labkode, line 55: In this case the json output cannot contain c.Capabilities.Dav.Chunking, in this case it will output an "empty" string but the key will still be there.

Doing:
```
c.Capabilities.Dav = nil
```

should work

labkode (changes requested):

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/uploads.go
ishank011: That would also remove some of the required DAV capabilities https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/handlers/cloud/capabilities/capabilities.go#L122-L127

I tested v2.7.0 with this code and it did use v1 chunking with this config, so I think that should be good.

labkode: We don't use the other dav parameters, so it should be also safe to remove it completly.

ishank011: Will do

labkode (approved):

PR #1332, opened 2020-11-23
Add import support to Mentix

Daniel-WWU-IT:
This update adds import support to Mentix, transforming it into a **Mesh Entity Exchanger**. To properly support vendor site management, a new connector that works on a local file has been added as well.


labkode (approved):
All the changes are under mentix.

PR #1335, opened 2020-11-24
Set "Not Found" log-line to debug level

IljaN:
404 is an expected signal during normal operation thus should not be logged as application error. We see this error spamming our logs because we are using a GET request to check if a file already exists.


labkode:
@IljaN I agree. Just add a changelog and CI should be green.

ishank011 (approved):

PR #1340, opened 2020-11-26
Add cache to store UID to UserID mapping in EOS

ishank011:


IljaN (dismissed):
Alternatively a sync map could be used: other than that LGTM :+1: 

https://golang.org/pkg/sync/#Map

ishank011:
@IljaN I remembered a discussion about locked maps being more performant than sync https://github.com/cs3org/reva/pull/909#issuecomment-651729867, but reading the spec for this

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys.

I think that would be a better option. I'll update that. Thanks!

IljaN (approved):

labkode (approved):

PR #1342, opened 2020-11-27
Fix home storage redirect for remote.php/dav/files

refs:
There was a regression in which we were not routing correctly to the right storage depending on the URL.

This is a pressing issue for oC mobile clients, the earlier we get it in the better :)

This sticks to webdav's RFC in the way that it's pretty ... unspecified what to do in this case:

> The following rules are the default behavior for any method that
   supports the Depth header.  A method may override these defaults by
   defining different behavior in its definition.

[sauce](https://tools.ietf.org/html/rfc4918#page-70)


kulmann (approved):
👍

ishank011 (approved):

refs:
@ishank011 CI 🍏 , merge? so we can start updating ocis to reva@master

PR #1343, opened 2020-11-27
fix ocis move

butonic:
Use the old node id to build the target path for xattr updates.

https://github.com/cs3org/reva/pull/1343
https://github.com/owncloud/ocis/issues/975

### DIFF
--- a/changelog/unreleased/fix-ocis-move.md
+++ b/changelog/unreleased/fix-ocis-move.md
@@ -1,0 +1,6 @@
+Bugfix: fix ocis move
+
+Use the old node id to build the target path for xattr updates.
+
+https://github.com/cs3org/reva/pull/1343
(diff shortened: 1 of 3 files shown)


kulmann (approved):
Awesome, this indeed fixes https://github.com/owncloud/ocis/issues/975
Code looks good and I ran it locally with ocis to confirm. Creating folders `a` and `b`, then moving `b` into `a` the parentid in the extended attributes of `b` is correct after the move (see screenshot) and I can now upload files into `/a/b` again. 🚀 💪 

Screenshot 2020-11-28 at 06 34 16

refs (approved):
Thank you for getting to the root of it, we (@kulmann and I) were close tho ^^

labkode (approved):

PR #1346, opened 2020-11-30
fix ocs api shared with me response

C0rby:
The path of the files shared with me was incorrect. It was missing the share directory.

Fixes: https://github.com/owncloud/product/issues/204

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -53,6 +53,7 @@ import (
 type Handler struct {
 	gatewayAddr      string
 	publicURL        string
+	sharePrefix      string
 	displayNameCache *ttlmap.TTLMap
 }
 
@@ -61,6 +62,7 @@ func (h *Handler) Init(c *config.Config) error {
 	h.gatewayAddr = c.GatewaySvc
 	h.publicURL = c.Config.Host
 	h.displayNameCache = ttlmap.New(1000, 60)
+	h.sharePrefix = c.SharePrefix
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -813,6 +815,9 @@ func (h *Handler) listSharesWithMe(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 		h.addDisplaynames(r.Context(), gwc, data)
+		// Needed because received shares can be jailed in a folder in the users home
+		data.FileTarget = path.Join(h.sharePrefix, path.Base(info.Path))
+		data.Path = path.Join(h.sharePrefix, path.Base(info.Path))
```
butonic, line 820: hm, the info.Path should already contain the share prefix ... ah wait, the request firts queries the share manager, which only knows file ids, and then sends a stat request to the storage provider to fill in the file details.

@labkode Another reason whd I think we need a service that keeps track of mountpoints. IMO that is the storage registry. It should know the file reference and the mount point. then it can cache the etag for every root, especially the shares. We can then change this hardcoded SharePrefix to an arbitrary location in the users home and calculate the etag based on the information that is cached in the storage registry.

For now this is LGTM

butonic (dismissed):

butonic:
@C0rby the tests seem to indicate that only accepted shares get the prefix.

phil-davis:
> @C0rby the tests seem to indicate that only accepted shares get the prefix.

Yes, in oC10 when a user asks for the shares that are waiting to be accepted (pending) `shared_with_me=true&state=pending` then the response has the folder or file name that the sharer is sharing. The response does not "pre-calculate" that the share will end up in the share receiver's `/Shares` folder.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
C0rby: > hm, the info.Path should already contain the share prefix ... ah wait, the request firts queries the share manager, which only knows file ids, and then sends a stat request to the storage provider to fill in the file details.

Yes, exactly that was the problem here. :see_no_evil:

ishank011 (approved):

PR #1347, opened 2020-11-30
Add a site property to explicitly enable health checks

Daniel-WWU-IT:
This update introduces a new service property called `ENABLE_HEALTH_CHECKS` that must be explicitly set to `true` if a service should be checked for its health status. This allows us to only enable these checks for partner sites only, skipping vendor sites.


Daniel-WWU-IT:
This also adds some more labels to the Prometheus targets which are necessary for health monitoring, too.

labkode (approved):

PR #1354, opened 2020-12-02
OCDAV: map bad request and unimplemented codes

butonic:
We now return a 400 bad request when a grpc call fails with an invalid argument status and a 501 not implemented when it fails with an unimplemented status. This prevents 500 errors when a user tries to add resources to the Share folder or a storage does not implement an action.

This is part of a series to get 500 errors returned as proper 400 errors: owncloud/product#7


ishank011:
@butonic can we make this a bit modular? So many switch cases make reading through the code a bit difficult.

butonic:
@ishank011 I moved the switch to a function and now use that throughout the ocdav handler. using subloggers this even allows logging additional fields, eg. `src` or `dst`.

ishank011 (approved):
Perfect! Will merge once CI is green

labkode:
@butonic maybe rebase

butonic:
@labkode I had swallowed a conflic status when refactoring ... good to merge now.

labkode (approved):

PR #1361, opened 2020-12-08
Remove Expired Link on Access

refs:
@labkode this is a rework of https://github.com/cs3org/reva/pull/959 with a little twist. Instead of doing this on the service layer, as you recommended, I'd rather leave the manager handle deletion. This can come in handy in the future if we decide to introduce [work queues](https://github.com/gocraft/work) to do some async jobs.


refs:
I kind of prefer this approach instead of letting the service layer handle deleted public shares. It could spare quite a few API calls with the added cost that new public share managers implementations have to be aware of this logic. But also there is less magic.

I'm in favor of services just calling the implementation instead of modifying behavior.

labkode (approved):

PR #1365, opened 2020-12-09
return the users username instead of the uuid in share responses

C0rby:
A while ago we decided to use the users uuid only internally and to only display or return the users username. The ocs share api did still return the uuid.

Closes https://github.com/owncloud/ocis/issues/990

### DIFF
--- a/changelog/unreleased/ocs-share-api-userid.md
+++ b/changelog/unreleased/ocs-share-api-userid.md
@@ -1,0 +1,6 @@
+Change: replace the user uuid with the username in ocs share responses
+
+The ocs api should not send the users uuid. Replaced the uuid with the username.
+
+https://github.com/owncloud/ocis/issues/990
+https://github.com/cs3org/reva/pull/1370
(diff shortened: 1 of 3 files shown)


refs (approved):
👍 my only concern is that `UserServiceUserIDResolver` returns a `(string, error)` tuple. I see how this is useful for consensus and have a `resp.User.Username` everywhere instead of a `UUID` I find that a more generic / higher-level API would fit better for a larger project. The conversions package by definition is convoluted.

C0rby:
This PR doesn't seem to trigger drone builds anymore...

phil-davis:
PR #1364 also has no drone.
I raised PR #1366 just now. No drone happened there either.
It seems a recent problem! I asked: https://github.com/cs3org/reva/pull/1366#issuecomment-741886865

Maybe also ask in chat somewhere?

C0rby:
So I ran the two of the failed pipelines locally using `drone exec` and they ran successfully. :)

C0rby:
Closing because of PR against the unsafe branch https://github.com/cs3org/reva/pull/1370

PR #1368, opened 2020-12-10
Enhancement: calculate and expose actual file permission set

butonic:
Instead of hardcoding the permissions set for every file and folder to `ListContainer:true, CreateContainer:true` and always reporting the hardcoded string `WCKDNVR` for the WebDAV permissions we now aggregate the actual cs3 resource permissions in the storage drivers and correctly map them to ocs permissions and webdav permissions using a common role struct that encapsulates the mapping logic.

- refactor the create*Share methods to use the same stat
- use a common permissions check to prevent users from sharing with permissions they don't have
- map between share role, cs3 resource permissionset, ocs permissions
  - [x] the mapping needs to be simplified by mapping everything to roles
  - [x] we can bring back legacy ocs permissions by mapping them to all legacy role combinations and finding corresponding mappings in cs3 permission sets.
- refactor error handling
  - [ ] ~~use ocdav.HandleErrorStatus in more places~~ left for another PR I already touched too many things.
- fix storages and calculate permissions by reading the grants in the path instead of hardcoding `ListContainer:true, CreateContainer:true`
  - [x] ocis
  - [ ] ~~eos~~ I hardcoded returning all permissions, similar to what the ocs api was doing prior to this PR. 
  - [ ] ~~local~~ I hardcoded returning all permissions, similar to what the ocs api was doing prior to this PR. 
  - [ ] ~~s3~~ I hardcoded returning all permissions, similar to what the ocs api was doing prior to this PR. 
  - [ ] ~~owncloud~~ I hardcoded returning all permissions, similar to what the ocs api was doing prior to this PR. 

should fix 
- [QA] Resharing is always allowed even when we do not add that permission https://github.com/owncloud/ocis/issues/552
- [QA] Re-Sharer can increase permissions https://github.com/owncloud/ocis/issues/893
- cannot get ocs:share-permissions via WebDAV https://github.com/owncloud/ocis-reva/issues/47
- [EOS] Cannot access public links https://github.com/owncloud/ocis-reva/issues/315
- [EOS] share not created with default permissions https://github.com/owncloud/ocis-reva/issues/316

related
- upload-only public link does not refer to files-drop page, nor are the permissions enforced https://github.com/owncloud/ocis/issues/723
- path and other information are not shown if a share does not have "read" permission https://github.com/owncloud/ocis/issues/762
- various sharing settings cannot be set https://github.com/owncloud/ocis-reva/issues/41
- Some failing tests with Webdav custom properties https://github.com/owncloud/ocis-reva/issues/217
- Different propfind response compared with core https://github.com/owncloud/ocis-reva/issues/236
- Sharing seems to work but does not work https://github.com/owncloud/ocis-reva/issues/243
- Accessing non-existing public link should return 404, not 500 https://github.com/owncloud/ocis-reva/issues/290
- Public link enforce permissions https://github.com/owncloud/ocis-reva/issues/292
- [OCIS-Storage] etags don't change for a share receiver https://github.com/owncloud/product/issues/243
- [OCIS-storage] resharing doesn't works https://github.com/owncloud/product/issues/265
- [OCIS] share permissions not enforced https://github.com/owncloud/product/issues/270
- [OC Storage] etags are not updated on file operations https://github.com/owncloud/product/issues/279

@labkode eos, localfs, and the other storages need to return a proper PermissionSet for the logged in user so the ocs api can actually return the right ocs permissions. For the ocis I implemented a `ReadUserPermissions()` that aggregates the permissions on all segments of the path. `ReadUserPermissions()` is called in `GetMD()` and `ListContainer()`.   It should be possible to aggregate the permission set for the other storages in a similar manner. Can eos return the acl or permissions for the current user in the fileinfo or is that another call. ... well it might not be necessary if eos really cannot store acls on files directly (only via fuse richacls). 

Another option might be to push setting the permission set to the gateway. it could use the storage registry to fill in the permissions without having to stat every path segment, because it can look at the mount points of shares.

Anyway, this PR currently tries to check if the current user is the owner and returns all permissions if that is the case. This works on eos, ocis, local and owncloud AFAICT. not on s3, yet. For shared folders the ocis driver aggregates the permissions. I did not do thet for the other storages yet, but would start implementing it in the same way, unless you have a better idea how to do it more efficiently. For now I want to make it work. Profiling to make it fast comes later.


phil-davis:
Note: drone CI is not running for some reason today. So we are not getting results to know if the behaviour is improved or not.

labkode:
@phil-davis https://discourse.drone.io/t/new-builds-not-appearing-on-drone/8479

labkode (dismissed):

labkode:
@butonic looks good to me. We'll take care of the EOS implementation. There is a conflict to be resolved, maybe just rebase on top of master.

butonic:
@labkode looking at public links I see that we need to add permission checks and permission set filters to the publicstorageprovider.go from @refs because in the case of public shares ... the storage providers currently cannot determine the permissions ... because only the publicshare manager persists them.

We could implement a publicshare manager that persists shares in the storage ... which would be trivial for ocis. I asume you could persist additional metadata like expiry etc in eos as well ... the question is where should the code live that limits the cs3 permissions set to what the currently logged in user is allowed to do?

We should let the underlying storage system handle as many permission checks as possible, which is would allow a tight eos integration. On the other hand permissions for public links need to be handled in the cs3 publicstorageprovider, based on the share permissions... For the ocis driver we handle permission checks in the driver because we don't integrate with the underlying posix filesystem on purpose.

@labkode do you see my problem with the sharing responsibilities?

butonic:
anyway ... this PR con go in as is

C0rby (dismissed):
Awesome! :+1:

butonic:
converted to draft while implementing public share permissions

butonic:
@labkode @ishank011 hm, ci did not run for https://github.com/cs3org/reva/pull/1368/commits/2dab733487904ed9e6bc098953a9bca0c497e221 and https://github.com/cs3org/reva/pull/1368/commits/a2a2f2bb2ab5b753493137a160cf0c608dea0ec8  🤔 any idea why? did https://drone.cernbox.cern.ch/cs3org/reva go on vacation?

butonic:
fixing a ton of things:
```
62 | .../expected-failures-on-EOS-storage.txt           \|   8 ---
63 | .../expected-failures-on-OCIS-storage.txt          \| 139 -------
64 | .../expected-failures-on-OWNCLOUD-storage.txt      \|  34 --
65 | ...areOperations-getWebDAVSharePermissions.feature \|  23 --
66 | ...harePublicLink2-uploadToPublicLinkShare.feature \|  17 -
 ```

**Comment by butonic** [comment]
@phil-davis hm the `apiWebdavEtagPropagation1/moveFileFolder.feature:99 / 98` tests seem to be unstable:
- https://drone.cernbox.cern.ch/cs3org/reva/90/18/7
```
5366 | runsh: There were no unexpected failures.
5367 | runsh: There were no unexpected success.
```

- https://drone.cernbox.cern.ch/cs3org/reva/92/18/7
```
5365 | runsh: Total unexpected failed scenarios throughout the test run:
5366 | apiWebdavEtagPropagation1/moveFileFolder.feature:99
5367 | runsh: Total unexpected passed scenarios throughout the test run:
5368 | apiWebdavEtagPropagation1/moveFileFolder.feature:98
```

- https://drone.cernbox.cern.ch/cs3org/reva/94/18/7
```
5367 | runsh: Total unexpected failed scenarios throughout the test run:
5368 | apiWebdavEtagPropagation1/moveFileFolder.feature:99
5369 | runsh: There were no unexpected success.
```

might be latency / load related because it seems to be about etag propagation?

**Comment by butonic** [comment]
@phil-davis ah I need to rebase again ... nevermind

**Comment by butonic** [comment]
hm no. `apiWebdavEtagPropagation1/moveFileFolder.feature:98` reliably fails now on owncloud. updating the expected failures accordingly.

**Comment by butonic** [comment]
Ok, I fixed the test cases and removed the resulting expected test failures. There are still tons left, but getting fixing 100+ features feels pretty good. The new expected failure for the owncloud storage driver is I think now failing reliably because the ocs sharing logic now uses real permissions.

File: internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go [hunk 1]
```diff
@@ -19,18 +19,147 @@
...
+			}
+			role = conversions.RoleFromOCSPermissions(permissions)
+		}
+	}
+
+	if statInfo != nil && statInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE {
+		// Single file shares should never have delete or create permissions
+		permissions := role.OCSPermissions()
+		permissions &^= conversions.PermissionCreate
+		permissions &^= conversions.PermissionDelete
```
phil-davis, line 81: Here is this logic again, and was in other code a few times above.

pkg/storage/fs/ocis/node.go
```diff
@@ -273,32 +277,60 @@ func (n *Node) Parent() (p *Node, err error) {
...
+		n.owner.Idp = string(attrBytes)
 	} else {
 		return
 	}
-	return n.ownerID, n.ownerIDP, err
+	return n.owner, err
+}
+
+// PermissionSet returns the permission set for the current user
+// the parent nodes are not takeen into account
```
phil-davis, line 314:
```suggestion
// the parent nodes are not taken into account
```

pkg/storage/fs/ocis/node.go
```diff
@@ -452,6 +480,85 @@ func (n *Node) UnsetTempEtag() (err error) {
...
+	if grantees, err = n.ListGrantees(ctx); err != nil {
+		appctx.GetLogger(ctx).Error().Err(err).Interface("node", n).Msg("error listing grantees")
+		return nil, err
+	}
+
+	// instead of making n getxattr syscalls we are going to list the acls and filter them here
+	// we have two options here:
+	// 1. we can start iterating over the acls / grants on the node or
+	// 2. we can iterate over the number of groups
+	// The current implementation tries to be defensive for cases where users have hundreds or thousants of groups, so we iterate over the existing acls.
```
phil-davis, line 524:
```suggestion
	// The current implementation tries to be defensive for cases where users have hundreds or thousands of groups, so we iterate over the existing acls.
```

tests/acceptance/expected-failures-on-OWNCLOUD-storage.txt
```diff
@@ -1456,6 +1426,7 @@ apiWebdavEtagPropagation1/moveFileFolder.feature:62
 apiWebdavEtagPropagation1/moveFileFolder.feature:78
 apiWebdavEtagPropagation1/moveFileFolder.feature:79
 apiWebdavEtagPropagation1/moveFileFolder.feature:98
+apiWebdavEtagPropagation1/moveFileFolder.feature:99
```
phil-davis, line 1429: Note: this line got deleted in a PR yesterday (somehow it passed in the CI of PR #1385 )
So it is kind-of expected that it fails. (Possibly it is intermittent)

internal/http/services/owncloud/ocs/conversions/role.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+	}
+	if p == PermissionCreate {
+		return NewUploaderRole()
+	}
+	// legacy
+	return NewLegacyRoleFromOCSPermissions(p)
+}
+
+// NewLegacyRoleFromOCSPermissions tries to map ocs a lagecy combination of ocs permissions to cs3 resource permissions as a legacy role
```
phil-davis, line 301:
```suggestion
// NewLegacyRoleFromOCSPermissions tries to map a legacy combination of ocs permissions to cs3 resource permissions as a legacy role
```

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -196,24 +204,26 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
-					w.WriteHeader(http.StatusNotFound)
-				case rpc.Code_CODE_PERMISSION_DENIED:
-					log.Debug().Str("token", token).Interface("status", res.Status).Msg("permission denied")
-					w.WriteHeader(http.StatusForbidden)
-				default:
-					log.Error().Str("token", token).Interface("status", res.Status).Msg("grpc stat request failed")
-					w.WriteHeader(http.StatusInternalServerError)
-				}
+			case sRes.Status.Code == rpc.Code_CODE_PERMISSION_DENIED:
+				fallthrough
```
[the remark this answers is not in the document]
butonic: agreed. github does it for private repos. For the ocis api we need to follow whatever oc10 does ... we can create issues for cases that we notice and they need to be tackled in oc10 and in ocis.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/remote.go
```diff
@@ -82,77 +70,61 @@ func (h *Handler) createFederatedCloudShare(w http.ResponseWriter, r *http.Reque
...
-			response.WriteOCSError(w, r, response.MetaNotFound.StatusCode, "not found", nil)
-			return
-		}
-		response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "grpc stat request failed", err)
-		return
+	if statInfo != nil && statInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE {
+		// Single file shares should never have delete or create permissions
+		permissions := role.OCSPermissions()
+		permissions &^= conversions.PermissionCreate
+		permissions &^= conversions.PermissionDelete
```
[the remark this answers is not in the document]
butonic: yeah, the conversions package needs to take into account the file type. so it can calculate the correct set of ocs permissions and cs3 permission flags for a role. As the comment says files never have an OCS Create and Delete permission.

phil-davis:
https://drone.cernbox.cern.ch/cs3org/reva/105/18/7
```
apiWebdavEtagPropagation1/moveFileFolder.feature:98
```

IMO those etag propagation scenarios are intermittent. Many have already been skipped on owncloud storage. Issue https://github.com/owncloud/product/issues/280 - I will make a PR to skip these examples at line 98 and 99 that seem to be giving pain now.

Restart CI here and it might come green :)

butonic:
green again

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -471,3 +447,75 @@ func (h *Handler) removePublicShare(w http.ResponseWriter, r *http.Request, shar
 
 	response.WriteOCSSuccess(w, r, nil)
 }
+
+func ocPublicPermToCs3(permKey int, h *Handler) (*provider.ResourcePermissions, error) {
+	role, ok := ocPublicPermToRole[permKey]
+	if !ok {
+		log.Error().Str("ocPublicPermToCs3", "shares").Msgf("invalid oC permission: %s", role)
```
C0rby, line 454: role would be nil here wouldn't it?

pkg/storage/fs/ocis/node.go
```diff
@@ -273,32 +277,60 @@ func (n *Node) Parent() (p *Node, err error) {
 
 // Owner returns the cached owner id or reads it from the extended attributes
 // TODO can be private as only the AsResourceInfo uses it
-func (n *Node) Owner() (id string, idp string, err error) {
-	if n.ownerID != "" && n.ownerIDP != "" {
-		return n.ownerID, n.ownerIDP, nil
+func (n *Node) Owner() (o *userpb.UserId, err error) {
```
C0rby, line 280: Am I blind or is the UserId param never being used?

phil-davis:
@butonic PR #1382 has been merged and that changed the core commit id and some lines in expected-failures files.
It would be good to rebase this PR to be sure the the combination of new core commit id and merged expected-failures changes passes cleanly.

C0rby (approved):
Looks good. Some small remarks. :+1:

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -530,14 +645,14 @@ func (s *service) trimMountPrefix(fn string) (string, error) {
 	if strings.HasPrefix(fn, s.mountPath) {
 		return path.Join("/", strings.TrimPrefix(fn, s.mountPath)), nil
 	}
-	return "", errors.New(fmt.Sprintf("path=%q does not belong to this storage provider mount path=%q"+fn, s.mountPath))
+	return "", errors.Errorf("path=%q does not belong to this storage provider mount path=%q"+fn, s.mountPath)
```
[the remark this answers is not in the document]
butonic: The error handling should be handled in another pr ... IIRC hound or a lint complained about this ...

internal/grpc/services/authprovider/authprovider.go
```diff
@@ -108,18 +109,26 @@ func (s *service) Authenticate(ctx context.Context, req *provider.AuthenticateRe
 	password := req.ClientSecret
 
 	u, err := s.authmgr.Authenticate(ctx, username, password)
-	if err != nil {
+	switch v := err.(type) {
+	case nil:
+		log.Info().Msgf("user %s authenticated", u.String())
```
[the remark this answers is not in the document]
butonic: it was info before. see https://github.com/cs3org/reva/pull/1368/files#diff-187ee2d4618b4a46049ec9c40c111e1420baecb2622be545ec0c9e8cab196a07L119

could you raise an issue or try linkin an issue that deals with log levels?

pkg/storage/fs/ocis/node.go
butonic: it is returned ;-)

phil-davis:
@butonic PR #1387 was merged. That removed the intermittent Etag Propagation tests from the test suite.
Rebase and CI should go green again.

phil-davis:
There are a few more "unexpected passed scenarios" for some newer API tests that are passing. More deleting from expected-failures file :rocket: 

https://drone.cernbox.cern.ch/cs3org/reva/127/16/7
```
apiShareOperationsToShares/uploadToShare.feature:280
```

https://drone.cernbox.cern.ch/cs3org/reva/127/17/7
```
apiShareUpdateToShares/updateShare.feature:384
apiShareUpdateToShares/updateShare.feature:385
```

phil-davis:
@butonic PR #1388 was merged. IMO it will be good to rebase this and get fresh CI.
There are no conflicts reported, and I don't think there will be any interaction between that "range request" code and this "permissions" code. But a fresh run of CI  will make sure.

refs:
bug while testing:


1. log in as admin:admin
2. create a folder and share it with Marie
3. as marie: upload a file to the folder
4. as admin: try to delete the file that Marie uploaded to the shared folder

expected: admin should be able to delete the file as she is the owner of the shared folder
got: admin cannot delete file uploaded to shared folder

butonic:
- [ ] moving a file into a shared folder fails with 400
```
2021-01-07T14:56:58Z DBG gateway: split: parts[1]:Invoice for MBA39A4.pdf != shareFolder:Shares pkg=rgrpc service=storage traceid=5108148c26c8de8e51abc9ab1f42f74c
2021-01-07T14:56:58Z DBG gateway: split: parts[1]:Invoice for MBA39A4.pdf != shareFolder:Shares pkg=rgrpc service=storage traceid=5108148c26c8de8e51abc9ab1f42f74c
2021-01-07T14:56:58Z DBG unary code=OK end="07/Jan/2021:14:56:58 +0000" from=tcp://127.0.0.1:56288 pkg=rgrpc service=storage start="07/Jan/2021:14:56:58 +0000" time_ns=268600 traceid=5108148c26c8de8e51abc9ab1f42f74c uri=/cs3.gateway.v1beta1.GatewayAPI/Move user-agent=grpc-go/1.26.0
2021-01-07T14:56:58Z DBG bad request dst="/home/Shares/Neuer Ordner/Invoice for MBA39A4.pdf" pkg=rhttp service=storage src="/home/Invoice for MBA39A4.pdf" status={"code":4,"message":"gateway: move:error: bad request: gateway: move called on unknown path: /home/Invoice for MBA39A4.pdf","trace":"5108148c26c8de8e51abc9ab1f42f74c"} traceid=5108148c26c8de8e51abc9ab1f42f74c
2021-01-07T14:56:58Z WRN http end="07/Jan/2021:14:56:58 +0000" host=127.0.0.1 method=MOVE pkg=rhttp proto=HTTP/1.1 service=storage size=0 start="07/Jan/2021:14:56:58 +0000" status=400 time_ns=39223900 traceid=5108148c26c8de8e51abc9ab1f42f74c uri=/remote.php/webdav/Invoice%20for%20MBA39A4.pdf url=/remote.php/webdav/Invoice%20for%20MBA39A4.pdf
```

butonic:
now at 200+ fixed features:
```
61 | .../expected-failures-on-EOS-storage.txt           \|   8 ---
62 | .../expected-failures-on-OCIS-storage.txt          \| 176 ---------
63 | .../expected-failures-on-OWNCLOUD-storage.txt      \|  47 +--
```

labkode:
@butonic  https://lgtm.com/projects/g/cs3org/reva/rev/pr-0625c1d67fb1de4e2463c285552d4ead77d61d9e

phil-davis (dismissed):
LGTM, the permissions are much better than before. IMO after merging this, people can work on more of the "ToDo"

phil-davis:
> @butonic https://lgtm.com/projects/g/cs3org/reva/rev/pr-0625c1d67fb1de4e2463c285552d4ead77d61d9e

A later analysis at https://lgtm.com/projects/g/cs3org/reva/rev/pr-6affa6881c70272cacb9a1b528350a524a4c180d no longer complains about that.

butonic:
> @butonic https://lgtm.com/projects/g/cs3org/reva/rev/pr-0625c1d67fb1de4e2463c285552d4ead77d61d9e

fixed in https://github.com/cs3org/reva/pull/1368/commits/c69f4705e19aa25c0e4488f63c05a8e2bfda1a09

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -550,16 +665,22 @@ func (s *service) pathFromToken(ctx context.Context, token string) (string, erro
 			},
 		},
 	)
-	if err != nil {
-		return "", err
+	switch {
+	case err != nil:
+		return "", nil, nil, err
+	case publicShareResponse.Status.Code != rpc.Code_CODE_OK:
+		return "", nil, publicShareResponse.Status, nil
```
ishank011, line 672: Returning the status only when it's not OK seems to make it a bit inconsistent for me. Not a big deal but I'd prefer if we always return the status and check if it's OK or not in the caller.

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -275,6 +280,26 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 		Propstat: []propstatXML{},
 	}
 
+	role := conversions.RoleFromResourcePermissions(md.PermissionSet)
+
+	// files never have the create container permission
+	if md.Type == provider.ResourceType_RESOURCE_TYPE_FILE && md.PermissionSet != nil {
+		// we need to copy the permission set in case it was reused when passing it in while listing a collection
+		md.PermissionSet = copyPermissionSet(md.PermissionSet)
```
ishank011, line 288: Since we're assigning the permissions to the same variable, we do lose the original value. Why are we making a separate copy then?

ishank011 (commented):
I think we can merge this. I'll add the ACL traversal for eos. Since we don't really have permissions support for localfs, we can just let it be.

internal/http/services/owncloud/ocdav/propfind.go
butonic: PermissionSet is a struct which is passed by reference. There are code paths that will change the PermissionSet  for files, because they never have delete or create permission IIRC. When listing a folder this would change the default PermissionSet and folders that are rendered after a file would retain the changed default.

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
butonic: We are doing the Status.Code checks here to determine if we can continue or if we need to return the Status so the caller can decide what to do. `resolveToken()` is called in three places and I think this method does exactly what you intend: resolve the token but give me the error or status code if something goes wrong.

labkode:
About the EOS ACLs, EOS is different to local/ocis filesystems, EOS does not ACL traversal, ACLs are examined on the node (not from parents). It is possible in EOS to not have any permissions on /a/b/ but have on /a/b/c.

internal/http/services/owncloud/ocdav/propfind.go
ishank011: Yes but this particular function is called only from one place and we're forgetting the original reference (since we're storing the returned struct there) so essentially there's no way to get that back.

butonic: ok, you are right ... I tried cleaning the permissions and roles stuff up. I think I was confusing this with the storage drivers ...

butonic: all tests green!

ishank011 (approved):

internal/http/services/owncloud/ocdav/propfind.go
ishank011: Perfect!

PR #1370, opened 2020-12-14
return the users username instead of the uuid in share responses

C0rby:
A while ago we decided to use the users uuid only internally and to only display or return the users username. The ocs share api did still return the uuid.

Closes https://github.com/owncloud/ocis/issues/990

### DIFF
--- a/changelog/unreleased/ocs-share-api-userid.md
+++ b/changelog/unreleased/ocs-share-api-userid.md
@@ -1,0 +1,6 @@
+Change: replace the user uuid with the username in ocs share responses
+
+The ocs api should not send the users uuid. Replaced the uuid with the username.
+
+https://github.com/owncloud/ocis/issues/990
+https://github.com/cs3org/reva/pull/1370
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -273,15 +275,38 @@ func AsCS3Permissions(p int, rp *provider.ResourcePermissions) *provider.Resourc
 	return rp
 }
 
+// UserServiceUserIDResolver returns a userId resolver using the UserService
+func UserServiceUserIDResolver(ctx context.Context, client gatewayv1beta1.GatewayAPIClient) func(*userpb.UserId) (string, error) {
+	return func(id *userpb.UserId) (string, error) {
+		resp, err := client.GetUser(ctx, &userpb.GetUserRequest{UserId: id})
```
butonic, line 281: this must be cached, otherwise we will make an rpc to the gateway and then to the user provider, probably to ldap ... and back, several times per file list entry ...

butonic (changes requested):
this must be cached, otherwise we will make an rpc to the gateway and then to the user provider, probably to ldap ... and back, several times per file list entry ...

internal/http/services/owncloud/ocs/conversions/main.go
C0rby: Ugh... I think this is also causing the tests to fail... The ldap server gets overwhelmed and then crashes...
Locally the pipelines work... :

C0rby: I implemented the username lookup similar to the display name lookup which you implemented. Including the caching.

kulmann (approved):

PR #1374, opened 2020-12-16
Delete public links on GET (#1364)

labkode:


refs:
might need a rebase with master, I see some scenarios that were expected to fail that passed.

labkode:
@refs just rebased, let's see

refs (approved):
👍

PR #1378, opened 2020-12-17
[tests-only] Bump core commit id for tests

phil-davis:
Similar to https://github.com/owncloud/ocis/pull/1111

Gets latest test code and scenarios from `owncloud/core` for TUS and other tests.


phil-davis:
@labkode @ishank011 if you are back online today, this can be safely merged. It passed drone CI before Christmas.
Then when the drone hardware is running again, I can move forward with more test-suite things.

ishank011 (approved):

PR #1381, opened 2020-12-22
[tests-only] Trigger CI

phil-davis:
to see if drone is running


phil-davis:
There is no status reported from drone CI. I guess it is on Christmas holiday.

PR #1391, opened 2021-01-07
New reva errors package

ishank011:


pkg/errors/errors.go
```diff
@@ -0,0 +1,49 @@
...
+
+package errors
+
+import (
+	"fmt"
+	"path"
+	"runtime"
+	"strings"
+
+	"github.com/pkg/errors"
```
labkode, line 27: Can we rely instead on the std lib errors package?

import "errors" ?

ishank011: The stdlib package doesn't have the wrap functionality.

labkode: Should be possible with fmt package together, check the documentation for the examples:

```
A simple way to create wrapped errors is to call fmt.Errorf and apply the %w verb to the error argument:

errors.Unwrap(fmt.Errorf("... %w ...", ..., err, ...))
```

PR #1400, opened 2021-01-14
initial checksum support for ocis

butonic:
_TL;dr_

Fixes https://github.com/owncloud/ocis/issues/1291

_Long version_

The ocis driver now calculates the sha1, md4 and adler32 checksums when finalizing uploads. While this adds significant io because every file that was written has to be read it ensures that the hashes we store in the metadata reflect what was wraitten to the storage. We may improve performance in two ways:
1. run a go routine to calculate the hashas asynchronously
2. for uploads that are finished within one request calculate the hash when writing the file (although this would miss the point of the checksum, but this is what we do in oc10)

For chunked uploads this already does the right thing: read the completed file from disk and calculate the hashes.

oc10 supports multiple checksums for a file, even if the oc10 db contains only one checksum column and the propfind returns sth like
```xml
        <oc:checksums>
          <oc:checksum>SHA1:941f6bb205d2e9f MD5:aeoaeoaeoaeoaeo ADLER32:iaeoioeao</oc:checksum>
        </oc:checksums>
```

that is a multi value tag badly encoded as a single string / xml tag value. 🤦  ... legacy reasons ... 😭 

Anyway, the cs3 Resource info can only transport one Checksum. @ishank011 @labkode We should let it hold multiple checksums.

History:
introduced with https://github.com/owncloud/core/pull/22199 ... here the idea still was:
```xml
<oc:checksums>
  <oc:checksum>TYPE1:CHECKSUM1</oc:checksum>
  <oc:checksum>TYPE2:CHECKSUM2</oc:checksum>
</oc:checksum>
```

no idea where it broke. tried the related oc10 issues:
https://github.com/owncloud/core/issues/18716
https://github.com/owncloud/client/issues/3735
https://github.com/owncloud/ios-legacy/issues/719
https://github.com/owncloud/core/issues/22711
https://github.com/owncloud/core/issues/11811
https://github.com/owncloud/documentation/issues/2964

I only learned that the desktop client onld reads and sends the `OC-Checksum header`.

Will continue digging tomorrow...


butonic:
todo:
```

runsh: Total unexpected failed scenarios throughout the test run:
--
983 | apiOcisSpecific/apiMain-checksums.feature:18
984 | apiOcisSpecific/apiMain-checksums.feature:19
985 | apiOcisSpecific/apiMain-checksums.feature:23
986 | apiOcisSpecific/apiMain-checksums.feature:61
987 | apiOcisSpecific/apiMain-checksums.feature:62


```

butonic:
The original pr that should have mades checksums [more future proof](https://github.com/owncloud/core/pull/22199) expects a [comma separated list of checksums](https://github.com/owncloud/core/pull/22199/files#diff-93a78b3254674a734c2c52faa21609f095658cea04537cf96c19ef0bc143d514R43). see https://github.com/owncloud/core/pull/21997#issuecomment-181230285 for the first mention of commas.

PR to fix this is owncloud core is: https://github.com/owncloud/core/pull/38304

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -384,14 +385,35 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 		lastModifiedString := t.Format(time.RFC1123Z)
 		response.Propstat[0].Prop = append(response.Propstat[0].Prop, s.newProp("d:getlastmodified", lastModifiedString))
 
+		// stay bug compatible with oc10, see https://github.com/owncloud/core/pull/38304#issuecomment-762185241
+		var checksums strings.Builder
```
refs, line 389: wouldn't it actually be easier by using a struct with XML tags and [xml.Marshal](https://golang.org/pkg/encoding/xml/#Marshal)?

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -204,6 +206,36 @@ func (s *svc) handlePutHelper(w http.ResponseWriter, r *http.Request, content io
...
+	// TUS Upload-Checksum header takes precedence
+	if checksum := r.Header.Get("Upload-Checksum"); checksum != "" {
+		cparts = strings.SplitN(checksum, " ", 2)
+		if len(cparts) != 2 {
+			sublog.Debug().Str("upload-checksum", checksum).Msg("invalid Upload-Checksum format, expected '[algorithm] [checksum]'")
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		// Then try owncloud header
+	} else if checksum := r.Header.Get("OC-Checksum"); checksum != "" {
```
refs, line 221: As a bandaid, you could implement the multi `algorithm:hash` value here, so we still have compatibility with oc10

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -249,8 +281,21 @@ func (s *svc) handlePutHelper(w http.ResponseWriter, r *http.Request, content io
 				w.WriteHeader(http.StatusPartialContent)
 				return
 			}
+			if httpRes.StatusCode == errtypes.StatusChecksumMismatch {
+				w.WriteHeader(http.StatusBadRequest)
+				xml := `<?xml version="1.0" encoding="utf-8"?>
+<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
+  <s:exception>Sabre\DAV\Exception\BadRequest</s:exception>
+  <s:message>The computed checksum does not match the one received from the client.</s:message>
+</d:error>`
```
refs, line 290: I believe we have a package for this kind of errors: https://github.com/cs3org/reva/blob/bd7234e5b29361f8a0047f620c72671aeca0ded7/internal/http/services/owncloud/ocdav/error.go, in particular: https://github.com/cs3org/reva/blob/bd7234e5b29361f8a0047f620c72671aeca0ded7/internal/http/services/owncloud/ocdav/error.go#L48-L55

pkg/storage/fs/ocis/node.go
```diff
@@ -342,7 +345,7 @@ func calculateEtag(nodeID string, tmTime time.Time) (string, error) {
 	} else {
 		return "", err
 	}
-	return fmt.Sprintf(`"%x"`, h.Sum(nil)), nil
```
refs, line 345: hmm? What's the change here? I would rather use the Printf verb. They both yield the same, [see this playground I just put together](https://play.golang.com/p/M-POh8s5rl2)

```golang
package main

import (
	"encoding/hex"
	"fmt"
)

func main() {
	fmt.Printf("raw strings concat:\t%v\nhex verb:\t\t%v",
		`"` + hex.EncodeToString([]byte("1234")) + `"`,
		fmt.Sprintf(`"%x"`, "1234"))
}
```

yields

    raw strings concat:	"31323334"
    hex verb:		    "31323334"
    Program exited.

pkg/storage/fs/ocis/ocis.go
```diff
@@ -67,9 +67,9 @@ const (
 	favPrefix string = ocisPrefix + "fav."
 
 	// a temporary etag for a folder that is removed when the mtime propagation happens
-	tmpEtagAttr   string = ocisPrefix + "tmp.etag"
-	referenceAttr string = ocisPrefix + "cs3.ref" // arbitrary metadata
-	//checksumPrefix    string = ocisPrefix + "cs."   // TODO add checksum support
+	tmpEtagAttr     string = ocisPrefix + "tmp.etag"
+	referenceAttr   string = ocisPrefix + "cs3.ref"      // arbitrary metadata
+	checksumPrefix  string = ocisPrefix + "cs."          // checksum support
```
refs, line 72: I think we can spare the comment here, the var name is descriptive enough

pkg/storage/fs/ocis/upload.go
```diff
@@ -41,6 +47,8 @@ import (
 
 var defaultFilePerm = os.FileMode(0664)
 
+// TODO Upload (and InitiateUpload) needs a way to receive the expected checksum.
+// Maybe in metadata as 'checksum' => 'sha1 aeosvp45w5xaeoe' = lowercase, space separated?
```
refs, line 51: worth tracking as technical debt? I fear this will get lost in the comments

refs: would be nice to end up with a list of new issues out of this PR that adds technical debt. I personally don't go through my TODO's so often, although there is great value in doing so, for the sake of the project. I feel there are lots of TODO comments unadressed.

pkg/storage/fs/ocis/upload.go
```diff
@@ -129,6 +138,18 @@ func (fs *ocisfs) InitiateUpload(ctx context.Context, ref *provider.Reference, u
 		if _, ok := metadata["sizedeferred"]; ok {
 			info.SizeIsDeferred = true
 		}
+		if metadata["checksum"] != "" {
+			parts := strings.SplitN(metadata["checksum"], " ", 2)
+			if len(parts) != 2 {
+				return nil, errtypes.BadRequest("invalid checksum format. must be '[algorithm] [checksum]'")
```
refs, line 144: this error is missleading, according to your comment:

```xml
<oc:checksums>
  <oc:checksum>SHA1:941f6bb205d2e9f MD5:aeoaeoaeoaeoaeo ADLER32:iaeoioeao</oc:checksum>
</oc:checksums>
```

there could be the case that a client sends more than one tuple of `<algo:hash>` that is not being accounted for. We should either:

- check that and throw an error because we don't support it yet
- add a check to deal with this case

pkg/storage/fs/ocis/node.go
butonic: @C0rby said `hex.EncodeToString` is faster

refs: probably, but, style? ^^

internal/http/services/owncloud/ocdav/put.go
butonic: `OC-Checksum` uses `:` as the separator. `Upload-Checksum` uses ` `. They both get forwarded with a lowercased algorithm and a ` ` separated checksum.

The OC10 code checks if the `OC-Checksum` headir is contained in the cheksum string that is stored in the database. So supporting more than one checksum in the header is currently not necessary imo.

pkg/storage/fs/ocis/node.go
butonic: k, for readability I'll go with `fmt.Sprintf(`"%x"`, h.Sum(nil)), nil`

internal/http/services/owncloud/ocdav/propfind.go
butonic: it would if we were rendering the whole propfind using an xml marshaler. Now we would be assembling a new property, running the xml marshaler on it, and assigning the bytes to the prop ... something for a different PR.

pkg/storage/fs/ocis/upload.go
butonic: tracked in https://github.com/cs3org/cs3apis/issues/103

pkg/storage/fs/ocis/upload.go
butonic: I am following the TUS `Upload-Checksum` header internally ... and that is single value ... anyway. Part of https://github.com/cs3org/cs3apis/issues/103 I'd say because currently this is all in Opaque headers.

butonic:
k, this fixes 22 checksum features. the remaining failures:
```

/drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:58 | 70s
-- | --
1264 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:67 | 70s
1265 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:119 | 70s
1266 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:129 | 70s
1267 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:138 | 70s
1268 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:147 | 70s
1269 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:158 | 70s
1270 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:174 | 70s
1271 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:192 | 70s
1272 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:258 | 70s
1273 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:312 | 70s
1274 | /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:324


```

are relvated to chunked upload or sharing and subject of a subsequent PR.

labkode:
@butonic is this compatible with existing sync clients?

butonic:
@labkode yes. I made sure that the rendering, as weird as it looks, matches what clients expect. Unfortunately, the desktop client would not properly handle correct xml like
```xml
<oc:checksums>
  <oc:checksum>TYPE1:CHECKSUM1</oc:checksum>
  <oc:checksum>TYPE2:CHECKSUM2</oc:checksum>
</oc:checksum>
```

so I implemented this bug compatible with oc10 🤷‍♂️ https://github.com/owncloud/core/pull/38304 contains the fix as well as discussion on what iOS and the desktop would do.

refs:
@butonic while trying this out I get tons of errors:

grafik

even after deleting everything under `/var/tmp/ocis`

refs:
I also don't see the `<oc:checksum>` header on the response XML:

```xml
<d:multistatus xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:s="http://sabredav.org/ns">
   <d:response>
      <d:href>/remote.php/webdav/</d:href>
      <d:propstat>
         <d:prop>
            <oc:permissions>DNVCKR</oc:permissions>
            <oc:favorite>0</oc:favorite>
            <oc:fileid>MTI4NGQyMzgtYWE5Mi00MmNlLWJkYzQtMGIwMDAwMDA5MTU3OmZhOGU1MjYyLTZhNjMtNDUzNS1hNzY4LTk0NzI1NDU3MzRiNw==</oc:fileid>
            <oc:owner-id>admin</oc:owner-id>
            <oc:owner-display-name>Admin</oc:owner-display-name>
            <oc:size>96</oc:size>
            <d:getlastmodified>Tue, 19 Jan 2021 16:23:43 +0000</d:getlastmodified>
            <d:getetag>"41aead0722284a3a739c9138996acd23"</d:getetag>
            <d:resourcetype>
               <d:collection />
            </d:resourcetype>
         </d:prop>
      </d:propstat>
      <d:propstat>
         <d:prop>
            <oc:share-types />
            <oc:privatelink />
            <d:getcontentlength />
         </d:prop>
         <d:status>HTTP/1.1 404 Not Found</d:status>
      </d:propstat>
   </d:response>
   <d:response>
      <d:href>/remote.php/webdav/ownCloud_Admin_Training_Handout.pdf</d:href>
      <d:propstat>
         <d:prop>
            <oc:permissions>DNVWR</oc:permissions>
            <oc:favorite>0</oc:favorite>
            <oc:fileid>MTI4NGQyMzgtYWE5Mi00MmNlLWJkYzQtMGIwMDAwMDA5MTU3OmFiZmViNTAzLTNjMzctNDRkMS05ZjJiLWM2ZWJhMWQ4MWIyNg==</oc:fileid>
            <oc:owner-id>admin</oc:owner-id>
            <oc:owner-display-name>Admin</oc:owner-display-name>
            <d:getcontentlength>395672</d:getcontentlength>
            <oc:size>395672</oc:size>
            <d:getlastmodified>Tue, 19 Jan 2021 16:23:43 +0000</d:getlastmodified>
            <d:getetag>"4958951fc0a36dd05d6e200849f54f94"</d:getetag>
            <d:resourcetype />
         </d:prop>
      </d:propstat>
      <d:propstat>
         <d:prop>
            <oc:share-types />
            <oc:privatelink />
         </d:prop>
         <d:status>HTTP/1.1 404 Not Found</d:status>
      </d:propstat>
   </d:response>
</d:multistatus>
```

butonic:
those are for existing files without checksums ;-)

refs:
> those are for existing files without checksums ;-)
> ...
> or might be folders 🤔

Hm, they also popup when I delete everything on `/var/tmp/ocis/*`

refs:
so correct me if I'm wrong but after an upload the uploaded file should have a checksum and this should be included in the `PROPFIND` request, is that right?

labkode:
@refs https://github.com/cernbox/smashbox/blob/master/protocol/protocol.md#checksumming-extensions

butonic:
@refs the errors are caused by the  metadata storage ... fixing and rebasing

butonic:
> so correct me if I'm wrong but after an upload the uploaded file should have a checksum and this should be included in the `PROPFIND` request, is that right?

only if it was requested in the PROPFIND:
```
curl 'https://cloud.ocis.test/remote.php/webdav/'   -X 'PROPFIND' -H 'depth: 1' -d $'<?xml version="1.0"?>\n<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n  <d:prop>\n    <oc:permissions />\n    <oc:favorite />\n    <oc:fileid />\n    <oc:owner-id />\n    <oc:owner-display-name />\n    <oc:share-types />\n    <oc:privatelink />\n <oc:checksums />   <d:getcontentlength />\n    <oc:size />\n    <d:getlastmodified />\n    <d:getetag />\n    <d:resourcetype />\n  </d:prop>\n</d:propfind>' -k -u einstein:relativity | xmllint -format -
```
`<oc:checksums />` is the magic key

refs:
> > so correct me if I'm wrong but after an upload, the uploaded file should have a checksum and this should be included in the `PROPFIND` request, is that right?
> 
> @refs only if it was requested in the PROPFIND:
> 
> ```
> curl 'https://demo.owncloud.com/remote.php/webdav/ownCloud Manual.pdf'   -X 'PROPFIND' -d $'<?xml version="1.0"?>
> > <d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
> >   <d:prop>
> >     <oc:checksums />
> >   </d:prop>
> > </d:propfind>' -k -u demo:demo | xmllint -format -
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100   593  100   451  100   142   1518    478 --:--:-- --:--:-- --:--:--  1996
> <?xml version="1.0"?>
> <d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns">
>   <d:response>
>     <d:href>/remote.php/webdav/ownCloud%20Manual.pdf</d:href>
>     <d:propstat>
>       <d:prop>
>         <oc:checksums>
>           <oc:checksum>SHA1:d4c9b536a6d082dcef9b96a9956986bd3472196f MD5:7dfd959493b7680b5a9e6dae7e2912f8 ADLER32:f7d46dfc</oc:checksum>
>         </oc:checksums>
>       </d:prop>
>       <d:status>HTTP/1.1 200 OK</d:status>
>     </d:propstat>
>   </d:response>
> </d:multistatus>
> ```
> 
> `<oc:checksums />` is the magic key

oooooh I see. Shows I'm still fresh flesh with the WebDAV spec. Thanks for explaining.

refs (approved):

pkg/storage/fs/ocis/upload.go
```diff
@@ -401,16 +425,67 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) (err error) {
...
+
+	r1 := io.TeeReader(f, sha1h)
+	r2 := io.TeeReader(r1, md5h)
+
+	if _, err := io.Copy(adler32h, r2); err != nil {
+		sublog.Err(err).Msg("ocisfs: could not copy bytes for checksumming")
+	}
+
+	// compare if they match the sent checksum
+	// TODO the tus checksum extension would do this on every chunk, but I currently don't see an easy way to pass in the requested checksum. for now we do it in FinishUpload which is also called for chunked uploads
```
ishank011, line 451: Would the client send the checksum only with the last chunk or with all of those?

pkg/storage/fs/ocis/upload.go
```diff
@@ -401,16 +425,67 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) (err error) {
...
+	if upload.info.MetaData["checksum"] != "" {
+		parts := strings.SplitN(upload.info.MetaData["checksum"], " ", 2)
+		if len(parts) != 2 {
+			return errtypes.BadRequest("invalid checksum format. must be '[algorithm] [checksum]'")
+		}
+		switch parts[0] {
+		case "sha1":
```
ishank011, line 458: Maybe make this a bit modular? Same comment for the `SetChecksum` part

ishank011 (commented):
@butonic mostly looks good. A few minor comments. And totally agree, we should return multiple checksums in `ResourceInfo` and not rely on storing those in the opaque map.

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -384,14 +384,35 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
...
+			if e, ok := md.Opaque.Map["adler32"]; ok {
+				if checksums.Len() == 0 {
+					checksums.WriteString("<oc:checksum>ADLER32:")
+				} else {
+					checksums.WriteString(" ADLER32:")
+				}
+				checksums.WriteString(string(e.Value))
+			}
+		}
+		if checksums.Len() > 13 {
```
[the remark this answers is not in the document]
butonic: we don't want to convert the string buffer to a string. `checksums.Len() > 0` should be more readable though

pkg/storage/fs/ocis/upload.go
butonic: it would send the checksum for a PATCH request. without the chunking extension that is the complete file. with the chunking extension it would be the checksum of the chunk.

ishank011: But when `FinishUpload` is called, `binPath` would have the complete file, right? In that case, the last chunk request would need to have have the checksum of the whole file, otherwise the asserts would fail.

butonic: The `metadata["checksum"]` is set in `InitiateUpload()`, which will store it in the info file. So the checksum (if one was sent) is available in FinishUpload(). Even if it was not requested, we still need to store it in the metadata, which is why we have to split checking them from persisting them. The latter of which has to happen always.

ishank011 (approved):

phil-davis:
:rocket: a list of expected-failures were removed - great stuff.

PR #1401, opened 2021-01-15
Use the user in request for deciding the layout for non-home DAV requests

### LINKED ISSUE #1215: Dav files namespace should use the user in request for deciding the layout
https://github.com/cs3org/reva/blob/5831dc09a1c39b770765b184d9efa13da09aa0e0/internal/http/services/owncloud/ocdav/dav.go#L59-L66

For the incoming `/dav/files/userID` requests, we have different namespaces depending on whether the request is for the logged-in user's namespace or not. Since in the storage drivers, we specify the layout depending only on the logged in user, this works in the first condition. However, when a user wants to access another user's namespace when the storage provider depends on the logged in user's namespace, it fails.

For example, consider the following case. The owncloud fs uses a layout 
`{{substr 0 1 .Id.OpaqueId}}/{{.Id.OpaqueId}}` The user einstein sends a request `/dav/files/marie`, which should be allowed. However, based on the current way we apply the layout, there's no way in which this can be translated to `/m/marie/`

https://github.com/cs3org/reva/blob/5831dc09a1c39b770765b184d9efa13da09aa0e0/internal/http/services/owncloud/ocdav/ocdav.go#L195-L197

It can only use the details of the user in context. This needs to be changed to handle such requests. @butonic @PVince81

ishank011:
For the incoming `/dav/files/userID` requests, we have different namespaces depending on whether the request is for the logged-in user's namespace or not. Since in the storage drivers, we specify the layout depending only on the user whose resources are to be accessed, this fails when a user wants to access another user's namespace when the storage provider depends on the logged in user's namespace.

For example, consider the following case. The owncloud fs uses a layout `{{substr 0 1 .Username}}/{{.Username}}` and it's mounted at `/users`. The user einstein sends a request to access a resource shared with him, say `/dav/files/marie/abcd`, which should be allowed. However, based on the way we applied the layout, this can only be translated to `/users/e/marie/abcd`.

Fixes #1215

### DIFF
--- a/changelog/unreleased/fix-dav-namespace.md
+++ b/changelog/unreleased/fix-dav-namespace.md
@@ -1,0 +1,16 @@
+Bugfix: Use the user in request for deciding the layout for non-home DAV requests
+
+For the incoming /dav/files/userID requests, we have different namespaces
+depending on whether the request is for the logged-in user's namespace or not.
+Since in the storage drivers, we specify the layout depending only on the user
+whose resources are to be accessed, this fails when a user wants to access
+another user's namespace when the storage provider depends on the logged in
+user's namespace. This PR fixes that.
+
+For example, consider the following case. The owncloud fs uses a layout
+{{substr 0 1 .Id.OpaqueId}}/{{.Id.OpaqueId}}. The user einstein sends a request
+to access a resource shared with him, say /dav/files/marie/abcd, which should be
+allowed. However, based on the way we applied the layout, there's no way in
+which this can be translated to /m/marie/.
+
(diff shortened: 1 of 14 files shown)


labkode (approved):

labkode:
The code and behaviour looks good to me, this will also allow oc web to access shared resources from the tab without having to mount them. @ishank011 some acceptance test are failing in CI

ishank011:
@labkode thanks for the quick review! That is just one test which is pretty unstable. I've retriggered the CI

butonic (commented):
LGTM

PR #1407, opened 2021-01-19
[tests-only] Bump commit id for API tests

phil-davis:
1) bump commit id and adjust expected-failures for the regular changes in expected-failures.
2) Rename expected-failures files to .md - the core API tests now have the expected-failures list in markdown format
3) Copy expected-failures markdown files from the ones that are being applied in `owncloud/ocis`
4) Bump the core commit id to the latest and adjust the expected-failures md files for what passes and fails in `cs3org/reva` (mostly removing entries from exxpected-failures because more scenarios pass here in `cs3org/reva` - and they will start passing in `owncloud/ocis` when the latest `reva` is updated into `owncloud/ocis`)


IljaN (approved):

phil-davis:
@labkode @ishank011 there are big diffs here, but is "just" the re-formatting of the long expected-failures files. I have verified that the expected-failures files do get processed. CI passes here. So you do not have to review every line of the expected-failures files!

Ready for review and approval.

ishank011 (approved):

PR #1415, opened 2021-01-21
Indicate in EOS containers that TUS is not supported

ishank011:


ishank011:
@phil-davis can you take a look what's happening here? https://drone.cernbox.cern.ch/cs3org/reva/265/

ishank011:
Never mind. My bad

butonic (approved):
ok it, it defaults to TUS being used and it has to be explicitly disabled. I don't know about all the implications for clients that still rely on the capabilities endpoint. but this is what will make ocis-web use normal PUT

ishank011:
Thanks a lot @butonic! For the capabilities endpoint, I already added the code to control the response using config envs in https://github.com/cs3org/reva/pull/1331

PR #1430, opened 2021-02-02
Undeterministic upload tests

refs:
## Why?

As a developer creating/maintaining a storage driver I want to be able to validate the atomicity of all my storage driver operations.

* Test for: Start 2 uploads, pause the first one, let the second one finish first, resume the first one at some point in time. Both uploads should finish. Needs to result in 2 versions, last finished is the most recent version.
* Test for: Start 2 MKCOL requests with the same path, one needs to fail.

This PR only shows faulty functionality. When working on a fix on a red/green basis, remove the test build tag and fix the code on the storage layer.


IljaN (approved):
LGTM :+1: Great research!

ishank011 (approved):
Nice!

PR #1451, opened 2021-02-04
Render additional share with info in ocs sharing api

butonic:
Recipients can now be distinguished by their email, which is rendered as additional info in the ocs api for share and file owners as well as share recipients.

Fixes https://github.com/owncloud/ocis/issues/1190

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
@@ -105,9 +105,9 @@ func (h *Handler) userAsMatch(u *userpb.User) *conversions.MatchData {
 		Label: u.DisplayName,
 		Value: &conversions.MatchValueData{
 			ShareType: int(conversions.ShareTypeUser),
-			// TODO(jfd) find more robust userid
-			// username might be ok as it is unique at a given point in time
-			ShareWith: u.Username,
+			// api compatibility with oc10: always use the username
+			ShareWith:               u.Username,
+			ShareWithAdditionalInfo: u.Mail,
 		},
 	}
 }
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -27,7 +27,6 @@ import (
 	"path"
 	"strconv"
 	"strings"
-	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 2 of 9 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
```diff
@@ -105,9 +105,9 @@ func (h *Handler) userAsMatch(u *userpb.User) *conversions.MatchData {
 		Label: u.DisplayName,
 		Value: &conversions.MatchValueData{
 			ShareType: int(conversions.ShareTypeUser),
-			// TODO(jfd) find more robust userid
-			// username might be ok as it is unique at a given point in time
-			ShareWith: u.Username,
+			// api compatability with oc10: always use the username
+			ShareWith:               u.Username,
+			ShareWithAdditionalInfo: u.Mail,
```
kulmann, line 110: oc10 lets the admin decide if the additional info should be `Mail` or `UserID`. For full compatibility you could make the attribute configurable here as well 🤷

butonic: Maybe later. For now, I just want to be able to distinguish the users. And email is what users know from their email client. Furthermore, for guests the username equals the email. Since we want to unify sharing with _people_ I prefer the mail.

Whin searching recipients, the username is available as the `shareWith`, anyway. The Displayname is in the Label ... which only leaves email ...

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -51,19 +49,24 @@ import (
...
 type Handler struct {
-	gatewayAddr      string
-	publicURL        string
-	sharePrefix      string
-	displayNameCache *ttlmap.TTLMap
-	userNameCache    *ttlmap.TTLMap
+	gatewayAddr         string
+	publicURL           string
+	sharePrefix         string
+	userIdentifierCache *ttlmap.TTLMap
```
ishank011, line 55: We can use `github.com/ReneKroon/ttlcache` which I used for the etag cache.

ishank011 (approved):

internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
kulmann: But with this PR we'll lose the option to hide it. Because ownCloud Web renders the `ShareWithAdditionalInfo` as soon as it exists. Unnecessary info for small instances... 👀

butonic: I don't think it is unnecessary. I think we should align on emails because they allow using the federated discovery mechanism described by openid connect. phoenix could actually do that instead of using the ocs api ... or ... well ... in addition ...

but that is a product management decision.

PR #1452, opened 2021-02-04
Separate blobs from metadata in the ocis driver

aduffeck:
This pr makes the ocis driver store the actual file content separately from the metadata, allowing for using a different (potentially faster) storage for the metadata.

Common code between the `ocis` and `s3ng` drivers is extracted into a `decomposedfs` package which leaves them with just a thin layer for the different blobstore implementations.


pkg/storage/utils/decomposedfs/decomposed.go
```diff
@@ -0,0 +1,437 @@
...
+type Decomposedfs struct {
+	lu           *Lookup
+	tp           Tree
+	o            *options.Options
+	p            PermissionsChecker
+	chunkHandler *chunking.ChunkHandler
+}
+
+// NewDefault returns an instance with default components
+func NewDefault(m map[string]interface{}, bs tree.Blobstore) (storage.FS, error) {
```
refs, line 85: This would be slightly easier to read if values like `Lookup` and `Tree` are declared as package global variables, as seen [here](https://github.com/asim/go-micro/blob/d94936f6c97c5b1c40d179b55d9458419464b0e3/network/router/options.go#L77-L85) for example.

pkg/storage/utils/decomposedfs/decomposed.go
```diff
@@ -0,0 +1,437 @@
...
+	lu := &Lookup{}
+	p := node.NewPermissions(lu)
+
+	lu.Options = o
+
+	tp := tree.New(o.Root, o.TreeTimeAccounting, o.TreeSizeAccounting, lu, bs)
+	return New(o, lu, p, tp)
+}
+
+// New returns an implementation to of the storage.FS interface that talk to
```
refs, line 100: > // New returns an implementation of the storage.FS interface that talk to

refs (commented):
some minimal preliminary comments. It makes sense to add another abstraction between the filesystem and a blob storage. I will try it locally later today and continue reviewing behavior on the meantime. 🚀 awesome awesome work

butonic:
@phil-davis the tests fail with `LdapErr: DSID-0C090400, comment: AcceptSecurityContext error` ... do we need to update tests?

aduffeck:
@phil-davis @butonic nevermind, apparently it was an intermittent infra problem. It worked fine now after retriggering it a few times.

butonic (changes requested):
Treesize accounting will now use the 0 bytes size of the metatdata file in the nodes dir instead of using the blobs size.

aduffeck:
@butonic @phil-davis the ldap problem seems to be back unfortunately... (e.g. in https://drone.cernbox.cern.ch/cs3org/reva/710)

butonic (approved):
nice!

refs (dismissed):

pkg/storage/fs/ocis/blobstore/blobstore.go
```diff
@@ -0,0 +1,83 @@
...
+func (bs *Blobstore) Delete(key string) error {
+	err := os.Remove(bs.path(key))
+	if err != nil {
+		return errors.Wrapf(err, "could not delete blob '%s'", key)
+	}
+	return nil
+}
+
+func (bs *Blobstore) path(key string) string {
+	return path.Join(bs.root, key)
```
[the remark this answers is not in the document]
C0rby: Ah also if we do
```golang
filepath.Join(bs.root, filepath.Clean(filepath.Join("/", key)))
```
We make sure that we don't traverse outside of `bs.root`.
See: 
```golang
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	// Without cleaning
	fmt.Println(filepath.Join("/asdf/", "../foo"))
	fmt.Println(filepath.Join("/asdf/", "/../foo"))
	// Traversal without leading slash
	fmt.Println(filepath.Join("/asdf/", filepath.Clean("../foo")))
	// Secure
	fmt.Println(filepath.Join("/asdf/", filepath.Clean("/../foo")))
}
```
Output:
```
/foo
/foo
/foo
/asdf/foo
```

pkg/storage/utils/decomposedfs/decomposed.go
```diff
@@ -0,0 +1,487 @@
...
+	ri, err := n.AsResourceInfo(ctx, rp, []string{"treesize", "quota"})
+	if err != nil {
+		return 0, 0, err
+	}
+
+	quotaStr := node.QuotaUnknown
+	if ri.Opaque != nil && ri.Opaque.Map != nil && ri.Opaque.Map["quota"] != nil && ri.Opaque.Map["quota"].Decoder == "plain" {
+		quotaStr = string(ri.Opaque.Map["quota"].Value)
+	}
+	stat := syscall.Statfs_t{}
```
C0rby, line 157: `syscall.Statfs_t` is not available in windows systems.

pkg/storage/fs/ocis/blobstore/blobstore.go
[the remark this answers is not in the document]
aduffeck: Thanks, good catch. I applied your suggestion.

C0rby (approved):
Awesome! :+1:

labkode (changes requested):
One minor nit: 


LGTM, just one minor nit, probably you forgot to rename the file:

` pkg/storage/utils/decomposedfs/decomposed.go` => ` pkg/storage/utils/decomposedfs/decomposedfs.go`


Same for the test files.

aduffeck:
@labkode Thanks for your comment. I renamed the file accordingly.

butonic (approved):

PR #1462, opened 2021-02-08
hash share passwords

C0rby:
Passwords should never be stored in plaintext. The manager stored them base64 encoded which is very easily decoded to plaintext. Now the passwords are hashed using the bcrypt algorithm with a default cost of 11. The cost can be configured.


ishank011 (approved):

ishank011:
@C0rby good to merge?

C0rby:
Yes, from my PoV it's ready.

PR #1464, opened 2021-02-09
LDAP driver for the groupprovider service

ishank011:


pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	BindPassword    string     `mapstructure:"bind_password"`
+	Idp             string     `mapstructure:"idp"`
+	Schema          attributes `mapstructure:"schema"`
+}
+
+type attributes struct {
+	// DN is the distinguished name in ldap, e.g. `cn=admins,ou=groups,dc=example,dc=org`
+	DN string `mapstructure:"dn"`
+	// UID is an immutable group id, see https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-design-concepts
+	GID string `mapstructure:"gid"`
```
butonic, line 70: Line 69 UID vs Line 70 GID.

pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	// Displayname is the Human readable name, e.g. `Database Admins`
+	DisplayName string `mapstructure:"displayName"`
+	// GIDNumber is a numeric id that maps to a filesystem gid, eg. 654321
+	GIDNumber string `mapstructure:"gidNumber"`
+}
+
+// Default attributes (Active Directory)
+var ldapDefaults = attributes{
+	DN:          "dn",
+	GID:         "objectGUID", // you can fall back to objectguid or even samaccountname but you will run into trouble when user names change. You have been warned.
```
butonic, line 84: This GID should be an immutable, non reassignable, unique and persistent identifier for a group. The `ms-DS-ConsistencyGuid` we use for a users `UID` is made for that purpose. In general the defaults used in the ldap user and group provider should follow the same schema: either be Active Directory compatible or just use posixaccount and posixgroup. The latter schemas don't really have a unique attribute, which is why we recommend using a dedicated `ownclouduuid` attribute: https://github.com/owncloud/ocis/blob/c8668e8cb171860c70fec29e5ae945bca44f1fb7/deployments/examples/cs3_users_ocis/config/ldap/ldif/10_owncloud_schema.ldif

pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	Schema          attributes `mapstructure:"schema"`
+}
+
+type attributes struct {
+	// DN is the distinguished name in ldap, e.g. `cn=admins,ou=groups,dc=example,dc=org`
+	DN string `mapstructure:"dn"`
+	// UID is an immutable group id, see https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-design-concepts
+	GID string `mapstructure:"gid"`
+	// CN is the group name, typically `cn`, `gid` or `samaccountname`
+	CN string `mapstructure:"cn"`
```
butonic, line 72: cn is posixgroup, samaccountname is AD

pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	// Mail is the email address of a group
+	Mail string `mapstructure:"mail"`
```
butonic, line 74: mail works for both. posixgroup requires another structural ldap schema which then might have a mail property ... should be optional anyway. for AD groups can have a mail property

pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	// DN is the distinguished name in ldap, e.g. `cn=admins,ou=groups,dc=example,dc=org`
+	DN string `mapstructure:"dn"`
+	// GID is an immutable group id, see https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-design-concepts
+	GID string `mapstructure:"gid"`
+	// CN is the group name, typically `cn`, `gid` or `samaccountname`
+	CN string `mapstructure:"cn"`
+	// Mail is the email address of a group
+	Mail string `mapstructure:"mail"`
+	// Displayname is the Human readable name, e.g. `Database Admins`
+	DisplayName string `mapstructure:"displayName"`
```
butonic, line 76: fine, but not necessarily present in a posixgroup

pkg/group/manager/ldap/ldap.go
```diff
@@ -0,0 +1,396 @@
...
+	// GIDNumber is a numeric id that maps to a filesystem gid, eg. 654321
+	GIDNumber string `mapstructure:"gidNumber"`
+}
+
+// Default attributes (Active Directory)
+var ldapDefaults = attributes{
+	DN:          "dn",
+	GID:         "objectGUID", // you can fall back to samaccountname but you will run into trouble when group names change. You have been warned.
```
butonic, line 84: I recommend defaulting to `ms-DS-ConsistencyGuid`: https://dirteam.com/sander/2017/07/12/azure-ad-connect-objectguid-vs-ms-ds-consistencyguid-part-1/

butonic: or we default to a more posix like schema. for all ... but we need to decide ... which is ldap server we are going to support by default.

butonic (approved):

pkg/group/manager/ldap/ldap.go
ishank011: I got this from https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-design-concepts#using-ms-ds-consistencyguid-as-sourceanchor.
> Use ms-DS-ConsistencyGuid as the sourceAnchor attribute for User objects. ObjectGUID is used for other object types.

But I can change it back to `ms-DS-ConsistencyGuid`.

pkg/group/manager/ldap/ldap.go
ishank011: I got this from https://docs.microsoft.com/en-us/azure/active-directory/hybrid/plan-connect-design-concepts#using-ms-ds-consistencyguid-as-sourceanchor.
> Use ms-DS-ConsistencyGuid as the sourceAnchor attribute for User objects. ObjectGUID is used for other object types.

But I can change it back to `ms-DS-ConsistencyGuid`.

labkode (approved):

PR #1471, opened 2021-02-11
EOS GRPC interface

ffurano:


ffurano:
Why these things in the build fail, as they are not testing eosgrpc ? Maybe I have to wait some time and rebase ?

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
 	github.com/mitchellh/copystructure v1.0.0 // indirect
 	github.com/mitchellh/mapstructure v1.4.1
+	github.com/myesui/uuid v1.0.0
```
ishank011, line 35: Can you remove this? It isn't being used anywhere

go.sum
```diff
@@ -700,6 +700,8 @@ github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826/go.mod h1:TaXosZuwd
 github.com/monoculum/formam v0.0.0-20180901015400-4e68be1d79ba/go.mod h1:RKgILGEJq24YyJ2ban8EO0RUVSJlF1pGsEvoLEACr/Q=
 github.com/moul/http2curl v0.0.0-20170919181001-9ac6cf4d929b/go.mod h1:8UbvGypXm98wA/IqH45anm5Y2Z6ep6O31QGOAZ3H0fQ=
 github.com/mwitkow/go-conntrack v0.0.0-20161129095857-cc309e4a2223/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
+github.com/myesui/uuid v1.0.0 h1:xCBmH4l5KuvLYc5L7AS7SZg9/jKdIFubM7OVoLqaQUI=
```
ishank011, line 703: Remove this as well?

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -35,9 +34,9 @@ import (
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/eosclient"
 	erpc "github.com/cs3org/reva/pkg/eosclient/eosgrpc/eos_grpc"
+	ehttp "github.com/cs3org/reva/pkg/eosclient/eosgrpc/eos_http"
```
ishank011, line 37: The proto files need to be added

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -99,12 +90,11 @@ type Options struct {
 	// SecProtocol is the comma separated list of security protocols used by xrootd.
 	// For example: "sss, unix"
 	SecProtocol string
+
+	httpopts ehttp.Options
```
ishank011, line 94: Capitalize the first letter

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -181,9 +180,28 @@ func New(opt *Options) *Client {
...
+	return err2
+}
+
 // Common code to create and initialize a NSRequest
-func (c *Client) initNSRequest(uid, gid string) (*erpc.NSRequest, error) {
+func (c *Client) initNSRequest(ctx context.Context, uid, gid string) (*erpc.NSRequest, error) {
 	// Stuff filename, uid, gid into the MDRequest type
+
+	log := appctx.GetLogger(ctx)
+	log.Debug().Str("(uid,gid)", "("+uid+","+gid+")").Msg("New grpcNS req")
```
ishank011, line 203: `log.Debug().Str("uid", uid).Str("gid", gid).Msg("New grpcNS req")` could be a cleaner alternative

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -446,15 +483,17 @@ func (c *Client) GetFileInfoByInode(ctx context.Context, uid, gid string, inode
 		info.Inode = inode
 	}
 
+	log.Info().Str("func", "GetFileInfoByInode").Uint64("inode", inode).Msg("grpc response")
```
ishank011, line 486: Change log level to debug. Ditto for all the places where you're logging details about the response

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -875,71 +936,100 @@ func (c *Client) List(ctx context.Context, uid, gid, dpath string) ([]*eosclient
 	for {
 		rsp, err := resp.Recv()
 		if err != nil {
-			if err == io.EOF {
+			log.Debug().Err(err).Str("func", "List").Str("path", dpath).Str("got err from EOS", err.Error()).Msg("grpc response")
+			if err == io.EOF || i > 0 {
```
ishank011, line 940: Will the logic work if it's an empty directory?

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -875,71 +936,100 @@ func (c *Client) List(ctx context.Context, uid, gid, dpath string) ([]*eosclient
...
-// Write writes a file to the mgm
-func (c *Client) Write(ctx context.Context, uid, gid, path string, stream io.ReadCloser) error {
-	fd, err := ioutil.TempFile(c.opt.CacheDirectory, "eoswrite-")
-	if err != nil {
-		return err
-	}
-	defer fd.Close()
-	defer os.RemoveAll(fd.Name())
+	var localfile io.WriteCloser
+	localfile = nil
```
ishank011, line 983: This can be removed. It's nil by default

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -875,71 +936,100 @@ func (c *Client) List(ctx context.Context, uid, gid, dpath string) ([]*eosclient
...
+	// it's more difficult to return a sound error in the case of troubles
+	//	localfile, err := os.Create(localTarget)
+	//	if err != nil {
+	//		log.Error().Str("func", "Read").Str("path", path).Str("uid,gid", uid+","+gid).Str("err", err.Error()).Msg("")
+	//		return nil, errtypes.InternalError(fmt.Sprintf("can't open local cache file '%s'", localTarget))
+	//	}
 
-	// copy stream to local temp file
-	_, err = io.Copy(fd, stream)
+	err, bodystream := c.GetHttpCl().GETFile(ctx, "", uid, gid, path, localfile)
```
ishank011, line 993: Why do we need to pass a WriteCloser to a GET call?

ffurano:
Hi Ishank, thank for the review. I'll go through all these points in the next days

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: What do you mean here? Which proto file?

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: I can, yet that was meant to be a private member. Do you think I have to make it public?

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: Not sure it's cleaner, at least for human reading. I did that way because this logger component does not even guarantee the ordering of the items it logs. Actually I find it a pretty bad logging component also for other reasons, yet...

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: Ok. This will do in a comprehensive logging tour

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -126,14 +120,19 @@ type Client struct {
 	cl  erpc.EosClient
 }
 
+// GetHttpCl creates an http client for immediate usage, using the already instantiated resources
+func (c *Client) GetHttpCl() *ehttp.EosHttpClient {
```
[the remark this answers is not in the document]
ffurano: Why?

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: This is temporary code, until we understand what to do there

ishank011:
@ffurano can you rebase instead of merging? It's impossible to review

ffurano:
> @ffurano can you rebase instead of merging? It's impossible to review

Uhm... I only did rebase so far

ffurano:
> > @ffurano can you rebase instead of merging? It's impossible to review
> 
> Uhm... I only did rebase so far

Actually I have the impression that you have put into master some previous stage of my work. I am reviewing all the new conflicts, yet now it's really a mess

ffurano:
> @ffurano can you rebase instead of merging? It's impossible to review

Rebase done

labkode:
@ffurano I see a dependency to a private repo of yours:
```
github.com/ffurano/grpc-proto
```

labkode (changes requested):
Remove private repos

ffurano:
> @ffurano I see a dependency to a private repo of yours:
> 
> ```
> github.com/ffurano/grpc-proto
> ```
> 
> Could you replace it with an upstream version?

There's no such a thing by now...

ffurano:
> > @ffurano I see a dependency to a private repo of yours:
> > ```
> > github.com/ffurano/grpc-proto
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you replace it with an upstream version?
> 
> There's no such a thing by now...

Also because reva must compile with a particular version of protoc

ishank011 (commented):
Did a quick pass. Will do a detailed review again

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -117,6 +120,12 @@ func (opt *Options) init() {
 	if opt.CacheDirectory == "" {
 		opt.CacheDirectory = os.TempDir()
 	}
+
+	if opt.httpopts.Init() != nil {
+		os.Exit(255)
```
ishank011, line 125: Add a panic instead of exiting

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -126,35 +135,37 @@ type Client struct {
...
 	if err != nil {
-		log.Debug().Str("Error connecting to ", "'"+opt.GrpcURI+"' ").Str("err:", err.Error()).Msg("")
-		return nil, err
+		log.Warn().Str("Error connecting to ", "'"+opt.GrpcURI+"' ").Str("err", err.Error()).Msg("")
 	}
 
 	log.Debug().Str("Going to ping ", "'"+opt.GrpcURI+"' ").Msg("")
 	ecl := erpc.NewEosClient(conn)
-	// If we can't ping... treat it as an error... we will see if this has to be kept, for now it's practical
+	// If we can't ping... just print ugly warnings. In the case EOS is down, grpc will take care of
```
ishank011, line 155: "just print warnings"

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -181,9 +188,36 @@ func New(opt *Options) *Client {
...
 }
 
+// If the error is not nil, take that
+// If there is an error coming from EOS, erturn a descriptive error
+func (c *Client) getRespError(rsp *erpc.NSResponse, err error) error {
+	if err != nil {
+		return err
+	}
+
+	if rsp == nil {
```
ishank011, line 198: Combine these three conditions using ||

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -372,24 +426,24 @@ func (c *Client) getACLForPath(ctx context.Context, uid, gid, path string) (*acl
...
-	log.Debug().Str("Exec ", "'"+path+"' ").Str("resp:", fmt.Sprintf("%#v", resp)).Msg("")
+	log.Debug().Str("func", "GetACLForPath").Str("path", path).Str("resp:", fmt.Sprintf("%#v", resp)).Msg("grpc response")
 
 	if resp.Acl == nil {
 		return nil, errtypes.InternalError(fmt.Sprintf("nil acl for uid: '%s' path: '%s'", uid, path))
 	}
 
 	if resp.GetError() != nil {
-		log.Info().Str("uid", uid).Str("path", path).Int64("errcode", resp.GetError().Code).Str("errmsg", resp.GetError().Msg).Msg("grpc response")
+		log.Info().Str("func", "GetACLForPath").Str("uid", uid).Str("path", path).Int64("errcode", resp.GetError().Code).Str("errmsg", resp.GetError().Msg).Msg("EOS negative resp")
```
ishank011, line 446: Changes this to Error()

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,500 @@
...
+	// in the go x509 implementation (why did they do that?!?!?)
+	// of course /etc/grid-security/certificates is NOT in those defaults!
+	ClientCADirs  string
+	ClientCAFiles string
+}
+
+// We want just one instance of these options in the whole app, as we don't want to
+// instantiate more than once the http client internals. For example to have
+// http keepalive, pools, etc...
+var httpTransport *http.Transport
```
ishank011, line 72: Don't keep httpTransport as a global variable. Make a function return it

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,500 @@
...
+			log.Error().Str("func", "GETFile").Str("url", finalurl).Int64("timeout", tdiff).Int("ntries", ntries).Msg("")
+			return nil, errtypes.InternalError("Timeout with url" + finalurl)
+		}
+
+		// Execute the request. I don't like that there is no explicit timeout or buffer control on the input stream
+		log.Debug().Str("func", "GETFile").Msg("sending req")
+		resp, err := c.cl.Do(req)
+
+		// Let's support redirections... and if we retry we have to retry at the same FST, avoid going back to the MGM
+		if resp != nil && (resp.StatusCode == 307 || resp.StatusCode == 302) {
```
ishank011, line 299: Why are you handling the redirect? The Go HTTP client does that itself

ishank011:
In general, I think we should reduce a bit of logging. We run reva in prod with the log level set to debug, this will make debugging more difficult

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -34,15 +34,17 @@ import (
...
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/eosclient"
-	erpc "github.com/cs3org/reva/pkg/eosclient/eosgrpc/eos_grpc"
+
+	// erpc "github.com/cs3org/reva/pkg/eosclient/eosgrpc/eos_grpc"
+	ehttp "github.com/cs3org/reva/pkg/eosclient/eosgrpc/eos_http"
 	"github.com/cs3org/reva/pkg/errtypes"
+	"github.com/cs3org/reva/pkg/logger"
 	"github.com/cs3org/reva/pkg/storage/utils/acl"
+	erpc "github.com/ffurano/grpc-proto/protobuf"
```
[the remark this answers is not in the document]
ishank011: We can also point this to https://github.com/cern-eos/grpc-proto if you can add CI to that repo to generate go bindings

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ffurano: It does not support it for all the cases needed by EOS

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ffurano: How do I make sure it's just one instance in that case?

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -768,12 +768,17 @@ func (fs *eosfs) GetQuota(ctx context.Context) (uint64, uint64, error) {
...
+		return 0, 0, errors.Wrap(err, "eos: no uid in ctx")
+	}
+
 	rootUID, rootGID, err := fs.getRootUIDAndGID(ctx)
 	if err != nil {
 		return 0, 0, err
 	}
 
-	qi, err := fs.c.GetQuota(ctx, u.Username, rootUID, rootGID, fs.conf.QuotaNode)
+	qi, err := fs.c.GetQuota(ctx, uid, rootUID, rootGID, fs.conf.QuotaNode)
```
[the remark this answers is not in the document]
ffurano: Why break? I use the binary client normally to do benchmarks

pkg/eosclient/eosgrpc/eosgrpc.go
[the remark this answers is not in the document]
ffurano: I'd say that the former seems easier to manage to me

pkg/storage/utils/eosfs/eosfs.go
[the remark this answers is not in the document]
ishank011: Ah okay. I didn't know the quota command works with the UID as well

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ishank011: You can just declare it when you're initializing the HTTP client. I think that's needed only in the New method but I see you're doing that in GETFile and PUTFile as well. Why do we need to initialize it again?

ffurano: The idea was to be able to refresh it if anything goes bad with connection pools or CRLs, and that was the reason for the mutex, to be able to refresh it from a different thread (possibly in comm troubles). Too bad that golang makes a singleton so awkward

pkg/eosclient/eosgrpc/eosgrpc.go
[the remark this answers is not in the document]
ffurano: done

labkode:
`go mod tidy` to remove unused deps

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ishank011: httpTransport is just a config object which stays constant except at init. Keep it as a member of the Client struct and reuse it

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,501 @@
...
+			// io.Copy(ioutil.Discard, resp.Body)
+			// resp.Body.Close()
+
+			loc, err := resp.Location()
+			if err != nil {
+				log.Error().Str("func", "GETFile").Str("url", finalurl).Str("err", err.Error()).Msg("can't get a new location for a redirection")
+				return nil, err
+			}
+
+			c.cl = &http.Client{
```
ishank011, line 311: This will leak a lot of file descriptors. Please remove this and add the `MaxConnsPerHost` and `MaxIdleConnsPerHost` parameters at init

ishank011: Also, the timeout. Uploads/downloads can take a long time

ffurano: > This will leak a lot of file descriptors. Please remove this and add the `MaxConnsPerHost` and `MaxIdleConnsPerHost` parameters at init

It should not, as http clients are pooled/cached in the Transport. I'll add the parms though

ishank011: Then there's no reason to store the client since you're creating a new one every time. So I think you can remove cl from the struct and add transport instead of having it as a global variable

ffurano:
"httpTransport is just a config object which stays constant except at init. Keep it as a member of the Client struct and reuse it"

no, it's not. httpTransport is a complex object that caches connections. If I keep it in the Client class then it will leak. It has to be unique in the whole app

https://golang.org/pkg/net/http/#Transport

"Transports should be reused instead of created as needed."

ishank011:
> If I keep it in the Client class then it will leak. It has to be unique in the whole app

Don't we have just one instance of the client struct in the whole app?

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ffurano: no, it's not. httpTransport is a complex object that caches connections. If I keep it in the Client class then it will leak. It has to be unique in the whole app

https://golang.org/pkg/net/http/#Transport

"Transports should be reused instead of created as needed."

ffurano:
> > If I keep it in the Client class then it will leak. It has to be unique in the whole app
> 
> Don't we have just one instance of the client struct (not the HTTP client, the one defined in eosgrpc) in the whole app?

Uhm, nothing enforces that in the Client, and it seems pretty awkward to assume that there is just one instance.

The creator can create as many clients it needs, as it goes through the config file:
func New(opt *Options) *Client

ffurano:
If you have a better way to implement a singleton in golang please let me know... so far I have done more or less like this: https://medium.com/@ishagirdhar/singleton-pattern-in-golang-9f60d7fdab23

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	err2 := errtypes.InternalError("Err from EOS: " + rspdesc(rsp))
+	return err2
+}
+
+// From the basepath and the file path... build an url
+func (c *Client) buildFullURL(urlpath, uid, gid string) (string, error) {
+	s := c.opt.BaseURL
+	if len(urlpath) > 0 && urlpath[0] != '/' {
```
ishank011, line 227: Can you do something like https://github.com/cs3org/reva/blob/master/pkg/ocm/share/manager/json/json.go#L321-L326?

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,503 @@
...
+		s += "/"
+	}
+	s += urlpath
+
+	// I feel safer putting here a check, to prohibit malicious users to
+	// inject a false uid/gid into the url
+	// Who knows, maybe it's redundant? Better more than nothing.
+	p1 := strings.Index(urlpath, "eos.ruid")
```
ishank011, line 235: https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/webdavstorageprovider.go#L281

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,503 @@
...
+		// TODO: the error reporting of http.transport is insufficient
+		// must check manually at least the existence of the certfiles
+		// The point is that also the error reporting of the context that calls this function
+		// is weak
+		httpTransport = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				Certificates: []tls.Certificate{cert},
+			},
+			MaxIdleConns:        10,
+			MaxConnsPerHost:     8,
```
ishank011, line 130: Can you make these configurable?

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
```diff
@@ -0,0 +1,503 @@
...
+			log.Error().Str("func", "GETFile").Str("url", finalurl).Int64("timeout", tdiff).Int("ntries", ntries).Msg("")
+			return nil, errtypes.InternalError("Timeout with url" + finalurl)
+		}
+
+		// Execute the request. I don't like that there is no explicit timeout or buffer control on the input stream
+		log.Debug().Str("func", "GETFile").Msg("sending req")
+		resp, err := c.cl.Do(req)
+
+		// Let's support redirections... and if we retry we have to retry at the same FST, avoid going back to the MGM
+		if resp != nil && (resp.StatusCode == 307 || resp.StatusCode == 302) {
```
ishank011, line 302: `(resp.StatusCode == http. StatusFound || resp.StatusCode == http. StatusTemporaryRedirect)`

ishank011 (commented):
I'm not sure this should be a singleton in the first place. We'll have one instance of the eos driver to connect to each of the EOS instances and I don't see any advantage of sharing the transport among them. I'd prefer it to be a part of the Client struct. We also need to increase the defaults and make transport configurable.

ffurano:
> I'm not sure this should be a singleton in the first place. We'll have one instance of the eos driver to connect to each of the EOS instances and I don't see any advantage of sharing the transport among them. I'd prefer it to be a part of the Client struct. We also need to increase the defaults and make transport configurable.

Oh, I think you make a good point here about sharing the transport, I'll have a look if it an eos instance can keep the same transport to cache the conns

pkg/eosclient/eosgrpc/eos_http/eoshttp.go
ffurano: Yes, of course

ishank011:
@ffurano `GetHTTPCl` is called on every PUT and GET call, which creates an instance of `*ehttp.Client`. 
I think here we can store the `*ehttp.Client` 

https://github.com/cs3org/reva/blob/338598b970f7f6cf8c4b6c5e825e75593e304aef/pkg/eosclient/eosgrpc/eosgrpc.go#L197-L201

and `GetHTTPCl` can just return that client.

ffurano:
> @ffurano `GetHTTPCl` is called on every PUT and GET call, which creates an instance of `*ehttp.Client`.
> I think here we can store the `*ehttp.Client`
> 
> https://github.com/cs3org/reva/blob/338598b970f7f6cf8c4b6c5e825e75593e304aef/pkg/eosclient/eosgrpc/eosgrpc.go#L197-L201
> 
> and `GetHTTPCl` can just return that client.

Ah, and now the transport class is not a singleton anymore, there is one instance per EOS endpoint configured

ishank011 (dismissed):

ishank011 (approved):

PR #1477, opened 2021-02-15
eosfs: set quota

labkode:
Testing...


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -64,6 +64,9 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 	// TODO(labkode): if both commits are enabled they could be done concurrently.
 	if s.c.CommitShareToStorageGrant {
 		addGrantStatus, err := s.addGrant(ctx, req.ResourceInfo.Id, req.Grant.Grantee, req.Grant.Permissions.Permissions)
+		if err != nil {
```
ishank011, line 67: Let's wrap the original error instead of creating a new one

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -913,6 +925,21 @@ func (fs *eosfs) createUserDir(ctx context.Context, u *userpb.User, path string,
...
+
+	// set quota for user
+	quotaInfo := &eosclient.SetQuotaInfo{
+		Username:  u.Username,
+		MaxBytes:  fs.conf.DefaultQuotaBytes,
+		MaxFiles:  fs.conf.DefaultQuotaFiles,
+		QuotaNode: fs.conf.QuotaNode,
+	}
+
+	err = fs.c.SetQuota(ctx, uid, gid, quotaInfo)
```
ishank011, line 937: This will be called every time a reference is created, since that uses `createUserDir` as well. We can move this to `createNominalHome`.

labkode: That also looks wrong, we should be creating the home on every reference.

ishank011: No we don't create a home directory. This is just a util method to create a directory whose owner is the user. It's used by multiple methods.

labkode: @ishank011 take a look at last commit. Calling createUserDir will set xattrs on each reference that are already inherited from the shadow home. Just creating the directory and setting the ref is enough

ishank011: But would the user still be able to delete references? Given that they don't own those

ishank011: Looks good

ishank011:
@labkode looks good. You added a duplicate file. Once that is removed, we can merge this

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -960,9 +987,11 @@ func (fs *eosfs) CreateReference(ctx context.Context, p string, targetURI *url.U
 	if err != nil {
 		return nil
 	}
-	if err := fs.createUserDir(ctx, u, tmp, false); err != nil {
-		err = errors.Wrapf(err, "eos: error creating temporary ref file")
-		return err
+
+	err = fs.c.CreateDir(ctx, uid, gid, tmp)
```
ishank011, line 991: We're using the root uid and gid here. The user won't be able to delete references. We'll need to use user credentials here but since we're setting system xattrs, we'll need root creds for that

labkode: You're right, the reference needs to be deleted, but the user should not own any of these areas, all these folders should be own by root

ishank011 (approved):

PR #1486, opened 2021-02-18
[tests-only] Use cn for gid schema

ishank011:


ishank011:
@butonic the reason a majority of the tests fail is that they expect the shared resource path to lie inside the '/Shares' directory. And this is just from the createShare POST call. The recipient doesn't even accept it yet.
```
Examples:
| ocs_api_version | ocs_status_code |
| 1               | 100             |
│ file_target has unexpected value '/textfile0.txt'
│
| file_target doesn't have value '/Shares/textfile0.txt'
| Failed asserting that false is true.
```

What do we do about this?

PR #1506, opened 2021-03-01
Site Accounts service for API keys

Daniel-WWU-IT:
This update adds a new service to Reva that handles site accounts creation and management. Registered sites can be assigned an API key through a simple web interface which is also part of this service. This API key can then be used to identify a user and his/her associated (vendor or partner) site.

Furthermore, Mentix was extended to make use of this new service. This way, all sites now have a stable and unique site ID that not only avoids ID collisions but also introduces a new layer of security (i.e., sites can only be modified or removed using the correct API key).

_Note: The old registration APIs were removed in this PR; it is thus mandatory to update the web applications as well, and somehow allow the users to create an account._


butonic:
We wanted to add account managemen to the cs3 apis as well as ocis has an accounts service, which duplicated responsibilities and made us configure revas userprovider to talk to glauth which would talk to the ocis accounts service.

Instead, we decided to use whatever user management is in place and rely on reva to reaid users from there. We will move the ui and persistence from our accounts service to glauth to have an embeddable user management with an ldap api.

Users and groups can then be managed using existing tools.

For guest accounts this means they need to be provisioned in an ldap server,  or in whatever backend is configured for revas userprovider (and groupprovider). CERN is using a rest api and they have a process to onboard guests.

@Daniel-WWU-IT does this pr also add a new backend for the user and group providers? Or do you want to be able to manage users using the CS3 api. For the letter you should add a PR to the CS3 apis repo. Ping me if you would like to collaborate on this. We have existing protobuf spects to manage accounts and groups in the existing ocis accounts service.

I just want to make sure we are not duplicating efforts.

Daniel-WWU-IT:
@butonic This is something entirely different. It has nothing to do with end users or groups. These accounts are to manage the registered sites and generate API keys for them, which in return are used (or will be used) in the web apps. The service name could be changed to make this a bit more clear, I am open to better names here ;)

butonic:
Hm, I already feared I did not fully understand what this is about ;-)

Naming is hard. `sites` is also not quite right. `siteacc` is a good compromise I think.

Daniel-WWU-IT:
Renamed. Indeed a much better name!

labkode (approved):

PR #1509, opened 2021-03-02
Introduce Named Services

refs:
## What?

Currently, the way that services are configured is very binary: if a key is present in the config file, use its value, or else default to the gateway address. This is inflexible in terms of scaling (in the cloud?) and we would rather use names instead of IP addresses.

When communicating with another service the caller needs to know beforehand the address of the target, the existing implementation defaults such address to the gateway. Every service needs to explicitly know the endpoints it needs to communicate with. This is a proposal to move away from hardcoding service IP addresses and rely upon name resolution instead. It delegates the address lookup to, for the sake of this PR, a static in-memory service registry, which can be re-implemented in multiple forms, since there is only an interface to satisfy.

### Technical Details

Reva configuration has been extended to support a top-level registry key / value pair. Its value looks like this:

```toml
## global registry initialization
[registry.services]
"com.owncloud.authregistry" = [{nodes = [{ address = "localhost:19000" }]}]
```

### Note

A decision was made not to support backward compatibility with the current IP addresses mapping in order to isolate the changes to a single service and fail if misconfigured; backward compatibility can be easily added to provide support for existing functionality by simply failing at querying the registry and defaulting to the current defaults.

### Implementation

The glue comes in the form of a new interface introduced in Reva, the Registry interface, that looks like:

```go
type Registry interface {
	Add(Service) error
	GetService(string) ([]*Service, error)
}
```

a Service type (that will be moved to be an interface)

```go
type Service struct {
	Name  string 	`mapstructure:"name"`
	Nodes []Node 	`mapstructure:"nodes"`
}
```

and Nodes:

```go
type Node struct {
	ID string 					`mapstructure:"id"`
	Address string 				`mapstructure:"address"`
	Metadata map[string]string 	`mapstructure:"metadata"`
}
```

This set of interfaces and types would theoretically allow any consumer of Reva that wants to start its own service to inject its own registry. Registry injection happens now as a functional option on `runtime.go`:

```go
func WithRegistry(r registry.Registry) Option {
	return func(o *Options) {
		o.Registry = r
	}
}
```

### Current Registry Implementation breakdown

The current registry relies on an in-memory implementation of the `registry.Registry` interface. It will parse the static config from the toml file. This has a set of drawbacks in the way that services outside the process boundary will not know of any registered service, but since the current changes are contained to the gateway, these effects are not yet seen. A good approach would be to have a dedicated service registry service running so it can be queried upon (for more on this, see https://www.nginx.com/blog/service-discovery-in-a-microservices-architecture/).

One drawback of this approach, as previously mentioned, is that we're using a global registry as default, to prove this hypothesis and showcase how it would work on Reva and with library consumers! We do not encourage the use of globals :)

### Ideas and further improvements

- With this in mind we could sanitize the set of defaults on the gateway config.
- Config would be much simpler as we don't have to carry IP addresses around every time a node is restarted, or we can provide support for dynamic IP addresses; configuration would just be a matter of writing the service name.
- Client-side load balancing; by adding client awareness of service registries and a selection strategy.

### DIFF
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -28,6 +28,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/cs3org/reva/pkg/registry/memory"
+
+	"github.com/cs3org/reva/pkg/utils"
+
 	"contrib.go.opencensus.io/exporter/jaeger"
 	"github.com/cs3org/reva/cmd/revad/internal/grace"
 	"github.com/cs3org/reva/pkg/logger"
@@ -56,6 +60,21 @@ func RunWithOptions(mainConf map[string]interface{}, pidFile string, opts ...Opt
 	parseSharedConfOrDie(mainConf["shared"])
 	coreConf := parseCoreConfOrDie(mainConf["core"])
 
+	// TODO: one can pass the options from the config file to registry.New() and initialize a registry based upon config files.
+	if options.Registry != nil {
+		utils.GlobalRegistry = options.Registry
+	} else if _, ok := mainConf["registry"]; ok {
+		for _, services := range mainConf["registry"].(map[string]interface{}) {
+			for sName, nodes := range services.(map[string]interface{}) {
+				for _, instance := range nodes.([]interface{}) {
+					if err := utils.GlobalRegistry.Add(memory.NewService(sName, instance.(map[string]interface{})["nodes"].([]interface{}))); err != nil {
+						panic(err)
+					}
+				}
+			}
+		}
+	}
+
 	run(mainConf, coreConf, options.Logger, pidFile)
 }
 
@@ -71,6 +90,7 @@ func run(mainConf map[string]interface{}, coreConf *coreConf, logger *zerolog.Lo
 	host, _ := os.Hostname()
 	logger.Info().Msgf("host info: %s", host)
 
+	// initRegistry()
 	initTracing(coreConf, logger)
 	initCPUCount(coreConf, logger)
 
--- a/internal/grpc/services/gateway/gateway.go
+++ b/internal/grpc/services/gateway/gateway.go
@@ -82,7 +82,10 @@ func (c *config) init() {
 
 	// if services address are not specified we used the shared conf
 	// for the gatewaysvc to have dev setups very quickly.
-	c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
+
+	// we're commenting this line to showcase the fact that now we don't want to point to an ip address but rather
+	// resolve an ip address from a name.
+	// c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
 	c.StorageRegistryEndpoint = sharedconf.GetGatewaySVC(c.StorageRegistryEndpoint)
 	c.AppRegistryEndpoint = sharedconf.GetGatewaySVC(c.AppRegistryEndpoint)
 	c.PreferencesEndpoint = sharedconf.GetGatewaySVC(c.PreferencesEndpoint)
--- a/internal/grpc/services/gateway/gateway.go
+++ b/internal/grpc/services/gateway/gateway.go
@@ -85,7 +85,7 @@ func (c *config) init() {
 
 	// we're commenting this line to showcase the fact that now we don't want to point to an ip address but rather
 	// resolve an ip address from a name.
-	// c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
+	c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
 	c.StorageRegistryEndpoint = sharedconf.GetGatewaySVC(c.StorageRegistryEndpoint)
 	c.AppRegistryEndpoint = sharedconf.GetGatewaySVC(c.AppRegistryEndpoint)
 	c.PreferencesEndpoint = sharedconf.GetGatewaySVC(c.PreferencesEndpoint)
--- a/pkg/registry/memory/memory.go
+++ b/pkg/registry/memory/memory.go
@@ -70,10 +70,10 @@ func (r *Registry) GetService(name string) (registry.Service, error) {
 
 // New returns an implementation of the Registry interface.
 func New(m map[string]interface{}) registry.Registry {
-	//c, err := registry.ParseConfig(m)
-	//if err != nil {
+	// c, err := registry.ParseConfig(m)
+	// if err != nil {
 	//	return nil
-	//}
+	// }
 
 	return &Registry{
 		services: map[string]registry.Service{},
--- a/pkg/registry/memory/memory.go
+++ b/pkg/registry/memory/memory.go
@@ -30,7 +30,7 @@ type Registry struct {
 	// m protects async access to the services map.
 	sync.Mutex
 	// services map a service name with a set of nodes.
-	services map[string][]registry.Service
+	services map[string]registry.Service
 }
 
 // Add implements the Registry interface. If the service is already known in this registry it will only update the nodes.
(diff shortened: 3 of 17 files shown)


refs:
CI tests are failing because the config is not yet updated :)

butonic:
For an overview of the basic go-micro interfaces have a look at https://github.com/owncloud/ocis/issues/1749

ishank011:
@refs @butonic this looks really cool!

One suggestion would be to have one more abstraction for the drivers of a particular service in the code itself. That way, we'll have only one service corresponding to a specific driver. The code would be cleaner and the load-balancing part would be more evident.

So instead of the current mapping `map[string][]*Service`, which can take the form
```
authprovider -> [{"auth-basic", [addr1, addr2]}, {"auth-bearer", [addr3, addr4]}, {"auth-basic", [addr5]}]
```
we can have  `map[string]map[string]*Service`
```
authprovider -> {"basic" -> {"auth-basic", [addr1, addr2, addr5]}, "bearer" -> {"auth-bearer", [addr3, addr4]}}
```

Also, didn't we decide on trying out the micro interfaces directly? We're adding similar interfaces here, right?

ishank011:
Also, I'd be definitely in favour of a dedicated service registry, where independent services can register themselves (over GRPC?) Does micro have that?

refs:
> Also, didn't we decide on trying out the micro interfaces directly? We're adding similar interfaces here, right?

We thought this way we wouldn't need to bring go-micro as a dependency and instead use our own interface definitions, a less aggressive approach ^^

butonic:
The go micro interfaces cover service discovery as well as clients and services. It even covers transport, codec, broker and what not. But one thing at a time.
1. A service registry could remain static in reva. Ocis could inject a dynamic one that is backed by a go micro implementation, eg. [consul, kubernetes, nats, mdns ...](https://github.com/asim/go-micro/tree/master/plugins/registry) That is what this PR shows
2. The go-micro interfaces use metadata to distinguish services, eg. the authprovider. For the static config we can do what you propose in https://github.com/cs3org/reva/pull/1509#issuecomment-788996699
3. go micro clients can have a selector strategy that chooses the next node in the list of service nodes that was returned by the registry.

If this makes sense, then we can add this in small steps. I'd say after fixing more business logic. But this would be a great way forward!

pkg/registry/registry.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// GetService retrieves a Service and all of its nodes by Service name. It returns []*Service because we can have
+	// multiple versions of the same Service running alongside each others.
+	GetService(string) ([]Service, error)
+}
+
+// Service defines a service.
+type Service interface {
+	GetName() string
+	SetName(string)
```
butonic, line 34: Do we need `SetName`? It is never used. Furthermore I prefer having `Name()` and `Nodes()` as well as passing in a name and a node via Options. See https://golang.org/doc/effective_go#Getters

AFAICT the interface only looks like this because the `memory.Node` implementation needs to have public `Name` and `Nodes` fields in order to unmarshal them using mapstructure. Maybe separating the config struct from the in memory representation makes sense, since we need to use the service names as map keys and `service.Name` fields anyway?

cmd/revad/runtime/runtime.go
```diff
@@ -56,6 +59,24 @@ func RunWithOptions(mainConf map[string]interface{}, pidFile string, opts ...Opt
...
+		utils.GlobalRegistry = r
+
+		// TODO move New to run and dedicated initRegistry
+		// TODO what if it was not initialized? iterate over existing http and grpc services?
+		// [grpc] and [http] both have an `address` property and iterate the services as [grpc.service.<servicename>]
+		// but the address is the listen address
+		// services like the gateway have host:port config for services
+		// micro uses an Address   option as the host:port to bind to and
+		//            an Advertise option as the host:port to use in the registry, if it is not set it will use the Address
+
```
butonic, line 77:
```suggestion
```

pkg/registry/memory/service.go
```diff
@@ -0,0 +1,43 @@
...
+	Name  string `mapstructure:"name"`
+	Nodes []node `mapstructure:"nodes"`
+}
+
+func (s *service) GetName() string {
+	return s.Name
+}
+
+func (s *service) GetNodes() []registry.Node {
+	ret := make([]registry.Node, 0)
```
butonic, line 34:
```suggestion
	ret := make([]registry.Node, len(s.Nodes))
```

pkg/registry/memory/node.go
```diff
@@ -0,0 +1,40 @@
...
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package memory
+
+import "fmt"
+
+// node implements the INode interface.
+type node struct {
+	id       string            `mapstructure:"id"`
```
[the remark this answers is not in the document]
refs: no they can't, but ID should be from the registry, we shouldn't rely in clients setting id's

butonic:
@ishank011 I think we should move forward with this kind of registry. We currently do not need other aspects of go micro.

labkode:
@butonic yes, we should not involve go-micro related code for now.

refs:
Indeed :) Let's fix the broken parts in this PR and get it merged. I will tackle it after my current task.

refs:
If these changes are okay, we could merge this one so we get at least the code in and could start using it in the short-term future. I evaluated and considered not to change the current config files because it would be a mess to update all occurrences in the config where we rely on the current way of configuring.

I'm in favor of providing backward compatibility with the current state of the codebase, and follow up with a PR in which we tackle the `pool.go` file to include the changes that use the new registry functionality.

/cc @ishank011 @labkode @butonic

refs:
ping @ishank011 shall we merge this one?

ishank011:
@refs will review today

refs:
Thanks!

pkg/registry/memory/memory.go
```diff
@@ -0,0 +1,116 @@
...
+		ns := make([]node, 0)
+		nodes := append(r.services[svc.Name()][i].Nodes(), svc.Nodes()...)
+		for n := range nodes {
+			ns = append(ns, node{
+				id:       nodes[n].ID(),
+				address:  nodes[n].Address(),
+				metadata: nodes[n].Metadata(),
+			})
+		}
+		r.services[svc.Name()] = []registry.Service{
```
ishank011, line 59: Here, we're ensuring that `services` always has a one-one `serviceName -> Service` mapping.

ishank011 (commented):
Hi @refs. One generic question. We always ensure that for a given service name, we have only one value in the map. We just append the nodes of the parameter passed in the `Add` call. Why not have the map as `services map[string]registry.Service` then instead of a slice?

Also, can we incorporate this comment?

> So instead of the current mapping `map[string][]*Service`, which can take the form
> 
> ```
> authprovider -> [{"auth-basic", [addr1, addr2]}, {"auth-bearer", [addr3, addr4]}, {"auth-basic", [addr5]}]
> ```
> 
> we can have `map[string]map[string]*Service`
> 
> ```
> authprovider -> {"basic" -> {"auth-basic", [addr1, addr2, addr5]}, "bearer" -> {"auth-bearer", [addr3, addr4]}}
> ```
> 

In this implementation, if I get a `bearer` auth request, I would need to traverse all nodes and check the metadata to find the node corresponding to the driver, plus load balancing would be a bit more cumbersome to implement. If we had the above implementation, the simplest strategy would just be selecting a random element from a slice.

refs:
> Why not have the map as map[string]registry.Service then instead of a slice?

You're right, and this was at the core of my concerns for the memory implementation. My implementation seems overly complicated to understand. The original idea was to provide a slight grim of flexibility but at some point along the way things got confusing. I'll simplify it and incorporate the changes in your comment, the arguments in favor are better than my mindless design 👯  I'll ping you when I get them in, thanks for the pair of eyes 👀 !

refs:
@ishank011 changes are in, could you review again? 🙏

ishank011:
@refs we aren't using the config anywhere, right? It has all the necessary structs but we're declaring them again in the memory driver.

I have a few thoughts about refactoring this:
- Remove `registry.Config` entirely.
- Change the `registry.Service` and `registry.Node` interfaces to structs. The drivers can just import this pkg and directly return these instead of having duplicate definitions.
- The map should be of type `map[string]map[string]*Service` in the memory driver instead of Config.

refs:
> Remove registry.Config entirely.

We do use it, I just commented it out for this temporary memory implementation:

```go
func New(m map[string]interface{}) registry.Registry {
	// c, err := registry.ParseConfig(m)
	// if err != nil {
	//	return nil
	// }

	return &Registry{
		services: map[string]registry.Service{},
	}
}
```

> Change the registry.Service and registry.Node interfaces to structs. The drivers can just import this pkg and directly return these instead of having duplicate definitions.

The point of using interfaces is not only that internal reva packages can work with them and import them, it serves as an integration between external service registries (such as go-micro) and Reva's internal. They are basically a contract, notice how the memory implementation even tho it implements internal types, it still works with the registry interfaces.

A benefit we get from having config parsing at the same level as the registry is that we can literally parse config from files and use the config object independently of the implementation, a bit of duplication there ensures that config does not change much, or if it changes, changes are propagated to all drivers since they all use the came config struct.

@ishank011

ishank011:
@refs ah okay, makes sense. Thanks for the explanation!

internal/grpc/services/gateway/gateway.go
```diff
@@ -82,7 +82,10 @@ func (c *config) init() {
 
 	// if services address are not specified we used the shared conf
 	// for the gatewaysvc to have dev setups very quickly.
-	c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
+
+	// we're commenting this line to showcase the fact that now we don't want to point to an ip address but rather
+	// resolve an ip address from a name.
+	// c.AuthRegistryEndpoint = sharedconf.GetGatewaySVC(c.AuthRegistryEndpoint)
```
ishank011, line 88: Can you uncomment this? We would need this for now

ishank011 (approved):

PR #1521, opened 2021-03-08
clarify expected failures

butonic:
Some features, while covered by the ownCloud 10 acceptance tests, will not be implmented for now:
- blacklisted / ignored files, because ocis/reva don't need to blacklist `.htaccess` files
- `OC-LazyOps` support was [removed from the clients](https://github.com/owncloud/client/pull/8398). We are thinking about [a state machine for uploads to properly solve that scenario and also list the state of files in progress in the web ui](https://github.com/owncloud/ocis/issues/214).

The expected failures files now have a dedicated _Won't fix_ section for these items.

Also see:
https://github.com/owncloud/client/pull/8398
https://github.com/owncloud/ocis/issues/214


phil-davis:
@butonic see #1522 for the fix for the local API tests. You can cherry-pick into here or...

ishank011 (approved):

PR #1526, opened 2021-03-09
Fix webdav file versions endpoint bugs

wkloucek:
Etag and error code related bugs have been fixed for the webdav file versions endpoint and removed from the expected failures file.

### DIFF
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -28,6 +28,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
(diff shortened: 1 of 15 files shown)


internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -216,15 +218,30 @@ func wrap(sid string, oid string) string {
 	return base64.URLEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", sid, oid)))
 }
 
+func isValidUUID(u string) bool {
```
labkode, line 221: @wkloucek why you enforce to have an UUID?

wkloucek: @labkode The ideas was to do a sanity check. But while you're asking it is nonsense to do it only here and it is not really needed too. Therefore I removed it.

butonic (dismissed):
❤️

labkode (approved):

Issue #1484, opened 2021-02-17
Create share by e-mail after invitation accepted


mirekys:
## Scenario

1. Let's assume we have user **A** who starts the invitation workflow by sending an invite link to user **B**'s e-mail b@example.org he knows.
2. User **B** accepts the invite link by going through mesh directory and logging into his provider of choice
3. User **A** now wants to share something to **B**
4. At this moment, user **A** still knows only the user **B**'s e-mail (b@example.org)

## The problem

User **A** has currently no means of determining `grantee.opaqueid` used in the [invitation sharing workflow](https://codimd.web.cern.ch/s/Sy1ndLaxv#Creating-shares), he doesn't even know which `provider.idp` was used by user **B** to accept the invite.

1. User **A** should be able to create a share for the user **B** by using the e-mail adress used in the invitation workflow:
```
$ cmd/reva/reva -insecure ocm-share-create -grantee b@example.org /home/test
```

_or/and_

2. User **A** should be able to list remote users that accepted the invites he sent. Such accepted invite records should contain some `label`, that helps user **A** to recognize, who actually accepted the invite using which invite link (the label would be the used e-mail address b@example.com in this case). User **A** could then find all necessary sharing details about user **B** (provider & opaqueid) in this listing based on the label he already knows and any additional user info exchanged at the moment the invite was accepted.

labkode:
@mirekys can you start implementing option 2)? This is the foundation

mirekys:
Implemented in #1527

PR #1527, opened 2021-03-09
Add FindAcceptedUsers method to OCM Invite API

ishank011:
@mirekys this is the current workflow. You can search for the user and then send a create share request using the results obtained. These steps will be merged in the UI. Please let me know your thoughts or if there are other workflows we can implement.

Depends on https://github.com/cs3org/cs3apis/pull/111

```
>> ocm-find-accepted-users
+--------------------------------------+-------------+---------------------+-----------------+
| OPAQUEID                             | IDP         | MAIL                | DISPLAYNAME     |
+--------------------------------------+-------------+---------------------+-----------------+
| 932b4540-8d16-481e-8ef4-588e4b6b151c | example.org | richard@example.org | Richard Feynman |
+--------------------------------------+-------------+---------------------+-----------------+
>>
>> ocm-find-accepted-users rich
+--------------------------------------+-------------+---------------------+-----------------+
| OPAQUEID                             | IDP         | MAIL                | DISPLAYNAME     |
+--------------------------------------+-------------+---------------------+-----------------+
| 932b4540-8d16-481e-8ef4-588e4b6b151c | example.org | richard@example.org | Richard Feynman |
+--------------------------------------+-------------+---------------------+-----------------+
>>
>> ocm-find-accepted-users example
+--------------------------------------+-------------+---------------------+-----------------+
| OPAQUEID                             | IDP         | MAIL                | DISPLAYNAME     |
+--------------------------------------+-------------+---------------------+-----------------+
| 932b4540-8d16-481e-8ef4-588e4b6b151c | example.org | richard@example.org | Richard Feynman |
+--------------------------------------+-------------+---------------------+-----------------+
>>
>> ocm-share-create -grantee 932b4540-8d16-481e-8ef4-588e4b6b151c -idp example.org /home/a.txt
create share done
+--------------------------------------+-----------------+--------------------------------------+----------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------+-------------------+-------------+--------------------------------------+-------------------------------+-------------------------------+
| #                                    | OWNER.IDP       | OWNER.OPAQUEID                       | RESOURCEID                                                                             | PERMISSIONS                                                                                                     | TYPE              | GRANTEE.IDP | GRANTEE.OPAQUEID                     | CREATED                       | UPDATED                       |
+--------------------------------------+-----------------+--------------------------------------+----------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------+-------------------+-------------+--------------------------------------+-------------------------------+-------------------------------+
| 2ea37028-0da5-4657-8ddc-c7e2014fb24b | cernbox.cern.ch | 4c510ada-c86b-4815-8820-42cdf82c3d51 | storage_id:"123e4567-e89b-12d3-a456-426655440000" opaque_id:"fileid-einstein%2Fa.txt"  | permissions:<get_path:true initiate_file_download:true list_container:true list_file_versions:true stat:true >  | GRANTEE_TYPE_USER | example.org | 932b4540-8d16-481e-8ef4-588e4b6b151c | 2021-03-09 14:14:24 +0100 CET | 2021-03-09 14:14:24 +0100 CET |
+--------------------------------------+-----------------+--------------------------------------+----------------------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------+-------------------+-------------+--------------------------------------+-------------------------------+-------------------------------+
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -4,11 +4,7 @@ issues:
       text: "SA1019:"
       linters:
         - staticcheck
-    - path: pkg/publicshare/manager/json/json.go
-      text: "SA1019:"
-      linters:
-        - staticcheck
-    - path: internal/grpc/services/gateway/publicshareprovider.go
+    - path: pkg/utils/utils.go
       text: "SA1019:"
       linters:
         - staticcheck
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ contrib:
 
 # for manual building only
 deps:
-	cd /tmp && rm -rf golangci-lint &&  git clone --quiet -b 'v1.26.0' --single-branch --depth 1 https://github.com/golangci/golangci-lint &> /dev/null && cd golangci-lint/cmd/golangci-lint && go install
+	cd /tmp && rm -rf golangci-lint &&  git clone --quiet -b 'v1.38.0' --single-branch --depth 1 https://github.com/golangci/golangci-lint &> /dev/null && cd golangci-lint/cmd/golangci-lint && go install
 	cd /tmp && go get golang.org/x/tools/cmd/goimports
 
 build-ci: off
--- a/changelog/unreleased/ocm-find-accepted-users.md
+++ b/changelog/unreleased/ocm-find-accepted-users.md
@@ -1,0 +1,3 @@
+Enhancement: Add FindAcceptedUsers method to OCM Invite API
+
+https://github.com/cs3org/reva/pull/1527
--- a/cmd/reva/main.go
+++ b/cmd/reva/main.go
@@ -55,6 +55,7 @@ var (
 		rmCommand(),
 		moveCommand(),
 		mkdirCommand(),
+		ocmFindAcceptedUsersCommand(),
 		ocmShareCreateCommand(),
 		ocmShareListCommand(),
 		ocmShareRemoveCommand(),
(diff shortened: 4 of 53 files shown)


mirekys (commented):
Thanks @ishank011, I think that this should be enough for what is needed [here](https://github.com/cs3org/reva/issues/1484) (atleast for 2)). It wouldn't exactly solve 1) (e-mail used for sending the invitation mail may be completely different from e-mails in accepted users model), but that shouldn't be much of a problem, given that the share originator always knows person's name AND e-mail or idp he used for accepting the invite. Only issue I could think of would be if the originator knows only a person's name and invite e-mail address, there could easily be multiple persons with the same name, even on the same idp, in his accepted users and it may not be easy to determine who is who. Maybe adding some timestamp when the invite token was accepted would help with this?

ishank011:
@mirekys I don't think the first case can be solved. Email is not the only way in which tokens are forwarded. So searching using that would fail in those cases.

I don't see how timestamps would help. If multiple such users with the same name accept the same token, how would the original user know which one to pick?

mirekys:
I was thinking that the original user approx. knows when he sent the invite to the person he wants to share something with, timestamp could help him to eliminate people with the same name, that accepted his invites much earlier, and find the latest accepted users faster (results could be sorted by that timestamp, with most recent on top). It doesn't help much with the case of sending a bunch of invites to same-named people at the same time, but it's another factor that could help with searching through accepted invites.

It's just a suggestion, I'm otherwise ok to go with what is already implemented in the PR 👍

labkode (approved):

PR #1532, opened 2021-03-10
[tests-only] Place temporary test directories in reva/tmp

aduffeck:
System /tmp directories are often tmpfs mounts which do not support user
extended attributes.


labkode:
@aduffeck where the tmp directories are getting created with this change? Inside the Reva git repo?

aduffeck:
@labkode yes, in `<gitroot>/tmp` which is part of the `.gitignore` file now.

labkode (approved):

PR #1541, opened 2021-03-12
Fix recycle to different location

aduffeck:
Restoring to a different location than the original one is specified in the CS3 APIs as the `restore_path` field of the `RestoreRecycleItemRequest` resource but wasn't implemented in the storage provider until now.

### DIFF
--- a/pkg/storage/fs/owncloud/owncloud.go
+++ b/pkg/storage/fs/owncloud/owncloud.go
@@ -2140,7 +2140,7 @@ func (fs *ocfs) ListRecycle(ctx context.Context) ([]*provider.RecycleItem, error
 	return items, nil
 }
 
-func (fs *ocfs) RestoreRecycleItem(ctx context.Context, key string) error {
+func (fs *ocfs) RestoreRecycleItem(ctx context.Context, key, restorePath string) error {
 	// TODO check permission? on what? user must be the owner?
 	log := appctx.GetLogger(ctx)
 	rp, err := fs.getRecyclePath(ctx)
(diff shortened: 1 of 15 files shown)


aduffeck:
@labkode @ishank011 I implemented the fix for the `localfs`, `owncloud`, `ocis` and `s3ng` drivers. Would it be possible that - should you agree to the change - one of you does it for `eosfs`?

**Note**: Integration tests are currently still failing because of https://github.com/cs3org/reva/pull/1576

ishank011:
@aduffeck sure, I'll take it up this week.

aduffeck:
@ishank011 awesome, thank you :+1:

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -2155,16 +2155,17 @@ func (fs *ocfs) RestoreRecycleItem(ctx context.Context, key string) error {
...
 
-	origin := "/"
-	if v, err := xattr.Get(src, trashOriginPrefix); err != nil {
-		log.Error().Err(err).Str("key", key).Str("path", src).Msg("could not read origin")
-	} else {
-		origin = filepath.Clean(string(v))
+	if restorePath == "" {
+		v, err := xattr.Get(src, trashOriginPrefix)
+		if err != nil {
+			log.Error().Err(err).Str("key", key).Str("path", src).Msg("could not read origin")
```
ishank011, line 2161: Shouldn't we return here? Because if `err != nil`, `restorePath` will be the base of src, restoring to which might create conflicts

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -63,7 +63,7 @@ type Tree interface {
 	// CreateReference(ctx context.Context, node *node.Node, targetURI *url.URL) error
 	Move(ctx context.Context, oldNode *node.Node, newNode *node.Node) (err error)
 	Delete(ctx context.Context, node *node.Node) (err error)
-	RestoreRecycleItemFunc(ctx context.Context, key string) (*node.Node, func() error, error)
+	RestoreRecycleItemFunc(ctx context.Context, key, targetPath string) (*node.Node, func() error, error)
```
ishank011, line 66: Can you change it to `restorePath` to maintain consistency?

pkg/storage/fs/owncloud/owncloud.go
aduffeck: It also restored to `/` before, I didn't want to change behaviour like that with this PR.
Having said that I think the origin xattr will only be missing in very bad exceptions anyway. In those cases it might be desirable to have the files restored to `/` (if possible), no?

pkg/storage/utils/decomposedfs/decomposedfs.go
aduffeck: @ishank011 :+1:, done

ishank011 (approved):

Issue #1502, opened 2021-02-26
Mesh Directory UI frontend redesign


mirekys:
Replace the temporary mesh directory web UI with a more complete-looking web with a proper project branding.

PR #1548, opened 2021-03-15
CI: bump EOS Dockerfile go to 1.16

mirekys:
Bump go in EOS docker CI step to support embed, which is needed since #1502


mirekys:
Duplicity of #1547

PR #1552, opened 2021-03-15
Reduce expected trashbin failures.

C0rby:
This PR fixes stuff in the trash-bin code and thus reduces the expected failures for the trash-bin tests.


refs (approved):

refs:
🙏 thank you for digging into this!

ishank011 (approved):

PR #1505, opened 2021-03-01
[meshdirectory] New UI frontend with project branding

mirekys:
Replaces the temporary version of web frontend of the mesh directory http service with
a new redesigned & branded one.

Because the new frontend is a more complex Vue SPA that contains image,
css and other assets, it is now served from a binary package distribution that was generated using the
[github.com/rakyll/statik](https://github.com/rakyll/statik) package. The `http.services.meshdirectory.static`
config option will be obsoleted by this change.

@labkode what should be the best place to put the Vue frontend sources into?
Is it ok to create e.g. `meshdirectory-web` repo under @sciencemesh or @cs3org for it?

Screenshot_2021-02-26 Mesh Directory ScienceMesh

Fixes https://github.com/cs3org/reva/issues/1502

PR #1556, opened 2021-03-16
[tests-only] Fix linter issues

aduffeck:


ishank011 (approved):

wkloucek:
@aduffeck @ishank011 what was the reason that an update to Go 1.16 was needed?

ishank011:
@wkloucek We wanted to embed a binary using the go1.16 package https://github.com/cs3org/reva/pull/1505

wkloucek:
Ah I see, its introduced by importing https://github.com/sciencemesh/meshdirectory-web/blob/main/meshdirectory_web.go

PR #1557, opened 2021-03-16
enforce quota

wkloucek:


wkloucek:
will not become green before https://github.com/cs3org/reva/pull/1570 is merged because of https://github.com/cs3org/cs3apis/pull/116
(see https://github.com/cs3org/cs3apis/pull/115#issuecomment-805730529)

ishank011 (approved):

PR #1559, opened 2021-03-16
[tests-only] Bump commit id for tests

phil-davis:
and adjust expected-failures

PR #1567, opened 2021-03-18
[tests-only] Implement .drone.env

phil-davis:
to store the CORE_BRANCH and CORE_COMMITID for tests, the same way as is done in https://github.com/owncloud/ocis/blob/master/.drone.env

The disadvantage of having the commit-id embedded in `.drone.star` is that a `.drone.star` in the PR branch gets used from the PR branch and if there are other, later, changes to `.drone.star` in `master` they do not take effect in drone CI. This can lead to confusion about exactly what pipelines ran with what settings in CI.

Having `.drone.env` means that `.drone.star` can be much less modified.

Note: I had to change to `owncloudci/alpine:latest` (like in `owncloud/ocis`) for `cloneOc10TestReposStep()` because it used the `source` command, and `owncloudci/php:7.4` complained that `source` did not exist!

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,0 +1,3 @@
+# The test runner source for API tests
+CORE_COMMITID=60444e34b52b7c22ea817375947c9b60e42cd7da
+CORE_BRANCH=master
--- a/.drone.env
+++ b/.drone.env
@@ -1,3 +1,3 @@
 # The test runner source for API tests
-CORE_COMMITID=60444e34b52b7c22ea817375947c9b60e42cd7da
+CORE_COMMITID=698a34c25001aa5b640cd3bbb24fdc376b110da2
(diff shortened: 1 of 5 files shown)


phil-davis:
https://drone.cernbox.cern.ch/cs3org/reva/960/5/6
```
+ git checkout $CORE_COMMITID
Note: switching to '60444e34b52b7c22ea817375947c9b60e42cd7da'.
```

Looks good - it switches to the correct commit id that is specified in `.drone.env`

phil-davis:
This needs #1559 merged first. Then I will rebase...

phil-davis:
@ishank011 IMO CI will pass now. Please review.

ishank011 (approved):

PR #1573, opened 2021-03-19
Use UidNumber and GidNumber fields in User objects

sudo-sturbia:


pkg/cbox/user/rest/rest.go
```diff
@@ -507,12 +497,8 @@ func (m *manager) IsInGroup(ctx context.Context, uid *userpb.UserId, group strin
...
-		if uidObj, ok := u.Opaque.Map["uid"]; ok {
-			if uidObj.Decoder == "plain" {
-				return string(uidObj.Value), nil
-			}
-		}
+	if u.UidNumber == 0 {
+		return "", errors.New("rest: could not retrieve UID from user")
 	}
-	return "", errors.New("rest: could not retrieve UID from user")
+	return fmt.Sprintf("%v", u.UidNumber), nil
```
sudo-sturbia, line 503: There are several parts of code similar to this, where the Opaque map was checked before returning a value. I chose to check if UidNumber and GidNumber are zeros. This only works if uid and gid can't be zeros, so I'd appreciate if someone can verify this.

ishank011: The (0, 0) values are reserved for root users, so we can assume that none of the ordinary users would have these. That being said, in cases where we're unable to retrieve the UID or GID, these might be treated as privileged users and granted access to resources they're not supposed to access. Can you find such instances in eosfs and add checks? If a user has these values, raise an error.

pkg/auth/manager/ldap/ldap.go
```diff
@@ -182,7 +183,14 @@ func (am *mgr) Authenticate(ctx context.Context, clientID, clientSecret string)
 	if getGroupsResp.Status.Code != rpc.Code_CODE_OK {
 		return nil, errors.Wrap(err, "ldap: grpc getting user groups failed")
 	}
-
+	gidNumber, err := strconv.ParseInt(sr.Entries[0].GetEqualFoldAttributeValue(am.c.Schema.GIDNumber), 10, 0)
```
ishank011, line 186: Bit size 0? Should be 64

pkg/auth/manager/ldap/ldap.go
```diff
@@ -191,17 +199,10 @@ func (am *mgr) Authenticate(ctx context.Context, clientID, clientSecret string)
 		Groups:      getGroupsResp.Groups,
 		Mail:        sr.Entries[0].GetEqualFoldAttributeValue(am.c.Schema.Mail),
 		DisplayName: sr.Entries[0].GetEqualFoldAttributeValue(am.c.Schema.DisplayName),
+		UidNumber:   uidNumber,
+		GidNumber:   gidNumber,
 		Opaque: &types.Opaque{
```
ishank011, line 204: Just don't initialize opaque here. It should be nil. Ditto in all the instances.

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1457,23 +1457,13 @@ func getResourceType(isDir bool) provider.ResourceType {
...
-			if gidObj.Decoder == "plain" {
-				gid = string(gidObj.Value)
-			}
-		}
+	if u.UidNumber == 0 {
+		return "", "", errors.New("eos: uid missing for user")
 	}
-	if uid == "" || gid == "" {
-		return "", "", errors.New("eos: uid or gid missing for user")
+	if u.GidNumber == 0 {
```
ishank011, line 1463: We need to make sure that UID = 0 or GID = 0 in user requests is blocked eosfs

pkg/auth/manager/ldap/ldap.go
sudo-sturbia: Done

pkg/auth/manager/ldap/ldap.go
sudo-sturbia: Done

pkg/cbox/user/rest/rest.go
sudo-sturbia: It seems that functions in `eosfs` use `getUser` function for any user related operations, so i updated it to check for uid and gid in fabd746. Please let me know if that's what you had in mind.

pkg/storage/utils/eosfs/eosfs.go
sudo-sturbia: I'm not sure i understand, if you mean that the statement should be `if u.UidNumber == 0 || u.GidNumber == 0`, then you will notice that i separated it into two if statements to provide a more specific error message.

ishank011:
@sudo-sturbia majorly looks good, nice work! I'll test it since you don't have access to an EOS instance so there might be a delay in merging. We're working on getting EOS tests integrated in the CI so we should have that soon

Meanwhile, please rebase and fix the failing integration tests.

sudo-sturbia:
@ishank011 Hi, I thought I'd let you know the latest updates: I looked at the integration tests and was able to find the errors, the changes are mainly in 999730a and e685753. As for the old commits, struct tags for `UIDNumber` and `GIDNumber` in `pkg/auth/manager/json` were updated to match `User` struct. Currently all tests pass, except for `LGTM: JavaScript`, which is odd as i haven't updated any JavaScript, so I'd appreciate it if you let me know how to recreate it locally so i can maybe find the problem.

ishank011 (approved):
@sudo-sturbia LGTM! Good work on finding the issues with the integration tests. The LGTM tool is not that mature yet, so we do see intermittent build failures sometimes; we can ignore that. I'll test the changes tomorrow and merge.

sudo-sturbia:
> @sudo-sturbia LGTM! Good work on finding the issues with the integration tests. The LGTM tool is not that mature yet, so we do see intermittent build failures sometimes; we can ignore that. I'll test the changes tomorrow and merge.

That sounds great. Thank you!

ishank011 (approved):

PR #1574, opened 2021-03-19
Modularized api token management in GRAPPA drivers

### LINKED ISSUE #1562: Modularize API token management in GRAPPA drivers
https://github.com/cs3org/reva/tree/master/pkg/cbox/group/rest and https://github.com/cs3org/reva/tree/master/pkg/cbox/user/rest have a lot of overlapping code which can be reused.

jimil749:
Fixes #1562 

This PR moves duplicated api token management methods:
- `renewAPIToken`
- `getAPIToken`
- `SendAPIRequest`

into a seperate `utils` package to prevent duplication. The above methods are used by `user/rest` and `group/rest` packages. 

PS: I am not 100% sure of the approach used here, so it'd be great if someone could suggest a better way to implement this. I'd be happy to work on the changes.

### DIFF
--- a/pkg/cbox/group/rest/rest.go
+++ b/pkg/cbox/group/rest/rest.go
@@ -125,10 +125,21 @@ func New(m map[string]interface{}) (group.Manager, error) {
 	}, nil
 }
 
+func (m *manager) newAPITokenManager() *utils.APITokenManager {
+	return &utils.APITokenManager{
+		TargetAPI:         m.conf.TargetAPI,
+		OIDCTokenEndpoint: m.conf.OIDCTokenEndpoint,
+		ClientID:          m.conf.ClientID,
+		ClientSecret:      m.conf.ClientSecret,
+		Client:            m.client,
+	}
+}
+
 func (m *manager) getGroupByParam(ctx context.Context, param, val string) (map[string]interface{}, error) {
 	url := fmt.Sprintf("%s/Group?filter=%s:%s&field=groupIdentifier&field=displayName&field=gid",
 		m.conf.APIBaseURL, param, val)
-	responseData, err := utils.SendAPIRequest(ctx, url, false, m.client, m.conf.TargetAPI, m.conf.OIDCTokenEndpoint, m.conf.ClientID, m.conf.ClientSecret)
+	apiTokenManager := m.newAPITokenManager()
+	responseData, err := apiTokenManager.SendAPIGetRequest(ctx, url, false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


ishank011:
Let's try to remove any references to token generation in the interface methods. We can have a struct `APITokenManager`, which can be instantiated by the rest packages with all the necessary config. It can have one exported method `SendAPIGetRequest` which takes a URL and performs a GET to that.

jimil749:
> Let's try to remove any references to token generation in the interface methods. We can have a struct `APITokenManager`, which can be instantiated by the rest packages with all the necessary config. It can have one exported method `SendAPIGetRequest` which takes a URL and performs a GET to that.

@ishank011 can you explain a bit about the "instantiated by the rest packages" part?

Also, are you implying that the interface methods in user and group rest packages should call `SendAPIGetRequest` providing just the URL (instead of all the variables that I'm passing now)?

jimil749:
Ah okay, so I should create a new struct in the `utils` package, namely `APITokenManager`, which would be initialized in the `user` and `group` rest packages and then call the `SendAPIGetRequest` method. Yes, that seems to make sense! :sweat_smile:

pkg/cbox/utils/tokenmanagement.go
```diff
@@ -0,0 +1,150 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+// APITokenManager stores config related to api management
+type APITokenManager struct {
```
ishank011, line 34: Use a struct `OIDCToken` as done previously, which should store the token, expiration and a mutex and the rest should go in a struct called `config`

pkg/cbox/group/rest/rest.go
```diff
@@ -135,116 +125,21 @@ func New(m map[string]interface{}) (group.Manager, error) {
...
-	}
-
-	return responseData, nil
 }
 
 func (m *manager) getGroupByParam(ctx context.Context, param, val string) (map[string]interface{}, error) {
 	url := fmt.Sprintf("%s/Group?filter=%s:%s&field=groupIdentifier&field=displayName&field=gid",
 		m.conf.APIBaseURL, param, val)
-	responseData, err := m.sendAPIRequest(ctx, url, false)
+	apiTokenManager := m.newAPITokenManager()
```
ishank011, line 141: apiTokenManager should be a member of manager and the same object should be reused instead of creating a new one on every request

pkg/cbox/utils/tokenmanagement.go
jimil749: Will do! :+1:

jimil749:
@ishank011 Can you take a look?

ishank011 (approved):
LGTM!

pkg/cbox/utils/tokenmanagement.go
```diff
@@ -0,0 +1,177 @@
...
+		a.oidcToken.apiToken = token
+		a.oidcToken.tokenExpirationTime = expiration
+	}
+	return nil
+}
+
+func (a *APITokenManager) getAPIToken(ctx context.Context) (string, time.Time, error) {
+
+	params := url.Values{
+		"grant_types": {"client_credentials"},
```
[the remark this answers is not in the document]
jimil749: Ah! My bad! I'll send a new pr with the change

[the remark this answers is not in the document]
ishank011: No worries, I fixed it in my PR

PR #1578, opened 2021-03-23
Datatx pull model createtransfer

redblom:
Implement `create transfer` for datatx pull model: a datatx type share is created
See datatx https://github.com/cs3org/reva/pull/1242

### DIFF
--- a/cmd/reva/transfer-create.go
+++ b/cmd/reva/transfer-create.go
@@ -20,23 +20,38 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"io"
+	"strings"
 
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	tx "github.com/cs3org/go-cs3apis/cs3/tx/v1beta1"
 )
 
 func transferCreateCommand() *command {
 	cmd := newCommand("transfer-create")
-	cmd.Description = func() string { return "create transfer between 2 remotes" }
-	cmd.Usage = func() string { return "Usage: transfer-create [-flags]" }
+	cmd.Description = func() string { return "create transfer between 2 sites" }
+	cmd.Usage = func() string { return "Usage: transfer-create [-flags] <path>" }
 	grantee := cmd.String("grantee", "", "the grantee, receiver of the transfer")
+	granteeType := cmd.String("granteeType", "user", "the grantee type, one of: user, group")
+	idp := cmd.String("idp", "", "the idp of the grantee, default to same idp as the user triggering the action")
 
 	cmd.Action = func(w ...io.Writer) error {
-		// validate flags
+		if cmd.NArg() < 1 {
+			return errors.New("Invalid arguments: " + cmd.Usage())
+		}
+
 		if *grantee == "" {
 			return errors.New("Grantee cannot be empty: use -grantee flag\n" + cmd.Usage())
 		}
+		if *idp == "" {
+			return errors.New("Idp cannot be empty: use -idp flag\n" + cmd.Usage())
+		}
+
+		// the resource to transfer; the path
+		fn := cmd.Args()[0]
 
 		ctx := getAuthContext()
 		client, err := getClient()
@@ -44,8 +59,30 @@ func transferCreateCommand() *command {
 			return err
 		}
 
-		transferRequest := &tx.CreateTransferRequest{}
+		gt := provider.GranteeType_GRANTEE_TYPE_USER
+		if strings.ToLower(*granteeType) == "group" {
+			gt = provider.GranteeType_GRANTEE_TYPE_GROUP
+		}
 
+		transferRequest := &tx.CreateTransferRequest{
+			Ref: &provider.Reference{
+				Spec: &provider.Reference_Path{
+					Path: fn,
+				},
+			},
+			Grantee: &provider.Grantee{
+				Type: gt,
+				Id: &provider.Grantee_UserId{
+					UserId: &userpb.UserId{
+						Idp:      *idp,
+						OpaqueId: *grantee,
+					},
+				},
+			},
+		}
+
+		fmt.Println("transfer-create:")
+		fmt.Println("------------------------------------------------------------------------")
 		transferResponse, err := client.CreateTransfer(ctx, transferRequest)
 		if err != nil {
 			return err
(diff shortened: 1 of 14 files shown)


cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,174 @@
...
+		})
+		if err != nil {
+			return err
+		}
+		if acceptedUserRes.Status.Code != rpc.Code_CODE_OK {
+			return formatError(acceptedUserRes.Status)
+		}
+
+		// verify resource stats
+		hRes, err := client.GetHome(ctx, &provider.GetHomeRequest{})
```
ishank011, line 83: User should specify the full path. We don't need this

cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,174 @@
...
+			return formatError(statRes.Status)
+		}
+
+		providerInfoResp, err := client.GetInfoByDomain(ctx, &ocmprovider.GetInfoByDomainRequest{
+			Domain: *idp,
+		})
+		if err != nil {
+			return err
+		}
+		permissions := conversions.PermissionWrite
```
ishank011, line 110: Add all permissions. The resource will lie in the recipient's namespace so they'll have all permissions

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -144,7 +149,31 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
 
-	share, err := s.sm.Share(ctx, req.ResourceId, req.Grant, name, req.RecipientMeshProvider, permissions, nil, "", ocm.Share_SHARE_TYPE_REGULAR)
+	// discover share type
+	sharetype := ocm.Share_SHARE_TYPE_REGULAR
+	protocol, ok := req.Opaque.Map["protocol"]
+	if ok {
+		var sp *shareProtocol
+		switch protocol.Decoder {
+		case "json":
+			err := json.Unmarshal(protocol.Value, &sp)
```
ishank011, line 159: You're not using sp anywhere. Why are you even reading it?

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -144,7 +149,31 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
 		}, nil
 	}
 
-	share, err := s.sm.Share(ctx, req.ResourceId, req.Grant, name, req.RecipientMeshProvider, permissions, nil, "", ocm.Share_SHARE_TYPE_REGULAR)
+	// discover share type
+	sharetype := ocm.Share_SHARE_TYPE_REGULAR
```
ishank011, line 153: This hack is wrong. The share type should be specified in all locations where `CreateOCMShare` is being called.

ishank011 (changes requested):

cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,174 @@
...
+		resourcePermissions := &provider.ResourcePermissions{
+			InitiateFileDownload: true,
+		}
+		datatxProtocol, err := json.Marshal(
+			map[string]interface{}{
+				"name": "datatx",
+				"options": map[string]string{
+					"desired-protocol": "webdav",
```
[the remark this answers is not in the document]
redblom: > 
> 
> Rename it to protocol

@ishank011 : Do you want that done too at: https://github.com/redblom/reva/blob/768129ee461a67d73a600e2d50ca573f8f29e254/pkg/ocm/share/manager/json/json.go#L280 ?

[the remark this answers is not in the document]
ishank011: Yes, please

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
redblom: The protocol value is read into it (line 159: `err := json.Unmarshal(protocol.Value, &sp)`). Then the sharetype is established from testing the protocol name (line 165: `if sp.Name == "datatx"`). I'll move line 156 to line 159.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
redblom: When a different share type is discovered (eg. 'datatx') the `sharetype` is set to represent that:
```
			if sp.Name == "datatx" {
				sharetype = ocm.Share_SHARE_TYPE_TRANSFER
			}
```

redblom:
> 
> 
> Once you've made the changes, please follow the steps in https://reva.link/docs/tutorials/share-tutorial/, followed by share manipulation as described in https://codimd.web.cern.ch/qBGUdbZhQHeXIw9FVCei6Q#Manipulating-shared-files, to make sure that everything works as expected. You might need to add the following lines to the config
> 
> ```
> [grpc.services.gateway]
> commit_share_to_storage_grant = true
> commit_share_to_storage_ref = true
> ```

@ishank011 : Done. There are no issues here but this https://github.com/cs3org/reva/pull/1242#issuecomment-808166218 is unclear to me. Currently in case of a transfer share, a ref is made in MyShares just like a regular share. Is this correct in case of a transfer share? 
What is exactly expected here in light of your comment https://github.com/cs3org/reva/pull/1242#issuecomment-808166218? 
What will be the expected path for rclone to use in the transfer?

cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,167 @@
...
+			return formatError(statRes.Status)
 		}
 
+		providerInfoResp, err := client.GetInfoByDomain(ctx, &ocmprovider.GetInfoByDomainRequest{
+			Domain: *idp,
+		})
+		if err != nil {
+			return err
+		}
+		permissions := conversions.PermissionAll
```
ishank011, line 103: Let's get the permissions like this https://github.com/cs3org/reva/blob/02dd990bf0567db4eb925d30ff6dc968fe6fcd25/cmd/reva/ocm-share-create.go#L102-L105

internal/http/services/ocmd/shares.go
```diff
@@ -155,6 +155,11 @@ func (h *sharesHandler) createShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	desiredProtocol, ok := options["desired-protocol"].(string)
```
ishank011, line 158: This will also need to be changed to `options["protocol"]`, right?

cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,172 @@
...
+		})
+		if err != nil {
+			return err
+		}
+
+		_, pint, err := getOCMSharePerm(rol)
+		if err != nil {
+			return err
+		}
+		resourcePermissions := &provider.ResourcePermissions{
```
ishank011, line 109: Just do 
```
resourcePermissions, pint, err := getOCMSharePerm(editorPermission)
if err != nil {
    return err
}
```

cmd/reva/transfer-create.go
```diff
@@ -19,42 +19,172 @@
...
+			InitiateFileDownload: true,
+		}
+		datatxProtocol, err := json.Marshal(
```
ishank011, line 112: Do we need the json struct? We're not using options anywhere and from the HTTP handler, you don't even pass that. Use
```
"protocol": &types.OpaqueEntry{
	Decoder: "plain",
	Value:   []byte("datatx")
},
```

PR #1583, opened 2021-03-25
Update example in README

sudo-sturbia:


ishank011:
Hey @sudo-sturbia. Change it to `storage-references`. We don't really support native integration with phoenix anymore and that is the most complete minimal example.

sudo-sturbia:
> Hey @sudo-sturbia. Change it to `storage-references`. We don't really support native integration with phoenix anymore and that is the most complete minimal example.

Updated. Thanks for the explanation!

ishank011 (approved):

PR #1584, opened 2021-03-25
Move shares folder out to a separate mount and handle ref conflicts

ishank011:


butonic:
@ishank011 I think this can be closed in favor of https://github.com/cs3org/reva/pull/2023

PR #1588, opened 2021-03-26
Make additional info attribute configureable

ishank011:


refs (dismissed):
👍

labkode:
@ishank011 can you make it configurable with a layout like we do for the home configuration, for example: ```{{.Username - .Mail}}```

ishank011:
@labkode done

refs:
> can you make it configurable with a layout like we do for the home configuration, for example: {{.Username - .Mail}}

@ishank011 could you document that functionality in the changelog? 😇

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -73,6 +76,8 @@ func (h *Handler) Init(c *config.Config) error {
 	h.sharePrefix = c.SharePrefix
 	h.homeNamespace = c.HomeNamespace
 
+	h.additionalInfoTemplate, _ = template.New("additionalInfo").Parse(c.AdditionalInfoAttribute)
```
refs, line 79: Are you not checking the error here because the only way it could fail is if the template has already been parsed? Otherwise, I think there is some value in error checking here, at least for debugging purposes.

changelog/unreleased/share-add-info-configureable.md
```diff
@@ -0,0 +1,3 @@
+Enhancement: Make the additional info attribute for shares configureable
+
```
refs, line 2:
```suggestion
AdditionalInfoAttribute can be configured via the `additional_info_attribute` key in the form of a Go template string. If not explicitly set, the default value is `{{.Mail}}`
```

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -52,17 +54,18 @@ import (
...
+	homeNamespace          string
+	additionalInfoTemplate *template.Template
+	userIdentifierCache    *ttlcache.Cache
 }
 
 // we only cache the minimal set of data instead of the full user metadata
 type userIdentifiers struct {
 	DisplayName string
-	UserName    string
+	Username    string
```
refs, line 68: Follow [Go's mixedCaps](https://golang.org/doc/effective_go#mixed-caps) style? 😇

refs (changes requested):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
ishank011: @refs I thought that we shouldn't fatal if the parsing fails. Similarly, in case executing the template fails, I didn't report an error but returned an empty string. I'll add a warning there

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
ishank011: I changed it so that we can use the same layout for shares and shares. Sharees uses executes the template on the CS3APIs user object, which has the 'Username' parameter (https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go#L119), so I changed it to make it compatible

refs: 0Oh thanks for clarifying :)

changelog/unreleased/share-add-info-configureable.md
```diff
@@ -0,0 +1,3 @@
+Enhancement: Make the additional info attribute for shares configureable
```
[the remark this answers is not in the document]
ishank011: Done

changelog/unreleased/share-add-info-configureable.md
ishank011: Done

refs (approved):

labkode (approved):

PR #1590, opened 2021-03-26
signature authentication for public links

C0rby:
Implemented the mechanism proposed here: https://github.com/cs3org/cs3apis/issues/110.
The signature authentication is limited to downloads.

### DIFF
--- a/internal/grpc/services/publicshareprovider/publicshareprovider.go
+++ b/internal/grpc/services/publicshareprovider/publicshareprovider.go
@@ -149,7 +149,7 @@ func (s *service) GetPublicShareByToken(ctx context.Context, req *link.GetPublic
 	log.Debug().Msg("getting public share by token")
 
 	// there are 2 passes here, and the second request has no password
-	found, err := s.sm.GetPublicShareByToken(ctx, req.GetToken(), req.GetPassword())
+	found, err := s.sm.GetPublicShareByToken(ctx, req.GetToken(), req.GetAuthentication(), req.GetSign())
 	switch v := err.(type) {
 	case nil:
 		return &link.GetPublicShareByTokenResponse{
@@ -180,7 +180,7 @@ func (s *service) GetPublicShare(ctx context.Context, req *link.GetPublicShareRe
 		log.Error().Msg("error getting user from context")
 	}
 
-	found, err := s.sm.GetPublicShare(ctx, u, req.Ref)
+	found, err := s.sm.GetPublicShare(ctx, u, req.Ref, req.GetSign())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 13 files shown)


micbar:
@ishank011 we would like to bring that into the release. Could you do a fastlane 🏎️ review?

internal/grpc/services/publicshareprovider/publicshareprovider.go
```diff
@@ -148,8 +149,17 @@ func (s *service) GetPublicShareByToken(ctx context.Context, req *link.GetPublic
 	log := appctx.GetLogger(ctx)
 	log.Debug().Msg("getting public share by token")
 
+	sig := req.GetAuthentication().GetSignature()
+	auth := publicshare.Authentication{
```
ishank011, line 153: Can you just pass `req.GetAuthentication()` to the package? This looks like we're unnecessarily complicating things

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -552,6 +552,7 @@ func (s *service) ListContainer(ctx context.Context, req *provider.ListContainer
 	for i := range listContainerR.Infos {
 		filterPermissions(listContainerR.Infos[i].PermissionSet, ls.GetPermissions().Permissions)
 		listContainerR.Infos[i].Path = path.Join(s.mountPath, "/", tkn, relativePath, path.Base(listContainerR.Infos[i].Path))
+		addShare(listContainerR.Infos[i], ls)
```
ishank011: Add error check

pkg/cbox/publicshare/sql/sql.go
```diff
@@ -348,6 +353,9 @@ func (m *manager) ListPublicShares(ctx context.Context, u *user.User, filters []
 		if expired(cs3Share) {
 			_ = m.cleanupExpiredShares()
 		} else {
+			if cs3Share.PasswordProtected && sign {
```
ishank011, line 356: This returns the password, not the sign

pkg/cbox/publicshare/sql/sql.go
```diff
@@ -455,3 +468,18 @@ func checkPasswordHash(password, hash string) bool {
 	err := bcrypt.CompareHashAndPassword([]byte(strings.TrimPrefix(hash, "1|")), []byte(password))
 	return err == nil
 }
+
+func authenticate(share *link.PublicShare, pw string, auth publicshare.Authentication) bool {
+	switch {
+	case auth.Password != "":
+		return checkPasswordHash(auth.Password, pw)
+	case auth.Signature != publicshare.Signature{}:
```
ishank011, line 476: Are you sure this condition isn't true when auth.Signature is nil?

ishank011 (changes requested):
@C0rby you're returning the password instead of the sign in the getter methods.

pkg/cbox/publicshare/sql/sql.go
C0rby: `auth.Signature` can't be nil.

pkg/cbox/publicshare/sql/sql.go
C0rby: What do you mean? The `cs3Share.PasswordProtected`? It's a boolean which is true when a password is set for the share.

https://github.com/cs3org/reva/blob/538c8bdafb0ceeca96f650e580885dfeb4035276/pkg/cbox/utils/conversions.go#L213-L215
https://github.com/cs3org/reva/blob/538c8bdafb0ceeca96f650e580885dfeb4035276/pkg/cbox/utils/conversions.go#L236

ishank011: Never mind

ishank011 (approved):

PR #1597, opened 2021-03-29
check ENOTDIR for readlink

butonic:
```json
{
  "level": "error",
  "service": "storage",
  "pkg": "rgrpc",
  "traceid": "24d2f1bb7b3cb88c68fd005ced336cce",
  "error": "Decomposedfs: Wrap: readlink error: readlink /var/tmp/ocis/storage/users/nodes/a5066dc8-8566-4aa4-9286-d63c30bdc9a6/somename: not a directory",
  "time": "2021-03-29T19:01:05Z",
  "caller": "/home/vscode/go/pkg/mod/github.com/cs3org/reva@v1.6.1-0.20210326165326-e8a00d9b2368/internal/grpc/services/storageprovider/storageprovider.go:571",
  "message": "error stating: path:\"/home/testfile.txt/somename\" "
}
```

readlink returns ENOTDIR when a path segment is a file and not a directory. `os.IsNotExist` won't cover this case: https://github.com/golang/go/issues/18974

part of https://github.com/owncloud/ocis/issues/1239

Links: https://www.man7.org/linux/man-pages/man2/readlink.2.html


butonic:
remaining apiWebdavProperties1/copyFile.feature tests require returning a new error type that we can use to map the `AlreadyExists` error ...

butonic:
@ishank011 I also removed the failing test from eos because I don't expect it to be an issue for eos ...

aduffeck (approved):
LGTM :+1: 

You could maybe consider adding a unit test to the `Describe("Child", ...` block in `node_test.go` for it, e.g.

```
		It("handles (broken) links including file segments by returning an non-existant node", func() {
			child, err := parent.Child(env.Ctx, "file1/broken")
			Expect(child).ToNot(BeNil())
			Expect(child.Exists).To(BeFalse())
		})
```

ishank011 (approved):

PR #1598, opened 2021-03-29
Introduce and map already exists status

butonic:
the next step after https://github.com/cs3org/reva/pull/1597 on the path to https://github.com/owncloud/ocis/issues/1239


butonic:
covered by https://github.com/cs3org/reva/pull/1725 as well

PR #1604, opened 2021-03-31
Privatelink support

butonic:
This pr adds initial support for private links, however due to the backwards compatible nature of the webdav api the `dav/meta` endpoint needs to return a path that is relative to the currently logged in users home. Currently shares are jailed into a subfolder inside the users home ... and we would need to reconstruct the path in the shares folder.

Maybe the full implementation should wait until we have a globally accessible endpoint.


butonic:
bitrotted and superseded by https://github.com/cs3org/reva/pull/2741

PR #1620, opened 2021-04-07
[tests-only] updated expected failures

kiranparajuli589:


kiranparajuli589:
previous(most recent) builds were behaving strangely failing most of the sharing tests. But at the date of the comment...CI seems stable (fails only on `licence-check` part of build suite)
So, expected failures need not be updated.

PR #1629, opened 2021-04-08
implement checksums in the owncloud storage driver

C0rby:


labkode:
@C0rby we usually abbreviate checksum as `xs` not `cs` in case you wanna change it.

C0rby:
@labkode, I don't really have a hard opinion on this.
The prefix was already there just commented so I just used it. Also in other places we use `cs`.
https://github.com/cs3org/reva/blob/a23b75020b0cae4adea44acfe767e75f7cdb1035/pkg/storage/utils/decomposedfs/xattrs/xattrs.go#L50

But I'm also fine with changing it to xs.
@butonic what do you think?

IljaN:
I`d say that we try to stay consistent as much as possible. :)

butonic:
I agree with @refs, we should try to be consistent. ... @labkode what could be the explanation behind using `xs` for checksums? Is that some legacy thing? Is `cs` taken by something else?

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -682,6 +684,14 @@ func (fs *ocfs) convertToResourceInfo(ctx context.Context, fi os.FileInfo, ip st
 
 	ri.PermissionSet = fs.permissionSet(ctx, ri.Owner)
 
+	// checksums
+	if _, ok := mdKeysMap[checksumsKey]; (!fi.IsDir()) && returnAllKeys || ok {
```
dragotin, line 688: I like to vote for more explicit error handling here. The handling of the IsDir() case should be done explicitely as that will ease debugging later big times. Also, would'nt a log message be helpful in the case that something is a dir that should not be or vise versa?

Also, I am only 38 years in programming and _still_ do not know quickly what `A && B || C` really evaluates to, how about helping me poor soul with additional parenthesis?

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -2228,5 +2238,44 @@ func (fs *ocfs) propagate(ctx context.Context, leafPath string) error {
...
+			Type: storageprovider.PKG2GRPCXS(algo),
+			Sum:  hex.EncodeToString(v),
+		}
+	case isNoData(err):
+		appctx.GetLogger(ctx).Debug().Err(err).Str("nodepath", nodePath).Str("algorithm", algo).Msg("checksum not set")
+	case isNotFound(err):
+		appctx.GetLogger(ctx).Error().Err(err).Str("nodepath", nodePath).Str("algorithm", algo).Msg("file not fount")
+	default:
+		appctx.GetLogger(ctx).Error().Err(err).Str("nodepath", nodePath).Str("algorithm", algo).Msg("could not read checksum")
+	}
```
dragotin, line 2255: How nice can it be :-)

pkg/storage/fs/owncloud/upload.go
```diff
@@ -375,6 +383,54 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) error {
...
+		ah := adler32.New()
+		f, err := os.Open(upload.binPath)
+		if err != nil {
+			log.Err(err).Msg("Decomposedfs: could not open file for checksumming")
+			// we can continue if no oc checksum header is set
+		}
+		defer f.Close()
+
+		r1 := io.TeeReader(f, sh)
+		r2 := io.TeeReader(r1, mh)
```
dragotin, line 403: Ok, bear with me, but.... short variable names have a long history of proving that they're not cool. We tried that in C already back in the dark days, paying back with lots of head scratching... So why not just spend a few more chars to get speaking names? Yes, also for dumb, short living local variables... Ok, I am not blaming you here for the golang way of doing things ;-)

pkg/storage/fs/owncloud/upload.go
```diff
@@ -375,6 +383,54 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) error {
...
+			err = upload.checkHash(parts[1], md5h)
+		case "adler32":
+			err = upload.checkHash(parts[1], adler32h)
+		default:
+			err = errtypes.BadRequest("unsupported checksum algorithm: " + parts[0])
+		}
+		if err != nil {
+			return err
+		}
+	}
```
dragotin, line 433: This means that we do not require checksums, right? But if  the feature "checksums" is activated, shouldn't it be an error/warning if the upload comes completely without checksum? Or is that checked elsewhere?

C0rby:
@dragotin, I must confess that I just copied the checksum code from the [decomposedfs](https://github.com/cs3org/reva/tree/master/pkg/storage/utils/decomposedfs). But I can still implement your feedback nonetheless.

pkg/storage/fs/owncloud/upload.go
C0rby: Yes, as it is right now checksums are optional. I don't know if we even have the flag to "activate" checksums.
Also I'm not sure how much sense it makes to spend too much effort in the owncloud storage driver since we are only using it for the webUI acceptance tests and even them will in the near future need to switch to the ocis storage driver.

This PR is just to make the webUI tests for the Thumbnails PR pass. https://github.com/owncloud/ocis/pull/1898

ishank011 (approved):

PR #1634, opened 2021-04-12
Mentix site authorization status fix

Daniel-WWU-IT:
If a site changes its authorization status, Mentix did not update its internal data to reflect this change. This PR fixes this issue.


changelog/1.6.0_2021-02-16/mentix-auth-fix.md
```diff
@@ -0,0 +1,5 @@
+Bugfix: Site authorization status changes
```
labkode, line 1: This file should not be changed, only the unreleased one.

Daniel-WWU-IT: That file actually went there by accident (and I wondered why the initial changelog had disappeared...). Removed the file (it never was released, it was just placed in the wrong directory).

labkode (approved):

PR #1637, opened 2021-04-13
Store transfers in separate directory

ishank011:
@redblom you can rebase your commits on top of this. The default path is `/home/Data-Transfers`.

You need to fix one line, the check for whether it's a transfer share or not.

https://github.com/cs3org/reva/blob/dd33392f6c176fbece84498db900863c43b441e6/internal/grpc/services/gateway/ocmshareprovider.go#L315

### DIFF
--- a/internal/grpc/services/gateway/ocmshareprovider.go
+++ b/internal/grpc/services/gateway/ocmshareprovider.go
@@ -311,17 +311,33 @@ func (s *svc) createWebdavReference(ctx context.Context, share *ocm.Share) (*rpc
 		return status.NewInternal(ctx, err, "error updating received share"), nil
 	}
 
-	// reference path is the home path + some name on the corresponding
-	// mesh provider (/home/MyShares/x)
-	// It is the responsibility of the gateway to resolve these references and merge the response back
-	// from the main request.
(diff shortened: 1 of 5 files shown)


internal/grpc/services/gateway/ocmshareprovider.go
```diff
@@ -313,15 +313,21 @@ func (s *svc) createWebdavReference(ctx context.Context, share *ocm.Share) (*rpc
...
 				},
 			},
 		})
+		if err != nil {
+			return status.NewInternal(ctx, err, "error creating transfers directory"), nil
+		}
+		if createTransferDir.Status.Code != rpc.Code_CODE_OK && createTransferDir.Status.Code != rpc.Code_CODE_ALREADY_EXISTS {
+			err := status.NewErrorFromCode(createTransferDir.Status.GetCode(), "gateway")
+			return status.NewInternal(ctx, err, "error creating transfers directory"), nil
+		}
```
redblom, line 329: This also returns `ALREADY_EXISTS` on the transfer folder itself.

ishank011: Check again. Reva logs it, doesn’t return that

PR #1646, opened 2021-04-19
Added dtu partner to ocm-partners example 

marpapdtu:


ishank011 (approved):
@marpapdtu please contact @Daniel-WWU-IT to add DTU to the mentix DB as well.

PR #1653, opened 2021-04-21
the facto -> a de facto

michielbdejong:
Ref https://en.wikipedia.org/wiki/De_facto_standard


michielbdejong:
Given this is just a minor edit in the docs and not a code change, do you still want changelog entry about it? Happy to create one if you do! :)

michielbdejong:
Given this is just a minor edit in the docs and not a code change, do you still want changelog entry about it? Happy to create one if you do! :)

ishank011 (approved):
@michielbdejong looks good. For future reference, we have a list of tags for which we skip the changelog check. Currently this includes `[build-deps]` and `[tests-only]`. I'll add `[docs-only]` to the list as well

Issue #1654, opened 2021-04-21
Trouble running `make` on master branch


michielbdejong:
```
[...]
go: downloading golang.org/x/crypto v0.0.0-20190911031432-227b76d455e7
`go env GOPATH`/bin/calens > /dev/null
go run tools/check-changelog/main.go
2021/04/21 13:51:51 No changelog added. Please create a changelog item based on your changes
exit status 1
make: *** [check-changelog] Error 1
```

I'm trying to work around this now with
```sh
go run tools/prepare-release/main.go --version unreleased
```

ishank011:
We expect a changelog to be added whenever any changes are made. You can run `make build` to skip this check.

PR #1657, opened 2021-04-22
make build instead of make

michielbdejong:
The current 'Build and run it yourself' instructions from the readme will fail if there are unreleased changes, as described in #1654.
As per https://github.com/cs3org/reva/issues/1654#issuecomment-824061617, it should however be enough to run `make build` instead of the `make` step.


michielbdejong:
Wait I'll recreate this with the [docs-only] tag in the commit message.

PR #1658, opened 2021-04-22
[docs-only] Run `make build` instead of `make`

michielbdejong:
The current 'Build and run it yourself' instructions from the readme will fail if there are unreleased changes, as described in #1654.
As per #1654 (comment), it should however be enough to run `make build` instead of the `make` step.

Continued from #1657.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,14 @@ $ git clone https://github.com/cs3org/reva
 $ cd reva
 $ make deps
 $ make build
-$ cd examples/storage-references
-$ ../../cmd/revad/revad -dev-dir .
+$ sudo mkdir -p /etc/revad
+$ sudo cp examples/storage-references/users.demo.json /etc/revad/users.json
+$ sudo cp examples/storage-references/groups.demo.json /etc/revad/groups.json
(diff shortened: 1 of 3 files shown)


michielbdejong:
Hm, seems that the `[docs-only]` tag didn't have the desired effect. Anyway, I think we don't want a changelog entry for this, right?

README.md
```diff
@@ -24,12 +24,15 @@ You need to have [Go](https://golang.org/doc/install), [git](https://git-scm.com
 $ git clone https://github.com/cs3org/reva
 $ cd reva
 $ make deps
-$ make
-$ cd examples/storage-references
-$ ../../cmd/revad/revad -dev-dir .
+$ make build
+$ sudo mkdir -p /etc/revad
```
ishank011, line 28: Can you remove sudo from the commands? We don't want to assume access control.

ishank011:
> Hm, seems that the `[docs-only]` tag didn't have the desired effect. 

Not merged yet https://github.com/cs3org/reva/pull/1655

ishank011 (approved):

PR #1664, opened 2021-04-26
Add cache warmup strategy for OCS resource infos

ishank011:
Recently, a TTL cache was added to OCS to store statted resource infos https://github.com/cs3org/reva/pull/1643. This PR adds an interface to define warmup strategies and also adds a cbox specific strategy which starts a goroutine to initialize the cache with all the valid shares present in the system.

@refs @butonic @labkode please take a look at this when you get the time. I'm not 100% sure about the design so would love any feedback. I'm also thinking about adding a strategy with the json share + OCIS drivers.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -50,6 +50,8 @@ import (
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/pkg/rhttp/router"
+	"github.com/cs3org/reva/pkg/share/cache"
+	"github.com/cs3org/reva/pkg/share/cache/registry"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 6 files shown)


butonic:
I don't feel fit for an in depth review today, but my mind immediately wandered towards caching the root etag for a storage space in the storage registry instead of the ocs or ocdav services. In any case, I think it makes sense to add a cache to the ocs service.

In my head a received share is also represented by a storage space. AFAICT the gateway and the storage providers have the ListStoragSpace API, which needs to still be implemented. So I am in favor of caching the results of the StorageSpace lists instead of adding a cach for stat results ... althoug one could argue that a storage space and a stat actually are the same thing.

I am trying to clarify the terminology. Maybe tomorrow a PR will be ready to clarify the terms ... just not feeling well ...

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -87,9 +96,25 @@ func (h *Handler) Init(c *config.Config) error {
 
 	h.resourceInfoCache = gcache.New(c.ResourceInfoCacheSize).LFU().Build()
 
+	if h.resourceInfoCacheTTL > 0 {
+		cwm, _ := getCacheWarmupManager(c)
```
refs, line 100:
```suggestion
		cwm, err := getCacheWarmupManager(c)
		if err == nil {
			go h.startCacheWarmup(cwm)
		}
```

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -87,9 +96,25 @@ func (h *Handler) Init(c *config.Config) error {
...
+		go h.startCacheWarmup(cwm)
```
refs, line 101:
```suggestion

```

refs (approved):
LGTM, just a nitpick

labkode (approved):

PR #1669, opened 2021-04-28
Mint scope-based access tokens for RBAC

### LINKED ISSUE #1549: Revamp public shares authentication model
## Why?
With the current workflow with which public shares are handled, we run the risk of exposing the resource owner's token to random users who access the shared link. Using only the token corresponding to the public shares, the `Authenticate` gateway method can be called, which will return the owner's token.

https://github.com/cs3org/reva/blob/3b7e1f0a50d6a5683f5ca18e4b6983608e6ab62d/internal/http/services/owncloud/ocdav/dav.go#L177-L183

## Anonymous users
I propose using anonymous users to grant share-specific access. When a public share is created, a temporary user is created and assigned the appropriate grant for the shared file.
- For a share corresponding to `#/s/RMxHPGhqbRySVOH`, we add an ACL `u:public_RMxHPGhqbRySVOH:rwx`.
- When this link is accessed, the `Authenticate` method of type `publicshares` returns this user if it's able to find the share corresponding to it.
- This user is set in the context, and the file systems correctly handle the requests as the ACLs are already set.

## Possible issues
- Users can misuse these anonymous users like they can in the current implementation. While they won't be able to access actual files as they would be protected by ACLs, the returned token can still be used to perform other operations (manipulating shares, accessing info about other users). This would have to be handled in the HTTP and GRPC auth interceptors with checks that these anonymous users should only be allowed access to the `/remote.php/dav/public-files/` endpoint and those exposed by the public storage provider service.
- For file systems which use local extended attributes, adding ACLs corresponding to non-existent users should be trivial. However, for EOS, this won't work as EOS expects the UID to actually exist when performing operations. Hence, it would have to be handled separately. While setting arbitrary ACLs does work, to actually access the resource, we'll need to impersonate the owner.
- For every public share created, we'll need to add an ACL. This does not seem to be too major an issue, as for one particular resource, we shouldn't expect more than ~10 links to be created. Especially when we have the janitor cleaning up expired links, we can remove the ACLs at the same time.

Please let me know your thoughts about this @labkode @butonic @refs @dragotin

ishank011:
Primarily, this PR is meant to introduce the concept of scopes into our tokens. At the moment, it addresses those cases where we impersonate other users without allowing the full scope of what the actual user has access to.

Going by owncloud/ocis#1971, it serves a similar purpose. We can modify the current design to incorporate both the things. I wanted to finish coding it so that I could verify its correctness. 

Here's a diagram detailing how the tokens are minted and additional scopes added https://codimd.web.cern.ch/XTib-1TzTyqx2IZJJOq5pA (you might have to zoom in a bit). Also a short explanation:
- We get the public share using the token provided by the client.
- In the public share, we know the resource ID, so we can add this to the allowed scope, but not the path.
- However, later OCDav tries to access by path as well. Now this is not allowed at the moment. However, from the allowed scope, we have the resource ID and we're allowed to stat that. We stat the resource ID, get the path and if the path matches the one passed by OCDav, we add it to the scope as well.

Serves as a better approach to https://github.com/cs3org/reva/issues/1549, closes #1549.

Depends on https://github.com/cs3org/cs3apis/pull/123


ishank011:
One caveat is that whenever we try to access a resource by an attribute that is not added to the token scope, we'll have to resolve it (resource ID -> path or vice versa). Even if we append the scope by adding the missing attribute, we can't pass the new token back to the caller.

labkode (approved):

PR #1676, opened 2021-04-30
[tests-only] Test against adjustUserDataCleanup core branch

phil-davis:
core branch `adjustUserDataCleanup` removes `deleteAllSharesForUser` from the `afterScenario` of tests.

This PR demonstrates if it is OK to stop doing that now.

When users are deleted, the shares of the user should be getting cleaned up by reva already.


phil-davis:
Passed - good

PR #1678, opened 2021-05-03
[WIP] introduce /dav/spaces endpoint && CS3 list spaces implementation (ocis driver only)

butonic:
initial implementation for listing storage spaces
- currently lists
  - user homes as type `personal` spaces
  - shares as type `share` spaces
  - if /spaces dir does not exist in the metadata it will iterate over all nodes to create the spaces for migration
- [ ] create a `share` space when creating a share
- [ ] create a `personal` space when creating a home
    - [ ] refactor CreateHome
- [ ] generic create space
- [ ] delete `share` space on unsharing
- [ ] delete `personal` in deprovisioning a user 
    - [ ] and all contained shares
- [ ] delete `share` space on deleting a folder for all children
- [ ] generic delete space
    - [ ] share is special because it should not delete the underlying items. Do we need to differentiate between spaces and sub spaces? eg when a user created a new `project` space from his `personal` space? or do we move it out of it and really manage it as a dedicated space? Should we manage `share` spaces in the same way? at least for ocis we could ... would be too theoretical imo.


butonic:
cc @dragotin

C0rby:
All method handlers are implemented. I guess the next step is to refactor/deduplicate the code.

butonic:
superseded by https://github.com/cs3org/reva/pull/1776

PR #1683, opened 2021-05-05
Filter shares based on type in OCS

ishank011:


ishank011:
@kulmann do we expect a list to be passed in the `share_types` query? Because the only usage I see is from the public links tab which sends `/ocs/v2.php/apps/files_sharing//api/v1/shares?format=json&share_types=3`

kulmann:
@ishank011 found this in oc10: https://github.com/owncloud/core/blob/799f107301e9835a46c3d1855f9a1a56f6706971/apps/files_sharing/js/sharedfilelist.js#L238 so it looks like it's supposed to be a comma separated list of share_types.

ishank011:
@kulmann perfect, thanks! Will change the code to handle this.

labkode (approved):

PR #1685, opened 2021-05-05
Refactor sharing

C0rby:
- extract share creation into its own method
- extract state conversion into its own method
- extract resource info retrieval
- some minor clean up
Also the responses to accepting or rejecting a share return the updated share now.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -61,10 +61,10 @@ type Handler struct {
 	publicURL              string
 	sharePrefix            string
 	homeNamespace          string
-	resourceInfoCacheTTL   time.Duration
 	additionalInfoTemplate *template.Template
 	userIdentifierCache    *ttlcache.Cache
 	resourceInfoCache      gcache.Cache
+	resourceInfoCacheTTL   time.Duration
 }
 
 // we only cache the minimal set of data instead of the full user metadata
@@ -87,14 +87,13 @@ func (h *Handler) Init(c *config.Config) error {
 	h.publicURL = c.Config.Host
 	h.sharePrefix = c.SharePrefix
 	h.homeNamespace = c.HomeNamespace
-	h.resourceInfoCacheTTL = time.Duration(c.ResourceInfoCacheTTL)
+	h.resourceInfoCache = gcache.New(c.ResourceInfoCacheSize).LFU().Build()
+	h.resourceInfoCacheTTL = time.Second * time.Duration(c.ResourceInfoCacheTTL)
 
 	h.additionalInfoTemplate, _ = template.New("additionalInfo").Parse(c.AdditionalInfoAttribute)
 
 	h.userIdentifierCache = ttlcache.NewCache()
-	_ = h.userIdentifierCache.SetTTL(60 * time.Second)
-
-	h.resourceInfoCache = gcache.New(c.ResourceInfoCacheSize).LFU().Build()
+	_ = h.userIdentifierCache.SetTTL(time.Second * 60)
 
 	if h.resourceInfoCacheTTL > 0 {
 		cwm, err := getCacheWarmupManager(c)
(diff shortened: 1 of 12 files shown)


C0rby:
I don't quite get why 
```
apiSharePublicLink1/deletePublicLinkShare.feature:37
apiSharePublicLink1/deletePublicLinkShare.feature:38
```
fail...
Locally they run just fine...
@individual-it or @phil-davis  do you have an idea?

Also @ishank011 or @labkode, I don't have access to the cs3org in snyk so I can't see what test is failing there. Could you tell me what I need to fix?

C0rby:
> I don't quite get why 
> ```
> apiSharePublicLink1/deletePublicLinkShare.feature:37
> apiSharePublicLink1/deletePublicLinkShare.feature:38
> ```
> fail...

I think I found the issue.

C0rby:
Listing a public link after renaming it fails for some reason....
Need to investigate further...

individual-it:
sadly the logs of reva are cut off in drone

individual-it:
I rebased on master here https://github.com/cs3org/reva/pull/1690 let's see
Next I will try to bump the oc-core commit id

phil-davis:
> sadly the logs of reva are cut off in drone

In owncloud drone CI we put a flush of the file-system and a 30-second wait at the end of acceptance test runs. That improved the situation - pipeline step output seemed to get from the drone agent to drone server more often! I am not sure what happens currently in cs3org/reva CI.

C0rby:
@individual-it, my refactoring is causing the issue. I just haven't found the cause yet...

C0rby:
Arrgh I found it...

When we are doing a stat request on a share like here https://github.com/cs3org/reva/blob/28500a8e0bb49f52f176c18b26f59d598c7cf0ab/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go#L178

we eventually end up here https://github.com/cs3org/reva/blob/28500a8e0bb49f52f176c18b26f59d598c7cf0ab/pkg/storage/fs/owncloud/owncloud.go#L746
In that line we get the file path using the OpaqueId as the key. When we rename a file, that path will be outdated because it doesn't get updated on a file rename. Because of that the stat request fails with a `not found` error.

Back in the `listPublicShares` method we check the result of the stat https://github.com/cs3org/reva/blob/28500a8e0bb49f52f176c18b26f59d598c7cf0ab/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go#L178-L179

But if you look closely you see that it doesn't check the status code of the stat response but of the listPublicShare response.

My refactoring changed that to checking the stat response and that's why it doesn't return a share after the shared file gets renamed...

C0rby:
This is the underlying issue. https://github.com/cs3org/reva/issues/1693 Will work on it on monday

C0rby:
Waiting for this fix to be merged: https://github.com/cs3org/reva/pull/1696

labkode:
@ishank011 can you review this one?

butonic (approved):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -61,10 +61,11 @@ type Handler struct {
 	publicURL              string
 	sharePrefix            string
 	homeNamespace          string
-	resourceInfoCacheTTL   time.Duration
 	additionalInfoTemplate *template.Template
 	userIdentifierCache    *ttlcache.Cache
 	resourceInfoCache      gcache.Cache
+	resourceInfoCacheTTL   time.Duration
+	useResourceInfoCache   bool
```
ishank011, line 68: Should we use it in the user-configurable struct? If a user sets `useResourceInfoCache` to true but doesn't set `resourceInfoCacheTTL` assuming a default value, we won't use the cache. I'd suggest removing this and checking  the value of `resourceInfoCacheTTL` in all places

ishank011: Okay, no it's not user configurable. But I'd still vote for removing the extra variable. Doesn't really add any value

C0rby: It does add a bit of readability but I have no hard opinion on this, I can remove it again.

ishank011 (approved):

Issue #1693, opened 2021-05-07
owncloud storage driver doesn't update file path in redis


C0rby:
The ownCloud storage driver stores the file path in redis using the fileid as the key.
When the file gets renamed 'Moved' the filepath doesn't get updated which causes problems when we want to get the path of that file later.

The filepath must be updated on 'Moves' and remove on 'Delete'

PR #1696, opened 2021-05-10
fix move in the owncloud storage driver

C0rby:
Update the filepath in redis when moving a file. Didn't implement it in delete since delete is still a bit more broken. But since we don't actively use the owncloud storage driver except for in CI it's not too important.

Fixes https://github.com/cs3org/reva/issues/1693


C0rby:
Argh... cloning the oc10 tests failed...
```
Cloning into '/drone/src/tmp/testrunner'...
error: RPC failed; curl 56 OpenSSL SSL_read: Connection reset by peer, errno 104
error: 15 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: index-pack failed
```

I don't have permissions to restart the build...

C0rby:
The thing about deleting is that purging non empty folders does not work and purging versions of files doesn't work.

I can't really tell how long it will take to fix them. 
Not removing the id/path mapping from redis won't break anything. It will just take up some bytes of space.

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -1589,6 +1589,25 @@ func (fs *ocfs) Move(ctx context.Context, oldRef, newRef *provider.Reference) (e
...
+
+	log := appctx.GetLogger(ctx)
+	conn := fs.pool.Get()
+	defer conn.Close()
+	// Ideally if we encounter an error here we should rollback the Move/Rename.
+	// But since the owncloud storage driver is not being actively used by anyone other
+	// than the acceptance tests we should be fine by ignoring the errors.
+	_ = filepath.Walk(newIP, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.Error().Str("path", path).Err(err).Msg("error caching id")
```
ishank011, line 1601: Can you add a log about rolling back the whole op? And a TODO comment

ishank011 (changes requested):
Okay, but please create an issue about this

pkg/storage/fs/owncloud/owncloud.go
C0rby: Ok I opened an issue https://github.com/cs3org/reva/issues/1699
And added todos.

ishank011 (approved):

PR #1707, opened 2021-05-13
[tests-only] Add feature to be able to run only specific part(s) of test suites in CI

saw-jan:
This PR adds the feature to be able to run only the certain part(s) of the test suites.

E.g;
`ocisIntegrationTests(6, [1, 4])` will run only `1` and `4` parts out of 6

Part of https://github.com/owncloud/QA/issues/670

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -92,6 +92,9 @@ def cephService():
 
 # Pipeline definitions
 def main(ctx):
+  # In order to run specific parts only, specify the parts as
+  # ocisIntegrationTests(6, [1, 4])     - this will only run 1st and 4th parts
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
LGTM - that will let anyone choose to run just some pipelines of tests that are split into many parallel parts. That will be handy when trying to investigate some specific test failure(s) - no need to keep re-running all the pipelines.

phil-davis:
@labkode @ishank011 we have done similar in owncloud repos to make it easy to cut-down CI  when someone wants to run just some specific part in order to investigate test fails.

ishank011 (approved):

PR #1718, opened 2021-05-20
Fix: use the -static ldflag only for the 'build-ci' target

SamuAlfageme:
cc/ @jimil749


jimil749:
@SamuAlfageme I tested this locally and this works fine! Thank you. :)

ishank011 (approved):

Issue #578, opened 2020-03-19
Separation of concerns: storages should not know about their mount point


butonic:
Currently, the mount point of a storage needs to be configured in two places: the storage registry and the storage provider.

The gateway uses the storage registry to route requests to the correct storage.

The storage provider wraps and unwraps the path by removing or adding the mount point prefix (roughly speaking).

This allows configuring two storage providers with the same mount point, which might lead to confusion or conflicts. It is no longer clear which storage provider is responsible. Even if it is supposed to represent a load balancing configuration that configuration is better stored in the storage registry. 

The storage providers should only be concerned and know about the namespace within them.

How the global namespace is composed is the responsibility of the storage registry.

The mount point path wrapping should be moved from the storage providers to the gateway, which already has the mount point from storage registry (which it always fetches to determine which storage to use).

Related PRs:
https://github.com/cs3org/reva/pull/562
https://github.com/cs3org/reva/pull/577
https://github.com/butonic/reva/pull/1

labkode:
> This allows configuring two storage providers with the same mount point, which might lead to confusion or conflicts. It is no longer clear which storage provider is responsible. Even if it is supposed to represent a load balancing configuration that configuration is better stored in the storage registry.

Having two storage providers providing the same mount path and different mount ids is really important in sharding scenarios where you hide under the same namespace (/eos/user/) multiple storage providers. So for example, /eos/user/h/hugo will go to storage-h and /eos/user/j/jorn to storage-j, that they both present /eos/user as their path but different mount ids.

Of course, this model is also plausible in the model you are proposing, and it may even be simpler.

Once we have the MVP we can jump again to this, I find very interesting and what you are proposing it may simplify some logic.

:)

labkode:
As discussed in #601 it makes lot of sense

PR #1721, opened 2021-05-20
switch references

butonic:
A preview of the changes that need to be done when merging https://github.com/cs3org/cs3apis/pull/130

currently uses a go.mod replace to use the generated go-cs3libs at https://github.com/butonic/go-cs3apis/commit/2093bacf375338a7e4234791fd8aa449fd1ec49f

Initial notes
- After merging this GetPath becomes obsolete because it was only used to merge a relative path with an id based reference that is stored in the share manager. With the new Reference type we can represent the combination with a single struct. Deferred to a subsequent PR.
- When creating a stat response or listing the path needs to take into account the original reference: if it was an absolute path we need to return an absolute path. If it was a reference with a storage and node id we need to keep them, and just adjust the path.
  - The ResourceInfo needs to hold both, a Reference that includes the name AND an Id property that holds a Reference for the actual file id (composed of storage id and node id) of the node
- The storage provider must not know its mount point. That is the job of the registry/gateway. We adressed that in https://github.com/cs3org/reva/pull/1678/files#diff-198f1004a921b3627f7572a239452974429a7de6e4fa47f445c2ad35d2cd9026L1103 by resolving the resource using the storage.FS interface ... when really it should be done by the drivers, tracked in https://github.com/cs3org/reva/issues/578

### DIFF
--- a/internal/grpc/interceptors/auth/auth.go
+++ b/internal/grpc/interceptors/auth/auth.go
@@ -261,7 +261,11 @@ func dismantleToken(ctx context.Context, tkn string, req interface{}, mgr token.
 
 			// Since the public share is obtained from the scope, the current token
 			// has access to it.
-			statReq := &provider.StatRequest{Ref: share.ResourceId}
+			statReq := &provider.StatRequest{
+				Ref: &provider.Reference{
+					ResourceId: share.ResourceId,
+				},
+			}
 
 			statResponse, err := client.Stat(ctx, statReq)
 			if err != nil || statResponse.Status.Code != rpc.Code_CODE_OK {
--- a/internal/grpc/interceptors/auth/auth.go
+++ b/internal/grpc/interceptors/auth/auth.go
@@ -261,11 +261,7 @@ func dismantleToken(ctx context.Context, tkn string, req interface{}, mgr token.
 
 			// Since the public share is obtained from the scope, the current token
 			// has access to it.
-			statReq := &provider.StatRequest{
-				Ref: &provider.Reference{
-					Spec: &provider.Reference_Id{Id: share.ResourceId},
-				},
-			}
+			statReq := &provider.StatRequest{Ref: share.ResourceId}
 
 			statResponse, err := client.Stat(ctx, statReq)
 			if err != nil || statResponse.Status.Code != rpc.Code_CODE_OK {
@@ -280,7 +276,7 @@ func dismantleToken(ctx context.Context, tkn string, req interface{}, mgr token.
 		}
 	}
 
-	return nil, err
+	return nil, errtypes.PermissionDenied("access token has invalid scope")
 }
 
 func extractRef(req interface{}) (*provider.Reference, bool) {
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -112,7 +112,7 @@ func (h *Handler) startCacheWarmup(c cache.Warmup) {
 		return
 	}
 	for _, r := range infos {
-		key := wrapResourceID(r.Id)
+		key := wrapResourceID(&provider.Reference{ResourceId: r.Id})
 		_ = h.resourceInfoCache.SetWithExpire(key, r, time.Second*h.resourceInfoCacheTTL)
 	}
 }
@@ -331,7 +331,7 @@ type PublicShareContextName string
 
 func (h *Handler) getShare(w http.ResponseWriter, r *http.Request, shareID string) {
 	var share *conversions.ShareData
-	var resourceID *provider.Reference
+	var reference *provider.Reference
 	ctx := r.Context()
 	logger := appctx.GetLogger(r.Context())
 	logger.Debug().Str("shareID", shareID).Msg("get share by id")
@@ -371,7 +371,9 @@ func (h *Handler) getShare(w http.ResponseWriter, r *http.Request, shareID strin
 
 	if err == nil && psRes.GetShare() != nil {
 		share = conversions.PublicShare2ShareData(psRes.Share, r, h.publicURL)
-		resourceID = psRes.Share.ResourceId
+		reference = &provider.Reference{
+			ResourceId: psRes.Share.ResourceId,
+		}
 	}
 
 	if share == nil {
@@ -404,7 +406,9 @@ func (h *Handler) getShare(w http.ResponseWriter, r *http.Request, shareID strin
 		*/
 
 		if err == nil && uRes.GetShare() != nil {
-			resourceID = uRes.Share.ResourceId
+			reference = &provider.Reference{
+				ResourceId: uRes.Share.ResourceId,
+			}
 			share, err = conversions.CS3Share2ShareData(ctx, uRes.Share)
 			if err != nil {
 				response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "error mapping share data", err)
@@ -419,7 +423,7 @@ func (h *Handler) getShare(w http.ResponseWriter, r *http.Request, shareID strin
 		return
 	}
 
-	info, status, err := h.getResourceInfoByID(ctx, client, resourceID)
+	info, status, err := h.getResourceInfoByID(ctx, client, reference)
 	if err != nil {
 		log.Error().Err(err).Msg("error mapping share data")
 		response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "error mapping share data", err)
@@ -506,7 +510,9 @@ func (h *Handler) updateShare(w http.ResponseWriter, r *http.Request, shareID st
 		return
 	}
 
-	statReq := provider.StatRequest{Ref: uRes.Share.ResourceId}
+	statReq := provider.StatRequest{Ref: &provider.Reference{
+		ResourceId: uRes.Share.ResourceId,
+	}}
 
 	statRes, err := client.Stat(r.Context(), &statReq)
 	if err != nil {
@@ -616,7 +622,7 @@ func (h *Handler) listSharesWithMe(w http.ResponseWriter, r *http.Request) {
 		var info *provider.ResourceInfo
 		if pinfo != nil {
 			// check if the shared resource matches the path resource
-			if !utils.ResourceEqual(rs.Share.ResourceId, pinfo.Id) {
+			if !utils.ResourceEqual(&provider.Reference{ResourceId: rs.Share.ResourceId}, &provider.Reference{ResourceId: pinfo.Id}) {
 				// try next share
 				continue
 			}
@@ -624,7 +630,7 @@ func (h *Handler) listSharesWithMe(w http.ResponseWriter, r *http.Request) {
 			info = pinfo
 		} else {
 			var status *rpc.Status
-			info, status, err = h.getResourceInfoByID(ctx, client, rs.Share.ResourceId)
+			info, status, err = h.getResourceInfoByID(ctx, client, &provider.Reference{ResourceId: rs.Share.ResourceId})
 			if err != nil || status.Code != rpc.Code_CODE_OK {
 				h.logProblems(status, err, "could not stat, skipping")
 				continue
@@ -763,7 +769,7 @@ func (h *Handler) addFilters(w http.ResponseWriter, r *http.Request, prefix stri
 }
 
 func wrapResourceID(r *provider.Reference) string {
-	return wrap(r.StorageId, r.NodeId)
+	return wrap(r.ResourceId.StorageId, r.ResourceId.OpaqueId)
 }
 
 // The fileID must be encoded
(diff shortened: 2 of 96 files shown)


labkode:
* It needs to be possible to obtain an absolute path from any reference

butonic:
@labkode can you elaborate? The gateway is currently path based. The now reference would allow directly forwarding all requests if they have a storage_id set. The response for a stat or list container countains ResouceInfo elements, which should contain a path relative to the resource that was used to make the request.

That being said: when a request is made with an absolute path, all resulting references must (and will) be path based only again.

Is that what you mean?

refs:
Reminder to update `go.mod` removing this replace `github.com/cs3org/go-cs3apis => github.com/refs/go-cs3apis v0.0.0-20210520134557-97cdc479815b`

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -340,14 +341,15 @@ func (fs *eosfs) isShareFolderChild(ctx context.Context, p string) bool {
...
+func (fs *eosfs) GetPathByID(ctx context.Context, id *provider.Reference) (string, error) {
 	u, err := getUser(ctx)
 	if err != nil {
 		return "", errors.Wrap(err, "eos: no user in ctx")
 	}
 
 	// parts[0] = 868317, parts[1] = photos, ...
-	parts := strings.Split(id.OpaqueId, "/")
+	// FIXME @butonic REFERENCE ... umm ... 868317/photos? @ishank011 HELP?!?
+	parts := strings.Split(id.NodeId, "/")
```
ishank011, line 352: @butonic pretty sure we can change this to `fileID, err := strconv.ParseUint(id.NodeId, 10, 64)`. I can't remember any scenario where we append the path to the inode. Plus looking at the code, `GetPathByID` is never used anyway, so we should be good

labkode:
@butonic the share manager will persist a share like:
`share with id 123 points to resourceID abc:456.`

A call to `GetPathById(abc:456)` is needed to obtain the path of that share and present it to the user.

A user won't access a share by doing `ls /myshares/123`. That is not user friendly.
I hope is clear. Ping me on Gitter if is not for a quicker chat.

butonic:
@labkode right. I think the first goal should be to just replace the reference. No change in how the current path or id based references are used. `/myshares/someid` is not user friendly and AFAICT is nothing that this PR will touch or introduce.

The namespace that a user will see when using the CS3 api to list a constainer depends on the type of reference he uses. When he uses an absolute path all returned references will be absolute. When he uses a mixed reference (with id & path) the returned references will be relative to the id. But again, I think that is something that will become clearer when rebasing the spaces PR on this.

refs:
@labkode @ishank011 We've come full circle with a topic: Should `ResourceInfo` contain 2 different References ([see here](https://github.com/butonic/cs3apis/blob/simplify-references/cs3/storage/provider/v1beta1/resources.proto#L47-L68)), and if so, what would be a more descriptive name that carries semmantical meaning?

Some extracts from @butonic and I over RocketChat providing more context:

> [...] in a stat request, the id must always be a `Reference` with `storage_id` and `node_id` set, but the response also needs to carry the name of the file
>
> [...] think of [`RecourceInfo.Ref`](https://github.com/butonic/cs3apis/blob/simplify-references/cs3/storage/provider/v1beta1/resources.proto#L68) as Parent (either as id or as path) + Name
>
> [...] a `ResouceInfo` needs two References: one is the Id of the `Resource`, the other is a Path to the `Resource` which can be given either as an absolute path, or as a relative path to a storage space.

If we really need both `Reference` members, their names should carry more semantic meaning. We thought of renaming `ResourceInfo.Ref` to `ResourceInfo.Path`. WDYT? It is not a huge of a deal, but adding more semantic meaning to the API should be always a goal.

butonic:
note that `ResourceInfo.Path` was renamed to `ResourceInfo.Ref` in https://github.com/cs3org/cs3apis/pull/126/files#diff-b91600c359c2e479b2dbc31ff8dd5bd52bfaa48dc70c33a0de121215ff0c030eR68

I think we should leave the name `Path` but change the type from string to `Reference`

butonic:
@labkode @refs @ishank011 I think this can be merged (after review). But it needs to be updated with the newly generated go libs from the official cs3 api changes in https://github.com/cs3org/cs3apis/pull/126. Let me know what you think.

micbar:
@labkode @ishank011 can you review / merge?

labkode:
@butonic @refs we need to differentiate between a resource ID and a Reference. A resource ID is unique mapping 1 to 1 to a resource. On the other hand, multiple reference can point to the same ResourceID. They are used for different purposes.

The ResourceInfo needs to contain a ResourceID and Path as before (no change here).
The Reference should only be used as a method to getting information.

From a ResourceInfo:ResourceID and ResourceInfo.Path you can always construct a Refence.

Please change the PR accordingly.

butonic:
> @butonic @refs we need to differentiate between a resource ID and a Reference. A resource ID is unique mapping 1 to 1 to a resource. On the other hand, multiple reference can point to the same ResourceID. They are used for different purposes.

✔️ 

> The ResourceInfo needs to contain a ResourceID and Path as before (no change here).
> The Reference should only be used as a method to getting information.

✔️ 

> From a ResourceInfo:ResourceID and ResourceInfo.Path you can always construct a Refence.

Only when you know to which node the path is relative to. I wonder why we need a path at all, and not just the basename of the resource. When listing a resource the Reference used to make the determines the path to the resource, anyway. A client listing `/path/to/resource` already knows that all children are prefixed with that path, e.g. `/path/to/resource/child`. Why is a name property with `child` not enough? I know that the gateway currently always looks up the path for a reference. When the reference holds a spaceid we don't need to do that and can just forward the request to the correct storage provider. When listing the available spaces clients always will have references with a space id. 

I used a Reference for Path, because there is a corner case for the spaces api: when making an id based request we need to know the parent id so that we can build the full path relative to the storage space root. So I used a Reference to return both, the Id of the parent and the name of the child.

In any case the same reasoning applies to all cases: path, id based and mixed requests. The returned path should be relative to the reference that was used to make the request. But that makes lookung up the path for an id impossible.

> Please change the PR accordingly.
I will try to use a path and check with our spaces implementation. IIRC we need to be able to get to parent from an id only based reference.

internal/grpc/interceptors/auth/auth.go
```diff
@@ -280,7 +276,7 @@ func dismantleToken(ctx context.Context, tkn string, req interface{}, mgr token.
 		}
 	}
 
-	return nil, err
+	return nil, errtypes.PermissionDenied("access token has invalid scope")
```
butonic, line 279: @ishank011 I think this is a bug and we should return an explicit permission denied error here because `extractRef` might not return `ok` ... then err is still nil.

butonic:
@labkode @ishank011 I updated this PR with [the generated go libs](https://github.com/butonic/go-cs3apis/commit/0b2c079174c9) for https://github.com/cs3org/cs3apis/pull/130

I limited the changes to the Reference usage, save [one bugfix](https://github.com/cs3org/reva/pull/1721#pullrequestreview-678392617). ResourceInfo keeps the Id (now a `Reference`) and the `string` type `Path`. I'd like to get this in ASAP after merging the CS3 API changes.

Let me know what you think or go ahead and merge https://github.com/cs3org/cs3apis/pull/130.

pkg/storage/utils/eosfs/eosfs.go
butonic: ok, but maybe not as part of this pr?

internal/grpc/interceptors/auth/auth.go
ishank011: @butonic looks good, thanks!

labkode:
@butonic can you make LGTM happy as well? I think is important from a security perspective, specially when dealing with files.

pkg/sdk/action/download.go
```diff
@@ -89,13 +89,7 @@ func (action *DownloadAction) Download(fileInfo *storage.ResourceInfo) ([]byte,
...
 func (action *DownloadAction) initiateDownload(fileInfo *storage.ResourceInfo) (*gateway.InitiateFileDownloadResponse, error) {
 	// Initiating a download request gets us the download endpoint for the specified resource
-	req := &provider.InitiateFileDownloadRequest{
-		Ref: &provider.Reference{
-			Spec: &provider.Reference_Path{
-				Path: fileInfo.Path,
-			},
-		},
-	}
+	req := &provider.InitiateFileDownloadRequest{Ref: &provider.Reference{ResourceId: fileInfo.Id}}
```
ishank011, line 92: Let's leave this as `Path: fileInfo.Path`?

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -792,20 +792,20 @@ func (fs *ocfs) GetPathByID(ctx context.Context, id *provider.ResourceId) (strin
...
 
-	if ref.GetId() != nil {
-		ip, err := fs.getPath(ctx, ref.GetId())
+	// if storage id is set look up that
+	if ref.ResourceId != nil {
+		ip, err := fs.getPath(ctx, ref.ResourceId)
 		if err != nil {
 			return "", err
 		}
+		filepath.Join("/", ip, ref.Path)
```
ishank011, line 802: This should be assigned to `ip`, right?

pkg/storage/fs/s3/s3.go
```diff
@@ -115,12 +115,12 @@ func (fs *s3FS) addRoot(p string) string {
...
 
 func (fs *s3FS) resolve(ctx context.Context, ref *provider.Reference) (string, error) {
-	if ref.GetPath() != "" {
+	if ref.Path != "" {
 		return fs.addRoot(ref.GetPath()), nil
 	}
 
-	if ref.GetId() != nil {
-		fn := path.Join("/", strings.TrimPrefix(ref.GetId().OpaqueId, "fileid-"))
+	if ref.ResourceId.OpaqueId != "" {
```
ishank011, line 122: `ref.ResourceId` can be nil, which will result in a panic

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -625,8 +618,7 @@ func (h *Handler) listSharesWithMe(w http.ResponseWriter, r *http.Request) {
 		var info *provider.ResourceInfo
 		if pinfo != nil {
 			// check if the shared resource matches the path resource
-			if rs.Share.ResourceId.StorageId != pinfo.GetId().StorageId ||
-				rs.Share.ResourceId.OpaqueId != pinfo.GetId().OpaqueId {
+			if !utils.ResourceEqual(&provider.Reference{ResourceId: rs.Share.ResourceId}, &provider.Reference{ResourceId: pinfo.Id}) {
```
ishank011, line 621: We can directly use `ResourceIDEqual` here and other places but those can be fixed later

ishank011 (commented):
Hi @butonic. Just a couple of minor comments

butonic:
@labkode the file traversal cases are in progress by @C0rby 
@ishank011 will address your issues asap!

pkg/sdk/action/download.go
butonic: `initiateDownload` uses a `ResourceInfo`. The only stable identifier we get from that is the id. Maybe it should use a new `Reference` as the parameter? 
```
func (action *DownloadAction) Download(fileInfo *storage.ResourceInfo) ([]byte, error) {
```
also should use a Reference?

pkg/utils/utils.go
```diff
@@ -196,3 +197,46 @@ func UnmarshalJSONToProtoV1(b []byte, m proto.Message) error {
...
+
+	if p == "/" {
+		return "."
+	}
+	return "." + p
+}
+
+// IsSameUserID compares to UserIds for equallity.
+// TODO move to a comparison package
+func IsSameUserID(i *userpb.UserId, j *userpb.UserId) bool {
```
ishank011, line 233: There's already a method `UserEqual which does that`

pkg/storage/fs/owncloud/owncloud.go
```diff
@@ -263,7 +263,9 @@ func (fs *ocfs) toInternalPath(ctx context.Context, sp string) (ip string) {
 	if fs.c.EnableHome {
 		u := user.ContextMustGetUser(ctx)
 		layout := templates.WithUser(u, fs.c.UserLayout)
-		ip = filepath.Join(fs.c.DataDirectory, layout, "files", sp)
+		// The inner filepath.Join prevents the path from breaking out of
+		// <fs.c.DataDirectory>/<layout>/files/
+		ip = filepath.Join(fs.c.DataDirectory, layout, "files", filepath.Join("/", sp))
```
ishank011, line 268: Can you give an example where it would break without the inner join?

pkg/sdk/action/download.go
ishank011: I can't really say, haven't looked at this code

pkg/storage/fs/owncloud/owncloud.go
C0rby: If `sp` starts with `..` then it would 'break out' of our files directory.
See for example: https://play.golang.org/p/JVfjiNAYCR0

ishank011: Ah cool

ishank011 (approved):

PR #1724, opened 2021-05-21
fix response format

C0rby:
After accepting or declining a share the response needs to enclose the updated share in an array.
This is to make the API oc10 compatible.


C0rby:
Fixes: https://github.com/owncloud/web/issues/4946#issuecomment-845941525

ishank011 (approved):

PR #1739, opened 2021-05-27
Allow full paths

butonic:
A continuation of https://github.com/cs3org/reva/pull/1605, https://github.com/labkode/reva/pull/183 and https://github.com/cs3org/reva/pull/1691.

@phil-davis @individual-it I will add the failing tests to the expected failures because there is now a new scenario which is not covered by the oc10 tests. When using a global namespace on top of eos it is possible to do a PROPFIND on `/eos/users/e/` when setting the `share_prefix="/"`, `files_namespace="/"` and `webdav_namespace="/"` in the ocs/ocdav config sections.

The implentation in this PR allows using global paths, but when a share is not mounted it will leave the internal path, exactly like oc10. It leaks the internal path ... yes, just like oc10 ... so we need to fix it anyway. This will be addressed with the spaces api. For now I'll add the failing test cases to expected failures.

When the spaces API is in place we can clean this up more elegantly because shares will be accessed via the new `/dav/spaces` endpoint.

The faling tests expect the shares to be mounted in the users home. But the config is actually setting up a reva instance that jails users into home AND a shares folder, so I think unless an acceptance test contains:
```
    And the administrator has set the default folder for received shares to "Shares"
    And auto-accept shares has been disabled
```
it will fail, which makes sense.


butonic:
maybe we can use the `share_prefix` to trigger trimming the paths when they have not been accepted, yet. That might let some tests pass...

butonic:
After discussing this with @phil-davis and @individual-it we will introduce a new `expose_global_namespace` flag to trigger exposing global paths.

butonic:
Looking at the test results https://github.com/cs3org/reva/pull/1739/commits/81a3b1de4e843c37f48d81f28f9160229e6d8e74 actually sooms to work without an additional flag:

- When shares are jailed into the users home the webdav url to access them is `https://domain.tld/webdav/Shares/<sharename>`.  This is the default ocis configuration and we set `share_prefix='/Shares'` so the ocs list shares with me endpoint constructs paths by using the `/Shares` prefix and the basename of the shared resource.
- cernBox uses a global namespace where the webdav url to access a share is, eg. `https://domain.tld/webdav/users/m/marie/path/to/share`.  Setting `share_prefix='/'` will make the ocs list shares with me endpoint just use the path returned from the gateway as is: global.

Note that to jails users into their home ocis also has to set the ocdav `files_namespace=/users` and `webdav_namespace=/home`.
cernBox uses `files_namespace=/` (I think ... maybe they need `/users` here as well) and `webdav_namespace=/` to expose the global namespace on the webdav endpoints.

We are working on a owncloud share storage provider that will use an oc10 database to mount shares exactly like in oc10. We need this for migration purposes to run oc10 and ocis in parallel. We will then move towards the spaces api to give clients a more direct way to talk to the correct storage providers.

labkode:
@butonic CI is green! @ishank011 can you please give another pair of eyes?

labkode:
@labkode @ishank011 re-check if part of fork

labkode:
@butonic can you rebase this one?

butonic:
@labkode @ishank011 @micbar rebased and dropped cosmetic changes to the expected failures files

butonic:
@labkode can you check this out?

labkode (approved):

PR #1750, opened 2021-06-01
set Content-Type correctly

individual-it:
all headers need to be set before `w.WriteHeader(status code)` is called

fixes https://github.com/owncloud/ocis/issues/1779

### DIFF
--- a/changelog/unreleased/fix-content-type.md
+++ b/changelog/unreleased/fix-content-type.md
@@ -1,0 +1,5 @@
+Bugfix: set Content-Type header correctly for ocs requests
+
+Before this fix the `Content-Type` header was guessed by `w.Write` because `WriteHeader` was called to early. Now the `Content-Type` is set correctly and to the same values as in ownCloud 10
+
+https://github.com/owncloud/ocis/issues/1779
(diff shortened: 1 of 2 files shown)


labkode (approved):

individual-it:
Thanks for approving , I want to tweak it a bit so that reva responses the same as oC10

individual-it:
@labkode I've added some more changes, please review again

changelog/unreleased/fix-content-type.md
```diff
@@ -0,0 +1,5 @@
+Bugfix: set Content-Type header correctly for ocs requests
+
+Before this fix the `Conten-Type` header was guessed by `w.Write` because `WriteHeader` was called to early. Now the `Content-Type` is set correctly and to the same values as in ownCloud 10
```
phil-davis, line 3:
```suggestion
Before this fix the `Content-Type` header was guessed by `w.Write` because `WriteHeader` was called to early. Now the `Content-Type` is set correctly and to the same values as in ownCloud 10
```

phil-davis (dismissed):
LGTM - only one nit-pick in the changelog.

phil-davis (approved):

labkode (approved):

Issue #1765, opened 2021-06-04
Docker binaries are not statically linked, so using 'scratch' in the Docker files results in error


ylebre:
The docker images built using Dockerfile.reva and Dockerfile.revad end in an error:
```
standard_init_linux.go:211: exec user process caused "no such file or directory"
```
The cause seems to the compiled reva and revad binaries are not statically linked.

Adding env CGO_ENABLED lets the compiler generate a statically linked binary.
After this change, running the docker image results in
```
nothing to do, no grpc/http enabled_services declared in config
```
which makes sense because the passed configuration file is empty.

PR #1766, opened 2021-06-04
Build static binaries for Docker, so that we can use 'scratch' image

ylebre:
This should fix issue #1765

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,9 @@ ci: build-ci test  lint-ci
 
 # to be run in Docker build
 build-revad-docker: off
-	go build -ldflags ${BUILD_FLAGS} -o ./cmd/revad/revad ./cmd/revad
+	env CGO_ENABLED=0 go build -ldflags ${BUILD_FLAGS} -o ./cmd/revad/revad ./cmd/revad
 build-reva-docker: off
-	go build -ldflags ${BUILD_FLAGS} -o ./cmd/reva/reva ./cmd/reva
+	env CGO_ENABLED=0 go build -ldflags ${BUILD_FLAGS} -o ./cmd/reva/reva ./cmd/reva
 clean:
 	rm -rf dist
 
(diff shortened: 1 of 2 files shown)


IljaN:
There is probably more: https://github.com/owncloud/ocis/issues/375

ishank011:
@SamuAlfageme can you take a look?

michielbdejong:
I'm trying this out in https://github.com/cs3org/ocm-test-suite/pull/27.
So far it works, but I must say the new very slim image is hard to debug, for instance:
```sh
docker run -it --entrypoint /bin/sh revad
```
fails because there is no `/bin/sh`.

SamuAlfageme:
@ylebre great catch, I did forget to set the `CGO_ENABLED` variable on the `ENV`s  builder step from the Dockerfile(s). 

@IljaN I'll also review what cgo dependencies are we still using and how can we avoid them if that is the case.

SamuAlfageme:
@michielbdejong I totally empathize. Using the binary `FROM scratch` is great to reduce the attack surface and size of the image, but can make debugging inside the container **very** cumbersome. 

I have used this [scratch-debugger](https://github.com/kubernetes-retired/contrib/tree/master/scratch-debugger) script in the past. But if you're developing locally, I'll recommend you create an image with your OS of choice as base and then copying the revad binary `--from` the `cs3org/revad` image.  

In the future, things like ephemeral containers in k8s https://github.com/kubernetes/enhancements/issues/277 will make that task easier.

We can also reconsider using busybox or alpine if this becomes too annoying in the future

SamuAlfageme (approved):

SamuAlfageme:
@ylebre could you add a changelog entry for your fix?

ylebre:
@SamuAlfageme sure thing, added to the PR now.

SamuAlfageme (approved):

SamuAlfageme:
@ylebre thx a bunch!

PR #1774, opened 2021-06-08
Add user ID cache warmup to EOS storage driver

ishank011:


pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -119,6 +118,14 @@ func (c *Config) init() {
 		c.UserLayout = "{{.Username}}" // TODO set better layout
 	}
 
+	if c.UserIDCacheSize == 0 {
+		c.UserIDCacheSize = 1000000
+	}
+
+	if c.UserIDCacheWarmupDepth == 0 {
+		c.UserIDCacheWarmupDepth = 2
```
labkode, line 126: Perhaps put a comment about what depth means in terms on layout for example for CERN: `/eos/user/<letter>/<username>`...

ishank011: Will add

PR #1775, opened 2021-06-08
Unit test for EOS ACLs ordering

gmgigi96:


glpatcern:
This is (unfortunately) postponed until eos respects the order - the tests correctly fail because of that!

PR #1785, opened 2021-06-11
Extend app registry with AddProvider method and mimetype filters

### LINKED ISSUE #1779: Revamp Application Registry
Currently, the registry works by simply redirecting requests corresponding to the mimetype to the endpoint specified by a given set of rules.

For example, let's say we want to redirect office format requests to a reva appprovider running at localhost:16001 and codimd ones to one running at localhost:16002. We'll add the following rules to appregistry

```
[grpc.services.appregistry.static.rules]
"application/vnd.oasis.opendocument.text" = "localhost:16001"
"application/vnd.oasis.opendocument.spreadsheet" = "localhost:16001"
"application/vnd.oasis.opendocument.presentation" = "localhost:16001"
"application/compressed-markdown" = "localhost:16002"
```

Thus when we receive a request to open a file with the extension `.docx` in the app provider, the storage provider will resolve the mimetype to be an office text format and this request will be forwarded to the reva at localhost:16001.

@glpatcern can you elaborate what is the desired mechanism?

ishank011:
Closes #1779

### DIFF
--- a/examples/ocmd/ocmd-server-1.toml
+++ b/examples/ocmd/ocmd-server-1.toml
@@ -77,21 +77,18 @@ providers = "providers.demo.json"
 driver = "memory"
 
 [grpc.services.appprovider]
-driver = "demo"
-iopsecret = "testsecret"
-wopiurl = "http://0.0.0.0:8880/"
-wopibridgeurl = "http://localhost:8000/wopib"
+driver = "wopi"
+
+[grpc.services.appprovider.drivers.wopi]
+iop_secret = "hello"
+wopi_url = "http://0.0.0.0:8880/"
+wopi_bridge_url = "http://localhost:8000/wopib"
 
 [grpc.services.appregistry]
 driver = "static"
 
-[grpc.services.appregistry.static.rules]
-"text/plain" = "localhost:19000"
-"text/markdown" = "localhost:19000"
-"application/compressed-markdown" = "localhost:19000"
-"application/vnd.oasis.opendocument.text" = "localhost:19000"
-"application/vnd.oasis.opendocument.spreadsheet" = "localhost:19000"
-"application/vnd.oasis.opendocument.presentation" = "localhost:19000"
+[grpc.services.appregistry.drivers.static.providers]
+"localhost:19000" = {"address" = "localhost:19000", "name" = "WOPIServer", mimetypes = ["text/plain", "text/markdown", "application/compressed-markdown", "application/vnd.oasis.opendocument.text", "application/vnd.oasis.opendocument.spreadsheet", "application/vnd.oasis.opendocument.presentation"]}
 
 [grpc.services.storageprovider]
 driver = "localhome"
--- a/examples/ocmd/ocmd-server-1.toml
+++ b/examples/ocmd/ocmd-server-1.toml
@@ -76,19 +76,18 @@ providers = "providers.demo.json"
 [grpc.services.publicshareprovider]
 driver = "memory"
 
+[grpc.services.appregistry]
+driver = "static"
+
 [grpc.services.appprovider]
-driver = "wopi"
+driver = "demo"
+app_provider_url = "localhost:19000"
 
 [grpc.services.appprovider.drivers.wopi]
 iop_secret = "hello"
 wopi_url = "http://0.0.0.0:8880/"
-wopi_bridge_url = "http://localhost:8000/wopib"
-
-[grpc.services.appregistry]
-driver = "static"
-
-[grpc.services.appregistry.drivers.static.providers]
-"localhost:19000" = {"address" = "localhost:19000", "name" = "WOPIServer", mimetypes = ["text/plain", "text/markdown", "application/compressed-markdown", "application/vnd.oasis.opendocument.text", "application/vnd.oasis.opendocument.spreadsheet", "application/vnd.oasis.opendocument.presentation"]}
+app_name = "Collabora"
+app_url = "https://your-collabora-server.org:9980"
 
 [grpc.services.storageprovider]
 driver = "localhome"
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,7 @@ require (
 go 1.16
 
 replace (
-	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210611101519-1497ab8e8312
+	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210628124432-b0cb570ff3ee
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
 	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // temporary downgrade
--- a/go.mod
+++ b/go.mod
@@ -69,7 +69,7 @@ require (
 go 1.16
 
 replace (
-	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210628124432-b0cb570ff3ee
+	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210701122052-bb129518b297
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
 	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // temporary downgrade
--- a/go.mod
+++ b/go.mod
@@ -68,6 +68,8 @@ require (
 go 1.16
 
 replace (
+	//github.com/cs3org/go-cs3apis => ../cs3apis/build/go-cs3apis
+	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210611101519-1497ab8e8312
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
 	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // temporary downgrade
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/ReneKroon/ttlcache/v2 v2.7.0
 	github.com/asaskevich/govalidator v0.0.0-20200428143746-21a406dcc535 // indirect
 	github.com/aws/aws-sdk-go v1.40.0
+	github.com/beevik/etree v1.1.0
 	github.com/bluele/gcache v0.0.2
 	github.com/c-bata/go-prompt v0.2.5
 	github.com/cheggaaa/pb v1.0.29
--- a/go.mod
+++ b/go.mod
@@ -69,7 +69,7 @@ require (
 go 1.16
 
 replace (
-	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210701122052-bb129518b297
+	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210715114809-34729f68a479
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
 	google.golang.org/grpc => google.golang.org/grpc v1.26.0 // temporary downgrade
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,6 @@ require (
 go 1.16
 
 replace (
-	//github.com/cs3org/go-cs3apis => ../cs3apis/build/go-cs3apis
 	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210611101519-1497ab8e8312
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
--- a/internal/grpc/services/gateway/appregistry.go
+++ b/internal/grpc/services/gateway/appregistry.go
@@ -32,7 +32,7 @@ func (s *svc) GetAppProviders(ctx context.Context, req *registry.GetAppProviders
 	if err != nil {
 		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
 		return &registry.GetAppProvidersResponse{
-			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
+			Status: status.NewInternal(ctx, err, "error getting app registry client"),
 		}, nil
 	}
 
@@ -49,7 +49,7 @@ func (s *svc) AddAppProvider(ctx context.Context, req *registry.AddAppProviderRe
 	if err != nil {
 		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
 		return &registry.AddAppProviderResponse{
-			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
+			Status: status.NewInternal(ctx, err, "error getting app registry client"),
 		}, nil
 	}
 
@@ -66,7 +66,7 @@ func (s *svc) ListAppProviders(ctx context.Context, req *registry.ListAppProvide
 	if err != nil {
 		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
 		return &registry.ListAppProvidersResponse{
-			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
+			Status: status.NewInternal(ctx, err, "error getting app registry client"),
 		}, nil
 	}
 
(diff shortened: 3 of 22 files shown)


labkode:
Needs https://github.com/cs3org/cs3apis/pull/131 to be merged first. After, that:
* Point to upstream go-cs3apis
* Document breaking change in changelog

internal/http/services/appprovider/appprovider.go
```diff
@@ -0,0 +1,232 @@
...
+			return
+		}
+
+		s.handleWopiOpen(w, r)
+	})
+}
+
+// WopiResponse holds the various fields to be returned for a wopi open call
+type WopiResponse struct {
+	WopiClientURL  string `json:"wopiclienturl"`
```
glpatcern, line 110: If this is supposed to be a generic appprovider responding via HTTP, then the response should be similar to the GRPC one, just the `AppURL`. Why do we want to expose the different parts of a WOPI-enabled application address (which is WOPI specific)?

internal/http/services/appprovider/appprovider.go
```diff
@@ -0,0 +1,232 @@
...
+	// &revisionhistory=1
+	// &title=Hello.odt
+	u.RawQuery = q.Encode()
+
+	js, err := json.Marshal(
+		WopiResponse{
+			WopiClientURL: u.String(),
+			AccessToken:   accessToken,
+			// https://wopi.readthedocs.io/projects/wopirest/en/latest/concepts.html#term-access-token-ttl
+			AccessTokenTTL: time.Now().Add(time.Second*time.Duration(s.conf.AccessTokenTTL)).UnixNano() / 1e6,
```
glpatcern, line 167: (Related to the above) this configuration parameter is specified in the WOPI server config, and would need to be duplicated in some Reva config. Why is it important to provide this information to the web UI?

pkg/app/provider/wopi/wopi.go
```diff
@@ -0,0 +1,317 @@
...
+		// register the supported mimetypes in the AppRegistry: this is hardcoded for the time being
+		if c.AppName == "CodiMD" {
+			appURLs = getCodimdExtensions(c.AppURL)
+		} else if c.AppName == "Etherpad" {
+			appURLs = getEtherpadExtensions(c.AppURL)
+		}
+	} else if discRes.StatusCode == http.StatusOK {
+		var netZoneName string
+		if c.AppName == "Collabora" {
+			netZoneName = "external-http"
```
glpatcern, line 235: I fear this is very CERN-deployment specific... we'll probably need to come up with something better

examples/ocmd/ocmd-server-1.toml
```diff
@@ -77,21 +77,18 @@ providers = "providers.demo.json"
...
-driver = "demo"
-iopsecret = "testsecret"
-wopiurl = "http://0.0.0.0:8880/"
-wopibridgeurl = "http://localhost:8000/wopib"
+driver = "wopi"
+
+[grpc.services.appprovider.drivers.wopi]
+iop_secret = "hello"
+wopi_url = "http://0.0.0.0:8880/"
+wopi_bridge_url = "http://localhost:8000/wopib"
```
glpatcern, line 85: `wopi_bridge_url` is gone now!
Shall we add `ssl_verify = "True"`, cf. https://github.com/cs3org/reva/pull/1842 ?

internal/http/services/appprovider/appprovider.go
wkloucek: We are currently using this in the oCIS WOPI extension to not have the access token in the browsers' address bar. If you have the access token in the address bar it could easily leak to other people (copy and paste)

https://github.com/owncloud/ocis-wopiserver/blob/e23ff205ec4474e4297ba39248636dca5fbb9402/ui/app.js#L84-L105 shows the form push with the access token. Downside is that you cant copy the URL and open it in a new tab

wkloucek: The same form post method is also recommended for embedding into iframes: https://wopi.readthedocs.io/en/latest/hostpage.html

glpatcern: Makes sense, indeed we currently embed those URLs in an iframe so the user only sees a cernbox URL and can't really copy/paste it (yet of course by inspecting the iframe's content you'd see the access_token).

At this point I'd at least rename the struct to `AppResponse`, including the fields `AppURL` and `AccessToken`, just to be more generic and not WOPI specific. Other applications that do not implement an access token mechanism may leave it empty. Also, I'd still drop the TTL value - eventually, what rules is the TTL of the main JWT obtained by the user when logging in the first time, right?

internal/http/services/appprovider/appprovider.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	info, errCode, err := s.getStatInfo(ctx, r.URL.Query().Get("fileId"), client)
+	if err != nil {
+		ocmd.WriteError(w, r, errCode, "error statting file", err)
+	}
+
+	openReq := gateway.OpenInAppRequest{
+		Ref:      &provider.Reference{ResourceId: info.Id},
+		ViewMode: getViewMode(info),
```
wkloucek, line 131: From my viewpoint it would be desirable if somebody could also decide to open a file in readonly mode even if he has permissions to open it in readwrite mode.  (We didn't takle this in oCIS WOPI extension because we have only one open action in the Web UI until now, but this will be more flexible in the future https://github.com/owncloud/web/issues/5135)

internal/http/services/appprovider/appprovider.go
wkloucek: Depending on how you define a TTL, the access token ttl is WOPI specific since it is the UNIX timestamp of the moment when the storage / REVA JWT will expire.
In Collabora providing access_token_ttl information triggers a warning for the user that editing will be no longer possible after the expiry (because after that Collabora cannot save the document obviously). 

https://wopi.readthedocs.io/projects/wopirest/en/latest/concepts.html#term-access-token-ttl:
```
To prevent data loss for users, Office for the web will prompt users to save and refresh their sessions if the access token for their session is close to expiring. In order to do this, Office for the web needs to know when the access token will expire, which it determines based on the access_token_ttl value.
```

internal/http/services/appprovider/appprovider.go
glpatcern: That's a very valid point, and the `viewMode` should indeed come from the client (the web UI) exactly like today it is an external parameter passed to the GRPC `open-in-app` command.

wkloucek: The current logic could be used as a fallback if no `viewMode` was given. I originally used the logic to determine maximum permissions of a user with the stat call because we have no possibility to determine the ViewMode in ownCloud Web currently.

internal/http/services/appprovider/appprovider.go
glpatcern: ok, to be exact it's the `min(Reva_JWT_TTL, WOPI_access_token_TTL)`. I must say I had forgotten about this feature of the WOPI protocol, but then the value must be computed on the fly, not taken from the config (yeah, as admitted by the WOPI docs, the name is misleading, it should really be called `access_token_expiration_time`): can we extract the expiration time of a Reva JWT? If so, we can safely assume that the WOPI access_token's validity is never shorter (its default value in the WOPI server config is `24h`) and return the Reva JWT TTL.

internal/http/services/appprovider/appprovider.go
ishank011: Yep, that would be ideal. For now, the idea was to mimic the logic of ocis-wopiserver so we could use it directly with web. We also need to discuss changing the endpoint

internal/http/services/appprovider/appprovider.go
wkloucek: yeah of course we can take the `exp` property from the WOPI jwt access token (if we do so without validating the WOPI access token, we also don't need the WOPI jwt secret in here, which is fine for only taking the `exp`  from it)

internal/http/services/appprovider/appprovider.go
wkloucek: I also thought about how we could make this more generic, so that ownCloud Web just displays the available apps. Ideally this would be done by a extension (which is shipped with oCIS by default) or even integrated in Web directly. We probably should have a discussion with @kulmann and @dragotin about that topic. But I don't know if it is within the scope of this PR!?

ishank011: Yes that is incorporated in the PR! We had the `FindAppProviders` call which returns all apps for a specific mime type and now we also have a default app configured for each mime type. So we can expose the former to get a list of all apps and then the open call would have the app name as a parameter. Would be good to discuss this at the Friday meeting

pkg/app/provider/wopi/wopi.go
glpatcern: Much better, thanks! I hope this is generic enough for all possible deployments ;-)

ishank011:
Tested it with the wopiserver and everything works.

There's one workaround that I had to use. We expect the app provider to register itself to the app registry, but it can't happen in the `New` method as the rgrpc package first retrieves these methods of all the registered services and then starts the GRPC server. So I start a goroutine from the `New` method of the app provider service which waits for a couple of seconds to  send the `AddAppProvider` call, assuming that in this time, the server has been started (this can be made configurable). Thoughts? @labkode @glpatcern

https://github.com/cs3org/reva/blob/de61bcc4ad28480f46be622bdeeee25530e367eb/internal/grpc/services/appprovider/appprovider.go#L87-L106

ishank011:
The checks seemed to have disappeared for the PR, weird. Force pushing didn't work either

wkloucek:
> The checks seemed to have disappeared for the PR, weird. Force pushing didn't work either

@ishank011  https://www.githubstatus.com/ Webhook status is degraded

ishank011:
Ah okay. Thanks @wkloucek!

internal/http/services/appprovider/appprovider.go
glpatcern: If we have the expiration time of the REVA jwt access token, that's more than enough - we can assume the WOPI jwt access token is issued with a _longer_ expiration in all cases - at least it should be configured like that.

internal/grpc/services/gateway/appregistry.go
```diff
@@ -38,7 +38,24 @@ func (s *svc) GetAppProviders(ctx context.Context, req *registry.GetAppProviders
...
+
+	return res, nil
+}
+
+func (s *svc) AddAppProvider(ctx context.Context, req *registry.AddAppProviderRequest) (*registry.AddAppProviderResponse, error) {
+	c, err := pool.GetAppRegistryClient(s.c.AppRegistryEndpoint)
+	if err != nil {
+		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
+		return &registry.AddAppProviderResponse{
+			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
```
wkloucek, line 52: should be "error getting app registry client"!?

internal/grpc/services/gateway/appregistry.go
```diff
@@ -55,7 +72,41 @@ func (s *svc) ListAppProviders(ctx context.Context, req *registry.ListAppProvide
...
+
+	return res, nil
+}
+
+func (s *svc) GetDefaultAppProviderForMimeType(ctx context.Context, req *registry.GetDefaultAppProviderForMimeTypeRequest) (*registry.GetDefaultAppProviderForMimeTypeResponse, error) {
+	c, err := pool.GetAppRegistryClient(s.c.AppRegistryEndpoint)
+	if err != nil {
+		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
+		return &registry.GetDefaultAppProviderForMimeTypeResponse{
+			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
```
wkloucek, line 86: should be "error getting app registry client"!?

internal/grpc/services/gateway/appregistry.go
```diff
@@ -55,7 +72,41 @@ func (s *svc) ListAppProviders(ctx context.Context, req *registry.ListAppProvide
...
+
+	return res, nil
+}
+
+func (s *svc) SetDefaultAppProviderForMimeType(ctx context.Context, req *registry.SetDefaultAppProviderForMimeTypeRequest) (*registry.SetDefaultAppProviderForMimeTypeResponse, error) {
+	c, err := pool.GetAppRegistryClient(s.c.AppRegistryEndpoint)
+	if err != nil {
+		err = errors.Wrap(err, "gateway: error calling GetAppRegistryClient")
+		return &registry.SetDefaultAppProviderForMimeTypeResponse{
+			Status: status.NewInternal(ctx, err, "error getting user share provider client"),
```
wkloucek, line 103: should be "error getting app registry client"!?

internal/http/services/appprovider/appprovider.go
wkloucek: I also think that this part is too WOPI specific in order to be the http variant of the GRPC `OpenInAppRequest`. The input and output should match the GRPC counterpart.

internal/http/services/appprovider/appprovider.go
```diff
@@ -0,0 +1,232 @@
...
+
+func init() {
+	global.Register("appprovider", New)
+}
+
+// Config holds the config options that need to be passed down to all ocdav handlers
+type Config struct {
+	Prefix         string `mapstructure:"prefix"`
+	GatewaySvc     string `mapstructure:"gatewaysvc"`
+	AccessTokenTTL int    `mapstructure:"access_token_ttl"`
```
wkloucek, line 52: `AccessTokenTTL` is not really used for minting a token with that lifetime. We should use the actual lifetime of the token.

pkg/app/registry/static/static.go
```diff
@@ -22,39 +22,84 @@ import (
...
 )
 
-type registry struct {
-	rules map[string]string
+func init() {
+	registry.Register("static", New)
+}
+
+type config struct {
+	Providers map[string]*registrypb.ProviderInfo `mapstructure:"providers"`
```
wkloucek, line 38: @ishank011 I don't fully get the strategy how to get a list of supported mime types by client applications (not by the app provider itself). With the current implementation I would need to register an app provider multiple times (depending on the count of client applications) and change the description of each instance to the client application's name.

Example: 
The app provider "wopi" supports opening files in different client applications: "Collabora", "Microsoft Office Online Server", "CodiMD" and "Etherpad". As a user I don't want to see that I can open the file with "wopi" but I want to see if I can open the file in "Collabora", ... And I only want to see the file action if the open action is possible for that client application.

ishank011: Hi @wkloucek. Yes you would need multiple instances of the appprovider, each pointing to a different client app, all of which can be served by the same wopi server though. At initialization, each of the appproviders will call their discovery URLs, retrieve the mime types and register themselves as  "Collabora", "Microsoft Office Online Server", "CodiMD", and so on. From the web app, we can call `FindAppProviders` with the file name, which will give you a list of all apps supporting that mimetype. Then we'll call `OpenInApp` with that particular app name.

We need to finalize how the calls from web would look like and for that, we have a meeting scheduled tomorrow

internal/http/services/appprovider/appprovider.go
ishank011: Yes, but I don't want to add the token dismantling logic to an HTTP service. It's being done in multiple places in ocis and I would like to get rid of that there as well. I'll try to think of changes in the CS3APIs definitions through which this can be passed to this layer.

wkloucek: Sounds reasonable :-) I don't like this as well

pkg/app/registry/static/static.go
wkloucek: We said, that we first will go with an external extension for Web and later try to move it in to the Web repository. 

Currently there are limitations what a Web extension can do [when registering file handlers](https://github.com/owncloud/ocis-wopiserver/blob/main/ui/app.js), therefore I already wrote down requirements which also cover the ones for the AppProvider integration: https://github.com/owncloud/web/issues/5135

So basically we are currently limited by the topics mentioned in this issue.

In my opinion we need some simple map like this in the frondent:

``` json
{
  "application/vnd.oasis.opendocument.text": {
    "appprovider": [
      { "provider": "collabora.example.test", "displayname": "Collabora" },
      { "provider": "onlyoffice.example.test", "displayname": "Only Office" }
    ]
  },
  "application/vnd.oasis.opendocument.spreadsheet": {
    "appprovider": [
      { "provider": "collabora.example.test", "displayname": "Collabora" },
      { "provider": "onlyoffice.example.test", "displayname": "Only Office" }
    ]
  }
}
```

But we should consider to add more information to it, since some appproviders might support opening files from x while other do not. Shares, Federated Shares, ... come into my mind when thinking about that.
Also some appproviders might support multiple modes to open a file: open in read-write-mode and open in read-only-mode.

ishank011: So with the new workflow, we won’t need to list these extensions statically. There’s an endpoint called GetStorageProvides which can be exposed through a REST layer which will be called for each file and the list of supported apps should be populated from that response. And the response is similar to what you mentioned but instead of returning providers for all mime types, we return only those for the passed file.

Not opening shares is very reva specific so I don’t think it’ll fit in CS3APIs. That logic can be implemented in the app drivers and we can return the appropriate status code.

But the option to restrict view modes in apps itself makes a lot of sense. I’ll add it to the PR

labkode (approved):

labkode:
Merging as we discussed 1 week ago to move forward.

go.mod
```diff
@@ -68,6 +69,7 @@ require (
 go 1.16
 
 replace (
+	github.com/cs3org/go-cs3apis => github.com/ishank011/go-cs3apis v0.0.0-20210715114809-34729f68a479
```
wkloucek, line 72: @ishank011 @labkode this made it to master, could you please take care to remove it?

ishank011: @wkloucek yep, created a PR

PR #1798, opened 2021-06-16
Implement support for deny ACLs

### LINKED ISSUE #1792: Support full denial as permission
Following discussions with @ishank011, we're going to add support for full denials without changing the CS3 APIs, that is within the semantic already available in:

https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.ResourcePermissions

The idea is to consider the empty permission set (value = `0`) as a valid `None` permission, and make it work its way through the drivers and in the eosbinary client. Work in progress by @gmgigi96 and myself in https://github.com/glpatcern/reva/tree/denyacl

Closes https://github.com/cs3org/cs3apis/issues/129

glpatcern:
This is still work in progress, but we thought to expose it earlier to gather feedback. Fixes #1792.

~~The important aspect especially for ownCloud is that now an empty permission set (equivalent to `''` in WebDAV terms, and `0` for the integer representation) is considered valid. So the introduced change does not break the existing APIs but does represent a breaking change from a semantic point of view.~~

EDIT: to avoid breaking any existing assumption, the denial permission gets for now a different bit, `PermissionNone`. `PermissionInvalid` is still defined as `0`.


pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -258,6 +258,31 @@ func (c *Client) executeEOS(ctx context.Context, cmd *exec.Cmd) (string, string,
 	return outBuf.String(), errBuf.String(), err
 }
 
+func (c *Client) SetACLs(ctx context.Context, uid, gid, path string, acls []*acl.Entry) error {
```
labkode, line 261: make private and check for denial in AddACL

pkg/eosclient/eosclient.go
```diff
@@ -27,6 +27,7 @@ import (
 
 // EOSClient is the interface which enables access to EOS instances through various interfaces.
 type EOSClient interface {
+	SetACLs(ctx context.Context, uid, gid, path string, acls []*acl.Entry) error
```
labkode, line 30: make it private, no need to be in interface

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -224,6 +224,10 @@ func (c *Client) initMDRequest(uid, gid string) (*erpc.MDRequest, error) {
 	return mdrq, nil
 }
 
+func (c *Client) SetACLs(ctx context.Context, uid, gid, path string, acls []*acl.Entry) error {
```
labkode, line 227: same here, private

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -40,19 +40,21 @@ const (
 	PermissionShare
 	// PermissionAll grants all permissions on a resource
 	PermissionAll Permissions = (1 << (iota - 1)) - 1
+	// PermissionNone grants no permissions on a resource
+	PermissionNone = PermissionAll + 1
```
labkode, line 44: `PermissionNone = 1 << PermisionAll`
provides a safer namespace

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -427,29 +452,63 @@ func (fs *eosfs) AddGrant(ctx context.Context, ref *provider.Reference, g *provi
 
 	fn := fs.wrap(ctx, p)
```
labkode, line 454: Add here an if condition that behaves like this:

IF Grant is Deny:
    then do the logic below with `setAcls`
 ELSE 
     do the logic as before with `AddACL`

internal/http/services/owncloud/ocs/conversions/permissions.go
gmgigi96: Thanks for the review @labkode.
I think using my implementation is sufficient to ensure that all the bits of the previous permissions are zero, the only 1 bit is the one used only in PermissionNode.
It should work also for 1 << PermissionAll, but now PermissionAll = 31, so is a shift of 31 bits!! If you only add 2 more permissions you will overflow...

glpatcern: I think the idea is to leave some "space" for potential future regular permissions, which would still be included in `PermissionAll` "before" `PermissioNone`. But yes, shifting it by 31 bits is way too much. Yet, we'd need to define `PermissionNone` _independently_ from `PermissionAll`, otherwise it would get redefined if another permission is added.

Is it an option to break now (but then be safe for the future) all permissions by having `PermissionNone = 1` and the others following? As well as defining `PermissionAll = (1 << (iota - 1)) - 2` i.e. all bits set except the None?

If that's not possible, we're left with e.g.:
`PermissionNone = 1 << 10     # hardcoded, 5 bits away from the currently used bits`
and the checks for valid values cannot use a `PermissionRead, PermissionNone` range as many values are undefined in between.

glpatcern: Better yet:
```
PermissionShare
PermissionNone
PermissionMax Permissions = (1 << (iota - 1)) - 1
PermissionAll = PermissionMax - PermissionNone
```
Current values are all preserved, and validity checks can be based on the range `PermissionRead, PermissionMax`. `PermissionMax` is not to be used outside range checks.

pkg/storage/utils/eosfs/eosfs.go
labkode: @gmgigi96 can you add this logic?

labkode:
@gmgigi96 take a look at my comment about `setAcls`

pkg/storage/utils/eosfs/eosfs.go
gmgigi96: do we need it? for now is only for testing, when the --priority functionality will be implemented in EOS, we will use directly and only AddACL

labkode: Yes we need it like that we can merge this PR and not have a big one.
I've started already to add, will push by end of day. 
I take care of this one

glpatcern: And what about `GetOwners()`? Do we merge it as is for now?

gmgigi96: @labkode ok, didn't know i wanted to merge right now :)

cmd/reva/common.go
```diff
@@ -30,8 +30,11 @@ import (
 )
 
 const (
-	viewerPermission string = "viewer"
-	editorPermission string = "editor"
+	viewerPermission  string = "viewer"
```
glpatcern, line 33: Let's review this @labkode and in case we postpone to another PR. But the `coowner` role is good to have now.

labkode:
@glpatcern @gmgigi96 I'm looking for a better alternative for `GetOwners`.
This new method sets a new dependency with the gateway service that was not there before.
With this change the storage providers needs to know where the gateway service is to query group information.
This means to also add this configuration change to our deployment.

While I'm looking how to add this logic into gateway layer reusing the permission set from the FileInfo and pass it down from the gateway to the storage provider you can create a PR towards https://github.com/cernbox/reva@qa so we can deploy the change in our QA infra.

labkode:
Superseded by #1856

PR #1776, opened 2021-06-08
[WIP] introduce /dav/spaces endpoint && CS3 list spaces implementation (ocis driver only)

butonic:
rebase of https://github.com/cs3org/reva/pull/1678 on top of the new-references branch

PR #1802, opened 2021-06-18
[spaces 1/2] initial ListStorageSpaces implementation

butonic:
I split https://github.com/cs3org/reva/pull/1776 into two PRs:
1. the initial list spaces implementation
2. the new [/dav/spaces endpoint](https://github.com/cs3org/reva/pull/1803) to access storage with relative references

This is the first PR about spaces. It
- adds logic to the gateway to find or query the known storage providers and deduplicate the results
  - we currently have a storage provider mounted at `/home` and `/users`, both will return the the same spaces
  - the registry currently does not set the `ProviderId`, only the `ProviderPath`, which we use in a hacky way to ignore all storage providers that start with  a `/`, eg. `/public`, `/home`, etc. We then query only the remaining ones, which have a storage id set and can be routed to when using the `ResourceId` of a relative `Reference`.
- adds a `ListStorageSpaces` implementation for the decomposedfs
- adds `ListStorageSpaces` stubs for the other storage providers

### DIFF
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -115,37 +115,118 @@ func (s *svc) CreateStorageSpace(ctx context.Context, req *provider.CreateStorag
 
 func (s *svc) ListStorageSpaces(ctx context.Context, req *provider.ListStorageSpacesRequest) (*provider.ListStorageSpacesResponse, error) {
 	log := appctx.GetLogger(ctx)
-	// TODO: needs to be fixed
 	var id *provider.StorageSpaceId
 	for _, f := range req.Filters {
 		if f.Type == provider.ListStorageSpacesRequest_Filter_TYPE_ID {
 			id = f.GetId()
 		}
 	}
-	parts := strings.SplitN(id.OpaqueId, "!", 2)
-	if len(parts) != 2 {
-		return &provider.ListStorageSpacesResponse{
-			Status: status.NewInvalidArg(ctx, "space id must be separated by !"),
-		}, nil
-	}
-	c, err := s.find(ctx, &provider.Reference{ResourceId: &provider.ResourceId{
-		StorageId: parts[0], // FIXME REFERENCE the StorageSpaceId is a storageid + a opaqueid
-		OpaqueId:  parts[1],
-	}})
+
+	var providers []*registry.ProviderInfo
+	var err error
+	c, err := pool.GetStorageRegistryClient(s.c.StorageRegistryEndpoint)
 	if err != nil {
-		return &provider.ListStorageSpacesResponse{
-			Status: status.NewStatusFromErrType(ctx, "error finding path", err),
-		}, nil
+		return nil, errors.Wrap(err, "gateway: error getting storage registry client")
 	}
 
-	res, err := c.ListStorageSpaces(ctx, req)
-	if err != nil {
-		log.Err(err).Msg("gateway: error listing storage space on storage provider")
-		return &provider.ListStorageSpacesResponse{
-			Status: status.NewInternal(ctx, err, "error calling ListStorageSpaces"),
-		}, nil
+	if id != nil {
+		// query that specific story provider
+		parts := strings.SplitN(id.OpaqueId, "!", 2)
+		if len(parts) != 2 {
+			return &provider.ListStorageSpacesResponse{
+				Status: status.NewInvalidArg(ctx, "space id must be separated by !"),
+			}, nil
+		}
+		res, err := c.GetStorageProviders(ctx, &registry.GetStorageProvidersRequest{
+			Ref: &provider.Reference{ResourceId: &provider.ResourceId{
+				StorageId: parts[0], // FIXME REFERENCE the StorageSpaceId is a storageid + an opaqueid
+				OpaqueId:  parts[1],
+			}},
+		})
+		if err != nil {
+			return &provider.ListStorageSpacesResponse{
+				Status: status.NewStatusFromErrType(ctx, "ListStorageSpaces filters: req "+req.String(), err),
+			}, nil
+		}
+		if res.Status.Code != rpc.Code_CODE_OK {
+			return &provider.ListStorageSpacesResponse{
+				Status: res.Status,
+			}, nil
+		}
+		providers = res.Providers
+	} else {
+		// get list of all storage providers
+		res, err := c.ListStorageProviders(ctx, &registry.ListStorageProvidersRequest{})
+
+		if err != nil {
+			return &provider.ListStorageSpacesResponse{
+				Status: status.NewStatusFromErrType(ctx, "error listing providers", err),
+			}, nil
+		}
+		if res.Status.Code != rpc.Code_CODE_OK {
+			return &provider.ListStorageSpacesResponse{
+				Status: res.Status,
+			}, nil
+		}
+
+		providers = []*registry.ProviderInfo{}
+		// FIXME filter only providers that have an id set ... currently none have?
+		// bug? only ProviderPath is set
+		for i := range res.Providers {
+			// use only providers whose path does not start with a /?
+			if strings.HasPrefix(res.Providers[i].ProviderPath, "/") {
+				continue
+			}
+			providers = append(providers, res.Providers[i])
+		}
 	}
-	return res, nil
+
+	spacesFromProviders := make([][]*provider.StorageSpace, len(providers))
+	errors := make([]error, len(providers))
+	var wg sync.WaitGroup
+
+	for i, p := range providers {
+		wg.Add(1)
+		go s.listStorageSpacesOnProvider(ctx, req, &spacesFromProviders[i], p, &errors[i], &wg)
+	}
+	wg.Wait()
+
+	uniqueSpaces := map[string]*provider.StorageSpace{}
+	for i := range providers {
+		if errors[i] != nil {
+			log.Debug().Err(errors[i]).Msg("skipping provider")
+			continue
+		}
+		for j := range spacesFromProviders[i] {
+			uniqueSpaces[spacesFromProviders[i][j].Id.OpaqueId] = spacesFromProviders[i][j]
+		}
+	}
+	spaces := []*provider.StorageSpace{}
+	for spaceID := range uniqueSpaces {
+		spaces = append(spaces, uniqueSpaces[spaceID])
+	}
+
+	return &provider.ListStorageSpacesResponse{
+		Status:        status.NewOK(ctx),
+		StorageSpaces: spaces,
+	}, nil
+}
+
+func (s *svc) listStorageSpacesOnProvider(ctx context.Context, req *provider.ListStorageSpacesRequest, res *[]*provider.StorageSpace, p *registry.ProviderInfo, e *error, wg *sync.WaitGroup) {
+	defer wg.Done()
+	c, err := s.getStorageProviderClient(ctx, p)
+	if err != nil {
+		*e = errors.Wrap(err, "error connecting to storage provider="+p.Address)
+		return
+	}
+
+	r, err := c.ListStorageSpaces(ctx, req)
+	if err != nil {
+		*e = errors.Wrap(err, "gateway: error calling ListStorageSpaces")
+		return
+	}
+
+	*res = r.StorageSpaces
 }
 
 func (s *svc) UpdateStorageSpace(ctx context.Context, req *provider.UpdateStorageSpaceRequest) (*provider.UpdateStorageSpaceResponse, error) {
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -122,15 +122,17 @@ func (s *svc) ListStorageSpaces(ctx context.Context, req *provider.ListStorageSp
 		}
 	}
 
-	var providers []*registry.ProviderInfo
-	var err error
+	var (
+		providers []*registry.ProviderInfo
+		err       error
+	)
 	c, err := pool.GetStorageRegistryClient(s.c.StorageRegistryEndpoint)
 	if err != nil {
 		return nil, errors.Wrap(err, "gateway: error getting storage registry client")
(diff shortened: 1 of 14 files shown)


pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -465,6 +492,162 @@ func (fs *Decomposedfs) Download(ctx context.Context, ref *provider.Reference) (
...
+
+	// build the glob path, eg.
+	// /path/to/root/spaces/personal/nodeid
+	// /path/to/root/spaces/shared/nodeid
+	matches, err := filepath.Glob(filepath.Join(fs.o.Root, "spaces", spaceType, spaceID))
+	if err != nil {
+		return nil, err
+	}
+
+	var spaces []*provider.StorageSpace
```
C0rby, line 533: This could also be pre-allocated.

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -465,6 +492,162 @@ func (fs *Decomposedfs) Download(ctx context.Context, ref *provider.Reference) (
...
+
+			// quota
+			v, err := xattr.Get(matches[i], xattrs.QuotaAttr)
+			if err == nil {
+				// make sure we have a proper signed int
+				// we use the same magic numbers to indicate:
+				// -1 = uncalculated
+				// -2 = unknown
+				// -3 = unlimited
+				if quota, err := strconv.ParseInt(string(v), 10, 64); err == nil {
```
C0rby, line 631: `strconv.ParseInt(string(v), 10, 64)` can be replaced with `strconv.Itoa(string(v))` IIRC.

C0rby (approved):
Looks good. Just some small remarks.

C0rby (dismissed):

C0rby:
@butonic, can you restart the CI build or do I need `git push` something?

butonic:
@C0rby I can't, pushing should trigger it automatically.

refs (changes requested):
I would like for the Reva code base to be more approachable and part of it is to start being consistent. During this review I had this in mind:

- focus in consistency [(i.e `var` blocks vs inline declarations)](https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations)
- special emphasis on reducing the cyclomatic complexity of certain functions
  - this includes:
    - refactoring
    - abstracting function bodies away (not refactoring per-se...)
- clarity of language
  - this refers to comments as well as blocks of logic

While I was probably not 100% enforcing all this points, the areas I reviewed I believe serve as a good example where most of our logic goes awry. Just because we have a verbose API doesn't mean our blocks have to be verbose 🚀.

This review is by no means a blocker, but I would find value if the points are discussed collectively in the public chat.

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -115,37 +115,118 @@ func (s *svc) CreateStorageSpace(ctx context.Context, req *provider.CreateStorag
...
+
+	for i, p := range providers {
+		wg.Add(1)
+		go s.listStorageSpacesOnProvider(ctx, req, &spacesFromProviders[i], p, &errors[i], &wg)
+	}
+	wg.Wait()
+
+	uniqueSpaces := map[string]*provider.StorageSpace{}
+	for i := range providers {
+		if errors[i] != nil {
```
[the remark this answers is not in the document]
butonic: when a provider retuns an unsupported error we should continue with the others. expect failure.

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -115,37 +115,118 @@ func (s *svc) CreateStorageSpace(ctx context.Context, req *provider.CreateStorag
...
+			log.Debug().Err(errors[i]).Msg("skipping provider")
+			continue
+		}
+		for j := range spacesFromProviders[i] {
+			uniqueSpaces[spacesFromProviders[i][j].Id.OpaqueId] = spacesFromProviders[i][j]
```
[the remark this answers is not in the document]
butonic: actually it does dedupe: the home and user storage provider both return the same storage id ... we only need it once

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -428,9 +428,45 @@ func (s *service) CreateStorageSpace(ctx context.Context, req *provider.CreateSt
...
+		var st *rpc.Status
+		switch err.(type) {
+		case errtypes.IsNotFound:
+			st = status.NewNotFound(ctx, "not found when listing spaces")
+		case errtypes.PermissionDenied:
+			st = status.NewPermissionDenied(ctx, err, "permission denied")
+		case errtypes.NotSupported:
+			st = status.NewUnimplemented(ctx, err, "not implemented")
+		default:
+			st = status.NewInternal(ctx, err, "error listing spaces")
```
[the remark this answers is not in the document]
butonic: because the other cases are known errors. this is a catchall in case it is none of the above errors. These huge error handling blocks are ugly and I hope they can be refactored ... in a dedicated PR.

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -428,9 +428,45 @@ func (s *service) CreateStorageSpace(ctx context.Context, req *provider.CreateSt
...
+		return &provider.ListStorageSpacesResponse{
+			Status: st,
+		}, nil
+	}
+
+	for i := range spaces {
+		if hasNodeID(spaces[i]) {
+			// fill in storagespace id if it is not set
+			if spaces[i].Id == nil || spaces[i].Id.OpaqueId == "" {
+				spaces[i].Id = &provider.StorageSpaceId{OpaqueId: s.mountID + "!" + spaces[i].Root.OpaqueId}
```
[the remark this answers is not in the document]
butonic: which would make migration scenarios harder...

pkg/storage/registry/static/static.go
```diff
@@ -145,18 +145,25 @@ func (b *reg) FindProviders(ctx context.Context, ref *provider.Reference) ([]*re
...
+				}
 			}
-			// TODO(labkode): fill path info based on provider id, if path and storage id points to same id, take that.
-			if m := r.FindString(ref.ResourceId.StorageId); m != "" {
-				return []*registrypb.ProviderInfo{{
-					ProviderId: ref.ResourceId.StorageId,
-					Address:    addr,
-				}}, nil
+			// TODO if the storage id is not set but node id is set we could poll all storage providers to check if the node is known there
+			// for now, say the reference is invalid
```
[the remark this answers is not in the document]
butonic: It would be more interesting to get this to work with eos first ;-)

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -210,9 +215,31 @@ func (fs *Decomposedfs) CreateHome(ctx context.Context) (err error) {
...
+		return err
+	}
+
 	return
 }
 
+func (fs *Decomposedfs) createStorageSpace(spaceType, nodeID string) error {
+
+	// create space type dir
+	if err := os.MkdirAll(filepath.Join(fs.o.Root, "spaces", spaceType), 0700); err != nil {
```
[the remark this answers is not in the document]
butonic: it is as magic as the "nodes", "blobs", "trash" and "uploads" folders. ... but yeah ...

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -210,9 +215,31 @@ func (fs *Decomposedfs) CreateHome(ctx context.Context) (err error) {
...
+		return err
+	}
+
+	// we can reuse the node id as the space id
+	err := os.Symlink("../../nodes/"+nodeID, filepath.Join(fs.o.Root, "spaces", spaceType, nodeID))
+	if err != nil {
+		fmt.Printf("could not create symlink for personal space %s, %s\n", nodeID, err)
```
butonic, line 237:
```suggestion
		fmt.Printf("could not create symlink for '%s' space %s, %s\n", spaceType, nodeID, err)
```

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -465,6 +492,160 @@ func (fs *Decomposedfs) Download(ctx context.Context, ref *provider.Reference) (
...
+			}
+
+			// fill in user object if the current user is the owner
+			if utils.UserEqual(u.Id, owner) {
+				space.Owner = u
+			} else {
+				space.Owner = &userv1beta1.User{ // FIXME only return a UserID, not a full blown user object
+					Id: owner,
+				}
+			}
```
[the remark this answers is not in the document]
butonic: `space.Owner` currently expects a complete user object. here we only have the userid so I can only fill in the complete user when it is the owner, *else* I have to wrap the id in an otherwise empty user object.

refs (dismissed):

pkg/storage/utils/decomposedfs/decomposedfs.go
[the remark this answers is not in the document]
refs: yeah I knew you'd come up with this argument 🙈

refs (approved):

butonic:
@labkode @ishank011 this is ready for review

refs (dismissed):

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -428,9 +428,45 @@ func (s *service) CreateStorageSpace(ctx context.Context, req *provider.CreateSt
...
+		default:
+			st = status.NewInternal(ctx, err, "error listing spaces")
+		}
+		return &provider.ListStorageSpacesResponse{
+			Status: st,
+		}, nil
+	}
+
+	for i := range spaces {
+		if hasNodeID(spaces[i]) {
```
ishank011, line 455: If `hasNodeID` returns false and the storage space ID is still not set, should we return an error?

pkg/storage/registry/static/static.go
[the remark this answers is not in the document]
ishank011: Not sure about this one 😅

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -115,37 +115,120 @@ func (s *svc) CreateStorageSpace(ctx context.Context, req *provider.CreateStorag
...
+				Status: res.Status,
+			}, nil
+		}
+
+		providers = make([]*registry.ProviderInfo, 0, len(res.Providers))
+		// FIXME filter only providers that have an id set ... currently none have?
+		// bug? only ProviderPath is set
+		for i := range res.Providers {
+			// use only providers whose path does not start with a /?
+			if strings.HasPrefix(res.Providers[i].ProviderPath, "/") {
```
ishank011, line 179: I don't think this filter is needed. Since you're calling `GetStorageProviders` with the resource ID, their `ProviderPath` will always be an empty string and `ProviderId` will be set

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -210,9 +215,31 @@ func (fs *Decomposedfs) CreateHome(ctx context.Context) (err error) {
 			return
 		}
 	}
+
+	// add storage space
+	if err := fs.createStorageSpace("personal", h.ID); err != nil {
```
ishank011, line 220: `createStorageSpace` is called at the time of home and share creation and in `Tree.Setup`, you create spaces for existing nodes, but these won't be created when created directories and uploading files.

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -465,6 +492,159 @@ func (fs *Decomposedfs) Download(ctx context.Context, ref *provider.Reference) (
...
+				SpaceType: filepath.Base(filepath.Dir(matches[i])),
+				// Mtime is set either as node.tmtime or as fi.mtime below
+			}
+
+			if space.SpaceType == "share" {
+				if utils.UserEqual(u.Id, owner) {
+					// do not list shares as spaces for the owner
+					continue
+				}
+				space.Name = n.Name
```
ishank011, line 580: `space.Name = n.Name` this will already be set previously

internal/grpc/services/gateway/storageprovider.go
butonic: In this else branch I have to use `ListStorageProviders` because the request contains no filter properties. The `GetStorageProviders` is used when there is a filter in line 142 above.

pkg/storage/utils/decomposedfs/decomposedfs.go
butonic: That works as designed. A new directory is not automatically a new storage space. Neither is a file. They don't contain any new sharing information (ACLs) or a quota. Maybe we need to clarify the meaning of a space.

`Tree.Setup` checks if the spaces folder exists and if not will pick up all user homes and shared nodes to create `personal` and `share` type spaces. We do this to migrate an existing layout to now also keep track of spaces.

I consider `CreateHome` a legacy method to provision user homes. IMO the [admin `CreateUser` api](https://cs3org.github.io/cs3apis/#cs3.admin.user.v1beta1.CreateUserRequest) should, as part of creating the user, provision the users home or `personal` user spaces using the yet to implement `CreateStorageSpace` call. But that is part of a subsequent PR. For now we just want to make the different types of entrypoints (user homes, shares) discoverable as spaces.

other types could be `project`, `music` or `archive` ... far future though ... we'll see how it evolves.

pkg/storage/utils/decomposedfs/decomposedfs.go
butonic: good catch!

pkg/storage/utils/decomposedfs/decomposedfs.go
ishank011: Ah okay. For some reason, I thought each resource (or node in the ocis fs) is considered a space

ishank011 (approved):

PR #1803, opened 2021-06-18
[spaces 2/2] /dav/spaces endpoint

butonic:
I split https://github.com/cs3org/reva/pull/1776 into two PRs:
1. the initial list spaces implementation
2. the new /dav/spaces endpoint to access storage with relative references

This is the second PR about spaces. It
- adds a new ocdav handler for `/dav/spaces`
- refactors the ocdav handlers to make them reusable for `/dav/spaces` and `/(web)dav` requests
- adds logic to the gateway to forward relative `References` to the storage providers
- adds logic to the storageprovider to only prefix mountpoints for absolute `References`
- changes the `CreateDir` signature in the `storage.FS` interface to allow creating folders with relative `References`
- adds a new datatx spaces implementation
- update storage drivers to distinguish relative vs absolute `References` and at least handle absolute references for path and id only based access

## TODO
- [ ] move spaces migration code in decomposedfs tree to lis spaces PR
- [ ] revert storageprovider wrap returning ctx, no longer needed


ishank011:
@butonic can you also add sample expected requests and responses in the description once this PR is finished?

pkg/rhttp/datatx/utils/download/download.go
```diff
@@ -55,10 +54,10 @@ func GetOrHeadFile(w http.ResponseWriter, r *http.Request, fs storage.FS, spaceI
 		ref = &provider.Reference{Path: path.Join("/", fn)}
 	} else {
 		// build a storage space reference
-		parts := strings.SplitN(spaceID, "!", 2)
-		if len(parts) == 2 {
+		storageid, opaqeid, err := utils.SplitStorageSpaceID(spaceID)
+		if err != nil {
```
C0rby, line 58: I think this should be reversed: `if err == nil {`, right?

butonic:
@ishank011 I rebased, please review

ishank011:
@butonic I'm not sure about having the new protocol in `datatx` since it's not exactly a protocol. I think we can add the relavant checks in tus/simple and frame the requests accordingly. Also, won't spaces support TUS?

butonic:
I need to look into the spaces and TUS protocol. That should work.

butonic:
tus works as expected against the /dav/spaces endpoint. The upload URL is wrapped inside the datagateway token.

butonic:
- [ ] explain difference between `simple` and `spaces` protocol: should they be separate or can we combine them?

butonic:
@ish

> @butonic I'm not sure about having the new protocol in `datatx` since it's not exactly a protocol. I think we can add the relavant checks in tus/simple and frame the requests accordingly. Also, won't spaces support TUS?

How do we distinguish a `simple` request from a `spaces` request when the only difference is that for a `spaces` request the first path segment is the space id? The `simple` handler cannot just always take the first path segment as the space id. That is why we added a new datatx 'protocol': The path in the URL schema carries a different meaning.

C0rby (approved):

labkode:
@C0rby @butonic we need to address the LGTM errors, we are not sanitising the path, let's fix that.

C0rby:
@labkode, the lgtm-com comment says __This pull request fixes 30 alerts when merging__. Are there more alerts for this PR?
I find the lgtm UI a bit confusing so I might be missing some.

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -491,14 +500,42 @@ func (s *service) DeleteStorageSpace(ctx context.Context, req *provider.DeleteSt
...
+		}
+		parentRef = &provider.Reference{Path: path.Dir(ref.GetPath())}
+		name = path.Base(ref.GetPath())
+	default:
+		return &provider.CreateContainerResponse{
+			Status: status.NewInvalidArg(ctx, "invalid reference, name required"),
+		}, nil
+	}
+	var st *rpc.Status
 	if err != nil {
```
ishank011, line 525: `err` will always be nil here. We already check the error returned by `unwrap` in the switch case and return, and it is assigned to a local var anyway; so we can remove this block

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -573,9 +610,14 @@ func (s *service) Move(ctx context.Context, req *provider.MoveRequest) (*provide
 	}
 	targetRef, err := s.unwrap(ctx, req.Destination)
 	if err != nil {
-		return &provider.MoveResponse{
-			Status: status.NewInternal(ctx, err, "error unwrapping destination path"),
-		}, nil
+		switch err.(type) {
+		case errtypes.IsNotFound:
```
ishank011, line 614: `unwrap` always returns `errtypes.BadRequest` so this will never be true

pkg/auth/scope/publicshare.go
```diff
@@ -71,7 +71,8 @@ func publicshareScope(scope *authpb.Scope, resource interface{}) (bool, error) {
 }
 
 func checkStorageRef(s *link.PublicShare, r *provider.Reference) bool {
-	// r: <id:<storage_id:$storageID node_id:$nodeID path:$path > >
+	// TODO @ishank011 con you explain how this is used?
```
ishank011, line 74: For every incoming request of the above types, we extract the `Reference` object and go through all the objects that are in the current scope to see if access can be allowed to that particular reference. For example, we create a token with access to a particular public share and we store the resource ID that share points to.

pkg/auth/scope/resourceinfo.go
```diff
@@ -68,10 +68,18 @@ func resourceinfoScope(scope *authpb.Scope, resource interface{}) (bool, error)
...
+	// ref: <resource_id:<storage_id:$storageID opaque_id:$opaqueID path:$path> >
 	if ref.ResourceId != nil { // path can be empty or a relative path
-		// TODO what about the path?
-		return utils.ResourceIDEqual(inf.Id, ref.ResourceId)
+		if inf.Id.StorageId == ref.ResourceId.StorageId && inf.Id.OpaqueId == ref.ResourceId.OpaqueId {
+			if ref.Path == "" {
+				// id only reference
+				return true
+			}
+			// check path has same prefix below
```
ishank011, line 79: This looks good, should work!

ishank011 (commented):
Reviewed the storage provider changes. Will go through ocdav and decomposedfs as well

ishank011:
> How do we distinguish a simple request from a spaces request when the only difference is that for a spaces request the first path segment is the space id? The simple handler cannot just always take the first path segment as the space id. That is why we added a new datatx 'protocol': The path in the URL schema carries a different meaning.

I get the use case but I feel that it should be transparent to the clients. If a GRPC client wants to use relative paths, it should not know about the `spaces` concept, but just keep using it as it does for absolute paths.

Maybe we can add a check in the simple protocol. If the first routing in the URL path is of the type `storage-ID!opaque-ID`, then we frame the request reference accordingly?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1308,14 +1325,18 @@ func (s *svc) statOnProvider(ctx context.Context, req *provider.StatRequest, res
...
 	}
 
-	resPath := path.Clean(req.Ref.GetPath())
-	newPath := req.Ref.GetPath()
-	if resPath != "" && !strings.HasPrefix(resPath, p.ProviderPath) {
-		newPath = p.ProviderPath
+	if utils.IsAbsoluteReference(req.Ref) {
+		resPath := path.Clean(req.Ref.GetPath())
+		newPath := req.Ref.GetPath()
+		if resPath != "." && !strings.HasPrefix(resPath, p.ProviderPath) {
```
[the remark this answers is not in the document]
C0rby: But at this point `resPath` can never be empty because of `path.Clean`. `path.Clean` will return `"."` if `req.Ref.GetPath()` is empty.

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -307,6 +307,11 @@ func (s *svc) getHome(_ context.Context) string {
 
 func (s *svc) InitiateFileDownload(ctx context.Context, req *provider.InitiateFileDownloadRequest) (*gateway.InitiateFileDownloadResponse, error) {
 	log := appctx.GetLogger(ctx)
+
+	if utils.IsRelativeReference(req.Ref) {
```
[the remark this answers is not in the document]
butonic: > > How do we distinguish a simple request from a spaces request when the only difference is that for a spaces request the first path segment is the space id? The simple handler cannot just always take the first path segment as the space id. That is why we added a new datatx 'protocol': The path in the URL schema carries a different meaning.
> 
> I get the use case but I feel that it should be transparent to the clients. If a GRPC client wants to use relative paths, it should not know about the `spaces` concept, but just keep using it as it does for absolute paths.
> 
> Maybe we can add a check in the simple protocol. If the first routing in the URL path is of the type `storage-ID!opaque-ID`, then we frame the request reference accordingly?

Unfortunately, then any path that starts with that pattern will be interpreted as a relative spaces request. What about: `Hey! Cool Pics/2021-08-09/yay I can inject relative path references now.png` 

The only other way to reuse the `simple` datatx would be to use a dedicated path segment in the datatx urls:
- `/simple/absolute/path/to/file.ext`
- `/spaces/storage-ID!opaque-ID/relative/path/to/file.ext`

but it would break existing urls  ... well ... they are recerated whenever making an initiate download/upload call, so it should not be a problem. Would that be ok for you?

butonic:
> > How do we distinguish a simple request from a spaces request when the only difference is that for a spaces request the first path segment is the space id? The simple handler cannot just always take the first path segment as the space id. That is why we added a new datatx 'protocol': The path in the URL schema carries a different meaning.
> 
> I get the use case but I feel that it should be transparent to the clients. If a GRPC client wants to use relative paths, it should not know about the `spaces` concept, but just keep using it as it does for absolute paths.
> 
> Maybe we can add a check in the simple protocol. If the first routing in the URL path is of the type `storage-ID!opaque-ID`, then we frame the request reference accordingly?

Unfortunately, then any path that starts with that pattern will be interpreted as a relative spaces request. What about: `Hey! Cool Pics/2021-08-09/yay I can inject relative path references now.png` 

The only other way to reuse the `simple` datatx would be to use a dedicated path segment in the datatx urls:
- `/simple/absolute/path/to/file.ext`
- `/spaces/storage-ID!opaque-ID/relative/path/to/file.ext`

but it would break existing urls  ... well ... they are recerated whenever making an initiate download/upload call, so it should not be a problem. Would that be ok for you?

internal/grpc/services/gateway/storageprovider.go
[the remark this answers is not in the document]
butonic: Delete and Move both call getPath, which in turn calls stat. Together, they also do the necessary combined and absolute path handling.

pkg/storage/fs/owncloudsql/owncloudsql.go
```diff
@@ -706,7 +709,7 @@ func (fs *owncloudsqlfs) GetHome(ctx context.Context) (string, error) {
 	return "", nil
 }
 
-func (fs *owncloudsqlfs) CreateDir(ctx context.Context, sp string) (err error) {
+func (fs *owncloudsqlfs) CreateDir(ctx context.Context, ref *provider.Reference, sp string) (err error) {
```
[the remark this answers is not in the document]
butonic: Good catch! Actually I did not find a good reason why we changed the signature ... so I reverted that part ... and fixed resolving the reference here.

butonic:
@ishank011 @labkode rebased and adressed review. Good to merge?

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -328,6 +333,26 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	w.WriteHeader(http.StatusNoContent)
 }
 
+func (s *svc) handleSpacesPut(w http.ResponseWriter, r *http.Request, spaceID string) {
+	ctx := r.Context()
```
ishank011, line 337: Start the trace span here as well

ishank011 (commented):
@butonic couple of minor comments, we can merge then.

ishank011 (approved):

PR #1807, opened 2021-06-22
Add a shares storage provider for accessing shares in /home

aduffeck:
This PR implements a new storage provider which exposes all shares the user has access to.

When mounted to `/home/Shares` it makes the magic in the gateway code superfluous. See https://github.com/aduffeck/ocis/commits/sharestorageprovider for an example implementation of that scenario.


ishank011:
@aduffeck @butonic would it also be a good idea to move the shares directory out to its separate mount instead of restricting it inside the home directory in this PR as well? I started on that in https://github.com/cs3org/reva/pull/1584. The tests fail because in ocdav, we prepend `/home` to all incoming requests including `/Shares/...`, so such requests would need to be handled.

aduffeck:
@ishank011 that is exactly what we are currently doing in https://github.com/cs3org/reva/pull/1846 (which kinda supersedes this pr, so I will close this one now). The shares are currently mounted to [/home/Shares](https://github.com/owncloud/ocis/pull/2232/files#diff-6c18da689422bb4bf41ba8801992be636515458ad3af95905a79bc0efeb12508R38) but it could be any other path as well.

PR #1812, opened 2021-06-22
publicshare: check for more errors

labkode:
Now checks for errors when creating the hashes.

Next step will be to pass the ctx around and log when the (unlikely) hash creation fails to give a clue to the admin of the system about authentication failures when accessing a public share.


C0rby (approved):
Looks good to me. :+1:

PR #1816, opened 2021-06-23
[tests-only] Bump core commit id for tests

saw-jan:
Bumped commit ID of `core`
part of https://github.com/owncloud/QA/issues/662

PR #1817, opened 2021-06-23
[Tests-Only]Adjust remaining closed issues from the expected failures

jasson99:


phil-davis:
@jasson99 FYI we cherry-picked these changes into #1816 - so if that gets merged, you can close this PR.

PR #1819, opened 2021-06-23
Remove notifications key from ocs response

refs:


refs:
Again, the semantics are up to the clients to understand what it means to have a `notifications` key; for which we've seen clients (at least web) assumes that if the key is present, do notifications. This logic might be wrong because it could potentially break clients, the issue is that I'm unaware of that. @butonic do you have more input here?

refs:
cc @pascalwengerter is this still relevant enough to bring it up on monday in an impro meeting?

pascalwengerter:
> cc @pascalwengerter is this still relevant enough to bring it up on monday in an impro meeting?

Yes, please! Causes buggy behaviour in web & mobile clients, would be nice to have that streamlined and fixed :)

pascalwengerter:
@labkode can this get some attention soon? Reva incorrectly announcing capabilities causes errors in `web` and other clients

labkode:
@pascalwengerter I was waiting for the confirmation about your meeting. 
Do you need the current PR to be merged?

pascalwengerter:
> @pascalwengerter I was waiting for the confirmation about your meeting.
> Do you need the current PR to be merged?

Could you have a chat with @refs about it? We should get this merged to avoid unnecessary `notifications` requests in web & other clients due to Reva communicating its capabilities in an unexpected way 😉

butonic:
@refs  make this green again, then this can be merged! @pascalwengerter talked to @TheOneRing and the mobile teams: the notifications property should be omitted.

butonic:
then https://github.com/owncloud/ocis/blob/11722a3b4d2934475204c082ad2caaebd4161f2f/storage/pkg/command/frontend.go#L234-L236can go as well!

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/capabilities_test.go
```diff
@@ -35,14 +36,15 @@ func TestMarshal(t *testing.T) {
...
+	jsonExpect := `{"capabilities":{"core":null,"checksums":null,"files":null,"dav":null,"files_sharing":{"api_enabled":true,"resharing":false,"group_sharing":false,"auto_accept_share":false,"share_with_group_members_only":false,"share_with_membership_groups_only":false,"search_min_length":0,"default_permissions":0,"user_enumeration":null,"federation":null,"public":null,"user":null}},"version":null}`
 	xmlExpect := `<CapabilitiesData><capabilities><files_sharing><api_enabled>1</api_enabled><resharing>0</resharing><group_sharing>0</group_sharing><auto_accept_share>0</auto_accept_share><share_with_group_members_only>0</share_with_group_members_only><share_with_membership_groups_only>0</share_with_membership_groups_only><search_min_length>0</search_min_length><default_permissions>0</default_permissions></files_sharing></capabilities></CapabilitiesData>`
 
 	jsonData, err := json.Marshal(&cd)
 	if err != nil {
 		t.Fail()
 	}
 
+	fmt.Println(string(jsonData))
```
labkode, line 47: Probably this is not needed anymore?

labkode:
@refs one minor nit

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/capabilities_test.go
refs: sharp eyes, thanks!

labkode (approved):

PR #1743, opened 2021-05-28
Application passwords CLI commands

gmgigi96:
Related to #1719

This PR add the commands `token-list`, `token-remove` and `token-create` to manage the application passwords.

```
>> token-create -share bvqoPVphGpchsDd:r -label share-home -expiration 2021-05-29
+------------------+-------------------------------------+------------+--------------------------------+--------------------------------+--------------------------------+
| TOKEN            | SCOPE                               | LABEL      | EXPIRATION                     | CREATION TIME                  | LAST USED TIME                 |
+------------------+-------------------------------------+------------+--------------------------------+--------------------------------+--------------------------------+
| 074136IMU2yRGnY5 | share:"bvqoPVphGpchsDd" ROLE_VIEWER | share-home | 2021-05-29 02:00:00 +0200 CEST | 2021-05-28 18:28:55 +0200 CEST | 2021-05-28 18:28:55 +0200 CEST |
+------------------+-------------------------------------+------------+--------------------------------+--------------------------------+--------------------------------+
```
Multiple share and path options are possible:
```
>> token-create -path /home:r -path /home/M:r -share CgAeSOEDmGeBrpc:w -label new
+------------------+-------------------------------------------------------------------------------------------+-------+------------+--------------------------------+--------------------------------+
| TOKEN            | SCOPE                                                                                     | LABEL | EXPIRATION | CREATION TIME                  | LAST USED TIME                 |
+------------------+-------------------------------------------------------------------------------------------+-------+------------+--------------------------------+--------------------------------+
| I9J0bU7geBK28351 | path:"/home" ROLE_VIEWER, path:"/home/M" ROLE_VIEWER, share:"CgAeSOEDmGeBrpc" ROLE_EDITOR | new   |            | 2021-05-28 19:24:12 +0200 CEST | 2021-05-28 19:24:12 +0200 CEST |
+------------------+-------------------------------------------------------------------------------------------+-------+------------+--------------------------------+--------------------------------+
```

```
>> token-list
+--------------------------------------------------------------+------------+-------------------------------------+--------------------------------+
| TOKEN                                                        | LABEL      | SCOPE                               | EXPIRATION                     |
+--------------------------------------------------------------+------------+-------------------------------------+--------------------------------+
[...]

PR #1824, opened 2021-06-24
CS3 pulltransfer

redblom:
New CS3 datatx methods to support pull model. 
See https://github.com/cs3org/cs3apis/pull/134

### DIFF
--- a/changelog/unreleased/pull-transfer.md
+++ b/changelog/unreleased/pull-transfer.md
@@ -1,0 +1,4 @@
+Enhancement: New CS3API datatx methods
+
+CS3 datatx pull model methods: PullTransfer, RetryTransfer, ListTransfers
+Method CreateTransfer removed.
--- a/changelog/unreleased/pull-transfer.md
+++ b/changelog/unreleased/pull-transfer.md
@@ -2,3 +2,5 @@ Enhancement: New CS3API datatx methods
 
 CS3 datatx pull model methods: PullTransfer, RetryTransfer, ListTransfers
 Method CreateTransfer removed.
+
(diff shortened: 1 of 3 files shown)


ishank011:
@redblom do we have the complete workflow now? Did you run any end-to-end tests?

redblom:
@ishank011 : end-to-end create/accept share tests are successful (both regular and tx shares)

ishank011:
I don't see the implementation? Can you create a single PR for the changes? And add the command usage and workflow to the description? (Something like https://github.com/cs3org/reva/pull/1743#issue-656417871)

redblom:
@ishank011 : I'm struggling to get the dest webdav endpoint at service.datatx level that's needed for the rclone driver call. There's no access there to the gateway funcs.
Should I inject gateway access or ... inject the dest webdav endpoint?

Or maybe the CS3 PullTransferRequest is simply lacking the destination targetURI ... so maybe it should have targetURIs for both src and dest ?

redblom:
@ishank011 : the home dir is part of the src file path, eg. `/home/mybigfiles/files-to-transfer ...`
Upon transfer the files will then be uploaded at dest to `/home/Data-Transfers/home/mybigfiles/files-to-transfer ...`

To me this seems a bit odd. I'm actually wondering if a home dir should be part of a share path at all ?

ishank011:
Yep you just take the base path from the target and prepend the configurable data-transfers directory to get the full destination path.

redblom:
Superseded by https://github.com/cs3org/reva/pull/2052

PR #1828, opened 2021-06-27
Add a proper XML header to error replies.

dragotin:
See https://github.com/owncloud/product/issues/273 for more details.

### DIFF
--- a/internal/http/services/owncloud/ocdav/error.go
+++ b/internal/http/services/owncloud/ocdav/error.go
@@ -58,13 +58,17 @@ type exception struct {
 
 // Marshal just calls the xml marshaller for a given exception.
 func Marshal(e exception) ([]byte, error) {
-	return xml.Marshal(&errorXML{
+	xmlstring, err := xml.Marshal(&errorXML{
 		Xmlnsd:    "DAV",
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocdav/error.go
```diff
@@ -58,13 +58,17 @@ type exception struct {
...
 func Marshal(e exception) ([]byte, error) {
-	return xml.Marshal(&errorXML{
+	xmlstring, err := xml.Marshal(&errorXML{
 		Xmlnsd:    "DAV",
 		Xmlnss:    "http://sabredav.org/ns",
 		Exception: codesEnum[e.code],
 		Message:   e.message,
 		Header:    e.header,
 	})
+	if err == nil {
```
ishank011, line 68: @dragotin can you reverse the check? 😅
`if err != nil` and return `xmlstring` as the default? It's more go-idiomatic.

dragotin: Thanks for pushing me the right direction - changed!

ishank011: Thanks!

ishank011 (approved):

PR #1842, opened 2021-06-29
app provider: allow insecure connection to wopiserver / wopibridge

wkloucek:


glpatcern:
@wkloucek we could fold this in https://github.com/cs3org/reva/pull/1785 as the whole appprovider is being refactored and extended

wkloucek:
> @wkloucek we could fold this in #1785 as the whole appprovider is being refactored and extended

That would be fine for me. I needed it for as a intermediate step for owncloud/ocis#2204 and owncloud/ocis-wopiserver#16 and if I don't open a PR in time I will forget about it later... So feel free to close this one at any time in favor of yours ;-)

ishank011:
Implemented by #1785

PR #1849, opened 2021-06-30
Add read-only interceptor to use on storage providers

micbar:
# Description

For migration scenarios it can be handy to use a storage provider in read-only mode. For example if we use the ownCloud SQL driver to connect to a ownCloud Classic, readonly mode is a vital step in the migration process.

## Changes

- [x] Implement `readonly` Interceptor
- [x] Add proper WebDAV Response body also in some error cases

## How it works

- The `read-only` interceptor uses a list of known request types which are allowed.
- Known request types which change data on the storage are blocked and return a grpc error which will be forwarded to WebDAV.
- Unknown Request Types will be blocked and throw a GRPC error
- The interceptor changes the grants on the resources which will cause WebDAV to return read-only permissions.

## WebUI

- The webUI reflects the webdav permissions and "greys out" or hides all actions that would change the files on the storage

## Config

- To enable the interceptor, add it to the storage provider config.

### Example storage-home.toml
```
# This storage-home.toml config file will start a reva service that:
[shared]
jwt_secret = "Pive-Fumkiu4"
gatewaysvc = "localhost:19000"

# - authenticates grpc storage provider requests using the internal jwt token
# - authenticates http upload and download requests requests using basic auth
# - serves the home storage provider on grpc port 12000
# - serves http dataprovider for this storage on port 12001
#   - /data - dataprovider: file up and download
#
# The home storage will inject the username into the path and jail users into
# their home directory

[grpc]
address = "0.0.0.0:12000"
interceptors = [
  "readonly"
]

# This is a storage provider that grants direct access to the wrapped storage
# TODO same storage id as the /oc/ storage provider
# if we have an id, we can directly go to that storage, no need to wrap paths
# we have a locally running dataprovider
# this is where clients can find it
# the context path wrapper reads tho username from the context and prefixes the relative storage path with it
[grpc.services.storageprovider]
driver = "ocis"
mount_path = "/home"
mount_id = "123e4567-e89b-12d3-a456-426655440000"
expose_data_server = true
data_server_url = "http://localhost:12001/data"
enable_home_creation = true

[grpc.services.storageprovider.drivers.ocis]
root = "/var/tmp/reva/data"
enable_home = true
treetime_accounting = true
treesize_accounting = true
#user_layout = 
# do we need owner for users?
#owner = 95cb8724-03b2-11eb-a0a6-c33ef8ef53ad 


[http]
address = "0.0.0.0:12001"

[http.services.dataprovider]
driver = "ocis"
temp_folder = "/var/tmp/reva/tmp"

[http.services.dataprovider.drivers.ocis]
root = "/var/tmp/reva/data"
enable_home = true
treetime_accounting = true
treesize_accounting = true
```
### Known Issue

- Currently we need to add the interceptor to `storagehome` and `storageusers` which share the same physical storage (which is IMO weird and will be changed in the future)

### DIFF
--- a/changelog/unreleased/read-only-storageprovider-interceptor.md
+++ b/changelog/unreleased/read-only-storageprovider-interceptor.md
@@ -1,0 +1,5 @@
+Enhancement: Add readonly interceptor
+
+The readonly interceptor could be used to configure a storageprovider in readonly mode. This could be handy in some migration scenarios.
+
+https://github.com/cs3org/reva/pull/1849
--- a/internal/grpc/interceptors/loader/loader.go
+++ b/internal/grpc/interceptors/loader/loader.go
@@ -18,4 +18,8 @@
 
 package loader
 
-// Add your own.
+import (
+	// Load core GRPC services
+	_ "github.com/cs3org/reva/internal/grpc/interceptors/readonly"
+	// Add your own service here
+)
--- a/internal/grpc/interceptors/readonly/readonly.go
+++ b/internal/grpc/interceptors/readonly/readonly.go
@@ -1,0 +1,146 @@
+// Copyright 2018-2021 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package readonly
+
+import (
+	"context"
+
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	registry "github.com/cs3org/go-cs3apis/cs3/storage/registry/v1beta1"
+	"github.com/cs3org/reva/pkg/appctx"
+	"github.com/cs3org/reva/pkg/rgrpc"
+	rstatus "github.com/cs3org/reva/pkg/rgrpc/status"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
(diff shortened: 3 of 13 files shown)


micbar:
@labkode It took me a while to figure out the interceptors concept in reva. It works now when you add the interceptor to the storage provider config.

labkode:
@micbar awesome, this can be useful in many places, not only for a migration, for example to give access to read only storages.

Can you add in the PR description or in some example config how to enable it?

The CI complains about some tests.

micbar:
> The CI complains about some tests.

Yes. I am on it. Still working on the WebDAV response Codes and Response Body.

ishank011:
@micbar @labkode can this be combined with the scopes we introduced recently? It does a similar job and adding the read-only scope would be a good addition.

micbar:
@ishank011 Can you point me to some resources / docs / tickets regarding the scopes?

ishank011:
Hi @micbar. There's a brief description in the PR https://github.com/cs3org/reva/pull/1669 and the workflow is detailed here https://codimd.web.cern.ch/XTib-1TzTyqx2IZJJOq5pA. I'll add some proper documentation as well.

As a summary, the auth provider returns the scope for which the token is valid. For example, for basic auth and in [ocis](https://github.com/owncloud/ocis/blob/master/proxy/pkg/middleware/account_resolver.go#L95), this token has the ['owner' scope](https://github.com/cs3org/reva/blob/master/pkg/auth/scope/user.go#L35), i.e., unrestricted access to all resources. We can add a similar read-only scope, and make it configurable in the auth providers. The checks can be done like [this](https://github.com/cs3org/reva/blob/master/pkg/auth/scope/resourceinfo.go#L33-L68). I skipped the reader/editor checks because of issues with the WebDAV response codes as well. So it'll be good to fix those.

micbar:
@ishank011 
IMO your use case is different if I understand correctly

I am trying to protect a storage by keeping it in a forced read-only mode to make sure that nothing changes "on disk"  regardless of the accessing user and the auth method.

I would really like to keep the scope narrow in this PR. We need this for a migration scenario where you have a parallel usage of the same storage by two instances where only one instance has write access.

micbar:
@labkode I added an example config to the top post

butonic (dismissed):
omg so many fixes!

butonic (approved):

labkode:
@micbar @ishank011 let's keep the context of this PR for sysadmins to force a storage in read-only.
The scope for auth tokens is much more fine grained and I think this way is simpler for a sysadmins to configure the storage to be put in RO.

labkode (approved):

PR #1861, opened 2021-07-02
[WIP] Runtime plugins

jimil749:
## Description

This is my first try to enable runtime plugin support in the Reva ecosystem. This PR introduces a new `plugin` package, which is responsible for loading the plugins using the [hashicorp go-plugin](https://github.com/hashicorp/go-plugin) library. The current implementation uses Remote Procedure Call (RPC) to communicate between the host(reva core) and the plugin. I tried to move the *userprovider* drivers to the runtime paradigm, which implies that a user load userprovider drivers (like json, ldap etc) at runtime by just pointing to the binary in the configuration. 

## Working

In order to facilitate backward compatibility, there aren't any major changes to the configuration. Also, loading plugins from the in-memory is also supported. If the user wishes to load external plugins, he/she needs to add an extra shared config:
```
[shared]
plugin = true

[grpc.services.userprovider]
driver = '/path/to/binary'
```
This will configure the services to load plugins from the plugins package rather than the registry. The plugins package creates an rpc client, loads the required plugin and returns an `interface{}` value, which is then asserted into required type by the respective services. It is then responsibility of the services to call the `New` method to initialize the plugin with the configurations.

## Progress
- [x] Global Plugins flag: Option to load plugins from either registry or plugin package
- [x] Complementing the existing userprovider JSON driver with runtime paradigm.
- [x] Plugin Package
  - [x] Support for loading plugin *binaries*
  - [x] Support for compiling and loading the plugin from source
  - [ ] Support for plugins hosted as a version controlled repository
~~- [ ] Watcher module (Hot Reload)~~
~~- [ ] Separate dev and prod environment~~

 Since, there maybe a lot of changes involved with the approach I've taken, I am creating a draft PR. Any comments for improvement would be highly appreciated! :)


examples/plugin/plugin.toml
```diff
@@ -0,0 +1,28 @@
...
+[grpc]
+address = "0.0.0.0:19000"
+
+[grpc.services.gateway]
+userprovidersvc = "localhost:19000"
+
+
+[grpc.services.userprovider]
+# replace the driver with the plugin binary file
+driver = "/path/to/binary"
```
jimil749, line 25: Currently, I just have this dummy path. Should I also provide the actual binary for the purpose of testing?

internal/grpc/services/userprovider/userprovider.go
```diff
@@ -59,7 +70,37 @@ func parseConfig(m map[string]interface{}) (*config, error) {
...
+		return nil, err
+	}
+
+	// assert the loaded plugin into required interface
+	manager, ok := sym.(user.UserManager)
+	if !ok {
+		return nil, fmt.Errorf("could not assert the loaded plugin")
+	}
+
+	pluginConfig := filepath.Base(c.Driver)
```
jimil749, line 86: I'm using `filepath.Base` to retrieve the actual driver name from the config. Since, whenever we want load a binary, we're providing the absolute path (something like `/home/path/to/binary`), due to which we would not be able to access the plugin configuration via `c.Drivers[c.Driver]` (this would just result in an empty map) because, `c.Driver` would be `/home/path/to/binary` as opposed to just the driver name. So in order to access the plugin configuration I had to use `filepath.Base`

pkg/user/user.go
```diff
@@ -69,3 +69,11 @@ type Manager interface {
 	GetUserGroups(ctx context.Context, uid *userpb.UserId) ([]string, error)
 	FindUsers(ctx context.Context, query string) ([]*userpb.User, error)
 }
+
```
jimil749, line 72: I've used an extra interface to facilitate the rpc communication. The only diff is that this interface has an extra `New` method to initialize the plugins

examples/plugin/plugin.toml
labkode: Yep, the full path so is clear to the reader.

internal/grpc/services/userprovider/userprovider.go
labkode: That is okay! We can think of having some aliasing in the future.

pkg/user/user.go
labkode: Uhm, this in non ideal. I would expect concrete types to use New to create an object that implement the interface.

With this approach,  the concrete object (after has been created) has a New method, this is very confusing.

jimil749: @labkode I know this is a bit weird and confusing, but since we're using RPCs for communication, there is no way to initialize the object that implements the interface i.e no way to configure the plugin. The `New` method is to pass the configuration (the `user.demo.json` file in case of json plugin.) Without this method, we cannot pass the config, rendering the subsequent methods useless.

examples/plugin/plugin.toml
jimil749: Okay. I'll add the plugin binary in the example folder.

pkg/user/user.go
jimil749: To avoid confusion, we can maybe change the function name to something more appropriate like `Initialize()`?

ishank011:
@jimil749 can you also push the changes in the json pkg? It'll help me review the code

ishank011:
> @jimil749 can you also push the changes in the json pkg? It'll help me review the code

Or do we have different code for the plugin and the registry paths?

jimil749:
@ishank011 sure. The plugin code is slightly different than the existing json pkg. I'll add the source code in the example directory.

ishank011:
Really good progress!

One thought about managing the configuration part. Can we have an interface in the plugin pkg with one method `Configure(m map[string]interface{}) error` and the manager interfaces in the drivers will embed this? Their `New` methods will be sort of wrappers around this method. And we can get rid of the duplicate interfaces and assume that all managers will have the `Configure` method. If this works, we can have the same driver code for both paradigms.

Also, a question. The `ManagerRPC` would work for all drivers, right? If you have two plugins, json and demo, just changing the path in the configuration should work?

jimil749:
> Also, a question. The `ManagerRPC` would work for all drivers, right? If you have two plugins, json and demo, just changing the path in the configuration should work?

Just to quickly answer this: Yes! This method works for all the "userprovider" plugins. I tested it out with demo plugin and it works!

jimil749:
> One thought about managing the configuration part. Can we have an interface in the plugin pkg with one method `Configure(m map[string]interface{}) error` and the manager interfaces in the drivers will embed this? Their `New` methods will be sort of wrappers around this method. And we can get rid of the duplicate interfaces and assume that all managers will have the `Configure` method. If this works, we can have the same driver code for both paradigms.

That's a great suggestion! Let me try to implement this locally and see how it goes. This would really improve the implementation.

jimil749:
> One thought about managing the configuration part. Can we have an interface in the plugin pkg with one method `Configure(m map[string]interface{}) error` and the manager interfaces in the drivers will embed this? Their `New` methods will be sort of wrappers around this method. And we can get rid of the duplicate interfaces and assume that all managers will have the `Configure` method. If this works, we can have the same driver code for both paradigms.
>

@ishank011, why can't we just add the method to the manager interface, instead of embedding? I think that should also work? Because, embedding will add dependency on the plugins package. (which would cause cyclic imports)

We could simply add Configure method to all the managers and have the `New` methods as wrappers, which would use the Configure method to configure the plugins. Does that seem fine?

ishank011:
The cyclic dependency isn't ideal. Can we have a registry similar to what we have for the individual packages to populate `pluginMap`? The reason why I want to have this interface in the plugin package is to maintain consistency across packages. If we want to add new things sometime in the future, that would be a lot easier as well

jimil749:
@ishank011, thanks for the clarification! Makes perfect sense to me. :+1:

internal/grpc/services/userprovider/userprovider.go
```diff
@@ -59,7 +71,39 @@ func parseConfig(m map[string]interface{}) (*config, error) {
...
+		return nil, err
+	}
+	return manager, nil
+}
+
+func getDriver(c *config, plugin bool) (user.Manager, error) {
+	// if plugin flag is set, we fetch the driver from the plugin package via hashicorp go-plugin system
+	if plugin {
+		// populate the plugin registry
+		pluginregistry.Register("userprovider", &user.ProviderPlugin{})
```
ishank011, line 99: This should be called from `pkg/user.Init()`

ishank011:
Majorly looks good. I have a few nitpicks but we can get to those later on. We can proceed to the next step I guess? Let's discuss the compilation part in the issue

jimil749:
Sure!

pkg/user/user.go
jimil749: @labkode, this has been updated!

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return "", err
+	}
+	return pluginsDir, nil
+}
+
+// Load loads the plugin using the hashicorp go-plugin system
+func Load(driver string, pluginType string) (interface{}, error) {
+	bin := driver
+	if filepath.Ext(driver) == ".go" {
```
jimil749, line 59: Is this okay? I am currently assuming that if the user wants to compile, he/he can provide path to the source code ending with `.go`. So whenever the config ends with `.go`, we compile!

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,94 @@
...
+	MagicCookieValue: "hello",
+}
+
+func compile(path string, pluginType string) (string, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("could not find current directory: %v", err)
+	}
+	name := fmt.Sprintf("%d", rand.Int())
+	pluginsDir := filepath.Join(wd, "bin", pluginType, name)
```
jimil749, line 45: I store the binaries in `./bin/userprovider/<rand_int>`. Is this fine?

jimil749:
@ishank011, I'm a bit worried about one thing: with the current approach, loading a plugin binary works just fine. The hashicorp go-plugin spawns a new plugin process for us, which can be used over rpc. But we're not killing the process once we're done using it. So when I quit revad, the plugin process is not killed. We need to clean that up too.

internal/grpc/services/userprovider/userprovider.go
```diff
@@ -59,7 +71,38 @@ func parseConfig(m map[string]interface{}) (*config, error) {
...
+	}
+
+	// assert the loaded plugin into required interface
+	manager, ok := sym.(user.Manager)
+	if !ok {
+		return nil, fmt.Errorf("could not assert the loaded plugin")
+	}
+
+	basename := filepath.Base(c.Driver)
+	pluginConfig := strings.TrimSuffix(basename, filepath.Ext(basename))
```
jimil749, line 88: Again, I need to do this because, whenever a user provides configure like:
```
[grpc.services.userprovider]
driver = /home/path/to/json.go

[grpc.services.userprovider.json]
users = users.demo.json
```
so I need to strip off, `.go` from the driver field to access to users field!

pkg/plugin/plugin.go
ishank011: This path should point to a package as it can have multiple go files.

pkg/plugin/plugin.go
ishank011: Why a random integer? Let's use `./bin/userprovider/driver_name`, we can replace the old binaries and won't have to worry about cleaning those up.

pkg/plugin/plugin.go
jimil749: Ah okay. So then we can stat the path, and check if that's a directory, we compile else we directly load.

internal/grpc/services/userprovider/userprovider.go
ishank011: Pointing to the package should take care of that

ishank011:
The GRPC services have a `Close` method where we can call [`Kill`](https://github.com/hashicorp/go-plugin/blob/master/client.go#L407) if the plugin flag is set

pkg/plugin/plugin.go
jimil749: Yeah makes sense. I'll make the changes

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+
+	logger := hclog.New(&hclog.LoggerOptions{
+		Name:   "plugin",
+		Output: os.Stdout,
+		Level:  hclog.Trace,
+	})
+
+	client := plugin.NewClient(&plugin.ClientConfig{
```
ishank011, line 73: This client should be a singleton and whenever a service calls Load with a driver type, `Dispense` should be called and that instance of the plugin should be returned. So when shutting down reva, each plugin instance won't need to be killed, you can just call `Kill` on the singleton. I just glanced over the go-plugin code and we'll need to look into the specifics, but I think it should work

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+	MagicCookieValue: "hello",
+}
+
+func compile(path string, pluginType string) (string, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("could not find current directory: %v", err)
+	}
+	pluginsDir := filepath.Join(wd, "bin", pluginType, filepath.Base(path))
+	command := fmt.Sprintf("GO111MODULE=off go build -o %s %s", pluginsDir, path)
```
jimil749, line 44: Here, the `path` should be *relative*, so when we are configuring the driver, the path should be relative to the current working directory because `go build` does not work when we provide absolute path. So config should be like:
```
[grpc.services.userprovider]
driver = "../json"
```

jimil749: `go build -o json /home/path/to/source` is not supported

pkg/plugin/plugin.go
jimil749: Hmm, with the current pattern, we are essentially creating a *new* client for every service. So by having client as a singleton, we want every service to use that same client to connect to the drivers?

ishank011: Okay no it won't work, you need to pass the binary location when instantiating it

jimil749:
@ishank011, currently this approach works just fine:

1. The plugin package is able to load plugin binary via the hashicorp go-plugin.
2. The plugin package is also able to first compile the source code, store the binary at `./bin/userprovider/<driver>` and then load the binary. (Only constraint is that the user needs to provide relative path to the source code!). Package differentiates the binary file and source code using `os.Stat`. If the user points to a package/directory, we compile else we directly load the driver.
3. The older in-memory registry way loading is also supported, can be toggled by the `plugin` flag in the shared config.

But I'm not sure why the single integration test is failing! :sweat_smile:

ishank011 (commented):
@jimil749 good work! Mostly looks good, just added a few minor comments. But what I'd want to see in the example is something like
```
[grpc.services.userprovider]
driver = "../../pkg/user/json"
```

and that should work. What changes do we need to make that happen?

jimil749:
> @jimil749 good work! Mostly looks good, just added a few minor comments. But what I'd want to see in the example is something like
> 
> ```
> [grpc.services.userprovider]
> driver = "../../pkg/user/json"
> ```
> 
> and that should work. What changes do we need to make that happen?

If we want to point to the `../../pkg/user/manager/json`, we'd need to make that package compatible with the hashicorp go-plugin library. I think we need to add main method in the package, which starts up the hashicorp plugin server. Something like this:

```go
func main() {
	plugin.Serve(&plugin.ServeConfig{
		HandshakeConfig: Handshake,
		Plugins: map[string]plugin.Plugin{
			"userprovider": &user.ProviderPlugin{Impl: &Manager{}},
		},
	})
}
```
This *has* to be done for every runtime plugin!

ishank011:
Yep, and trivially we would have to name the package main as well because they won't have an entrypoint otherwise, but if we do, that would make the registry approach unusable.

So we can go for something like https://github.com/golang/go/issues/18124#issue-192683286

> This _has_ to be done for every runtime plugin!

Yep, that's the point

pkg/user/rpc_user.go
```diff
@@ -0,0 +1,196 @@
...
+	"context"
+	"net/rpc"
+
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	"github.com/cs3org/reva/pkg/pluginregistry"
+	"github.com/hashicorp/go-plugin"
+)
+
+func init() {
+	pluginregistry.Register("userprovider", &ProviderPlugin{})
```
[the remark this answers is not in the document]
jimil749: Unfortunately, we can't do that. Config is propagated after the init methods are called. So it would always be false!

jimil749:
> Yep, and trivially we would have to name the package main as well because they won't have an entrypoint otherwise, but if we do, that would make the registry approach unusable.
> 
> So we can go for something like [golang/go#18124 (comment)](https://github.com/golang/go/issues/18124#issue-192683286)
> 
> > This _has_ to be done for every runtime plugin!
> 
> Yep, that's the point

Ah okay! I'll take a look and see if we can incorporate the mentioned way here!

ishank011:
Yeah I don't think that would work. Read the second comment, it won't work with exports. But still, give it a try

jimil749:
@ishank011 I tried working on it and it kind of works but I feel like it is a bit ugly! So basically I had to add a new package in `./pkg/user`, wherein I have a single json.go file defined with the following content:

```go
package main

import (
	"github.com/cs3org/reva/pkg/user"
	json "github.com/cs3org/reva/pkg/user/manager/json"
	"github.com/hashicorp/go-plugin"
)

// Handshake hashicorp go-plugin handshake
var Handshake = plugin.HandshakeConfig{
	ProtocolVersion:  1,
	MagicCookieKey:   "BASIC_PLUGIN",
	MagicCookieValue: "hello",
}

func main() {
	plugin.Serve(&plugin.ServeConfig{
		HandshakeConfig: Handshake,
		Plugins: map[string]plugin.Plugin{
			"userprovider": &user.ProviderPlugin{Impl: &json.Manager{}},
		},
	})
}
```

This package just serves the json plugin(without actually modifying the existing driver). But the ugly part is that we would need to do this for all the drivers. So for example, talking about the userprovider drivers, we'd need to do this for ldap and demo drivers too!

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,119 @@
...
+	MagicCookieKey:   "BASIC_PLUGIN",
+	MagicCookieValue: "hello",
+}
+
+func compile(pluginType string, path string) (string, error) {
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("could not find current directory: %v", err)
+	}
+	binaryPath := filepath.Join(wd, "bin", pluginType, filepath.Base(path))
```
ishank011, line 53: We need to create this directory. Also, not sure about creating it in the directory where you're running reva from. Maybe `/var/tmp/reva/bin/`?

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,119 @@
...
+	command := fmt.Sprintf("GO111MODULE=off go build -o %s %s", binaryPath, path)
```
ishank011, line 54: Why are we using `GO111MODULE=off`? We have the go.mod file in the parent directory, so shouldn't it be fine? I got multiple errors with the flag set to off because the libraries weren't present in my path. And it will be deprecated from go 1.16 so we shouldn't use it in the first place.

ishank011 (commented):
Ran into a couple of issues when running the code.

pkg/plugin/plugin.go
ishank011: Okay no, this works. I only ran into the other issue

pkg/plugin/plugin.go
jimil749: Yes, I forgot to update this. I also got the error, I did this because I was testing it on go project without using go modules. I will remove it!

pkg/plugin/plugin.go
jimil749: Done!

internal/grpc/services/userprovider/userprovider.go
```diff
@@ -59,12 +70,40 @@ func parseConfig(m map[string]interface{}) (*config, error) {
...
+	plugin, err := c.load()
+	if err != nil {
+		return nil, err
 	}
+	return plugin, nil
+}
 
-	return nil, errtypes.NotFound(fmt.Sprintf("driver %s not found for user manager", c.Driver))
+func getDriver(c *config) (user.Manager, *plugin.RevaPlugin, error) {
+	isAlphaNum := regexp.MustCompile(`^[A-Za-z0-9]+$`).MatchString
```
jimil749, line 83: @ishank011, I am adding the regex check here. So when we have to load from the registry, the configuration should just be a single word (like `json`, `s3` etc). Otherwise, when we'd want to load from the plugin package the user will provide something like, `./json` or `github.com/reva-pkg/user`.

ishank011: Perfect! Will go through it

ishank011:
@jimil749 another thing. We should read the stderr from the exec command as well. Currently, we just get
```
2021-07-26 16:23:15.02 ERR ../../cmd/revad/runtime/runtime.go:205 > error starting the grpc server error="unable to register services: rgrpc: grpc service userprovider could not be started,: exit status 2" pid=71770
```

which makes it difficult to debug the extension.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,116 @@
...
+		bin, err = compile(pluginType, driver)
+		if err != nil {
+			return "", err
+		}
+	}
+	return bin, nil
+}
+
+// Load loads the plugin using the hashicorp go-plugin system
+func Load(driver string, pluginType string) (*RevaPlugin, error) {
```
ishank011, line 78: Reverse the parameter sequence here as well

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+var handshake = plugin.HandshakeConfig{
+	ProtocolVersion:  1,
+	MagicCookieKey:   "BASIC_PLUGIN",
+	MagicCookieValue: "hello",
+}
+
+func compile(pluginType string, path string) (string, error) {
+	dirname := "/var/tmp/reva"
```
ishank011, line 49: Declare as a global const

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,116 @@
...
+// Load loads the plugin using the hashicorp go-plugin system
+func Load(driver string, pluginType string) (*RevaPlugin, error) {
+	bin, err := checkDirAndCompile(pluginType, driver)
+	if err != nil {
+		return nil, err
+	}
+
+	logger := hclog.New(&hclog.LoggerOptions{
+		Name:   "plugin",
+		Output: os.Stdout,
```
ishank011, line 86: I don't see the plugin logs anywhere. What do we need to enable that?

pkg/plugin/plugin_config.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2018-2021 CERN
```
ishank011, line 1: Rename the file

internal/grpc/services/userprovider/userprovider.go
ishank011: Let's delegate this to the plugin pkg. We should do something like

```
func getDriver(c *config) (user.Manager, *plugin.RevaPlugin, error) {
	p, err := plugin.Load("userprovider", c.Driver)
	if err == nil {
	    // assert the loaded plugin into required interface
	    manager, ok := plugin.Plugin.(user.Manager)
	   ...
	} else if IsNotFound(err) {
	    // plugin not found, fetch the driver from the in-memory registry
	    if f, ok := registry.NewFuncs[c.Driver]; ok {
	    	    mgr, err := f(c.Drivers[c.Driver])
		    return mgr, nil, err
	    }
	} else {
	    return nil, nil, err
	}
	return nil, nil, errtypes.NotFound(fmt.Sprintf("driver %s not found for user manager", c.Driver))
}
```

Also, have the regex as a global var so we don't have to compile it every time

pkg/plugin/plugin.go
jimil749: Ideally you should see the log because I've set log level to `hclog.Trace`. I am getting the following output:
Screenshot from 2021-07-26 21-18-37

pkg/plugin/plugin_config.go
jimil749: Does it make sense to rename this to `plugin.go` and rename the current `plugin.go` to `loader.go`? because that's what the current `plugin.go` is doing! Not too sure though.

pkg/plugin/plugin.go
ishank011: I meant how do we log something from the plugin? We probably need to pass the client's context here as our logger is associated with it.

https://github.com/cs3org/reva/blob/968fd84137744f7dd3a52685bc8d0c1785a2e52c/pkg/user/rpc_user.go#L180-L184

jimil749: Done! :+1:

ishank011:
TODO:

- [ ] Return stderr from the exec call as part of the error from the `plugin.Load` method
- [ ] Delete the binary when `Kill` is called
- [ ] (Medium priority) Maintain the checksum of the repo as an xattr on the binary and only compile if it has changed

pkg/user/rpc_user.go
```diff
@@ -0,0 +1,211 @@
...
+	"net/rpc"
+
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	"github.com/cs3org/reva/pkg/appctx"
+	"github.com/cs3org/reva/pkg/plugin"
+	hcplugin "github.com/hashicorp/go-plugin"
+)
+
+func init() {
+	gob.Register(&userpb.User{})
```
jimil749, line 33: We can register interfaces (which we would like to pass as context) here. But I am not sure what all structs/interfaces that we'd like to add!

ishank011: Are we able to log from the plugins now?

jimil749: I tried passing the user object as a part of the context. I was able to access the user object (via the context.Value) on the plugin side. But here's the problem with the logger: the appctx's `WithLogger` attaches zerolog's `ctxKey` struct to the context, which is not exported, hence we cannot register that at init time.

ishank011: Can you look into ways of getting around that?

ishank011: In the auth pkg, we're passing `auth.Scope` without registering it. How does that work?

jimil749: I think this *might* be a roadblock as there is not way to `gob.Register(ctxkey{})` at init stage because ctxKey{} is private! Hence the RPC is always going to fail when attaching logger to context.

jimil749: > Can you look into ways of getting around that?

Yes. I was just investigating ways around this!

ishank011: I'll merge this for now but this needs to be addressed ASAP.

ishank011 (approved):

pkg/user/rpc_user.go
jimil749: Yes! So the reason we're getting the gob error is because we're passing `map[interface{}]interface{}`, If we are passing `interface{}` through gob, that NEEDS to be registered, passing structs is fine. That's why it does not complain when using `auth.Scope`. https://pkg.go.dev/encoding/gob#Register I'll further dig into this

ishank011: Then instead of the map, let's store all the keys that we know of in a struct. The ones we currently use are the logger, user object and the token.

PR #1870, opened 2021-07-07
be defensive about wrongly quoted etags

butonic:
When ocdav renders etags it will now try to correct them to the definition as *quoted strings* which do not contain `"`. This prevents double or triple quoted etags on the webdav api.

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -419,7 +419,7 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 			// etags must be enclosed in double quotes and cannot contain them.
 			// See https://tools.ietf.org/html/rfc7232#section-2.3 for details
 			// TODO(jfd) handle weak tags that start with 'W/'
-			propstatOK.Prop = append(propstatOK.Prop, s.newProp("d:getetag", md.Etag))
+			propstatOK.Prop = append(propstatOK.Prop, s.newProp("d:getetag", quoteEtag(md.Etag)))
 		}
 
 		if md.PermissionSet != nil {
@@ -712,7 +712,7 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 				switch pf.Prop[i].Local {
 				case "getetag": // both
 					if md.Etag != "" {
-						propstatOK.Prop = append(propstatOK.Prop, s.newProp("d:getetag", md.Etag))
+						propstatOK.Prop = append(propstatOK.Prop, s.newProp("d:getetag", quoteEtag(md.Etag)))
 					} else {
 						propstatNotFound.Prop = append(propstatNotFound.Prop, s.newProp("d:getetag", ""))
 					}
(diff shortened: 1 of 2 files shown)


ishank011 (approved):

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -816,6 +816,14 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 	return &response, nil
 }
 
+// be defensive about wrong encoded etags
+func quoteEtag(etag string) string {
+	if strings.HasPrefix(etag, "W/") {
+		return `W/"` + strings.Trim(etag[2:], `"`) + `"`
+	}
+	return `"` + strings.Trim(etag, `"`) + `"`
```
ishank011, line 824: @butonic is this not handled by the file systems? I think it should be added there so that other clients also don't face this issue. Also, what is the use case for W/

butonic: It should be handled by storage drivers, not filesystems (if the filesystem gives you an etag that is great, but POSIX eg. does not have them, so the driver needs to provide an etag). In oc10 we internally omitted the quotes, but the etag spec says they are a quoted string, so we have to enclose the etag in quotes in webdav responses. The PR is called *defensive* to protect against sloppy driver implementations.

The `W/` indicates a so called *weak* etag. It is only relevant for range requests. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag for a good explanation.

That being said all reva storage drivers I am aware of currently only produce strong etags, BUT in the future there might be a storage driver that fetches resources from another web server via whatever HTTP based protocol ... and that MIGHT use a weak etag ... which is why we have to properly deal with a weak etag here.

I propose to internally or even on the CS3 api omit the `"`. It is far more readable when debugging and in logs.

@labkode opinions?

ishank011: Yep, I meant storage drivers only, sorry about that.

I agree with removing the quotes in the driver implementations and the CS3APIs. We can add a comment there https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.ResourceInfo and remove those from the drivers. I think we can wait for @labkode to respond as well

labkode: I'm in favour as well to keep the storage drivers implementation cleaner to remove quoting/encoding.
The decoration should happen in the ocdav layer as it only concerns owncloud clients.

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -816,6 +816,14 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 	return &response, nil
 }
 
+// be defensive about wrong encoded etags
+func quoteEtag(etag string) string {
+	if strings.HasPrefix(etag, "W/") {
+		return `W/"` + strings.Trim(etag[2:], `"`) + `"`
```
labkode, line 822: This will panic when etag is empty.

butonic: no, see line above: `if strings.HasPrefix(etag, "W/") {`

ishank011 (approved):

PR #1887, opened 2021-07-13
Add "a" and "l" filters in grappa

### LINKED ISSUE #1773: Create filtering namespace for finding users and groups
In current CERNBox we have different namespaces to filters users and groups:

`hugo` will look into primary and e-groups
`a:hugo` will look into primary/secondary/service/e-groups

Given that we have now the possibility to share with lightweight accounts we should add a third option:
`l:hugo` to only look into lightweight accounts.

This prefixes need to be understood in our userprovider and groupprovider drivers so we query Grappa for what we need.

Screenshot 2021-06-08 at 16 32 54

gmgigi96:
This PR adds the namespace filters "a" and "l" for grappa queries.

- with no filter: look into primary and e-groups
- `"a"`: look into primary/secondary/service/e-groups
- `"l"`: look into lightweight accounts.

Queries are in the form `[filter:]name`.

Closes #1773

### DIFF
--- a/changelog/unreleased/filter-ns-users-groups.md
+++ b/changelog/unreleased/filter-ns-users-groups.md
@@ -1,0 +1,10 @@
+Enhancement: Add "a" and "l" filter for grappa queries
+
+This PR adds the namespace filters "a" and "l" for grappa queries.
+With no filter will look into primary and e-groups, with "a" will look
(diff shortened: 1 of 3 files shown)


labkode:
@gmgigi96 looks good! Can you rebase on top of master?

labkode (approved):

PR #1888, opened 2021-07-13
Refactor dav

C0rby:
I refactored the webdav code to make it cleaner and reusable.
This is a preparation for https://github.com/cs3org/reva/pull/1803.

Currently those changes are also in the spaces PR but once this PR is merged we can rebase the spaces PR and reduce its diff.


internal/http/services/owncloud/ocdav/move.go
```diff
@@ -27,27 +28,39 @@ import (
...
-	dst = path.Join(ns, dst)
+	dstPath = path.Join(ns, dstPath)
 
-	sublog := appctx.GetLogger(ctx).With().Str("src", src).Str("dst", dst).Logger()
-	sublog.Debug().Str("overwrite", overwrite).Msg("move")
+	sublog := appctx.GetLogger(ctx).With().Str("src", srcPath).Str("dst", dstPath).Logger()
+	src := &provider.Reference{Path: srcPath}
+	dst := &provider.Reference{Path: dstPath}
+
+	intermediateDirRefFunc := func() (*provider.Reference, *rpc.Status, error) {
```
ishank011, line 53: This looks like an overkill. If you want to modularize this, how about just creating a method that takes a ref and returns a ref having the dir, rather than passing the functor object?

internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -32,56 +33,47 @@ import (
...
 
-	if depth != "infinity" && depth != "0" {
-		w.WriteHeader(http.StatusBadRequest)
-		m := fmt.Sprintf("Depth header is set to incorrect value %v", depth)
-		b, err := Marshal(exception{
-			code:    SabredavBadRequest,
-			message: m,
-		})
-		HandleWebdavError(&sublog, w, b, err)
+	cp := s.prepareCopy(ctx, w, r, srcRef, dstRef, intermediateDirRefFunc, &sublog)
```
ishank011, line 75: Can we call `prepareCopy` from `executePathCopy` (maybe rename it)?

ishank011 (commented):
@C0rby a couple of minor comments

internal/http/services/owncloud/ocdav/copy.go
C0rby: Not really, executePathCopy is a recursive function and we would need to introduce some condition to only execute prepareCopy once in executePathCopy.

C0rby: Also later there will be a `executeSpacesCopy` which implements spaces specific stuff, that is why `prepareCopy` is a function on its own.

internal/http/services/owncloud/ocdav/move.go
C0rby: Yeah it looks overkill but for the spaces the way to get the intermediate directory is different. That is why I'm passing the function so that the caller can decide how to get the intermediate dir.

butonic (approved):
Thx for carving the refactoring out of https://github.com/cs3org/reva/pull/1803. That will make that PR a lot easier to digest!

internal/http/services/owncloud/ocdav/move.go
ishank011: I'd still prefer an if condition to determine how to get this intermediate directory but we can discuss it in the spaces PR

ishank011 (approved):

PR #1891, opened 2021-07-14
Support for Deleting Shared Folders as a Share Receiver

refs:
closes https://github.com/owncloud/ocis/issues/2148

### DIFF
--- a/changelog/unreleased/delete-shared-resources.md
+++ b/changelog/unreleased/delete-shared-resources.md
@@ -1,0 +1,5 @@
+Bugfix: Delete Shared Resources as Receiver
+
+It is now possible to delete a shared resource as a receiver and not having the data ending up in the receiver's trash bin, causing a possible leak.
+
+https://github.com/cs3org/reva/pull/1891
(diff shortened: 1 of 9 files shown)


butonic (approved):
#nice ❤️

labkode:
@refs what about putting this logic in the gateway?
If the gateway can catch a Remove request to share it will call the share manager and set is as decline (not pending, see attached video).

@ishank011 what do you think?

ishank011:
> If the gateway can catch a Remove request to a share it will call the share manager and set is as decline (not pending, see attached video)

+1

Instead of calling `ListReceivedShares` for every delete request, we can instead add that logic here 

https://github.com/cs3org/reva/blob/905fa4c17f9b7f6ed95232a0f29a0d11340652b1/internal/grpc/services/gateway/storageprovider.go#L850-L857

refs:
> @refs what about putting this logic in the gateway?
> If the gateway can catch a Remove request to a share it will call the share manager and set is as decline (not pending, see attached video).
> 
> @ishank011 what do you think?
> unsharing.mov

Agree, that'd save some bandwidth 👍 !

refs:
@labkode decline seems like a more appropriate state for the share. I'll update the PR to reflect it.

refs:
@ishank011, @labkode if this matches the criteria, it should be ready to be merged 🎉

butonic (approved):

ishank011:
@refs looks good but I think we can add this parameter to `DeleteRequest` if we want to skip moving the resource to recycle bin. What do you think? @labkode @butonic

ishank011 (approved):

PR #1899, opened 2021-07-19
fix: chunked upload of existing file creates new version

C0rby:
Chunked uploads didn't create a new version, when the file to upload already existed.

### DIFF
--- a/pkg/storage/utils/decomposedfs/upload.go
+++ b/pkg/storage/utils/decomposedfs/upload.go
@@ -114,16 +114,14 @@ func (fs *Decomposedfs) InitiateUpload(ctx context.Context, ref *provider.Refere
 
 	log := appctx.GetLogger(ctx)
 
-	var relative string // the internal path of the file node
-
-	n, err := fs.lu.NodeFromResource(ctx, ref)
+	n, err := fs.lookupNode(ctx, ref.Path)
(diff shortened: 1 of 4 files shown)


pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -360,6 +358,33 @@ func (fs *Decomposedfs) GetUpload(ctx context.Context, id string) (tusd.Upload,
 	}, nil
 }
 
+// lookupNode looks up nodes by path.
+// This method can also handle lookups for paths which contain chunking information.
```
butonic, line 362: could you add a reason how this function differs from the one in the lookup interface? Do you think other implementations of the decomposedfs need the same handling / detection of chunked paths? should `fs.lu.NodeFromPath` maybe get a parameter to handle chunked paths? why not add a `fs.lu.NodeFromChunkedPath` and refactor the two functions to share more code?

labkode (approved):

pkg/storage/utils/decomposedfs/upload.go
butonic: Ah, now I see that you are calling `fs.lu.NodeFromPath` in `lookupNode` ... nevermind my comment ;-)

PR #1900, opened 2021-07-19
check name for illegal values during uploads and moves

C0rby:
The code was not checking for invalid file names during uploads and moves.


C0rby:
@phil-davis, are those supposed to be issue demonstrations?
```
apiOcisSpecific/apiWebdavMove1-moveFolder.feature:20
apiOcisSpecific/apiWebdavMove1-moveFolder.feature:31
apiOcisSpecific/apiWebdavMove1-moveFolder.feature:42
apiOcisSpecific/apiWebdavMove1-moveFolder.feature:53
apiOcisSpecific/apiWebdavMove1-moveFolder.feature:64
apiOcisSpecific/apiWebdavMove2-moveFile.feature:22
```

What should I do with those tests now? Delete them?

phil-davis:
> What should I do with those tests now? Delete them?

https://drone.cernbox.cern.ch/cs3org/reva/2304/6/6
`HTTP status code 400 is not one of the expected values 201 or 500`

I guess those tests were returning 500 and now return 400 - an improvement. But it looks like they should be able to really work and return 201.

IMO adjust the expectation to `Then the HTTP status code should be "400"` to "record" the current behavior.

Then someone needs to review all those "local" API tests to see what use they are. We have expected-failures files these days, that keep track of what core tests are known to be failing.

C0rby:
> `HTTP status code 400 is not one of the expected values 201 or 500`
> 
> I guess those tests were returning 500 and now return 400 - an improvement. But it looks like they should be able to really work and return 201.

Not quite right. They were returning 201 (ocis storage) or 500 (eos storage) but should have been returning 400 because the scenarios are testing illegal behavior like renaming files/folders to contain invalid characters.

So I think those tests can be removed now, right?

phil-davis:
> So I think those tests can be removed now, right?

True - there are core scenarios in core `apiWebdavMove1/moveFolder.feature` like:
`Scenario Outline: Renaming a folder to a backslash should return an error`
Those get run in the usual core API test pipelines.

Remove these "local" API tests.

phil-davis (dismissed):

ishank011:
@C0rby should we also add these checks for copy?

C0rby:
That's a good point. Let me check if copy is also affected.

C0rby:
Yes, I added the check to `COPY` and `MKCOL`.

ishank011 (approved):

PR #1903, opened 2021-07-20
check uploader during TUS uploads

C0rby:
Previously when a user initiated an upload and another user sent data to the initialized endpoint the request was not rejected.


C0rby:
See https://github.com/owncloud/ocis/issues/1141 for more information.

PR #1904, opened 2021-07-20
datagateway: zero content-length when swallowing body

butonic:
When swallowing the body the Content-Lenght needs to be set to 0 to prevent proxies from reading the body.

part of https://github.com/owncloud/ocis/pull/2248


butonic:
@C0rby @micbar this might affect response codes when using the ocis proxy.

labkode (approved):

PR #1911, opened 2021-07-22
fix versions order in propfind response

C0rby:
The order of the file versions in propfind responses was incorrect.


ishank011:
@C0rby can we do it in the storageprovider service?

C0rby:
Sure, I moved it to the storage provider. :thumbsup:

ishank011 (approved):

PR #1912, opened 2021-07-22
fix response when listing another user's file versions

C0rby:
The OCS API returned the wrong response when a user tried to list the versions of another user's file.

### DIFF
--- a/internal/http/services/owncloud/ocdav/versions.go
+++ b/internal/http/services/owncloud/ocdav/versions.go
@@ -107,6 +107,15 @@ func (h *VersionsHandler) doListVersions(w http.ResponseWriter, r *http.Request,
 		return
 	}
 	if res.Status.Code != rpc.Code_CODE_OK {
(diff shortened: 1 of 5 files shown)


internal/http/services/owncloud/ocdav/versions.go
```diff
@@ -120,6 +130,7 @@ func (h *VersionsHandler) doListVersions(w http.ResponseWriter, r *http.Request,
 		return
 	}
 	if lvRes.Status.Code != rpc.Code_CODE_OK {
+		fmt.Printf("\n\n status %v \n\n", lvRes.Status)
```
ishank011, line 133: This crept in from debugging?

C0rby: Ah damn it...
Yes, forgot to remove it...

ishank011 (dismissed):

ishank011 (approved):

PR #1918, opened 2021-07-23
fix share path when listing a specific share

C0rby:
The file path and target of shares where missing the share jail path when a specific share was listed.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
@@ -54,6 +54,7 @@ import (
 	"github.com/cs3org/reva/pkg/share"
 	"github.com/cs3org/reva/pkg/share/cache"
 	"github.com/cs3org/reva/pkg/share/cache/registry"
+	"github.com/cs3org/reva/pkg/user"
 	"github.com/cs3org/reva/pkg/utils"
 	"github.com/pkg/errors"
 )
@@ -299,6 +300,7 @@ func (h *Handler) GetShare(w http.ResponseWriter, r *http.Request) {
 		resourceID = psRes.Share.ResourceId
 	}
 
+	var receivedShare *collaboration.ReceivedShare
 	if share == nil {
 		// check if we have a user share
 		logger.Debug().Str("shareID", shareID).Msg("get user share by id")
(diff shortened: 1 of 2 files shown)


labkode:
@C0rby I think this one is related with https://github.com/cs3org/reva/pull/1739
I need @phil-davis @butonic to give us some help on this part as we maintain that PR as part of our fork and
I'll like to have it settled.

ishank011:
@C0rby the share prefix should only be prepended if the user is a recipient of the share. For share creators who want to get a share by ID, it should be skipped.

micbar:
@C0rby can you update?

C0rby:
> @C0rby the share prefix should only be prepended if the user is a recipient of the share. For share creators who want to get a share by ID, it should be skipped.

Yes, you're right. I'll update the code.

ishank011:
The thing is this method calls `GetShare` which doesn't return any info about whether the share was accepted by the recipient or not. And we don't call `mapState` here either (since we don't even have that info), so the state is always set to 0 which corresponds to `ocsStateAccepted`.

C0rby:
Ah, I didn't notice that.. Yeah, that makes it a bit more difficult...

C0rby:
Now I added an extra request to get the received share in the case if the current user is not the owner. The received share contains the state information. It is one additional request but shouldn't be too costly IMO.

refs:
> Now I added an extra request to get the received share in the case if the current user is not the owner. The received share contains the state information. It is one additional request but shouldn't be too costly IMO.

Exactly. I did the same in a previous PR. 👍

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -438,7 +460,11 @@ func (h *Handler) getShare(w http.ResponseWriter, r *http.Request, shareID strin
 		response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "error mapping share data", err)
 	}
 	h.mapUserIds(ctx, client, share)
-
+	if receivedShare != nil && receivedShare.State == collaboration.ShareState_SHARE_STATE_ACCEPTED {
+		// Needed because received shares can be jailed in a folder in the users home
+		share.FileTarget = path.Join(h.sharePrefix, path.Base(info.Path))
```
labkode, line 465: This breaks global urls for accessing shares as you're forcing a basePath for `FileTarget` and `Path`. 
It needs https://github.com/cs3org/reva/pull/1739 to be merged first and apply the same logic if I didn't miss anything.

labkode: @C0rby I just merged https://github.com/cs3org/reva/commit/afa06f7d83d5751b94e37e7bc3e1c5ef9580e06a, can you update this part with the same logic. If you're in the mood, it would make sense to refactor that `switch` logic into a function

C0rby:
AFAICT this PR is no longer necessary since the changes in `addFileInfo` already set the correct `FileTarget`.

PR #1924, opened 2021-07-26
use seconds as unit for transfer token ttl

wkloucek:


ishank011 (approved):
Okay previously the expiration was 100 minutes, nice XD

PR #1926, opened 2021-07-26
Trash Bin Restore Expected Failures

refs:
| test | fixed? |
|----------------------------|--------------------------|
| apiTrashbinRestore/trashbinRestore.feature:309   | ✅ |
| apiTrashbinRestore/trashbinRestore.feature:310   | ✅ |
| apiTrashbinRestore/trashbinRestore.feature:478   | ✅ |
| apiTrashbinRestore/trashbinRestore.feature:479   | ✅ |

### DIFF
--- a/internal/http/services/owncloud/ocdav/trashbin.go
+++ b/internal/http/services/owncloud/ocdav/trashbin.go
@@ -471,7 +471,7 @@ func (h *TrashbinHandler) restore(w http.ResponseWriter, r *http.Request, s *svc
 
 	// Restoring to a non-existent location is not supported by the WebDAV spec. The following block ensures the target
 	// restore location exists, and if it doesn't returns a conflict error code.
-	if dstStatRes.Status.Code == rpc.Code_CODE_NOT_FOUND && strings.HasSuffix(dst, itemPath) && isNested(dst) {
+	if dstStatRes.Status.Code == rpc.Code_CODE_NOT_FOUND && isNested(dst) {
 		parentStatReq := &provider.StatRequest{
 			Ref: &provider.Reference{Path: path.Join(getHomeRes.Path, filepath.Dir(dst))},
 		}
--- a/internal/http/services/owncloud/ocdav/trashbin.go
+++ b/internal/http/services/owncloud/ocdav/trashbin.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


butonic (approved):

internal/http/services/owncloud/ocdav/trashbin.go
```diff
@@ -469,6 +469,26 @@ func (h *TrashbinHandler) restore(w http.ResponseWriter, r *http.Request, s *svc
...
+
+		parentStatResponse, err := client.Stat(ctx, parentStatReq)
+		if err != nil {
+			sublog.Error().Err(err).Msg("error sending grpc stat request")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if parentStatResponse.Status.Code == rpc.Code_CODE_NOT_FOUND {
+			HandleErrorStatus(&sublog, w, &rpc.Status{Code: rpc.Code_CODE_ALREADY_EXISTS})
```
ishank011, line 487: I'm not sure I follow this. If the destination's parent directory doesn't exist, why are we reporting already exists?

butonic: we can go with
> (c) Use `FAILED_PRECONDITION` if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, `FAILED_PRECONDITION` should be returned since the client should not retry unless the files are deleted from the directory.

internal/http/services/owncloud/ocdav/trashbin.go
```diff
@@ -469,6 +469,26 @@ func (h *TrashbinHandler) restore(w http.ResponseWriter, r *http.Request, s *svc
 		return
 	}
 
+	// Restoring to a non-existent location is not supported by the WebDAV spec. The following block ensures the target
+	// restore location exists, and if it doesn't returns a conflict error code.
+	if dstStatRes.Status.Code == rpc.Code_CODE_NOT_FOUND && strings.HasSuffix(dst, itemPath) && isNested(dst) {
```
ishank011, line 474: `strings.HasSuffix(dst, itemPath)` do we always need to check this? Shouldn't we check that the parent dir exists even if the name of the file is to be changed?

ishank011 (commented):
@refs thanks for the fix! I just have one question

internal/http/services/owncloud/ocdav/trashbin.go
refs: That sounds correct; I think this was a leftover. Let me update it and run CI as soon as possible 🥇

ishank011 (approved):

PR #1927, opened 2021-07-27
Implement rollback to version

ffurano:


labkode (approved):
Looks good to me. I haven't tested it locally, in case there is some incompatibility with the new proto generated files we can always fix it in another PR.

PR #1935, opened 2021-07-28
JWT token mananger now returns the hash of the token

ishank011:


pkg/token/manager/jwt/jwt.go
```diff
@@ -105,11 +109,16 @@ func (m *manager) MintToken(ctx context.Context, u *user.User, scope map[string]
...
+	return m.cacheAndReturnHash(tkn)
 }
 
 func (m *manager) DismantleToken(ctx context.Context, tkn string) (*user.User, map[string]*auth.Scope, error) {
-	token, err := jwt.ParseWithClaims(tkn, &claims{}, func(token *jwt.Token) (interface{}, error) {
+	cachedToken, err := m.getCachedToken(tkn)
+	if err != nil {
+		return nil, nil, err
+	}
+
```
butonic, line 120:
```suggestion
// TODO fetch token from .. somewhere ?
```

This PR would make the tokenmanager only work when all services are running in the same process. Do you plan to make this more flexble? The `x-access-token` JWT is an internal token that should not leave the boundaries of the microservices. For extarnal access there is the openid connect auth, refresh and id token.

You should never have to encode the x-access-token in a url. We should consider proper API keys for that, IMO.

butonic (changes requested):

pkg/token/manager/jwt/jwt.go
ishank011: I tried with a persistent kv store as well (https://github.com/xujiajun/nutsdb), the problem was that the data wasn't synced until the connection was closed, so I faced the same issue of access from a different process. I'll try to benchmark how much of a drop in performance we face because of opening and closing a new file descriptor every time.

Agreed that the access token shouldn't be present in the URL, currently it's required by wopiserver though. We embed it in an iframe so the user doesn't see it but it can be easily extracted. But the issue is not just that; even adding it as a header might hit the upper limits (in our deployment, my token was 2.9k characters long and the [nginx limit](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers) is 4-8k)

butonic: For external access OpenID Connect should be used. I'm not sure kow the wopiserver works. Could we treat it like an openid connect client? We could generate a secret key for it ...

glpatcern: Currently, the wopiserver needs a token with full access rights to create/rename/delete files and folders on behalf of the user. If a different openID connect based token could be used, that's fine a priori. But I understand the issue is to carry over wherever needed the Reva token as it can be too large even for headers...

butonic: The internal access token is used to authenticate requests between micreservices and carry the some user data and the roles of a user. In OCIS we do it that way and the token remains small enough to cause problems. The services can then verify the tokens and fetch the permissions that are attached to the roles the user has. The permissions rarely change, so they can be cached. I haven't had the time to fully digest the roles and scopes concept in reva, but if it causes the token to become so large I think we need a different solution. It might depend on what we actually add to the token. maybe using abbreviations in the json keys and values will shrink it to a reasonable size?

Furthermore, I don't think the internal access token should ever leave the boundaries of the microservices. OpenID Connect has an id token that carries the user data, a refresh token that is longer lived and can be used to retrieve a new access token (which should then be a hash, not a jwt). access tokens should never be part of urls as they will be logged in intermediate proxies.

ishank011:
Superseded by https://github.com/cs3org/reva/pull/2085

PR #1937, opened 2021-07-30
Do not overwrite more specific matches when finding storage providers

aduffeck:
Depending on the order of rules in the registry it could happend that more specific matches (e.g. /home/Shares) were
overwritten by more general ones (e.g. /home). This PR makes sure that the registry always returns the most specific
match.


butonic (dismissed):

ishank011 (commented):
@aduffeck looks good. Can you just remove the extra line in the changelog?

aduffeck:
@ishank011 sure, done

ishank011 (approved):

PR #1938, opened 2021-07-30
Added methods to get and put data into context

jimil749:

### DIFF
--- a/pkg/appctx/ctxmap.go
+++ b/pkg/appctx/ctxmap.go
@@ -1,0 +1,63 @@
+// Copyright 2018-2021 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


ishank011 (commented):
Put this code in a separate file in the same package. Maybe `ctxmap.go`. And can you also add some unit tests?

jimil749:
@ishank011 I tried to add the unit tests but it fails when I run with `go test -race` (Seems to work fine when run it without the `race` flag). Not totally sure why.
It fails with the following error:
```
fatal error: checkptr: pointer arithmetic result points to invalid allocation
```

pkg/appctx/ctxmap.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return ctx
+}
+
+func getKeyValue(ctx context.Context, m map[interface{}]interface{}) {
+	ictx := *(*iface)(unsafe.Pointer(&ctx))
+	if ictx.data == 0 || int(*(*emptyCtx)(unsafe.Pointer(ictx.data))) == 0 {
+		return
+	}
+	valCtx := (*valueCtx)(unsafe.Pointer(ictx.data))
```
jimil749, line 58: This seems a bit dangerous! (VSCode complains here w `possible misuse of unsafe.Pointer`). We're dereferencing the value stored at `ictx.data` to `valCtx` type.

labkode (approved):

PR #1939, opened 2021-07-30
fix permissions for share jail

micbar:
# Description

The shared jail should not be writable.

closes https://github.com/owncloud/ocis/issues/2322

@aduffeck @refs @ishank011 This will only be temporary, until we remove all "Share" awareness and logic from the storage.

### DIFF
--- a/changelog/unreleased/fix-share-jail-perms
+++ b/changelog/unreleased/fix-share-jail-perms
@@ -1,0 +1,5 @@
+Bugfix: fix the share jail permissions in the decomposedfs
+
+The share jail should be not writable
+
+https://github.com/cs3org/reva/pull/1939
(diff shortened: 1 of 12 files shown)


ishank011:
It should already be handled, no? https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/storageprovider.go#L531-L657

micbar:
> It should already be handled, no? https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/storageprovider.go#L531-L657

yes, this is already forbidden.

But to get the information to the Client, the webDAV permissions should not indicate write possibilities.

So  we need to return the permissions also from the decomposedfs

micbar:
@aduffeck @butonic It turned into a small refactoring. Happy to discuss this in general how we work with pointer variables.

@dragotin FYI

butonic (approved):

ishank011 (approved):

PR #1940, opened 2021-07-30
s3ng: Provide objectSize when uploading to reduce memory usage

### LINKED ISSUE #1933: Provide objectSize for S3 upload to reduce memory usage
When using OCIS with the S3ng storage enabled, I noticed relatively high memory usage when uploading files. This could be more than 1GB for a single user. I did a bit of digging and noticed that in https://github.com/cs3org/reva/blob/7d251bffd9e699af55f811c1dbb54f4e9a14f377/pkg/storage/fs/s3ng/blobstore/blobstore.go#L63 no objectSize is provided, but rather just a [`-1`](https://github.com/minio/minio-go/blob/410eb9e1392ce96323f1ecc11915141a51d0dc72/api-put-object.go#L208). According to https://github.com/minio/minio-go/issues/1496 this can result in the observed memory usage. 
I did some testing on my own and checked what will happen if the correct size is provided. In my testing, the memory usage of OCIS was reduced significantly - from >1GB down to <300MB. 
If you like, I'm happy to open a PR to fix this issue. The question is: What would be the best way? In https://github.com/AndreasSko/reva/commit/be1e7de45db315cca9ec00ec38f1fc5d44977437 I changed the interfaces for https://github.com/cs3org/reva/blob/aad6c519983bb12ae892214ffc4bbef9c7276d46/pkg/storage/utils/decomposedfs/decomposedfs.go#L74 and https://github.com/cs3org/reva/blob/aad6c519983bb12ae892214ffc4bbef9c7276d46/pkg/storage/utils/decomposedfs/tree/tree.go#L48 to expect an `*os.File` instead of an `io.Reader`. But I could imagine to also simply provide the fileSize as an int64 directly. I'm open for feedback here :slightly_smiling_face: 

Related issue: https://github.com/owncloud/ocis/issues/2326

AndreasSko:
According to the [documentation](https://github.com/minio/minio-go/blob/410eb9e1392ce96323f1ecc11915141a51d0dc72/api-put-object.go#L200) for `PutObject`, setting `objectSize` to `-1` causes a multipart Put operation until the input stream reaches EOF, which can result in high memory usage. Something similar has already been reported in https://github.com/minio/minio-go/issues/1496

This PR changes the behavior so it tries to determine the file size before uploading it. Should the given `io.Reader` not be a file, it falls back to using `-1`.

In my testing this change has reduced the memory usage when uploading files to OICS quite a bit: From over 1GB for a single user, down to <300MB.

Closes #1933

### DIFF
--- a/changelog/unreleased/provide-size-for-s3-upload.md
+++ b/changelog/unreleased/provide-size-for-s3-upload.md
@@ -1,0 +1,7 @@
+Enhancement: Reduce memory usage when uploading with S3ng storage
+
+The memory usage could be high when uploading files using the S3ng storage.
+By providing the actual file size when triggering `PutObject`, 
+the overall memory usage is reduced.
+
+https://github.com/cs3org/reva/pull/1940
--- a/pkg/storage/fs/s3ng/blobstore/blobstore.go
+++ b/pkg/storage/fs/s3ng/blobstore/blobstore.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"io"
 	"net/url"
+	"os"
 
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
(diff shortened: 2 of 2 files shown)


labkode (approved):

labkode:
Thanks @AndreasSko for this nice fix!

PR #1941, opened 2021-07-30
fix TUS uploads with transfer token only

wkloucek:
TUS uploads had been stopped when the user JWT token expired, even if only the transfer token should be validated. Now uploads will continue as intended.

fixes owncloud/ocis#1205

### DIFF
--- a/cmd/reva/upload.go
+++ b/cmd/reva/upload.go
@@ -176,9 +176,6 @@ func uploadCommand() *command {
 			if err != nil {
 				return err
 			}
-			if token, ok := ctxpkg.ContextGetToken(ctx); ok {
-				c.Header.Add(ctxpkg.TokenHeader, token)
-			}
 			c.Header.Add(datagateway.TokenTransportHeader, p.Token)
 			tusc, err := tus.NewClient(dataServerURL, c)
 			if err != nil {
--- a/changelog/unreleased/tus-upload-with-transfer-token-only.md
+++ b/changelog/unreleased/tus-upload-with-transfer-token-only.md
@@ -1,0 +1,5 @@
+Bugfix: Fix TUS uploads with transfer token only
+
+TUS uploads had been stopped when the user JWT token expired, even if only the transfer token should be validated. Now uploads will continue as intended.
+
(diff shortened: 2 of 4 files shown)


cmd/reva/upload.go
```diff
@@ -176,9 +176,6 @@ func uploadCommand() *command {
 			if err != nil {
 				return err
 			}
-			if token, ok := tokenpkg.ContextGetToken(ctx); ok {
```
butonic, line 179: @ishank011 @labkode for TUS we do not need to mint an `X-Access-Token`, because the user information is stored on the server side in a metadata file for the upload. I am now wondering how the simple upload can determine which user should be used to access a file:
1. the `simple` code path in https://github.com/cs3org/reva/pull/1941/files#diff-38538775a8b4fe2f49a53d1d5700c8efb3270fd88a2743c92fe5dcd396372afdL150-L169 does not mint an `X-Access-Token`
2. The datagateway only verifies the `X-Transfer-Token` which does not contain any user information (just the path, expiry and the audience)
3. The dataprovider never receives an `X-Access-Token`, the auth middleware cannot put a user in the context.
4. The storage driver has no way to find the numericuid or username to use for the storage operation.

Could it be that for eos that only works because `revad` is running with a sudo user? Could you double check that you can use the `reva` cli to upload a file in `simple` mode and that it is created with the right permissions?
Or does that work because a new file derives the owner from the parent folder?
What if you upload a file into a shared folder as the recipient: who becomes the owner of the file?

AFAICT we need the userid in the transfer token so that the dataprovider can correctly become the user that tries to up/download the file. What am I missing?

wkloucek: @labkode @ishank011 could you have a look at this please?

ishank011: In the EOS driver, we do have the user set correctly, I've verified it from the logs. https://github.com/cs3org/reva/blob/master/pkg/storage/utils/eosfs/upload.go#L66

Let me go through the code once and then we can merge it

labkode (approved):

labkode (approved):

PR #1946, opened 2021-08-02
Add a share manager that connects to oc10 databases

aduffeck:


pkg/share/manager/sql/conversions.go
```diff
@@ -0,0 +1,259 @@
+// Copyright 2018-2021 CERN
```
ishank011, line 1: Maybe we can refactor `pkg/cbox/utils/conversions` to reduce duplicity?

pkg/share/manager/sql/sql.go
```diff
@@ -0,0 +1,502 @@
...
+	"github.com/cs3org/reva/pkg/utils"
+	"github.com/mitchellh/mapstructure"
+	"github.com/pkg/errors"
+
+	// Provides mysql drivers
+	_ "github.com/go-sql-driver/mysql"
+)
+
+func init() {
+	registry.Register("sql", NewMysql)
```
ishank011, line 46: @aduffeck can you register it as `oc10-sql` maybe? We already have a share driver with this name and there'll be a race condition which would affect which of the two drivers is served

aduffeck: done

pkg/share/manager/sql/conversions.go
aduffeck: The differences are big enough that my feeling is it's easier to keep them separate to be honest. Let me know if you want to go for it anyway and I'll see what I can do.

butonic (approved):

labkode (approved):

PR #1957, opened 2021-08-05
fix etag propagation on deletes

C0rby:
When deleting a file the etag propagation would skip the parent of the deleted file.


phil-davis (approved):
LGTM - simple change, lots of scenarios start passing, no regressions elsewhere.

labkode (approved):

PR #1960, opened 2021-08-05
fix updating shares in the memory share manager

C0rby:
This is a small change but if I'm correct, than this should "fix" a lot of tests in the expected failures file.


phil-davis (approved):
:medal_sports: 1 line of code added and 18 test scenarios start passing!

ishank011 (approved):

PR #1969, opened 2021-08-06
[WIP] connect ocs users endpoint to CS3 admin api

butonic:
We want to move user management out of the core of ocis as we focus on an external identity management. That being said, the tests need a working provisioning API, which we will move from the ocis ocs implementation to reva, dropping an ocs implementation from ocis altogether. It has to live with the owncloud flavoured webdav anyway and splitting these services up is no longer necessary and makes configuration of the proxy unnecessarily complex and error prone.

- [ ] add stubs for endpoints and methods that should be implemented
- signing-key
  - [ ] ~~detemine where to store it when it was generated? put it in a storage? user properties?~~
  -  no longer needed https://github.com/owncloud/ocis/issues/2374
  - [ ] how can the ocis proxy access it?
- [ ] how can we differentiate users from admins in reva (ocis has roles) @ishank011 maybe? Some endpoints need to check if the user has adimn powers as not all user provisioning endpoints are allowed to be used by mere mortals.
- [ ] unit tests!

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/cloud/user/user.go
+++ b/internal/http/services/owncloud/ocs/handlers/cloud/user/user.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 
 	"github.com/cs3org/reva/internal/http/services/owncloud/ocs/response"
+	"github.com/cs3org/reva/pkg/rhttp/router"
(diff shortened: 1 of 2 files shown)


butonic:
closing until the admin api exists

PR #1970, opened 2021-08-07
Enable fetching plugins from github repository

jimil749:
This PR adds following changes:

1. **Feature**: Enables fetching reva-plugins from github, Users can provide following configuration:
```
[grpc.services.userprovider]
driver=https://github.com/jimil749/json
```
This configuration downloads the plugin source code into `/var/reva/tmp/ext/<pluginType>/<driver>`, compiles the code and loads it onto reva at runtime. Repo used for testing: [https://github.com/jimil749/json](https://github.com/jimil749/json)

2. **Fix**: Use Struct based approach to encode and decode context kv pairs. Instead of sending `map[interface{}]interface{}` over RPC, we extract kv pairs into following struct:

```go
type Ctx struct {
    User   []*userpb.User
    Token  []string
}
```
and send this struct over RPC.

PS: @ishank011 I did not test having github repo w/o go.mod because the traditional `go get` would be deprecated in the further go versions. If we still want to support plugins w/o go modules, we'd have to download the files into `GOPATH` for successful compilation.


pkg/plugin/loader.go
```diff
@@ -82,14 +87,59 @@ func checkDirAndCompile(pluginType, driver string) (string, error) {
 	return bin, nil
 }
 
+// downloadPlugin downloads the plugin and stores it into local filesystem
+func downloadAndCompilePlugin(pluginType, driver string) (string, error) {
+	driverURL := strings.TrimPrefix(driver, "https://")
```
jimil749, line 92: This is needed because of go-getter. It errors when url is of the type `https://github.com/jimil749/...`, it should be of the type: `github.com/jimil749/...`

pkg/plugin/context.go
```diff
@@ -0,0 +1,29 @@
...
+	User  []*userpb.User
+	Token []string
+}
+
+// GetContextKV retrieves context KV pairs and stores it into Ctx
+func GetContextKV(ctx context.Context) *Ctx {
+	ctxVal := &Ctx{}
+	m := appctx.GetKeyValuesFromCtx(ctx)
+	for _, v := range m {
+		switch c := v.(type) {
```
jimil749, line 21: Does this work? Not too sure. Just checking the type to populate `User`/`Token`

pkg/auth/rpc_auth.go
```diff
@@ -112,7 +113,18 @@ func (m *RPCServer) Configure(args ConfigureArg, resp *ConfigureReply) error {
 
 // Authenticate RPCServer Authenticate method
 func (m *RPCServer) Authenticate(args AuthenticateArgs, resp *AuthenticateReply) error {
-	ctx := appctx.PutKeyValuesToCtx(args.Ctx)
+	ctx := setContext(args.Ctx.Token, args.Ctx.User)
 	resp.User, resp.Auth, resp.Error = m.Impl.Authenticate(ctx, args.ClientID, args.ClientSecret)
 	return nil
 }
+
+func setContext(ctxToken []string, userCtx []*userpb.User) context.Context {
```
ishank011, line 121: Move this to plugin pkg

pkg/plugin/context.go
ishank011: I'd rather just use `ContextGetToken` and `ContextGetUser`

pkg/auth/rpc_auth.go
jimil749: Thought of doing that, but it leads to cyclic dependencies b/w the user and plugin package. (User imports the plugin package for registering and plugin package would import the user package for the `ContextGetUser` and `ContextSetUser` methods). What do you think about moving all the user-context methods into a separate package, maybe `userctx`? This would involve updating all the packages that use usercontext!

pkg/plugin/context.go
jimil749: :+1:

pkg/auth/rpc_auth.go
ishank011: Let's have the methods to read and set ctx in `pkg/plugin/ctx`

go.sum
```diff
@@ -1,5 +1,6 @@
 bou.ke/monkey v1.0.2 h1:kWcnsrCNUatbxncxR/ThdYqbytgOIArtYWqcQLQzKLI=
```
ishank011, line 1: Please run `go mod tidy` once

pkg/plugin/loader.go
```diff
@@ -82,14 +87,59 @@ func checkDirAndCompile(pluginType, driver string) (string, error) {
...
+	}
+
+	return true
+}
+
 // Load loads the plugin using the hashicorp go-plugin system
 func Load(pluginType, driver string) (*RevaPlugin, error) {
+	var bin string
+	var err error
 	if isAlphaNum(driver) {
```
ishank011, line 129: The in-memory plugins can have hyphens and underscores, so let's add that to the regex

pkg/plugin/loader.go
```diff
@@ -82,14 +87,59 @@ func checkDirAndCompile(pluginType, driver string) (string, error) {
...
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return false
+	}
+
+	return true
+}
+
 // Load loads the plugin using the hashicorp go-plugin system
 func Load(pluginType, driver string) (*RevaPlugin, error) {
+	var bin string
```
ishank011, line 127: Move lines 127-142 to a different method maybe?

pkg/plugin/loader.go
```diff
@@ -82,14 +87,59 @@ func checkDirAndCompile(pluginType, driver string) (string, error) {
...
+}
+
+// isValidURL tests a string to determine if it is a well-structure URL
+func isValidURL(driver string) bool {
+	_, err := url.ParseRequestURI(driver)
+	if err != nil {
+		return false
+	}
+
+	u, err := url.Parse(driver)
```
ishank011, line 117: Why do we need the Parse check after we call ParseRequestURI?

pkg/auth/rpc_auth.go
jimil749: That works, but wouldn't lead to some code duplication?

pkg/plugin/loader.go
ishank011: This isn't really good. We need to support other protocols such as git as well. It even says there in the README.

> Protocols are used to download files/directories using a specific mechanism. Example protocols are Git and HTTP.

We can pass the getters to the client https://github.com/hashicorp/go-getter/blob/main/client.go#L144 https://github.com/hashicorp/go-getter/blob/main/get.go#L67-L75

pkg/auth/rpc_auth.go
ishank011: How so?

jimil749: Are you suggesting to *move* the methods from user pkg to `pkg/plugin/ctx`? Because we already have methods to Get and Set contexts in `user` pkg and having the same methods in `pkg/plugin/ctx` would lead to duplication!

pkg/plugin/loader.go
jimil749: With the current mechanism, we are supporting Git protocol. The URL of the type `github.com/jimil749/json` would be turned into Git URL by go-getter. The go-getter uses Git protocol to fetch the repo.

> The concept of a detector automatically turns invalid URLs into proper URLs. For example: "github.com/hashicorp/go-getter" would turn into a Git URL. Or "./foo" would turn into a file URL

Reason to trim `https://` was because `github.com/jimil/...` is not a valid URL (w/o the https:// part), hence the url checker would fail if one entered URL of that kind.

We could get rid of this by changing the way we check the URL.

pkg/plugin/loader.go
jimil749: We use Parse to check if scheme and host exist. `ParseRequestURI` allows strings like `test:test` to pass.

pkg/plugin/loader.go
ishank011: The user can pass paths beginning with other protocols, eg. http, git, file. We'll have to trim all these separately. Let's try the approach in the above comment

pkg/auth/rpc_auth.go
ishank011: No it'll have two methods.

```
func GetContextStruct(ctx context.Context) *Ctx
```

which calls ContextGetToken and ContextGetUser, and

```
func GetContext(ctxStruct *Ctx) context.Context
```

which calls ContextSetToken and ContextSetUser.

jimil749: Ahh yes! Was a bit confused. Sorry about that :sweat_smile:

pkg/plugin/loader.go
jimil749: Ahhh, so we're supporting protocols other than git! User can just add a prefix as per their choices.

ishank011: Also, I didn't go through how the library's clients store the transport, but let's ensure that we don't leak fds.

examples/plugin/plugin.toml
```diff
@@ -9,7 +9,7 @@ address = "0.0.0.0:19000"
 userprovidersvc = "localhost:19000"
 
 [grpc.services.userprovider]
-driver = "./json"
+driver = "https://github.com/jimil749/json"
```
ishank011, line 12: Also, just to demo the multiple ways of adding plugins, let's have one local and one hosted remotely

jimil749: Sure, I'll do that.

pkg/auth/rpc_auth.go
jimil749: There needs to be a small change in the methods. Because `rpc_user.go`, which belongs to the **user** package, will import `GetContextStruct` and `SetContext` methods (for encoding and decoding contexts), we can't have these methods call `ContextSetUser` and `ContextGetUser`. (Cyclic dependencies!). That's why I thought of moving the user context methods out in separate package.

jimil749: Unfortunately having methods to set and read ctx in `pkg/plugin/ctx` won't do us good because of `rpc_user.go`'s dependency on it. So if it is feasible I can send a PR to move out user context methods into a separate pkg.

pkg/plugin/loader.go
jimil749: So @ishank011, I don't think we'd need to pass our own Getter, because all the protocols like git, http, file are supported out of the box and are detected based on the `Src` field. For instance, if the url is `github.com/abc/xyz` -> git protocol or if url is `./foo` -> file protocol. This is done by the go-getter client. We just need to pass the url and rest would be taken care by the `getter.Get()` method!

ishank011: I didn't mean writing our own getter implementation, but just passing the in-built ones when instantiating the client. The problem with the current method is that we need to anticipate the protocols and trim all of those. If we didn't have to do that, then it would have been perfectly fine. So rather than the trimming, I'd prefer that we just make it known to the client.

pkg/auth/rpc_auth.go
ishank011: We can do that then

jimil749: Is it fine if I send a separate PR for that?

ishank011: Yes please

jimil749: #1982

pkg/plugin/loader.go
jimil749: We don't actually need to pass the in-built getter because by default `go-getter` already uses those if not specified in the client (https://github.com/hashicorp/go-getter/blob/main/client_option.go#L29). So if we need to support multiple protocols and also not use trimming, we could ask the users to specify urls like: `git::https://github.com/repo/plugin` which would force the protocol desired by the user.

jimil749:
I've pushed a couple of changes, let me know what do you think about them:

1. Go-getter: Now, instead of trimming, we ask user to specify the protocol that they want to use. They just need to append the url with the required protocol. So for git protocol, correct configuration would be `git::https://github.com/jimil749/json`.
2. Context: Use `SetContext` and `GetContextStruct` from `plugin` pkg to set and get contexts.
3. Compilation: Add `cmd.Dir` field to run the build command in the directory where the source code lies. This is because building a go project with `go.mod` from the current Reva directory is not possible. 

Due to (3) the user needs to provide absolute path to the source in order to compile local plugins.

pkg/plugin/loader.go
ishank011: Doesn't look like a good option to me. We ask the use to specify the protocol, and then specify another to override it. I'd still prefer passing the getters at init.

jimil749: Sure, just want to know that how would the user configure the plugins then. Could you provide an example of diff types of URLs/Protocol configuration?

ishank011: Sure so
```
https://github.com/repo/plugin
http://github.com/repo/plugin
s3://my-s3-deployment.com/bucket/repo/plugin
file:///root/repo/plugin.git
```

jimil749: Ok, so the only ambiguity is b/w `git` and `http` protocol. Even if I pass the getters as mentioned [here](https://github.com/hashicorp/go-getter/blob/main/get.go#L67-L75) (which I think we don't need to because those are used by default if no getter is passed while instantiating the client), there would be ambiguity b/w git and http protocol. The correct way to use `git` protocol is by passing url like `github.com/jimil749/json` and not `http://github.com/jimil749/json` (wherein the getter would use http protocol). Sorry, but I am still confused as in what would passing the getter to Client achieve?

PR #1971, opened 2021-08-07
Documentation for runtime-plugin

jimil749:
This PR adds basic developer manual/documentation for runtime plugins. Let me know if I missed something or if there's any scope for improvement! :sweat_smile: 

### DIFF
--- a/docs/content/en/docs/concepts/plugins.md
+++ b/docs/content/en/docs/concepts/plugins.md
@@ -1,0 +1,132 @@
+---
+title: "Runtime Plugins"
+linkTitle: "Runtime Plugins"
+weight: 10
+description: >
+  Guide for developing runtime plugin drivers for Reva
+---
+
+## Reva Plugins
+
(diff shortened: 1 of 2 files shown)


docs/content/en/docs/concepts/plugins.md
```diff
@@ -0,0 +1,132 @@
...
+    "github.com/hashicorp/go-plugin"
+)
+
+// Assume this implements the user.Manager interface
+type Manager struct{}
+
+// Handshake hashicorp go-plugin handshake
+var Handshake = plugin.HandshakeConfig{
+	ProtocolVersion:  1,
+	MagicCookieKey:   "BASIC_PLUGIN",
```
labkode, line 58: @jimil749 would make sense that the protocol version and Magic* contents are constants defined in the Reva plugin package that can be used?

jimil749: Yes, that would be in-fact better alternative. The developer can then just import the package and use the exported consts. :+1:

labkode: Awesome

jimil749: I'll update the documentation once we get #1970 merged. (There might be more changes)

labkode (approved):

PR #1982, opened 2021-08-10
Move user context methods into separate package

jimil749:
This PR moves the user-context methods out of the user pkg into a separate `userctx` package. 


ishank011 (commented):
Let's add the token methods to this pkg as well? Maybe we'll have to rename it

jimil749:
@ishank011 sorry for the late turnaround, VSCode gave me a lot of trouble this afternoon!

jimil749:
I see that the tests are failing. Will take a look

PR #1984, opened 2021-08-11
Traces: Use OpenTelemetry

refs:
What is [OpenTelemetry](https://opentelemetry.io/docs/concepts/what-is-opentelemetry/)?

- [open standard](https://github.com/open-telemetry/opentelemetry-specification)
- sandbox CNCF project.
- formed through a merger of the OpenTracing and OpenCensus.

> OpenCensus and OpenTracing have merged to form OpenTelemetry, which serves as the next major version of OpenCensus and OpenTracing. OpenTelemetry will offer backwards compatibility with existing OpenCensus integrations, and we will continue to make security patches to existing OpenCensus libraries for two years.

There is a lot of outdated documentation as a result of this merger, and we think adopting the latest standard and libraries will benefit the project.


internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -63,10 +65,12 @@ const (
 
 // ns is the namespace that is prefixed to the path in the cs3 namespace
 func (s *svc) handlePathPropfind(w http.ResponseWriter, r *http.Request, ns string) {
-	ctx := r.Context()
-	ctx, span := trace.StartSpan(ctx, "propfind")
+	ctx := rtrace.Propagator.Extract(r.Context(), propagation.HeaderCarrier(r.Header))
```
butonic, line 68: isn't this done by https://github.com/cs3org/reva/pull/1984/files#diff-e970327f0efe1f62011a0842762a6c652f53979a5c0fcec42b72897e798ac646R305 already?

refs:
@labkode @ishank011 does it make sense I update all traces occurrences that use the `go.opencensus.io` lib? I see the EOS storage provider is rich with traces and metrics, we could tackle that in this PR or in a follow-up one. How does that sound?

There are still a few technicalities to solve, but I'd say this PR is in good track.

internal/http/services/owncloud/ocdav/propfind.go
refs: yes 🙈

labkode:
Hi @refs, yes we can take the EOS part later. I think we should go for OT as default.

refs:
@labkode I think the current state should be ready for review. However I still need to wrap my head around the metrics instrumentation.

labkode (approved):

PR #1985, opened 2021-08-11
Quota stubs

butonic:
The `owncloud` and `owncloudsql` drivers now read the available quota from disk to no longer always return 0, which causes the web UI to disable uploads.

It is just a *stub* because we currently cannot pass the actual folder that we want to query for quota.


wkloucek (approved):
LGTM

refs (approved):

ishank011 (approved):

PR #1988, opened 2021-08-12
remove fmt.Print statement from decomposedfs

wkloucek:
fixes https://github.com/owncloud/ocis/issues/2290

### DIFF
--- a/changelog/unreleased/fix-spaces-print-log.md
+++ b/changelog/unreleased/fix-spaces-print-log.md
@@ -1,0 +1,5 @@
+Bugfix: Remove fmt.Print statement from decomposedfs
+
+The logs have been spammed with messages like `could not create symlink for ...` when using the decomposedfs, eg. with the oCIS storage.
+
+https://github.com/cs3org/reva/pull/1988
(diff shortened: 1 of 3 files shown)


wkloucek:
@butonic what should we do here? There is no logging available. Can we just remove the Print statements?

butonic:
The correct implementation would check
1. if the symlink already exists
2. if it points to the correct node (if not overwrite the existing symlink)
3. if the symlink cannot be created ... log an error ... hah the trick is https://github.com/cs3org/reva/blob/b1d57b9313d743e3afdf0490b93c6f45e57559d0/pkg/auth/manager/ldap/ldap.go#L113

it will use the default log configuration, but at least it is a structured log message.

and the left over case really is an error the admin should tackle, because the correct symlink cannot be created.

wkloucek:
> The correct implementation would check
> 
> 1. if the symlink already exists
> 2. if it points to the correct node (if not overwrite the existing symlink)
> 3. if the symlink cannot be created ... log an error ... hah the trick is https://github.com/cs3org/reva/blob/b1d57b9313d743e3afdf0490b93c6f45e57559d0/pkg/auth/manager/ldap/ldap.go#L113
> 
> it will use the default log configuration, but at least it is a structured log message.
> 
> and the left over case really is an error the admin should tackle, because the correct symlink cannot be created.

Doesn't sound like just replacing the Print with a Log. Could you please take care?

butonic:
closing in favor of https://github.com/cs3org/reva/pull/1992

PR #1989, opened 2021-08-12
implement url translation for legacy urls

C0rby:
First draft of the URL translation for OC10 URLS.
This implements a redirect from the old OC10 URL format to the current format we use in oCIS web. Currently only the public link and the old files app URL work because we don't do any file lookups by id yet. If this needs to be supported we would need to figure out how to lookup file based on their OC10 id.

To test this you need to apply this diff to oCIS and add a replace for reva to the go.mod file:

```diff
diff --git a/proxy/pkg/proxy/proxy.go b/proxy/pkg/proxy/proxy.go
index b066e9631..1d73b08b5 100644
--- a/proxy/pkg/proxy/proxy.go
+++ b/proxy/pkg/proxy/proxy.go
@@ -328,6 +328,10 @@ func defaultPolicies() []config.Policy {
                                        Endpoint: "/index.php/",
                                        Backend:  "http://localhost:9140",
                                },
+                               {
+                                       Endpoint: "/apps/files/",
+                                       Backend:  "http://localhost:9140",
+                               },
                                {
                                        Endpoint: "/data",
                                        Backend:  "http://localhost:9140",
```

### DIFF
--- a/changelog/unreleased/legacy-path-translations.md
+++ b/changelog/unreleased/legacy-path-translations.md
@@ -1,0 +1,5 @@
+Enhancement: Add redirects from OC10 URL formats
+
+Added redirectors for ownCloud 10 URLs. This allows users to continue to use their bookmarks from ownCloud 10 in ocis.
+
+https://github.com/cs3org/reva/pull/1989
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -153,7 +153,7 @@ func (s *svc) Close() error {
 }
 
 func (s *svc) Unprotected() []string {
-	return []string{"/status.php", "/remote.php/dav/public-files/"}
+	return []string{"/status.php", "/remote.php/dav/public-files/", "/apps/files/", "/index.php/f/", "/index.php/s/"}
 }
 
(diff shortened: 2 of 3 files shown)


C0rby:
@labkode @butonic, this PR implements the OC10 URL redirect to oCIS.
Currently only  `index.php/s/<token> (public links)` and  `apps/files/?dir=/Documents&fileid=32` are supported. But for the second URL only the `dir` parameter is used.

Is the lookup by `fileid` required? If so, how should we implement this? Do we need to access the OC10 database or is that information accessible  somewhere in reva?

C0rby:
To move forward with this I would propose to merge this PR in this state and if necessary implement the fileid based lookup later on.

This PR will allow the following:

* `oc10.example.com/apps/files/?dir=/Photos&fileid=10` -> `ocis.example.com/#/files/list/all/Photos`
* `oc10.example.com/s/TttylAos50jEApg` -> `ocis.example.com/#/files/public/list/TttylAos50jEApg`

Also I think the lgtm bot alert can be ignored since this is not an open redirect.
The target URL is always fixed to the configured public URL.
So the user will either get redirected to:
* a login
* the requested path
* a not found page

refs (approved):

labkode (approved):

PR #1991, opened 2021-08-13
Remove references when declining shares

C0rby:
Implemented the removal of share references when a share gets declined.
Now when a user declines a share it will no longer be listed in their `Shares` directory.

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -26,6 +26,7 @@ import (
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	collaboration "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	typesv1beta1 "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/errtypes"
 	"github.com/cs3org/reva/pkg/rgrpc/status"
@@ -126,6 +127,8 @@ func (s *svc) RemoveShare(ctx context.Context, req *collaboration.RemoveShareReq
 	if err != nil {
 		return nil, errors.Wrap(err, "gateway: error calling RemoveShare")
 	}
+
+	s.removeReference(ctx, share.ResourceId)
 
 	// if we don't need to commit we return earlier
 	if !s.c.CommitShareToStorageGrant && !s.c.CommitShareToStorageRef {
@@ -324,7 +327,7 @@ func (s *svc) UpdateReceivedShare(ctx context.Context, req *collaboration.Update
 			}
 			return rsp, nil
 		} else if req.Field.GetState() == collaboration.ShareState_SHARE_STATE_REJECTED {
-			// Nothing more to do, return the original result
+			s.removeReference(ctx, res.Share.Share.ResourceId)
 			return res, nil
 		}
 	}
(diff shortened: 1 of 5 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -336,7 +340,7 @@ func (s *svc) UpdateReceivedShare(ctx context.Context, req *collaboration.Update
 	}, nil
 }
 
-func (s *svc) createReference(ctx context.Context, resourceID *provider.ResourceId) *rpc.Status {
+func (s *svc) createReference(ctx context.Context, resourceID *provider.ResourceId, delete string) *rpc.Status {
```
refs, line 343: What are the semantics of the `delete` parameter?

refs: ohh I see, it signals the storageprovider to remove the reference that we just created? I think I need more context to fully grasp this one ^^

C0rby: That was just a hacky way to implement the Deletion without having a `RemoveReference` method in the CS3 API. This is not the finished code. :sweat_smile:

refs: oooooooooohh oh I see, sorry I was too fast on having a look at this while in wip

butonic (dismissed):

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -336,6 +339,70 @@ func (s *svc) UpdateReceivedShare(ctx context.Context, req *collaboration.Update
...
+		Opaque: &typesv1beta1.Opaque{
+			Map: map[string]*typesv1beta1.OpaqueEntry{
+				// This signals the storageprovider that we want to delete the share reference and not the underlying file.
+				"deleting_shared_resource": {},
+			},
+		},
+		Ref: &provider.Reference{Path: sharePath},
+	}
+
+	deleteResp, err := homeProvider.Delete(ctx, deleteReq)
```
refs, line 391: Just a question, couldn't we use the gateway client here instead of the `homeProvider`? This implies we know beforehand the reference is handled by the home storage provider, but that could not be the case, as there could be multiple storage providers in play at any given point in the future 🤔 

Also the gateway client will take care of resolve the reference and find the right storage provider for any give request.

refs (changes requested):
Just a remark that I would like clarified before merging, but conceptually 👍 LGTM

internal/grpc/services/gateway/usershareprovider.go
C0rby: Since we are always requesting the home path and then lookup the provider by the path prefixed with the home path we will always have the correct provider. It might not be called home provider in the future but it will be the one which can handle the request.

We could theoretically use the gateway client but we still would have to lookup the home path and then build the share path (/home/Shares/<sharename>) since we want to delete this reference and not the underlying file/folder. So using the gateway client would work it just wouldn't save us anything IMO.

refs: Good point; I've been messing around with the storage and understand the problem at hand.

I was trying to tackle this from another angle and therefore saving some requests. We could list the [received shares that yields a ListReceivedSharesResponse](https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.ListReceivedSharesResponse), iterating over the [received shares](https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.ReceivedShare) we then have access to the `resource_id` [from the share element](https://cs3org.github.io/cs3apis/#cs3.sharing.collaboration.v1beta1.Share), but the problem is that the `resource_id` is the original file, which does not belong to the user, and, as per the oCIS storage, does not have any mean to find any reference node. So we're back to your solution being the correct one and getting the node based on the path, which explains why we need the home mount point and where the shares folder is located.

LGTM 👍

refs (approved):

labkode (approved):

PR #1992, opened 2021-08-13
Check if symlink exists instead of spamming the console

butonic:
The logs have been spammed with messages like `could not create symlink for ...` when using the decomposedfs, eg. with the oCIS storage. We now check if the link exists before trying to create it.

deprecates https://github.com/cs3org/reva/pull/1988

### DIFF
--- a/pkg/storage/utils/decomposedfs/tree/tree.go
+++ b/pkg/storage/utils/decomposedfs/tree/tree.go
@@ -33,6 +33,7 @@ import (
 	"github.com/cs3org/reva/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/pkg/ctx"
 	"github.com/cs3org/reva/pkg/errtypes"
+	"github.com/cs3org/reva/pkg/logger"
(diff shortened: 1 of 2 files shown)


pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -169,6 +169,40 @@ func (t *Tree) Setup(owner string) error {
 	return nil
 }
 
+// linkSpace creates a new symbolic link for a space with the given type st, and node id
+func (t *Tree) linkSpace(spaceType, spaceID, nodeID string) {
+	spacesPath := filepath.Join(t.root, "spaces/"+spaceType, spaceID)
```
C0rby, line 174:
```suggestion
	spacesPath := filepath.Join(t.root, "spaces", spaceType, spaceID)
```

C0rby (changes requested):

pkg/storage/utils/decomposedfs/tree/tree.go
butonic: 🤦

labkode (approved):

PR #1993, opened 2021-08-13
fix owncloudsql GetMD

butonic:
The GetMD call internally was not prefixing the path when looking up resources by id.

cc @aduffeck


glpatcern (approved):
Looks like a trivial fix, let's go ahead :)

PR #1994, opened 2021-08-13
userprovider owncloudsql

butonic:
I am currently using glauth and the graphapi app to use my owncloud10 instance as a user provider for ocis using the ldap user provider ... while that works, it adds significant lag, to a point where listing shares takes ages when a lot of user have to be fetched by claim.

So, I sat down and implemented a small owncloudsql user provider. It works ok, and way faster then my current setup.

> We added a new backend for the userprovider that is backed by an owncloud 10 database. By default the `user_id` column is used as the reva user username and reva user opaque id. When setting `join_username=true` the reva user username is joined from the `oc_preferences` table (`appid='core' AND configkey='username'`) instead. When setting `join_ownclouduuid=true` the reva user opaqueid is joined from the `oc_preferences` table (`appid='core' AND configkey='ownclouduuid'`) instead. This allows more flexible migration strategies.


pkg/user/manager/owncloudsql/accounts/accounts.go
```diff
@@ -0,0 +1,225 @@
...
+
+// GetAccountGroups lasts the groups for an account
+func (as *Accounts) GetAccountGroups(ctx context.Context, uid string) ([]string, error) {
+	rows, err := as.db.QueryContext(ctx, "SELECT gid FROM oc_group_user WHERE uid=?", uid)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var group string
```
C0rby, line 212: This can be moved into the loop to reduce the scope.

pkg/user/manager/owncloudsql/owncloudsql.go
```diff
@@ -0,0 +1,180 @@
...
+func (m *manager) FindUsers(ctx context.Context, query string) ([]*userpb.User, error) {
+
+	accounts, err := m.db.FindAccounts(ctx, query)
+	if err == sql.ErrNoRows {
+		return nil, errtypes.NotFound("no users found for " + query)
+	} else if err != nil {
+		return nil, err
+	}
+
+	users := []*userpb.User{}
```
C0rby, line 133: This should be pre-allocated.
```suggestion
	users := make([]*userpb.User, 0, len(accounts))
```

pkg/user/manager/owncloudsql/owncloudsql.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, err
+	}
+	return groups, nil
+}
+
+func (m *manager) convertToCS3User(ctx context.Context, a *accounts.Account) (*userpb.User, error) {
+	u := &userpb.User{
+		Id: &userpb.UserId{
+			Idp:      m.c.Idp,
+			OpaqueId: a.UserID, // TODO rea
```
C0rby, line 160: Incomplete TODO?

C0rby (commented):
Kudos for writing tests!

pkg/user/manager/owncloudsql/accounts/accounts.go
butonic: wouldn't that add an allocation for every iteration? I wanted to reuse the variable, because it will be copied when appending it to groups.

pkg/user/manager/owncloudsql/owncloudsql.go
butonic: leftover

butonic:
remove leftover todo

pkg/user/manager/owncloudsql/accounts/accounts.go
C0rby: No, IIRC there will be only one allocation and the compiler can apply more optimizations since it knows that the value will not be referenced outside of the loop.

C0rby: Here this is not exactly the same code but should behave the same:
```golang
package loop

import (
	"strconv"
)

func declareOutside() []string {
	results := make([]string, 0, 32)
	var s string
	for i := 0; i < 32; i++ {
		s = strconv.Itoa(i)
		results = append(results, s)
	}
	return results
}

func declareInside() []string {
	results := make([]string, 0, 32)
	for i := 0; i < 32; i++ {
		var s string
		s = strconv.Itoa(i)
		results = append(results, s)
	}
	return results
}
```
```
go test -bench=. -benchmem
goos: linux
goarch: amd64
pkg: example.com/loop
cpu: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
BenchmarkOutside-8   	 3056689	       395.5 ns/op	     512 B/op	       1 allocs/op
BenchmarkInside-8    	 3067557	       393.0 ns/op	     512 B/op	       1 allocs/op
PASS
ok  	example.com/loop	3.214s
```

butonic: Oh cool!

C0rby (approved):

labkode (approved):

PR #2006, opened 2021-08-18
owncloud/ocs: Move to go-chi/chi based URL routing

### LINKED ISSUE #1986: use chi router instead of ShiftPath
I stumbled over https://benhoyt.com/writings/go-routing/ which examines the different ways routing can be done in golang.

IMO implementing the ocs / ocdav and other http endpoints with the ShiftPath approach makes it hard to understand what code is going to be executed for a given URL. We are using chi in some ocis services, eg the [graph](https://github.com/owncloud/ocis/blob/master/graph/pkg/service/v0/service.go#L31-L54) and our [ocs](https://github.com/owncloud/ocis/blob/d63a79a0ba73a5b75f0a381362aa52f7c6ae7c25/ocs/pkg/service/v0/service.go#L77-L135) implementation. We plan to fully move the latter into reva. An overview of what only the additional routing part looks like with shift path can be found in https://github.com/cs3org/reva/pull/1969/files#diff-6462df320ad761b651f59bdd74b3bd9123b0630538b7af253378a039a822b7f6R72-R151.

For readability (and performance) I'd vote to switch to chi. It has no further dependencies, supports quite a few middlewares and wrapping things like the opentelemetry api [is easy](https://fernando-bandeira.medium.com/building-apis-with-go-part-3-instrumentation-and-error-handling-daba9385e3ec#9d1d).

But I'd like to get an opinion from @labkode @ishank011 @refs @C0rby and @rhafer before moving the user provisioning part from ocis to reva. We could make the switch as part of that work as it will touch a lot of the code anyway.

rhafer:
This should help with code readability and make it easier to figure out
 which code is executed for which URL.
    
Closes #1986


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/cs3org/go-cs3apis v0.0.0-20210802070913-970eec344e59
 	github.com/eventials/go-tus v0.0.0-20200718001131-45c7ec8f5d59
 	github.com/gdexlab/go-render v1.0.1
+	github.com/go-chi/chi v1.5.4
```
refs, line 23: could you fetch the latest version of Chi?

`go get github.com/go-chi/chi/v5` should do it

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -537,6 +467,23 @@ func (h *Handler) updateShare(w http.ResponseWriter, r *http.Request, shareID st
...
+	shareID := chi.URLParam(r, "shareid")
+	if h.isPublicShare(r, shareID) {
+		h.removePublicShare(w, r, shareID)
+		return
+	}
+	h.removeUserShare(w, r, shareID)
+}
+
+func (h *Handler) ListShares(w http.ResponseWriter, r *http.Request) {
+	if h.isListSharesWithMe(w, r) {
```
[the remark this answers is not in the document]
rhafer: Done

go.mod
rhafer: Done. Which makes me wonder.  Should we update ocis to chi/v5 as well? It's at some v4 release currently.

refs: Yes, we [should update it as well](https://github.com/owncloud/ocis/blob/master/go.mod#L26), good catch

C0rby: Yes, we should do that but we would need to update the protobuf plugins first, they are still using the old go-chi version.
I already tried. :see_no_evil:

rhafer: Done -> https://github.com/owncloud/ocis/pull/2429

internal/http/services/owncloud/ocs/response/response.go
```diff
@@ -155,7 +156,7 @@ func WriteOCSResponse(w http.ResponseWriter, r *http.Request, res Response, err
 	version := APIVersion(r.Context())
 	m := statusCodeMapper(version)
 	statusCode := m(res.OCS.Meta)
-	if version == "v2.php" && statusCode == http.StatusOK {
+	if version == "2" && statusCode == http.StatusOK {
```
labkode, line 159: is this expected?

rhafer: Yeah, at least that is how the counterpart on ocis behave(s|d). From https://github.com/owncloud/ocis/blob/master/ocs/pkg/service/v0/response/response.go#L77 : 

```
    if version == ocsVersion2 && statusCode == http.StatusOK {
        rsp.OCS.Meta.StatusCode = statusCode
    }
    return nil
```
(where ocsVersion2 is defined as `"2"`)

butonic: yep ocs v2 maps the application level status codes to transport level status codes: the ocs ok status will become a 200 Ok HTTP status.

butonic (approved):
Yep, makes sense!

refs (approved):

labkode (approved):
Looks good to me.
Let's another approval for the OC team to be merged.

refs (approved):

refs:
LGTM

Issue #2001, opened 2021-08-17
make share manager single source of truth


butonic:
# Sharing
This document proposes changes on managing and accessing shares in reva. The goal is to streamline the information and make the process more robust.

## Summary
- The share managers become the single source for all share information, including names / mountpoints
- Shares are no longer recorded on the storage using the storage providers
- Accepting shares no longer creates a reference in a storage provider but in the share manager
- The sharesstorageprovider provides both a file tree exposing all accepted shares as well as a separate space for each received share

## Current implementation

Basic information about shares - i.e. who shared what with whom - are currently managed by the share manager. Additional information such as the names and the mountpoints of the shares are physically kept on the storage as references in the according locations and those references are being maintained by the storage providers.

This split of information makes it difficult to gather the mountpoints of all shares of user and is prone to inconcistencies between the data in the share manager and on the storage.
- There is no way around looking up the name of the accepted share, either in the storage provider or in the share manager: we need it to deduplicate shares with the same name.
- There is significant overhead when having to stat all shares to get the actual name (or path) of the accepted share

## Proposed change

With this proposal shares will no longer be "mounted" on the storage. Instead the names and mountpoints move into the share manager alongside the basic share information and all references disappear from the storage (they can optionally be sent there as well and the storage driver can decide how to represent them, eg. as symlinks).

A `SharesStorageProvider` provides a file tree of all shares the currently logged in user has received and that tree can be mounted by the gateway at an appropriate location, eg.: `/shares` or `/home/shares`. In addition the `SharesStorageProvider` also takes care of exposing received shares as spaces.

With this change the manager becomes the single source of truth regarding shares. It can thus answer both:
- the question which resources a specific user shared with others, as well as 
- the question which shares a specific user received, including their acceptance state and mount points.

The storage providers and drivers no longer have to keep track of shares, in fact they do not have to know shares exist in the first place.

This becomes optional, as was intended from the beginning, by configuring the gateway to (not) forward grant requests and create reference requests to the storage providers. See `commit_share_to_storage_grant` and `commit_share_to_storage_ref`. Administrators can still configure the gateway to forward grants to storage providers and they can still apply acls to share a resource on the os layer. The source of truth however is the share manager (think of it as periodically setting the ACLs on the storage, overwriting existing ACLs).

## Expected Results
The `SharesStorageProvider` reads all metadate from the share manager and can provide access to all received shares by returning CS3 references. 

### Share indicators require asking the share manager
[...]

PR #1846, opened 2021-06-30
[WIP] Sharestorageprovider oc10 sm

aduffeck:
This is WIP to access oc10 shares from ocis.

\cc @butonic

PR #2023, opened 2021-08-31
Sharestorageprovider oc10 sm

butonic:
continuation of https://github.com/cs3org/reva/pull/1846 and https://github.com/cs3org/reva/pull/1584
- [x] rebased and fixed mocks
- [x] prototype share manager changes as described in https://github.com/cs3org/reva/issues/2001 
- [ ] check if the deleted webdav / ocm code is breaking things ... and what it was supposed to handle


aduffeck:
Following our discussion these are the remaining tasks to implement:

- [ ] Change the ocs API to check for and handle collisions with already existing shares or other nodes on the storage when accepting shares
- [ ] Extend the sharesstorageprovider to consider all shares pointing to the mountpoint in question, e.g. when renaming or deleting (declining) a share
- [ ] Make the sql share manager completely hide the fact that accepted group shares can be represented as child shares in the database

We also concluded that spaces support will not be implemented as part of this PR.

refs:
[`2b9adac` (#2023)](https://github.com/cs3org/reva/pull/2023/commits/2b9adac6e630ffcd5c4f84837d2bc5f38549f52b) should address points 1 and 2 from the last comment.

There is however a tradeoff that we made. When there are 2 shares to the same resource, one is a group share and the other a user share, the group share is excluded from the results, so there are not 2 folders to the same location, this is as we discussed.

However, when listing all shares (in the sahred with me tab) we should list both received shares, as the user might want to decline them (i.e update the share state). This is currently not possible since listing the shares (by navigating to `/Shares/`) uses the same endpoint as the shared with me endpoint, and there is no filtering options to hide shares which point to the same resource.

@aduffeck could you udate the todo list with a checkmark for both tasks? I for some reason am not able to do so :(

//cc @butonic

refs:
also please notice the replaces in the `go.mod` files. Pulling will most likely break your local instance unless you update the replaces 🙈

aduffeck:
@refs I'm not really following the problem with the same endpoint being used in `/Shares` and the shared-with-me endpoint. `/Shares` uses the sharesstorageprovider which talks to the share manager directly and can - and in fact does - merge the two shares to the same resource together, no?

refs:
> @refs I'm not really following the problem with the same endpoint being used in `/Shares` and the shared-with-me endpoint. `/Shares` uses the sharesstorageprovider which talks to the share manager directly and can - and in fact does - merge the two shares to the same resource together, no?

What I meant is when we list shares on "Shared with me":

grafik

I think it would be better if we list **all** of the shares, group and user shares alongside, even if the mount point is the same. And I'm comparing it to this view:

grafik

navigating to `/Shares` from the web ui actually renders the mount point, but here we should hide the group share (based in: user shares are more explicit, and just hide the less explicit shares, totally my choice, we can decide on this).

The issue comes that these 2 different pages use the same endpoint, and there are currently no filters in the CS3APIS to capture this behavior. What I want is to use the same endpoint and fetch different results depending on where the request was made. Maybe I'm missing an essential part elsewhere 🤷

refs:
> @refs I'm not really following the problem with the same endpoint being used in `/Shares` and the shared-with-me endpoint. `/Shares` uses the sharesstorageprovider which talks to the share manager directly and can - and in fact does - merge the two shares to the same resource together, no?

@aduffeck What I meant is when we list shares on "Shared with me":

grafik

I think it would be better if we list **all** of the shares, group and user shares alongside, even if the mount point is the same. And I'm comparing it to this view:

grafik

navigating to `/Shares` from the web ui actually renders the mount point, but here we should hide the group share (based in: user shares are more explicit, and just hide the less explicit shares, totally my choice, we can decide on this).

The issue comes that these 2 different pages use the same endpoint, and there are currently no filters in the CS3APIS to capture this behavior. What I want is to use the same endpoint and fetch different results depending on where the request was made. Maybe I'm missing an essential part elsewhere 🤷

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1299,11 +672,37 @@ func (s *svc) stat(ctx context.Context, req *provider.StatRequest) (*provider.St
...
+				_, _ = io.WriteString(etagHash, res.Info.Etag)
+			}
+			for _, child := range embeddedMounts {
+				childStatRes, err := s.stat(ctx, &provider.StatRequest{Ref: &provider.Reference{Path: child}})
+				if err != nil {
+					return &provider.StatResponse{
+						Status: status.NewStatusFromErrType(ctx, "stat ref: "+req.Ref.String(), err),
+					}, nil
+				}
+				_, _ = io.WriteString(etagHash, childStatRes.Info.Etag)
```
butonic, line 696: need to check childStatRes

butonic:
```
storage-sharing_1        | {"level":"error","service":"storage","pkg":"rgrpc","error":"Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'results AS\n\t\t(\n\t\t\tSELECT s.*, storages.numeric_id FROM oc_share s\n\t\t\tLEFT JOIN o' at line 1","time":"2021-09-28T10:41:09.102292283Z","caller":"/home/jfd/Repositories/reva/internal/grpc/services/usershareprovider/usershareprovider.go:193","message":"error listing received shares"}
```

mysql hates us ;-)

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -308,166 +305,30 @@ func (s *svc) getHome(_ context.Context) string {
...
-				Status: status.NewOK(ctx),
-				Protocols: []*gateway.FileDownloadProtocol{
-					{
-						Opaque:           opaque,
-						Protocol:         "simple",
-						DownloadEndpoint: ep,
-					},
-				},
-			}, nil
-		}
```
butonic, line 398: @ishank011 there are several places in the gateway that have special "webdav"  handling ... what for? AFAICT it is related to OCM. Unfortunately, there are no tests covering that right now. Well the oc10 core testsuite does cover federated sharing, but the drone ci does not set up a second instance ... 

While there are ways forward to cover OCM let's start with what these webdav handling is meant to do? Could you enlighten us?

ishank011: @butonic yep, the OCM share targets are saved as `webdav://$token@$idp?name=$path`. When OCM shares are accepted, the references have these targets instead of `cs3` for other shares.
When `checkRef` is called and it detects that the reference is of type 'webdav', the requests are then forwarded to the methods in the `webdavstorageprovider.go` file, and they actually make the webdav calls to the idp of the share creator.

butonic: @ishank ... um `$path`? so the references will break when the path changes? Why are we not using ids? Is OCM necessary for your current deployment?

butonic: IMO the share storage provider will return a cs3 reference which resolves to a yet to be implemented OCM storage provider. That can encapsulate the token and idp data necessary to access the remote instance.

butonic: I am currently looking at the `/examples/ocmd` setup. AFAICT it is deployed next to an existing reva deployment ... is it only starts the ocmd service on the http port ... so normal webdav is not available without an additional reva deployment.

butonic: I started a PR to clean up the toml configs and add an ocm config set in https://github.com/cs3org/reva/pull/2239

pkg/share/manager/json/json.go
```diff
@@ -393,7 +398,29 @@ func (m *mgr) ListReceivedShares(ctx context.Context, filters []*collaboration.F
...
+	// to the same resource. Leave the more explicit and hide the more explicit. In this case we hide the group shares
+	// and return the user share to the user.
+	filtered := make([]*collaboration.ReceivedShare, 0)
+	filtered = append(filtered, rss...)
+
+	for i := range rss {
+		for j := range rss {
+			if rss[i].Share.ResourceId.GetOpaqueId() == rss[j].Share.ResourceId.GetOpaqueId() {
+				if rss[i].Share.GetGrantee().GetType() == provider.GranteeType_GRANTEE_TYPE_GROUP && rss[j].Share.GetGrantee().GetType() == provider.GranteeType_GRANTEE_TYPE_USER {
+					if rss[i].State == rss[j].State {
```
ishank011, line 412: Not sure about this. If the permissions differ for the shares, shouldn't be return both?

internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -235,20 +237,55 @@ func (s *svc) getClient() (gateway.GatewayAPIClient, error) {
...
 	// the FilesHandler method invoked by a /dav/files/fileOwner where fileOwner
 	// is not the same as the logged in user. In that case, we'll treat fileOwner
 	// as the username whose files are to be accessed and use that in the
 	// namespace template.
 	u, ok := ctxpkg.ContextGetUser(ctx)
 	if !ok || !useLoggedInUserNS {
-		requestUserID, _ := router.ShiftPath(requestPath)
-		u = &userpb.User{
-			Username: requestUserID,
+		var requestUsernameOrID string
```
ishank011, line 248: I left a couple of comments in the ldap refactor PR https://github.com/cs3org/reva/pull/2133. Can you take a look at those?

butonic:
merged in edge as part of https://github.com/cs3org/reva/pull/2234

PR #2026, opened 2021-09-01
fix moving of a shared file between shares

C0rby:
As the share receiver, moving a shared file to another share was not possible.

### DIFF
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -1029,37 +1029,61 @@ func (s *svc) Move(ctx context.Context, req *provider.MoveRequest) (*provider.Mo
 		dshareName, dshareChild := s.splitShare(ctx, dp)
 		log.Debug().Msgf("srcpath:%s dstpath:%s srcsharename:%s srcsharechild: %s dstsharename:%s dstsharechild:%s ", p, dp, shareName, shareChild, dshareName, dshareChild)
 
-		if shareName != dshareName {
-			err := errtypes.BadRequest("gateway: move: src and dst points to different targets")
-			return &provider.MoveResponse{
-				Status: status.NewStatusFromErrType(ctx, "gateway: error moving", err),
-			}, nil
-
-		}
-
-		statReq := &provider.StatRequest{Ref: &provider.Reference{Path: shareName}}
-		statRes, err := s.stat(ctx, statReq)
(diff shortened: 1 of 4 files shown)


phil-davis:
Related issue https://github.com/owncloud/ocis/issues/2442

phil-davis (approved):
LGTM - fixes issue https://github.com/owncloud/ocis/issues/2442 and various test scenarios that depend on that issue.

butonic (changes requested):
@labkode correct me if I am wrong. The storageprovider is not supposed to deal with cross storageprovider move requests, right? Cross storage moves are a COPY and DELETE operation that will create a new resource with a new fileid. A cross storage move needs to be handled by the client, so that it can notify the user about the change.

That being said, we are planning to be able to move spaces between storage providers if they a able to import ids, which depends on the implementation.

@C0rby can you fix this in the ocdav layer instead of the CS3 gateway?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1089,42 +1113,42 @@ func (s *svc) Move(ctx context.Context, req *provider.MoveRequest) (*provider.Mo
...
 		}
 		return res, nil
 	}
 
-	srcP, dstP := srcList[0], dstList[0]
+	srcProvider, dstProvider := srcProviders[0], dstProviders[0]
 
 	// if providers are not the same we do not implement cross storage copy yet.
-	if srcP.Address != dstP.Address {
+	if srcProvider.Address != dstProvider.Address {
```
C0rby, line 1141: @butonic, If I understand correctly than this shouldn't be a cross storage copy/move. I didn't remove this check here.
Or is a move between two shares always considered a cross storage copy/move

butonic: ok, now I get it. This fixes the case for moving a file from one share into another share from the same storage, eg.
`/home/Shares/foo/readme.txt` -> `/home/Shares/bar/readme.md`, where `foo` and `bar` are located on the same storage provider.

@labkode then this is good to merge IMO.

butonic (approved):

internal/grpc/services/gateway/storageprovider.go
C0rby: Yes, exactly. :+1:

labkode (approved):

PR #2031, opened 2021-09-01
Add authentication provider field in the token

gmgigi96:


butonic:
Why do you need the auth_provider in the token?

labkode:
Hi @butonic, the goal with this is to be able to know from the token payload how the user was authenticated. We need this as we have some auth mechanism (like the `machine` type) that allows to impersonate the user and today you cannot differentiate if the token was generated by the physical user or because of an impersonation without scrapping logs. This helps for auditing purposes.

labkode:
@butonic please review as I'll like this one to be part of the new release for OCIS.

micbar:
@butonic

butonic:
> Hi @butonic, the goal with this is to be able to know from the token payload how the user was authenticated. We need this as we have some auth mechanism (like the `machine` type) that allows to impersonate the user and today you cannot differentiate if the token was generated by the physical user or because of an impersonation without scrapping logs. This helps for auditing purposes.
Hm, especially for impersonating users I would expect the identities of both: the impersonator and the user. That way every service can log both identities in an audit log. Furthermore, in the web ui you may want to show that the user is being impersonated, which also requires both identities in the token ... or a second token.
> 
> The auth_type value is just the driver used to authenticate: `basic`, `bearer` and so on.
This will require aggregating logs via spans.

What do you think about putting both identities in the token?

ishank011:
> @labkode What do you think about putting both identities in the token?

I agree with this. We already have the user object in the token and I think that should indicate whether an admin is impersonating some other user. Maybe we can have a custom `Id.OpaqueId` for such cases, and also indicate this in the `UserType` field?

gmgigi96:
Sounds good, but probably modifying the `UserType` with a custom one can
break something in the code, because this should be the one of the
impersonated user. Why don't we do the same you proposed, but using the
`User.Opaque` field, in which insert the user that impersonate?

PR #2034, opened 2021-09-03
Fail initialization if app is unsupported

glpatcern:
A mini PR to improve error handling when an app is not WOPI-compliant nor it is supported by the WOPI bridge extensions


labkode (approved):

labkode:
@glpatcern if this is ready just mark the PR as "Ready for Review" so it moves out of draft state.

glpatcern:
Indeed, it can be merged right away - then we still have your patches Hugo ;-)

PR #2050, opened 2021-09-07
add filter by sharetype in the ocs API

C0rby:
Added a query parameter to the OCS API to filter the received shares by type.


labkode (dismissed):
Looks good to me but I need @butonic, @refs to approve as well.

butonic:
The `share_type` parameter was introduced to oc10 with https://github.com/owncloud/core/pull/38053

butonic (dismissed):
could you add the filter to the sql driver in https://github.com/cs3org/reva/blob/eacfb8a0f3cbe9451b9863495b43bb10b07b709b/pkg/share/manager/sql/sql.go#L278
and maybe in https://github.com/cs3org/reva/blob/eacfb8a0f3cbe9451b9863495b43bb10b07b709b/pkg/cbox/share/sql/sql.go#L283 as well?

if not track it in a new issue.

C0rby:
@butonic @labkode, I added the filters to the sql share managers. Please have a look at my changes. Also I couldn't test these changes since I don't know how/what to setup for the sql share managers. If you have some documentation or some tips, that would be great.

C0rby:
> This pull request **introduces 2 alerts** when merging [8f0de6e](https://github.com/cs3org/reva/commit/8f0de6ee3723c705f619ca72c71246ce015ed3ec) into [f2109fc](https://github.com/cs3org/reva/commit/f2109fce3e7da12648149b0b86b4c952577bd20c) - [view on LGTM.com](https://lgtm.com/projects/g/cs3org/reva/rev/pr-d6c2804598135fb18b8106bec88e9d722ac6887a)
> 
> **new alerts:**
> 
>     * 2 for Useless assignment to local variable

I resolved this but the comment doesn't disappear and I also can't remove it.

rhafer (dismissed):
LGTM. At least I didn't find something obviously wrong here. (Take this with a grain of salt, as I am still pretty new to the code base)

butonic (approved):

pkg/cbox/share/sql/sql.go
```diff
@@ -464,3 +486,69 @@ func (m *mgr) UpdateReceivedShare(ctx context.Context, ref *collaboration.ShareR
 	rs.State = f.GetState()
 	return rs, nil
 }
+
+func groupFiltersByType(filters []*collaboration.Filter) map[collaboration.Filter_Type][]*collaboration.Filter {
+	sorted := make(map[collaboration.Filter_Type][]*collaboration.Filter)
```
refs, line 491: I believe `sorted` is a misnomer. The end result is not sorted but rather "bundled", where all values in `filters` have been put into boxes and maps are unsorted by design. In order to avoid false assumptions in the future here I'd rename `sorted` to `b` (for bundled) 🤷

pkg/cbox/share/sql/sql.go
```diff
@@ -464,3 +486,69 @@ func (m *mgr) UpdateReceivedShare(ctx context.Context, ref *collaboration.ShareR
...
+}
+
+func granteeTypeToShareType(granteeType provider.GranteeType) int {
+	switch granteeType {
+	case provider.GranteeType_GRANTEE_TYPE_USER:
+		return shareTypeUser
+	case provider.GranteeType_GRANTEE_TYPE_GROUP:
+		return shareTypeGroup
+	}
+	return -1
```
refs, line 505: why not using multiple returns and actually return an error here?

refs: or is this `-1` [here](https://github.com/cs3org/reva/pull/2050/files#diff-bee286ddbc20ada3dd35cb24990f6044dc9809ef6243519e28697ea3ca7be184R538) accepted?

refs (dismissed):
Just a pedantic comment from mine, nice PR and nice expected failures cleanup! Thanks!

pkg/cbox/share/sql/sql.go
C0rby: Yeah, true. I forgot to rename it because previously the method name was `sort*`.

butonic (dismissed):

refs (dismissed):

C0rby:
The tests fail now because of this PR https://github.com/cs3org/reva/pull/2072. Because of that PR rejected shares are not shown in `ListReceivedShares` responses and there is no way to re-accept or even list a rejected share. 

@labkode, how do we want to go forward with this? I think setting the filter in the ocdav layer would be better and then making that behavior configurable. Then an admin could choose if rejected shares should be shown or not.

cc @micbar @butonic

labkode:
Hi @C0rby that PR should do nothing about accepting or denying shares as we know them today. That PR handles a specific type of share that denies access (DENIAL). That type of share cannot be accepted or rejected by the receiver, the receiver basically can't do anything about it.

The filter set in that PR is to not shown this type of special shares to the receiving user, NOT to filter out reject shares.

Let me know if you need further clarifications.

C0rby:
> The filter set in that PR is to not shown this type of special shares to the receiving user, NOT to filter out reject shares.

I see, then I misunderstood this filter. I will rewrite my code then to reflect the correct behavior.

C0rby:
@labkode, how do I recognize such deny shares? 
Is a share a deny share when `DenyGrant` is set to true or how can I filter them out in the json and memory manager?

labkode:
Hi @C0rby, in the SQL manager we set a DenyShare as having `permissions=0`, so you can do the same on the other drivers, I think if semantically meaningful.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -545,7 +545,36 @@ func (h *Handler) listSharesWithMe(w http.ResponseWriter, r *http.Request) {
...
+		case int(conversions.ShareTypeUser):
+			filters = append(filters, share.UserGranteeFilter())
+		case int(conversions.ShareTypeGroup):
+			filters = append(filters, share.GroupGranteeFilter())
+		}
+	}
+
+	if len(shareTypes) != 0 && len(filters) == 0 {
+		// If a share_types filter was set for anything other than user or group shares just return an empty response
+		response.WriteOCSSuccess(w, r, []*conversions.ShareData{})
```
ishank011, line 573: Should we return an error? Might prove to be helpful to an admin

pkg/cbox/share/sql/sql.go
```diff
@@ -476,3 +502,72 @@ func (m *mgr) UpdateReceivedShare(ctx context.Context, ref *collaboration.ShareR
...
+	// That is why the filters got grouped by type.
+	// For every given filter type, iterate over the filters and if there are more than one combine them.
+	// Combine the different filter types using `AND`
+	var filterCounter = 0
+	for filterType, filters := range groupedFilters {
+		switch filterType {
+		case collaboration.Filter_TYPE_RESOURCE_ID:
+			filterQuery += "("
+			for i, f := range filters {
+				filterQuery += "item_source=?"
```
ishank011, line 542: `fileid_prefix` should be equal to `f.GetResourceId().StorageId` https://github.com/cs3org/reva/blob/master/pkg/cbox/share/sql/sql.go#L286

pkg/share/manager/json/json.go
```diff
@@ -360,19 +354,16 @@ func (m *mgr) ListShares(ctx context.Context, filters []*collaboration.Filter) (
...
-					if f.Type == collaboration.Filter_TYPE_RESOURCE_ID {
-						if utils.ResourceIDEqual(s.ResourceId, f.GetResourceId()) {
-							ss = append(ss, s)
-						}
-					}
+				continue
+			}
+			// check filters
+			for _, f := range filters {
+				if share.MatchesFilter(s, f) {
```
ishank011, line 365: Here we return the share if even a single filter matches (or relation) whereas in `ListReceivedShares`, it's an `and` relation. Why the difference?

ishank011 (commented):
Hi @C0rby this is really nice! Just left a couple of comments

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
ishank011: Okay, makes sense

pkg/share/manager/json/json.go
C0rby: Good catch, I rewrote the `ListReceivedShares` part and must have missed the other one. I'll change that.

pkg/cbox/share/sql/sql.go
ishank011: I think the resource ID filter is being applied twice in `ListReceivedShares`, once in the method itself and once in `translateFilters`. But in `translateFilters`, we don't check the storage ID. So that needs to be added there and the redundant check should be removed from the `ListReceivedShares` method

C0rby: Ah, now I got what you mean! Fixed it. :+1:

pkg/share/manager/json/json.go
C0rby: Ok, I think now I got the correct behavior done.
Filters of the same type form a disjuntion, a logical OR. Filters of separate type form a conjunction, a logical AND.
For example: (resource_id=1 OR resource_id=2) AND (grantee_type=USER OR grantee_type=GROUP)

C0rby:
Let me just add some unit tests for the new functions and then this PR is good to go. :)

C0rby:
It's ready now.

ishank011 (approved):

PR #2051, opened 2021-09-07
wopi: fix discovery urls

labkode:


glpatcern (approved):
It all makes sense, including the need for logging facility at init time...

wkloucek (approved):
Thanks, comes in very handy. I currently try to use OnlyOffice and it also has these query params...

PR #2053, opened 2021-09-08
Use safer defaults for TLS verification on LDAP connections

rhafer:
The LDAP client connections where hardcoded to ignore certificate validation
errors everywhere. This commit changes that to uses a secure default,
which can be overridden by the new config parameter 'insecure'.
Also the LDAP related test configs are updated to set that override for
the tests.

### DIFF
--- a/changelog/unreleased/ldap-tls-insecure.md
+++ b/changelog/unreleased/ldap-tls-insecure.md
@@ -1,0 +1,8 @@
+Enhancement: Safer defaults for TLS verification on LDAP connections
+
+The LDAP client connections were hardcoded to ignore certificate validation
+errors. Now verification is enabled by default and a new config parameter 'insecure'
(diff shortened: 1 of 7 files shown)


labkode (approved):
Thanks, great improvement!
We should probably run a `grep -i insecure` around the codebase :)

rhafer:
> We should probably run a `grep -i insecure` around the codebase :)

AFAICS those were the only remaining places where i was hardcoded.

C0rby (approved):
Awesome improvement!

rhafer:
Converted to draft for now. As I think this needs some adjustments in ocis.

labkode:
@rhafer please remove the draft state if you think is completed.

refs (dismissed):

labkode (approved):

PR #2054, opened 2021-09-08
fix response after deleting a share

C0rby:
Added the deleted share to the response after deleting it.


phil-davis:
Fixes https://github.com/owncloud/ocis/issues/721

labkode (approved):

Issue #2056, opened 2021-09-08
Stop build if build-only step fails


labkode:
To reduce resource consumption in the CI the running steps should be short-circuited if the build-only phase fails.
@wkloucek ?

It usually happens that we create a PR and then we need to provide the changelog. By having this we have a very quick feedback and we avoid keeping a CI job that will never make it to green.

phil-davis:
@dpakach can you get someone to have a look?

wkloucek:
Or just do it as a separate step and make all others depending on it. It takes only 7 seconds and has no dependencies.

Though one might work on a PR first and see if its green and add the change log before marking the PR as ready. This workflow would then definitely break.

PR #2057, opened 2021-09-09
[tests-only] [CI] Stop build if changelog step fails 

SwikritiT:
This PR creates changelog into a separate pipeline and makes other pipeline depend on it so that when this pipeline fails others won't run .
fixes: https://github.com/cs3org/reva/issues/2056

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -96,6 +96,7 @@ def main(ctx):
   # ocisIntegrationTests(6, [1, 4])     - this will only run 1st and 4th parts
   # implemented for: ocisIntegrationTests and s3ngIntegrationTests
   return [
+    changelog(),
     coverage(),
     buildAndPublishDocker(),
     buildOnly(),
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -266,6 +267,23 @@ def coverage():
     ]
   }
 
+def changeLog():
+  return {
+    "kind": "pipeline",
+    "type": "docker",
+    "name": "changelog",
+    "image": "registry.cern.ch/docker.io/library/golang:1.16",
```
phil-davis, line 275: IMO this needs a `"steps": [` section, which will need a "name" and the "image", "commands" and "environment".

phil-davis:
@SwikritiT I think that you can edit the PR title so it includes `[tests-only]` and then rebase and force push. Then the CI will no force there to be a changelog. The CI should keep running and go green.

The code looks good.

labkode (approved):

PR #2067, opened 2021-09-14
Extend AppRegistry and AppProvider

glpatcern:
Implementation in Reva of the extensions in https://github.com/cs3org/cs3apis/pull/145.

Some values have to be filled in from config, this will be targeted in a separate PR.


glpatcern:
@ishank011 this is now (finally) complete. I will test it tomorrow in our QA before merging (can't validate the tests yet as we need to merge the CS3APIs PR first).

ishank011:
Merged cs3org/cs3apis#145

pkg/app/registry/static/static.go
```diff
@@ -148,23 +148,31 @@ func (b *reg) ListProviders(ctx context.Context) ([]*registrypb.ProviderInfo, er
...
-			if _, ok := mimeTypes[m]; ok {
-				mimeTypes[m].AppProviders = append(mimeTypes[m].AppProviders, &t)
+			if _, ok := mtmap[m]; ok {
+				mtmap[m].AppProviders = append(mtmap[m].AppProviders, &t)
 			} else {
-				mimeTypes[m] = &registrypb.AppProviderList{AppProviders: []*registrypb.ProviderInfo{&t}}
+				mtmap[m] = &registrypb.MimeTypeInfo{
+					AppProviders: []*registrypb.ProviderInfo{&t},
+					Ext:          "", // TODO fetch from config
+					Name:         "",
```
ishank011, line 167: `Name` should be `m`

pkg/app/registry/static/static.go
```diff
@@ -148,23 +148,31 @@ func (b *reg) ListProviders(ctx context.Context) ([]*registrypb.ProviderInfo, er
...
+					Description:  "",
+					Icon:         "",
+				}
+				res = append(res, mtmap[m])
```
ishank011, line 171: Are you sure this works? As far as I can say `res` would just have a list of empty `MimeTypeInfo` objects, each containing only a single app in the `AppProviders` field. The `append` should happen after we've iterated over all the apps

internal/http/services/appprovider/appprovider.go
```diff
@@ -189,39 +293,24 @@ func (s *svc) handleOpen(w http.ResponseWriter, r *http.Request) {
...
+			// also filter Desktop-only apps if ua is not Desktop
+			if p.Name != "" && (ua.Desktop || !p.DesktopOnly) {
 				apps = append(apps, p)
 			}
 		}
 		if len(apps) > 0 {
-			mimeTypes[m] = &appregistry.AppProviderList{AppProviders: apps}
-		} else {
-			delete(mimeTypes, m)
+			res = append(res, m)
```
ishank011, line 310: We're adding all the apps for that mimetype, not using the filtered list from above.

internal/http/services/appprovider/appprovider.go
```diff
@@ -106,34 +116,127 @@ func (s *svc) Handler() http.Handler {
...
+	}
+
+	target := r.URL.Query().Get("filename")
+	if target == "" {
+		ocmd.WriteError(w, r, ocmd.APIErrorInvalidParameter, "Missing filename",
+			errtypes.UserRequired("Missing filename"))
+		return
+	}
+	// stat the container
+	_, ocmderr, err := statRef(ctx, provider.Reference{Path: r.URL.Query().Get("container")}, client)
```
ishank011, line 142: I'm not sure about separate `filename` and `container` queries. The parent name can be extracted from the full path

Do we even want to perform this check? Because we allow `PUT`ting at paths where parents don't exist

internal/http/services/appprovider/appprovider.go
```diff
@@ -106,34 +116,127 @@ func (s *svc) Handler() http.Handler {
...
-		ocmd.WriteError(w, r, ocmd.APIErrorServerError, "error listing supported mime types", status.NewErrorFromCode(listRes.Status.Code, "appprovider"))
+	// Create empty file via storageprovider: obtain the HTTP URL for a PUT
+	target = path.Join(r.URL.Query().Get("container"), target)
+	createReq := &provider.InitiateFileUploadRequest{
+		Ref: &provider.Reference{Path: target},
+		Opaque: &typespb.Opaque{
+			Map: map[string]*typespb.OpaqueEntry{
+				"Upload-Length": {
+					Decoder: "plain",
+					Value:   []byte(strconv.FormatInt(0, 10)),
```
ishank011, line 156: We can just do `[]byte("0")`

internal/http/services/appprovider/appprovider.go
```diff
@@ -106,34 +116,127 @@ func (s *svc) Handler() http.Handler {
...
+		return
+	}
+	if createRes.Status.Code != rpc.Code_CODE_OK {
+		ocmd.WriteError(w, r, ocmd.APIErrorServerError, "error creating resource", status.NewErrorFromCode(createRes.Status.Code, "appprovider"))
 		return
 	}
 
-	mimeTypes := listRes.MimeTypes
-	filterAppsByUserAgent(mimeTypes, r.UserAgent())
+	// Do a HTTP PUT with an empty body
```
ishank011, line 171: Don't need to do this. `InitiateFileUpload` fully handles zero-size uploads

pkg/app/registry/static/static.go
```diff
@@ -148,23 +148,34 @@ func (b *reg) ListProviders(ctx context.Context) ([]*registrypb.ProviderInfo, er
...
+					Ext:          "", // TODO fetch from config
+					Name:         "",
+					Description:  "",
+					Icon:         "",
+				}
 			}
 		}
 	}
-	return mimeTypes, nil
+	for _, m := range mtmap {
```
gmgigi96, line 175: Why add the cached elements to the result list at the end?
Just do it when creating the new MimeTypeInfo

glpatcern: That's right, I had changed approach following discussions with @ishank011 and trusting you guys as Go experts ;-) but as a matter of fact `mtmap` is a cache of pointers, so once allocated they can be pushed to the result even if the structs they point to have their `AppProviders` lists being modified.

ishank011 (approved):

PR #2071, opened 2021-09-14
Implement listing favorites via the dav report API

C0rby:
I added filter-files to the dav REPORT API. This enables the listing of
favorites.

This PR also includes a new in-memory `storage.FS` implementation which I use for the unit tests of the favorite package but it can also be extended and used for other unit tests depending on a storage.

### DIFF
--- a/pkg/storage/favorite/favorite.go
+++ b/pkg/storage/favorite/favorite.go
@@ -1,0 +1,78 @@
+// Copyright 2018-2021 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package favorite
+
+import (
+	"context"
+
+	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/pkg/storage"
+)
+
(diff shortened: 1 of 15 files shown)


C0rby:
@butonic, could you have a look at this approach?

I think we still need some system to notify the several services about certain events like _file created_, _file deleted_, etc.
Here for example if a file is marked as favorite and then deleted, the favorite manager doesn't know that the files was deleted and will still return it.

Apart from that is that the expected approach? Do we still need to list all file from the storage provider and filter them for the favorite flag?

pkg/storage/favorite/favorite.go
```diff
@@ -0,0 +1,67 @@
...
+	return InMemoryManager{favorites: make(map[string]map[string]*provider.ResourceInfo)}
+}
+
+// InMemoryManager implements the Manager interface to manage favorites using an in-memory storage.
+type InMemoryManager struct {
+	favorites map[string]map[string]*provider.ResourceInfo
+}
+
+func (m InMemoryManager) ListFavorites(userID *user.UserId) ([]*provider.ResourceInfo, error) {
+	favorites := make([]*provider.ResourceInfo, 0, len(m.favorites[userID.OpaqueId]))
```
butonic, line 47: I think the favorites manager should only have a list of ResourceIDs. The ocdav service (and later the gateway) needs to stat (and cache the actual resources). IIRC that is how the shares are handled as well. Hmmm, I can see the point of having the full Resource info in the favorites manager.

C0rby: I see, storing the `ResourceInfo` has the disadvantage that it will eventually be outdated...

I implemented it this way to not have to stat all the resources but maybe a compromise could be to use the storage driver in the favorite manager to be able to more efficiently stat the resources (without the grpc and auth interceptors overhead)

labkode:
@C0rby @butonic 
For every single operation we have to contact the favourite manager, which in a prod scenario it won't be in memory but a SQL or another storage that adds latency. I think this is a good starting point and I'm merging it as is to unblock the UI development.

However, this needs to be revisited. Either adding a cache or store this info in the web client and the web client will add the fav icon where needed by matching fileids from the webdav response.

labkode (approved):

labkode:
@butonic @C0rby I've reverted this PR. I didn't realize of the hack of using the direct fs in the ocdav layer that bypass all the grpc logic, plus it breaks any deployment just by running it with another driver.

Please remove the dependency on the `storage.FS`, there is no reason to have it there and  FS interactions can happen via GRPC.

C0rby:
Right, I'll change the implementation to make the calls via GRPC and then we'll solve the potential performance problem later on.
Sorry for the inconvenience.

PR #2073, opened 2021-09-15
Fix opening a readonly filetype with WOPI

wkloucek:
Eg. JPG or PNG are supported to be opened by Collabora - but in view mode only. "/app/open" request currently fails because the WOPI driver assuems that an edit vie mode always exists.

### DIFF
--- a/pkg/app/provider/wopi/wopi.go
+++ b/pkg/app/provider/wopi/wopi.go
@@ -149,7 +149,7 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
 			q.Add("appviewurl", viewAppURL)
 		}
 	}
-	if editAppURLs, ok := p.appURLs["edit"]; !ok {
+	if editAppURLs, ok := p.appURLs["edit"]; ok {
 		if editAppURL, ok := editAppURLs[ext]; ok {
 			q.Add("appurl", editAppURL)
 		}
--- a/pkg/app/provider/wopi/wopi.go
+++ b/pkg/app/provider/wopi/wopi.go
@@ -142,16 +142,29 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
 	}
 
 	q.Add("appname", p.conf.AppName)
-	q.Add("appurl", p.appURLs["edit"][ext])
+
+	var viewAppURL string
+	if viewAppURLs, ok := p.appURLs["view"]; ok {
+		if viewAppURL, ok = viewAppURLs[ext]; ok {
+			q.Add("appviewurl", viewAppURL)
+		}
+	}
+	if editAppURLs, ok := p.appURLs["edit"]; !ok {
+		if editAppURL, ok := editAppURLs[ext]; ok {
+			q.Add("appurl", editAppURL)
+		}
+	} else {
+		// assuming that an view action is always available in the /hosting/discovery manifest
+		// eg. Collabora does support viewing jpgs but no editing
+		// eg. OnlyOffice does support viewing pdfs but no editing
+		// there is no known case of supporting edit only without view
+		q.Add("appurl", viewAppURL)
+	}
 
 	if p.conf.AppIntURL != "" {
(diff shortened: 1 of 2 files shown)


glpatcern (approved):
Looks good, and also in line with the discussion in https://github.com/cs3org/cs3apis/pull/145.

Just fix the reference in the changelog (2073 instead of 2051), and once the above PR is included we can also expose the `capability` for such cases where only a viewURL exists.

pkg/app/provider/wopi/wopi.go
```diff
@@ -142,16 +142,29 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
...
-			q.Add("appviewurl", url)
+	if editAppURLs, ok := p.appURLs["edit"]; !ok {
+		if editAppURL, ok := editAppURLs[ext]; ok {
+			q.Add("appurl", editAppURL)
 		}
+	} else {
+		// assuming that an view action is always available in the /hosting/discovery manifest
+		// eg. Collabora does support viewing jpgs but no editing
+		// eg. OnlyOffice does support viewing pdfs but no editing
+		// there is no known case of supporting edit only without view
```
labkode, line 160: DrawIO I think is edit online but so far is not part of this model (yet).

wkloucek: It could be a possibility to just pass 
`appurl` to the CS3 WOPI server, which is then the edit or open url depending on what action should be done (specified in the user request or supported by the app). Currently we also always set `appviewurl` which will not be used when the action is editing.

glpatcern: As a matter of fact the full `CheckFileInfo` metadata does include both an edit and a view URL, and MS Office expects those. This is why both are expected from Reva regardless the action.

wkloucek: Could we then omit the ones not specified? Looks to me like I should better fix this in the WOPI server? Didn't know that before

glpatcern: As discussed offline, it should not be a problem to keep the whole discovery response as provided by WOPI.

glpatcern (dismissed):
I believe this can go ahead so we bundle it with the other ongoing modifications in the AppProvider area

pkg/app/provider/wopi/wopi.go
```diff
@@ -142,16 +142,29 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
...
+	var viewAppURL string
+	if viewAppURLs, ok := p.appURLs["view"]; ok {
+		if viewAppURL, ok = viewAppURLs[ext]; ok {
+			q.Add("appviewurl", viewAppURL)
+		}
 	}
-	if viewExts, ok := p.appURLs["view"]; ok {
-		if url, ok := viewExts[ext]; ok {
-			q.Add("appviewurl", url)
+	if editAppURLs, ok := p.appURLs["edit"]; !ok {
```
ishank011, line 152: This should be `if editAppURLs, ok := p.appURLs["edit"]; ok {`

pkg/app/provider/wopi/wopi.go
```diff
@@ -142,16 +142,29 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
...
+		if editAppURL, ok := editAppURLs[ext]; ok {
```
ishank011, line 153: If this `ok` is false, we won't add the `appurl` query

pkg/app/provider/wopi/wopi.go
wkloucek: good catch :+1:

pkg/app/provider/wopi/wopi.go
wkloucek: It will now raise an error if neither "appurl" nor "appviewurl" is set. And it will set the "appurl" header to the same value as "appviewurl" if it is empty.

ishank011 (approved):

PR #2074, opened 2021-09-16
Fix Stat() for eos storage provider

gmgigi96:


pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1697,6 +1697,9 @@ func (fs *eosfs) convert(ctx context.Context, eosFileInfo *eosclient.FileInfo) (
 				},
 			},
 		},
+		ArbitraryMetadata: &provider.ArbitraryMetadata{
+			Metadata: eosFileInfo.Attrs,
+		},
```
labkode, line 1702: @gmgigi96 are `sys.` attributes discarded in this field ?

gmgigi96: now yes ;)

labkode (approved):

PR #2076, opened 2021-09-16
Fix chi routing

butonic:
Chi routes based on the URL.RawPath, which is not updated by the shiftPath based routing used in reva. By setting the RawPath to an empty string chi will fall pack to URL.Path, allowing it to match percent encoded path segments, e.g. when trying to match emails or multibyte characters.


labkode (approved):

wkloucek:
@labkode could you please merge this?

C0rby (approved):

Issue #2084, opened 2021-09-20
Reduce JWT token size


ishank011:
Based on the problems described in https://github.com/cs3org/reva/pull/1935 and the proxy workaround proving to be extremely non-performant, we'll look into reducing the size of the constructed tokens.

The size inflation is because of two objects:
- The user groups: We'll remove the groups from the user object encoded in the token and the services requiring these will make the GRPC call to the userprovider service and maintain a local cache.
- Scopes: We'll try to use some abbreviations for the scopes. Also, to improve performance, I added some code to prematurely expand these for public shares and lightweight accounts, but this can be reverted. I'll run some benchmarks to arrive at an acceptable tradeoff.

Related https://github.com/owncloud/ocis/issues/2525

PR #2085, opened 2021-09-20
Make encoding user groups in access tokens configurable

ishank011:
This PR adds a shared config parameter `skip_user_groups_in_token` which can be used to control whether user groups should be encoded in the reva access tokens or not. If these are skipped, these are retrieved from the `userprovider` service and cached in the authentication interceptors, where these tokens are dismantled.

Closes https://github.com/cs3org/reva/issues/2084


ishank011:
Size of token for user einstein (member of three groups):
- Before the change: 620 bytes
- After the change: 538 bytes

glpatcern (approved):
Looks good to me, as far as I can understand. Hopefully we'll see the immediate benefit with the token provided to WOPI :)

Issue #2036, opened 2021-09-03
add pragmatic favorites implementation


butonic:
The web ui needs an efficient way to list all files that have been marked as favorite by a user. We already persist the faverite flag on a per user basis in the extended attributes when using the decomposedfs. But for querying we need an efficient lookup, similar to the share manager.

Searching by favorite is one case of searching by metadata. While the bigger picture is tracked in https://github.com/cs3org/reva/issues/41 and https://github.com/cs3org/cs3apis/issues/8 we propose an intermediate solution.

As discussed in person we propose to introduce a go interface and implementation that the ocdav api can use to implement listing of favorite files. The ocdav service can then persist the favorites using that api, optionally also set them on the storage using extended attributes and read them from the api.

When this is done we can either add a `favorites manager` to the CS3 api or directly work on https://github.com/cs3org/reva/issues/41

@labkode @ishank011

butonic:
from [gitter](https://gitter.im/cs3org/REVA?at=6135d56b7cd57813a8ad5628):
> #2036: is what I had in mind, just replicate the model for sharing. We'll need to extend the CS3APIS to provide that functionality. I suggest you start with the "metadata manager" driver and interface
> We hook it in the ocdav later as a workaround and when the CS3APIS plus the GRPC service are ready we port the logic from ocdav to this service

labkode:
Initial work done here: https://github.com/cs3org/reva/pull/2071

kulmann:
For enabling favorites in web we need a boolean capability `capabilities.files.favorites`.
Needs to be defined here: https://github.com/owncloud/ocis/blob/bbcea23e85b705aafee17e0fbf05d67cc961b819/ocs/pkg/service/v0/data/capabilities.go#L79
And then added with value `true` here:
https://github.com/owncloud/ocis/blob/dbb6367e92effdfde258c1fb03a14cbb60126e47/storage/pkg/command/frontend.go#L57

labkode:
@kulmann does that enables the sidebar "Favourites" only or also the start icon in the row rendering?

kulmann:
Enables everything related to favorites:
- nav item in the left sidebar with dedicated favorites view
- favorites star in the right sidebar
- context menu (right click) action for toggling favorites

PR #2086, opened 2021-09-20
Implement listing favorites via the dav report API

C0rby:
I added filter-files to the dav REPORT API. This enables the listing of
favorites.

This is the rework of https://github.com/cs3org/reva/pull/2071
See https://github.com/cs3org/reva/pull/2071#issuecomment-921858204.

### DIFF
--- a/changelog/unreleased/list-favorites.md
+++ b/changelog/unreleased/list-favorites.md
@@ -1,0 +1,7 @@
+Enhancement: Implement listing favorites via the dav report API 
+
+Added filter-files to the dav REPORT API. This enables the listing of
+favorites.
+
+https://github.com/cs3org/reva/pull/2071
+https://github.com/cs3org/reva/pull/2086
--- a/internal/http/services/owncloud/ocdav/report.go
+++ b/internal/http/services/owncloud/ocdav/report.go
@@ -22,8 +22,18 @@ import (
 	"encoding/xml"
 	"io"
 	"net/http"
+	"strings"
 
+	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	providerv1beta1 "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/pkg/appctx"
+	ctxpkg "github.com/cs3org/reva/pkg/ctx"
+)
+
+const (
+	elementNameSearchFiles = "search-files"
+	elementNameFilterFiles = "filter-files"
 )
 
 func (s *svc) handleReport(w http.ResponseWriter, r *http.Request, ns string) {
@@ -39,6 +49,11 @@ func (s *svc) handleReport(w http.ResponseWriter, r *http.Request, ns string) {
 	}
 	if rep.SearchFiles != nil {
 		s.doSearchFiles(w, r, rep.SearchFiles)
+		return
+	}
+
+	if rep.FilterFiles != nil {
+		s.doFilterFiles(w, r, rep.FilterFiles, ns)
 		return
 	}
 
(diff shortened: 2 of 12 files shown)


changelog/unreleased/list-favorites.md
```diff
@@ -0,0 +1,6 @@
+Enhancement: Implement listing favorites via the dav report API 
+
+Added filter-files to the dav REPORT API. This enables the listing of
+favorites.
+
+https://github.com/cs3org/reva/pull/2071
```
refs, line 6:
```suggestion
https://github.com/cs3org/reva/pull/2086
```

internal/http/services/owncloud/ocdav/report.go
```diff
@@ -59,9 +74,67 @@ func (s *svc) doSearchFiles(w http.ResponseWriter, r *http.Request, sf *reportSe
...
 }
 
+func (s *svc) doFilterFiles(w http.ResponseWriter, r *http.Request, ff *reportFilterFiles, namespace string) {
+	ctx := r.Context()
+	log := appctx.GetLogger(ctx)
+
+	if ff.Rules.Favorite {
+		// List the users favorite resources.
+		currentUser := ctxpkg.ContextMustGetUser(ctx)
+		favorites, err := s.favoritesManager.ListFavorites(ctx, currentUser.Id)
```
refs, line 84: couldn't we reuse the user in the context on `ListFavorites` instead of also use the user id as argument? Seems redundant, and it is the only place `currentUser` is used within this scope.

pkg/storage/favorite/favorite.go
```diff
@@ -0,0 +1,70 @@
...
+	// ListFavorites returns all resources that were favorited by a user.
+	ListFavorites(ctx context.Context, userID *user.UserId) ([]*provider.ResourceId, error)
+	// SetFavorite marks a resource as favorited by a user.
+	SetFavorite(ctx context.Context, userID *user.UserId, resourceID *provider.ResourceId) error
+	// UnsetFavorite unmarks a resource as favorited by a user.
+	UnsetFavorite(ctx context.Context, userID *user.UserId, resourceID *provider.ResourceId) error
+}
+
+// NewInMemoryManager returns an instance of a favorites manager using an in-memory storage.
+func NewInMemoryManager() Manager {
```
refs, line 39: In memory manager means favorites are wipe out on each restart 🙈 can we have a file based favorite manager at least? So we at least introduce no external dependencies and keep inline with what the shares-manager (first example I came up with) uses persistence?

pkg/storage/favorite/favorite.go
```diff
@@ -0,0 +1,70 @@
...
+	return InMemoryManager{favorites: make(map[string]map[string]*provider.ResourceId)}
+}
+
+// InMemoryManager implements the Manager interface to manage favorites using an in-memory storage.
+type InMemoryManager struct {
+	favorites map[string]map[string]*provider.ResourceId
```
refs, line 45: also mind that a map is not thread safe and will panic if [concurrent writes](https://medium.com/@luanrubensf/concurrent-map-access-in-go-a6a733c5ffd1). A mutex should guard against this.

refs (changes requested):
I think having persistence here and not only a memory implementation will take this PR further 👯

changelog/unreleased/list-favorites.md
C0rby: Ah good catch

internal/http/services/owncloud/ocdav/report.go
C0rby: I wanted to make the API a bit more explicit concerning the user. Implicitly passing the user around in the context doesn't feel right. But I'm open to change this.

pkg/storage/favorite/favorite.go
C0rby: Right. :thinking:

butonic (dismissed):
This is the first step. @labkode @ishank011 please review.

internal/http/services/owncloud/ocdav/report.go
```diff
@@ -59,9 +74,67 @@ func (s *svc) doSearchFiles(w http.ResponseWriter, r *http.Request, sf *reportSe
...
+			}
+			if statRes.Status.Code != rpcv1beta1.Code_CODE_OK {
+				HandleErrorStatus(log, w, statRes.Status)
+				continue
+			}
+
+			// The paths we receive has the format /user/<username>/<filepath>
+			// We only want the `<filepath>` part. Thus we remove the /user/<username>/ part.
+			parts := strings.SplitN(statRes.Info.Path, "/", 4)
+			statRes.Info.Path = parts[3]
```
ishank011, line 114: Maybe add a check to ensure `len(parts) == 4`

ishank011 (commented):
@butonic @C0rby looks okay for a short-term solution, but we need to discuss how to have a better implementation

pkg/storage/favorite/favorite.go
ishank011: Yep, it won't work without a lock

internal/http/services/owncloud/ocdav/report.go
ishank011: We do assume the user to be passed around like that in the other interfaces, but it's not an absolute necessity

C0rby:
@ishank011, it was agreed that this is just the short-term solution and a more future proof solution will follow. See https://github.com/cs3org/reva/issues/2036

ishank011:
@C0rby thanks! Can you add the mutex protection? We can merge it then

ishank011 (approved):

refs (approved):

ishank011:
@C0rby can you take a look at the failing tests, please?

C0rby:
Yeah, I'm on it.

PR #2093, opened 2021-09-22
limit the data exposed to resourceinfo and publicshare scopes

C0rby:
We changed the publicshare and resourceinfo scopes to contain only necessary values.
This reduces the size of the resulting token and also limits the amount of data which can be leaked.

Needed for: https://github.com/owncloud/ocis/issues/2479


ishank011 (approved):
Awesome!

PR #2095, opened 2021-09-23
Whitelist apps

gmgigi96:
This PR introduces the following enhancements:
1. Overrides the AppProvider supported mimetypes from the configuration
2. Add a configuration for the mime types in the AppRegistry (a friendly name, default AppProvider, description, icon, ...)

### DIFF
--- a/internal/grpc/services/appregistry/appregistry_test.go
+++ b/internal/grpc/services/appregistry/appregistry_test.go
@@ -40,6 +40,7 @@ func Test_ListAppProviders(t *testing.T) {
 	tests := []struct {
 		name      string
 		providers map[string]interface{}
+		mimeTypes map[string]map[string]string
 		want      *registrypb.ListAppProvidersResponse
 	}{
 		{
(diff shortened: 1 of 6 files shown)


glpatcern (commented):
Looks all good, I just want to prepare a good list of default mime types to be included. Will do that on Monday ;-)

internal/grpc/services/appregistry/appregistry_test.go
```diff
@@ -168,6 +186,45 @@ func Test_GetAppProviders(t *testing.T) {
...
+			"icon":        "https://example.org/icons&file=json.png",
+			"default_app": "some Address",
+		},
+		"text/xml": {
+			"extension":   "xml",
+			"name":        "XML File",
+			"icon":        "https://example.org/icons&file=xml.png",
+			"default_app": "some Address",
+		},
+		"application/vnd.openxmlformats-officedocument.wordprocessingml.document": {
```
glpatcern, line 202: I propose to keep the test examples smaller and remove all except JSON and XML. Then we should populate the example configuration file with some sensible defaults.

glpatcern (changes requested):
As quickly discussed, for the example config file I'd include the mime types for: docx, xlsx, pptx, odt, ods, odp, txt, md, root, ipynb.

internal/grpc/services/appregistry/appregistry_test.go
gmgigi96: As discussed, the mime types in the test are the minimum necessary to pass the tests.
For the example config, I've added all the extensions you mentioned except the root as it does not have a defined mime type.

glpatcern (approved):
All good for me!

Issue #2097, opened 2021-09-23
Change archiver to work resource id based


kulmann:
As I've learned in https://github.com/cs3org/reva/issues/2096 the archiver service currently works with absolute paths. That brings at least two limitations:

1: we need configuration for the web ui for defining a path prefix. In the CERN-Box environment this is not needed (`/`), for any standard ownCloud environment it would be `/home`. This kind of knowledge should not be needed in the web UI (+ this will not work with our spaces implementation)
2: similar to (1) folder downloads will not work for public links

Switching to resource id based file/folder selection for the archive will resolve (1) and (2). Additionally it will allow us to have support folder downloads on password protected public links without extra effort after https://github.com/owncloud/ocis/pull/2536 is merged.

kulmann:
cc'ing @wkloucek for more details if needed

gmgigi96:
The current archiver implementation has the same interface of an old archiver service, we used, and yes, it is limited.
So, I will change the interface to support both paths and resources ids, providing to the archiver a list of `path` and/or a list of `id`.

kulmann:
Awesome, thank you 🤩

butonic:
Related: https://github.com/owncloud/ocis/issues/2521

Cc @kulmann I don't know if @jvillafanez looked into this as well.

PR #2100, opened 2021-09-23
Add support for resource id to the archiver

gmgigi96:
Before the archiver only supported resources provided by a path.
Now also the resources ID are supported in order to specify the content of the archive to download. The parameters accepted by the archiver are two:
- an optional list of `path` (containing the paths of the resources)
- an optional list of `id` (containing the resources IDs of the resources)

Closes https://github.com/cs3org/reva/issues/2097

### DIFF
--- a/internal/http/services/archiver/archiver.go
+++ b/internal/http/services/archiver/archiver.go
@@ -22,401 +22,67 @@ import (
 	"archive/tar"
 	"archive/zip"
 	"context"
-	"encoding/base64"
-	"fmt"
 	"io"
-	"net/http"
 	"path"
 	"path/filepath"
-	"strings"
 	"time"
-	"unicode/utf8"
 
-	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
-	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
-	"github.com/cs3org/reva/internal/http/services/datagateway"
 	"github.com/cs3org/reva/pkg/errtypes"
-	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
-	"github.com/cs3org/reva/pkg/rhttp"
-	"github.com/cs3org/reva/pkg/rhttp/global"
-	"github.com/cs3org/reva/pkg/sharedconf"
+	"github.com/cs3org/reva/pkg/storage/utils/downloader"
 	"github.com/cs3org/reva/pkg/storage/utils/walker"
-	"github.com/gdexlab/go-render/render"
-	ua "github.com/mileusna/useragent"
-	"github.com/mitchellh/mapstructure"
-	"github.com/pkg/errors"
-	"github.com/rs/zerolog"
 )
-
-type svc struct {
-	config     *Config
-	httpClient *http.Client
-	gtwClient  gateway.GatewayAPIClient
-	log        *zerolog.Logger
-}
-
-// Config holds the config options that need to be passed down to all ocdav handlers
-type Config struct {
-	Prefix      string `mapstructure:"prefix"`
-	GatewaySvc  string `mapstructure:"gatewaysvc"`
-	Timeout     int64  `mapstructure:"timeout"`
-	Insecure    bool   `mapstructure:"insecure"`
-	MaxNumFiles int64  `mapstructure:"max_num_files"`
-	MaxSize     int64  `mapstructure:"max_size"`
-}
 
 const (
-	errMaxFileCount = errtypes.InternalError("reached max files count")
-	errMaxSize      = errtypes.InternalError("reached max total files size")
+	// ErrMaxFileCount is the error returned when the max files count specified in the config has reached
+	ErrMaxFileCount = errtypes.InternalError("reached max files count")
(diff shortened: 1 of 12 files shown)


internal/http/services/archiver/archiver.go
```diff
@@ -105,38 +106,73 @@ func (c *Config) init() {
...
+	if len(files) == 0 && len(ids) == 0 {
+		return nil, errtypes.BadRequest("file and id lists are both empty")
+	}
+
+	var f []string
+
+	for _, id := range ids {
+		// an id must have the form <storage_id>:<id>
+
+		split := strings.Split(id, ":")
```
kulmann, line 119: The resource ids arrive in the web ui as base64 encoded strings. With this implementation I'd need to decode the fileId in the web ui and send it in decoded form. Just knowing this I'd prefer if you could the base64 decoding here instead. But I'm not sure if I'm seeing the full picture here. Please let me know if the decoding should happen in the clients or here.

gmgigi96: Thanks for the review, fixed to receive a base64 encoded resource ID to be consistent with the rest of the backend.

glpatcern (dismissed):
With the latest fixes it looks good to go for me.

kulmann:
On a mac I get the following error when trying to extract the downloaded tarball with `7z x download.tar`:

```
7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
p7zip Version 16.02 (locale=utf8,Utf16=on,HugeFiles=on,64 bits,12 CPUs x64)

Scanning the drive for archives:
1 file, 453241 bytes (443 KiB)

Extracting archive: download.tar

ERRORS:
Headers Error


WARNINGS:
There are data after the end of archive

--
Path = download.tar
Type = tar
ERRORS:
Headers Error
WARNINGS:
There are data after the end of archive
Physical Size = 250880
Tail Size = 202361
Headers Size = 1536
Code Page = UTF-8



Archives with Errors: 1

Warnings: 1

Open Errors: 1
```

It extracts it, the folder structure is correct, but the files are corrupted. Does this work for you?

pascalwengerter:
Can confirm that on Linux/Ubuntu archives that contain `PNG` and `JPG` files don't work as expected - top level files can't be opened, while a folder containing images didn't even make it into the `*.tar`

gmgigi96:
@kulmann @pascalwengerter I tried on both Mac and Arch and I didn't manage to reproduce the errors you had. For me it works on both systems, could you please provide me the steps to reproduce it? Thanks

gmgigi96:
Can you please try doing an HTTP GET request directly to the archiver service?

gmgigi96:
@kulmann @pascalwengerter I added some unit tests into the archiver and all are passing, are you having the same issues?

internal/http/services/archiver/handler/handler.go
```diff
@@ -0,0 +1,237 @@
...
+type Config struct {
+	Prefix     string `mapstructure:"prefix"`
+	GatewaySvc string `mapstructure:"gatewaysvc"`
+	Timeout    int64  `mapstructure:"timeout"`
+	Insecure   bool   `mapstructure:"insecure"`
+	Name       string `mapstructure:"name"`
+	archiver.Config
+}
+
+func init() {
```
kulmann, line 67: This `init` is not invoked because you're importing the `archiver` package in the loader instead of the `handler` package. Importing the `archiver/handler` package in the `http/services/loader/loader.go` looks really wrong though 🤔

kulmann (changes requested):

internal/http/services/archiver/handler/handler.go
gmgigi96: Thanks for the review. Yes it's true. Now it should be resolved :)

kulmann (approved):
The id based version now works like a charm 🥳

kulmann:
> @kulmann @pascalwengerter I added some unit tests into the archiver and all are passing, are you having the same issues?

Resolved in the web PR. Sorry for the noise 😅

gmgigi96:
> > @kulmann @pascalwengerter I added some unit tests into the archiver and all are passing, are you having the same issues?
> 
> Resolved in the web PR. Sorry for the noise 😅

Glad it works! @labkode @glpatcern can we merge it now?

glpatcern:
Great, thanks all! Merging right away

glpatcern (approved):

individual-it:
@gmgigi96 when using both id and path in the same request e.g. `curl -k -u admin:admin 'https://localhost:9200/archiver?path=/home/image.jpg&id=MTI4NGQyMzgtYWE5Mi00MmNlLWJkYzQtMGIwMDAwMDA5MTU3OmZlZTVhNjZlLWI4MzYtNDY2OS05ZDJhLTliNGEwMDQwYmY0OQ==' --output /tmp/6.tar`

I get whatever is requested by path  in the folder called `home` and whatever is requested by id in the  folder `users/<uuid>/`
is that intended?

```
$ tar -tvf /tmp/6.tar 
drwxr-xr-x 0/0               0 2021-11-11 16:10 users/ddc2004c-0977-11eb-9d3f-a793888cd0f8/my_data
-rw-r--r-- 0/0          472800 2011-03-07 21:41 users/ddc2004c-0977-11eb-9d3f-a793888cd0f8/my_data/11-02.JPG
-rw-r--r-- 0/0         1278511 2020-08-17 15:39 home/image.jpg

PR #2103, opened 2021-09-24
AppProvider: report errors from WOPI and return base64-encoded fileids

glpatcern:
This is a small PR to log and report back from the AppProvider's `/app/open` to the UI any error reported by the WOPI server.

It also includes another fix in the AppProvider to return a base64-encoded fileid on `/app/new`.

### DIFF
--- a/internal/http/services/appprovider/appprovider.go
+++ b/internal/http/services/appprovider/appprovider.go
@@ -49,6 +49,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	idDelimiter string = ":"
+)
+
 func init() {
 	global.Register("appprovider", New)
 }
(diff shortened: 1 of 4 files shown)


internal/http/services/appprovider/appprovider.go
```diff
@@ -49,6 +49,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	idDelimiter string = ":"
```
ishank011, line 53: Can you please change `getStatInfo` to use this constant as well?

pkg/app/provider/wopi/wopi.go
```diff
@@ -167,15 +167,16 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
...
-	if openRes.StatusCode != http.StatusOK {
-		return nil, errtypes.InternalError("wopi: unexpected status from WOPI server: " + openRes.Status)
-	}
-
 	body, err := ioutil.ReadAll(openRes.Body)
 	if err != nil {
 		return nil, err
 	}
 
+	if openRes.StatusCode != http.StatusOK {
```
ishank011, line 175: Will `body` always be non-nil when status isn't OK? I'm not sure how `errors.New` will behave if it's nil. We should use `openRes.Status` as well

internal/http/services/appprovider/appprovider.go
glpatcern: sure, good point

pkg/app/provider/wopi/wopi.go
glpatcern: Yes, we are guaranteed from the WOPI server. I could still add code to convert `nil` to `""` or check this behavior; anyway will add a comment stating that.

glpatcern: Added a comment (and yes, WOPI always returns a string if status code != 200). Do you think we still need a check `if body == nil { body = "" }` or something like that?

ishank011: Yep, I'd prefer if we added that check. And that we also add `openRes.Status` in the error messages, like the log statement just above it.

glpatcern: All right, I'll add the sanity check. As for `openRes.Status`, I'd rather not include it in the message: it's fine to have it in the logs, it's useless for the user.

The whole point of this PR is to propagate back a useful message to the user (though it may not respect the locale), exactly like e.g. Collabora shows dialogs saying `Unable to lock file: the server reported "<bla bla>"` where the `<bla bla>` comes directly from the WOPI server. Likewise, here the UI could show a message (in the correct locale) like `Unable to open the file in <appName>: <bla bla>`, where `<bla bla>` is verbatim the error string coming from WOPI.

ishank011 (approved):

PR #2114, opened 2021-09-28
add as default app while registering and skip unset mimetypes

wkloucek:
https://github.com/cs3org/reva/pull/2095 didn't set default apps for me. Also listing supported mimetypes always failed for me unless I configured _all_ supported mimetypes of eg. Collabora.

My fix sets default apps while registering the app providers according to the config.
My fix skips all mimetypes not listed in the config.


glpatcern (approved):
Thanks, will integrate it right away

PR #2118, opened 2021-09-29
fix app open when multiple app providers are present

wkloucek:
Bugfix: Fix app open when multiple app providers are present

We've fixed the gateway behavior, that when multiple app providers are present, it always returned that we have duplicate names for app providers.
This was due the call to GetAllProviders() without any subsequent filtering by name. Now this filter mechanism is in place and the duplicate app providers error will only appear if a real duplicate is found.

Introduced in https://github.com/cs3org/reva/issues/2095

### DIFF
--- a/changelog/unreleased/app-registry-multiple-providers.md
+++ b/changelog/unreleased/app-registry-multiple-providers.md
@@ -1,0 +1,7 @@
+Bugfix: Fix app open when multiple app providers are present
+
+We've fixed the gateway behavior, that when multiple app providers are present, it always returned that we have duplicate names for app providers.
+This was due the call to GetAllProviders() without any subsequent filtering by name. Now this filter mechanism is in place and the duplicate app providers error will only appear if a real duplicate is found.
+
+https://github.com/cs3org/reva/issues/2095
+https://github.com/cs3org/reva/pull/2117
--- a/internal/grpc/services/gateway/appprovider.go
+++ b/internal/grpc/services/gateway/appprovider.go
@@ -295,13 +295,7 @@ func (s *svc) findAppProvider(ctx context.Context, ri *storageprovider.ResourceI
 
 	// if we only have one app provider we verify that it matches the requested app name
 	if len(res.Providers) == 1 {
-		p := res.Providers[0]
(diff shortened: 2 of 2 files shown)


glpatcern:
I had a look at the code but there's something I don't understand, will discuss it with @ishank011 and @gmgigi96.

wkloucek:
> I had a look at the code but there's something I don't understand, will discuss it with @ishank011 and @gmgigi96.

Is there something I can help you?

ishank011:
@wkloucek looks good. Ideally `filteredProviders` should always have one element only but in case multiple apps are registered with the same name, should we raise an error?

wkloucek:
> @wkloucek looks good. Ideally `filteredProviders` should always have one element only but in case multiple apps are registered with the same name, should we raise an error?

That was already there (https://github.com/cs3org/reva/blob/9c827f825516ef18b54b2b0cd62de810272e1722/internal/grpc/services/gateway/appprovider.go#L307-L309). Only the filter was missing

ishank011:
Okay yes, my bad. Can you just move the code below the `res.Status` check please?

ishank011:
Also, sorry for the repeated reviews, but can you also remove the check `if p.Name == app {` inside the block where we check that there's only one provider? It's redundant

wkloucek:
> Okay yes, my bad. Can you just move the code you added below the `res.Status` check please?

changed :+1: 

> Also, sorry for the repeated reviews, but can you also remove the check `if p.Name == app {` inside the block where we check that there's only one provider? It's redundant

That should stay... Image the user requesting an app that doesn't exist. The user would then get the only app existing, which is _not_ the app, the user requested.

ishank011:
> That should stay... Image the user requesting an app that doesn't exist. The user would then get the only app existing, which is not the app, the user requested.

But your code should take care of that, right?

wkloucek:
> > That should stay... Image the user requesting an app that doesn't exist. The user would then get the only app existing, which is not the app, the user requested.
> 
> But your code should take care of that, right?

You're right, sorry...

ishank011 (approved):
Perfect, thanks!

PR #130, opened 2019-06-28
oidc provider

butonic:
This adds a hardcoded oidc provider that can be used to log in `aaliyah_abernathy`, `aaliyah_adams`, and `aaliyah_anderson`. These are the three users that are used in nexus. But this PR alllows you to do local development, without any docker containers! Create a `/data` dir and create subfolders for `aaliyah_abernathy`, `aaliyah_adams`, and `aaliyah_anderson`. Start reva with the below configuration, eg: `cmd/revad/revad -c local.toml -p revad.pid | jq` with local.toml looking like this:
```toml
[core]
max_cpus = "2"

[log]
output = "stdout"
mode = "dev"
level = "debug"

[http]
enabled_services = ["datasvc", "ocdavsvc", "ocssvc", "oidcprovider", "wellknown"]
#enabled_middlewares = ["log", "trace", "auth", "cors"]
enabled_middlewares = ["log", "trace", "auth", "cors"]
network = "tcp"
address = "0.0.0.0:9998"

[http.middlewares.trace]
priority = 100
header = "x-trace"

[http.middlewares.log]
priority = 200

[http.middlewares.auth]
priority = 300
authsvc = "127.0.0.1:9999"
credential_strategy = "oidc"
#credential_strategy = "basic"
token_strategy = "header"
token_writer = "header"
token_manager = "jwt"
skip_methods = [
    "/status.php",
    "/oauth2",
    "/oauth2/auth", 
    "/oauth2/token", 
    "/oauth2/introspect",
    "/oauth2/userinfo", 
    "/oauth2/sessions", 
    "/.well-known/openid-configuration"
]

[http.middlewares.cors]
priority = 400
allowed_origins = ["*"]
allow_credentials = true
allowed_methods = ["OPTIONS", "GET", "PUT", "POST", "DELETE", "MKCOL", "PROPFIND", "PROPPATCH", "MOVE", "COPY", "REPORT", "SEARCH"]
allowed_headers = ["Origin", "Accept", "Content-Type", "X-Requested-With", "Authorization", "Ocs-Apirequest"]
options_passthrough = true

[http.middlewares.auth.token_managers.jwt]
secret = "Pive-Fumkiu4"

[http.middlewares.auth.token_strategies.header]
header = "X-Access-Token"

[http.middlewares.auth.token_writers.header]
header = "X-Access-Token"


[http.services.ocdavsvc]
prefix = ""
chunk_folder = "/var/tmp/owncloud/chunks"
[...]

PR #2120, opened 2021-09-30
[docs] Add code comments to RestoreRecycleItem

michielbdejong:
* Looking at the [reva-cli function for RestoreRecycleItem](https://github.com/cs3org/reva/blob/6823079c17027133bba040dcbdc8d29731a4349d/cmd/reva/recycle-restore.go#L32), it asks for `key` but for `Ref.Path` it just uses the result from `client.GetHome`.
* Looking at the [gateway service storage provider](https://github.com/cs3org/reva/blob/6823079c17027133bba040dcbdc8d29731a4349d/internal/grpc/services/gateway/storageprovider.go#L2105) it uses Ref (which reva-cli sets to the current user's home) to find the storage provider.
* Looking at the [storage service storage provider/](https://github.com/cs3org/reva/blob/6823079c17027133bba040dcbdc8d29731a4349d/internal/grpc/services/storageprovider/storageprovider.go#L893) it passes:
    * `req.Key` as `key`
    * `req.Ref.Path` as `path`
    * `req.RestoreRef` as `restoreRef`
* Looking at e.g. the ownCloud storage driver, [it uses](https://github.com/cs3org/reva/blob/6823079c17027133bba040dcbdc8d29731a4349d/pkg/storage/fs/owncloud/owncloud.go#L2199) `key` (not `path`!) as the original path of the recycle item to be restored.
* This makes sense because the end-user who issues the command through reva-cli can influence the value of `key` but not the value of `path`.
* It's confusing though, that it's called `key`.

So based on this spelonking, it looks like storage drivers should:
* look at `key` to know the path (string) of the file to be restored
* ignore `path`
* _if_ `restoreRef` is defined, restore to `restoreRef.Path`, otherwise just restore to `key` (i.e. where the recycle item originally came from).

### DIFF
--- a/pkg/storage/fs/nextcloud/nextcloud_server_mock.go
+++ b/pkg/storage/fs/nextcloud/nextcloud_server_mock.go
@@ -116,7 +116,7 @@ var responses = map[string]Response{
 	`POST /apps/sciencemesh/~einstein/api/storage/ListGrants {"path":"/subdir"} GRANT-REMOVED`: {200, `[]`, serverStateEmpty},
 
 	`POST /apps/sciencemesh/~einstein/api/storage/ListRecycle {"key":"","path":"/"} EMPTY`:   {200, `[]`, serverStateEmpty},
-	`POST /apps/sciencemesh/~einstein/api/storage/ListRecycle {"key":"","path":"/"} RECYCLE`: {200, `[{"opaque":{},"key":"some-deleted-version","ref":{"resource_id":{},"path":"/subdir"},"size":12345,"deletion_time":{"seconds":1234567890}}]`, serverStateRecycle},
+	`POST /apps/sciencemesh/~einstein/api/storage/ListRecycle {"key":"","path":"/"} RECYCLE`: {200, `[{"opaque":{},"key":"some-recycle-item","ref":{"resource_id":{},"path":"/subdir"},"size":12345,"deletion_time":{"seconds":1234567890}}]`, serverStateRecycle},
 
 	`POST /apps/sciencemesh/~einstein/api/storage/ListRevisions {"path":"/versionedFile"} EMPTY`:         {200, `[{"opaque":{"map":{"some":{"value":"ZGF0YQ=="}}},"key":"version-12","size":1,"mtime":1234567890,"etag":"deadb00f"}]`, serverStateEmpty},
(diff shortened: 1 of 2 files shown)


michielbdejong:
Ah wait, it's not as confusing as I first thought, because the key comes from the RecycleItem object that was returned from ListRecycle, so it is actually a key that the recycle bin implementation can choose freely. So the variable name `key` does make sense.

What I still don't understand is why there is a second parameter `path`, and it points to the user's home dir.
In this context, I would expect a variable called `path` to be the [RecycleItem](https://github.com/cs3org/cs3apis/blob/6eab4643f5113a54f4ce4cd8cb462685d0cdd2ef/cs3/storage/provider/v1beta1/resources.proto#L321)'s required field `Path`?

refs:
> What I still don't understand is why there is a second parameter path

I think the `path` is legacy, but someone correct me if I'm wrong. The `Reference` was reworked on [#130](https://github.com/cs3org/cs3apis/pull/130/files#diff-b91600c359c2e479b2dbc31ff8dd5bd52bfaa48dc70c33a0de121215ff0c030eR199) and that is the only reason I can point to as per why we don't use the `path` parameter but instead check the `restoreRef`'s `path`:

```go
...
	if restoreRef.Path == "" {
		v, err := xattr.Get(src, trashOriginPrefix)
		if err != nil {
			log.Error().Err(err).Str("key", key).Str("path", src).Msg("could not read origin")
		}
		restoreRef.Path = filepath.Join("/", filepath.Clean(string(v)), strings.TrimSuffix(filepath.Base(src), suffix))
	}
...
```

In that function, the path parameter is as good as gone :). Perhaps the entire interface can be refactored to reflect such changes actually. I'm referring to this interface:

```go
...
RestoreRecycleItem(ctx context.Context, key, path string, restoreRef *provider.Reference) error
...
```

Also for references I suggest having a look at DecomposedFS, since it is where we are dealing with most of the storage layer related code at ownCloud.

```go
...
func (fs *Decomposedfs) RestoreRecycleItem(ctx context.Context, key, path string, restoreRef *provider.Reference) error {
	if restoreRef == nil {
		restoreRef = &provider.Reference{}
	}
	rn, p, restoreFunc, err := fs.tp.RestoreRecycleItemFunc(ctx, key, path, restoreRef.Path)
	if err != nil {
		return err
	}
...
```

In which, funny enough, the `path` takes yet again another semantic turn, making matters more confusing :/

michielbdejong:
Will reopen this PR with the branch on my own fork, sorry.

PR #2121, opened 2021-09-30
Sharemanager api changes

butonic:
This PR introduces the filed mask to the UpdateReceivedShare APIs. This is a breaking change but we tried to change all user and ocm shareprovider managers to use tha field mask.

Unfortunately, the nextcloud driver internally seems to pass on the internal struct. @michielbdejong could you check we correctly implemented a backwards compatible change? 

This PR is based on https://github.com/cs3org/reva/pull/2023 but reverts everything but the api change.

cc @aduffeck @labkode


pkg/share/manager/nextcloud/nextcloud_test.go
```diff
@@ -726,96 +726,4 @@ var _ = Describe("Nextcloud", func() {
...
-						Seconds:              1234567890,
-						Nanos:                0,
-						XXX_NoUnkeyedLiteral: struct{}{},
-						XXX_unrecognized:     nil,
-						XXX_sizecache:        0,
-					},
-				},
-				State: collaboration.ShareState_SHARE_STATE_ACCEPTED,
-			}))
-			Expect(called[0]).To(Equal(`POST /apps/sciencemesh/~tester/api/share/UpdateReceivedShare {"ref":{"Spec":{"Id":{"opaque_id":"some-share-id"}}},"f":{"Field":{"DisplayName":"some new name for this received share"}}}`))
```
butonic, line 817: We deleted the test, because the implementing a test for this properly (using the fileld mask) requires changes that are out of our control. Please keep in mind that the CS3 API is not out of beta and still subject to changes. Modeling the sciencemesh app api after it may require you to sometimes change it as well.

michielbdejong: We'll make the corresponding change in the sciencemesh app api, no problem!
Please update the test instead of deleting it.
Is it failing in your branch now?

michielbdejong:
@butonic thanks for the heads-up!
Yeah, definitely please don't delete that test. :)
Here is a way to make it pass: https://github.com/butonic/reva/pull/193
This also makes the function simpler by keeping the `paramsObj` closer to the function arguments, and just JSON-encoding whatever comes in.

michielbdejong (changes requested):
Please have a look at https://github.com/butonic/reva/pull/193

pkg/share/manager/nextcloud/nextcloud.go
```diff
@@ -403,13 +404,29 @@ func (sm *mgr) GetReceivedShare(ctx context.Context, ref *collaboration.ShareRef
 }
 
 // UpdateReceivedShare updates the received share with share state.
-func (sm *mgr) UpdateReceivedShare(ctx context.Context, ref *collaboration.ShareReference, f *collaboration.UpdateReceivedShareRequest_UpdateField) (*collaboration.ReceivedShare, error) {
+func (sm *mgr) UpdateReceivedShare(ctx context.Context, receivedShare *collaboration.ReceivedShare, fieldMask *field_mask.FieldMask) (*collaboration.ReceivedShare, error) {
+	type updateField struct {
```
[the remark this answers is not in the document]
michielbdejong: Thanks for the effort, but we don't want to mimic the old API. In all other places, the Nextcloud-ScienceMesh app API exactly follows the CS3 API, just with a different encoding (JSON instead of protobuf). So when the CS3 API changes, the Nextcloud-ScienceMesh app API just goes along with exactly the same change.

pkg/share/manager/nextcloud/nextcloud_test.go
michielbdejong: Re-introduced and fixed the test in my PR to your PR :)

butonic:
@michielbdejong Great! Thx!

michielbdejong (dismissed):
Changes in pkg/share/manager/nextcloud LGTM now!

butonic:
Note that this PR still needs https://github.com/cs3org/cs3apis/pull/144 ... which has a failing CI ... urgh ... I'll take care of that...

PR #2143, opened 2021-10-07
Enabling apps to work in public shares.

C0rby:
I extended the scope checking to make it possible to use Apps (and the App Provider) in public shares.
I don't quite like this solution but we also haven't found a better alternative yet. Public share scopes work as long as the requests are path based but that is not always the case.

Maybe we need to rethink this scope implementation. It feels a bit to hacky and unflexible.

This solution is by far not pretty. But this is the smallest change I
could find to support apps in public shares without completely rewriting
the scopes code.
If we want to use any app in a public share the app needs an
authenticated context to access the files in the share. This is working
fine since we can authenticate using the share token. But the resulting
JWT contains a scope refering to the shared resource. That means for a
folder share it contains the resource ID of the shared folder. But if
the app wants to access a specific file in the share it will request
that file by ID. Because of that the old way of the public scope checker
doesn't work because it would compare or requested ID with the ID of
the shared folder and this would fail.
For this reason I added two additional Stat requests so that I
can check if the requested file is inside the shared folder.

This is the reva part of the required changes. The ocis part can be found in this PR: https://github.com/owncloud/ocis/pull/2536

### DIFF
--- a/internal/grpc/interceptors/auth/scope.go
+++ b/internal/grpc/interceptors/auth/scope.go
@@ -22,6 +22,8 @@ import (
 	"context"
 	"strings"
 
+	appprovider "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
+	appregistry "github.com/cs3org/go-cs3apis/cs3/app/registry/v1beta1"
 	authpb "github.com/cs3org/go-cs3apis/cs3/auth/provider/v1beta1"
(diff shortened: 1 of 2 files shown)


ishank011:
@C0rby please take a look at https://github.com/C0rby/reva/pull/209
@wkloucek can you please check that apps still work after this? My tests seem to pass

PR #2153, opened 2021-10-11
Restrict EOS project spaces sharing permissions to admins

ishank011:


labkode:
@ishank011 we might need to check if the user has the privileges when setting the permissions, but for time being, from the UI this will be hidden, so is good to go for first stage.

ishank011:
@labkode yes, I check that at least one of these has to be set when executing this code https://github.com/cs3org/reva/blob/55515341de154311a69dc14e0780f2a382892e56/pkg/cbox/storage/eoswrapper/eoswrapper.go#L149

The issue is that EOS never really sets the 'm' permission, so the permission's never set for any directory that isn't owned by the user. As a workaround, I'm adding a check that if 'w' and 'x' are set, we allow resharing as well. Does that sound okay?

PR #2162, opened 2021-10-12
implement the UpdateStorageSpace method

C0rby:
Added the UpdateStorageSpace method to the decomposedfs.
Users can now set a quota to their spaces (personal and project).


C0rby:
Ok I have implemented the permission check in the graph service for now.

PR #2163, opened 2021-10-13
Nextcloud-based OCM Share Manager

michielbdejong:


michielbdejong:
@labkode please review

Issue #2160, opened 2021-10-12
Enable recycle bin for arbitrary paths


ishank011:
Currently, all recycle bin operations in oCIS are made available via the `/remote.php/dav/trash-bin/$username/$key/$path` endpoints. These are restricted to only the logged-in user's home, but access for shared spaces is not yet available.

Considering a user case, in CERNBox, we have service accounts associated with project spaces and users having access to a specific project space should be able to list and restore deleted entries. This is supported by the [ListRecycleRequest](https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.ListRecycleRequest) message in CS3APIs, so now, it needs to be added to clients.

Similar to the `/remote.php/webdav/$path` and `/remote.php/dav/$user/$path` parallel,
I propose to add a new API `/remote.php/webdav/trash-bin/$base-path/$key/$path`,
which will allow access to the recycle spin associated with other accounts, given that the user has access to `$base-path`. 
For example, `/remote.php/webdav/trash-bin/projects/physics-simulations`

Both the implementations can share the same code, as is the case with the old and new files endpoints.

Please let me know your thoughts about this @labkode @butonic @refs @C0rby @dragotin

ishank011:
> Regarding a spaces implementation for eos: that can be done incrementally. The easiest step is to list spaces, which is what we started with for the decomposedfs: glob all user homes, stat them and fetch the eos fileid, use it as the spaceid.

Yes, I would be very interested in taking this up for EOS but I'm afraid we don't have the bandwidth at the moment. Once we're done with our rollouts, I'd be happy to get started with this, but it won't be possible for the next 2-3 months at least.

PR #2165, opened 2021-10-13
Allow access to recycle bin for paths outside user homes

ishank011:
Uses https://github.com/cs3org/cs3apis/pull/152
Closes https://github.com/cs3org/reva/issues/2160


labkode:
@ishank011 just point to upstream merged go-cs3apis.

PR #2187, opened 2021-10-18
Harden decomposedfs and add unit tests

micbar:
Follow up of #2152


micbar:
@butonic i moved the storage space lookup to the node. pls review again

PR #2210, opened 2021-10-27
fix app provider new file action and improve app provider error codes

wkloucek:
## Changelog
Change: Fix app provider new file creation and improved error codes

We've fixed the behavior for the app provider when creating new files.
Previously the app provider would overwrite already existing files when creating a new file, this is now handled and prevented.
The new file endpoint accepted a path to a file, but this does not work for spaces. Therefore we now use the resource id of the folder where the file should be created and a filename to create the new file.
Also the app provider returns more useful error codes in a lot of cases.

## Related issues
- needs change in https://github.com/owncloud/web/pull/5890 after merge

### DIFF
--- a/internal/http/services/appprovider/appprovider.go
+++ b/internal/http/services/appprovider/appprovider.go
@@ -182,8 +182,10 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
 	}
 
 	fileRef := &provider.Reference{
-		ResourceId: parentContainer.Info.Id,
-		Path:       utils.MakeRelativePath(filename),
+		Path: path.Join(parentContainer.Info.Path, utils.MakeRelativePath(filename)),
+		// TODO: switch to a relative reference as soon as all storage providers support this
+		// ResourceId: parentContainer.Info.Id,
+		// Path:       utils.MakeRelativePath(filename),
 	}
 
 	statFileReq := &provider.StatRequest{
--- a/internal/http/services/appprovider/appprovider.go
+++ b/internal/http/services/appprovider/appprovider.go
@@ -19,13 +19,9 @@
 package appprovider
 
 import (
-	"context"
-	"encoding/base64"
 	"encoding/json"
-	"fmt"
 	"net/http"
-	"strings"
-	"unicode/utf8"
+	"path"
 
 	appregistry "github.com/cs3org/go-cs3apis/cs3/app/registry/v1beta1"
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
@@ -33,8 +29,6 @@ import (
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	typespb "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/internal/http/services/datagateway"
-	"github.com/cs3org/reva/internal/http/services/ocmd"
-	"github.com/cs3org/reva/pkg/errtypes"
 	"github.com/cs3org/reva/pkg/rgrpc/status"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/pkg/rhttp"
@@ -42,15 +36,11 @@ import (
 	"github.com/cs3org/reva/pkg/rhttp/router"
 	"github.com/cs3org/reva/pkg/sharedconf"
 	"github.com/cs3org/reva/pkg/utils"
+	"github.com/cs3org/reva/pkg/utils/resourceid"
 	ua "github.com/mileusna/useragent"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
-)
-
-const (
-	idDelimiter string = ":"
 )
 
 func init() {
@@ -61,6 +51,7 @@ func init() {
 type Config struct {
 	Prefix     string `mapstructure:"prefix"`
 	GatewaySvc string `mapstructure:"gatewaysvc"`
+	Insecure   bool   `mapstructure:"insecure"`
 }
 
 func (c *Config) init() {
@@ -115,17 +106,17 @@ func (s *svc) Handler() http.Handler {
 			case "open":
 				s.handleOpen(w, r)
 			default:
-				ocmd.WriteError(w, r, ocmd.APIErrorServerError, "unsupported POST endpoint", nil)
+				writeError(w, r, appErrorUnimplemented, "unsupported POST endpoint", nil)
 			}
 		case "GET":
 			switch head {
 			case "list":
 				s.handleList(w, r)
 			default:
-				ocmd.WriteError(w, r, ocmd.APIErrorServerError, "unsupported GET endpoint", nil)
+				writeError(w, r, appErrorUnimplemented, "unsupported GET endpoint", nil)
 			}
 		default:
-			ocmd.WriteError(w, r, ocmd.APIErrorServerError, "unsupported method", nil)
+			writeError(w, r, appErrorUnimplemented, "unsupported method", nil)
 		}
 	})
 }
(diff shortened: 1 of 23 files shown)


labkode:
@glpatcern can you review this one/

internal/http/services/appprovider/appprovider.go
```diff
@@ -146,16 +142,80 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	target := r.URL.Query().Get("filename")
-	if target == "" {
+	rootID := r.URL.Query().Get("file_root_id")
```
glpatcern, line 145: As this is now a new required parameter, can you reference the corresponding PR in web to pass it?
Also, a cosmetic comment: would it make more sense to call the parameter `root_file_id`?

glpatcern (commented):
Looks all good to me, including the wrap/unwrap code refactoring.

internal/http/services/appprovider/appprovider.go
glpatcern: Yes, even better!

wkloucek:
@glpatcern I went over the error handling / logging. Could you please have a look again?

glpatcern (dismissed):
Thanks, looks good - will merge once the build is complete

wkloucek:
> Thanks, looks good - will merge once the build is complete

Sorry, I found some more cases where the http status code didn't fit. It now looks good

C0rby (dismissed):

glpatcern (dismissed):
Looks all ok, only thing is I'm changing some user-visible error messages in https://github.com/cs3org/reva/pull/2258, and some messages may need further polishing - will rebase that PR on top and polish as needed.

pkg/utils/resourceid/owncloud.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+
+	return &provider.ResourceId{
+		StorageId: parts[0],
+		OpaqueId:  parts[1],
+	}
+}
+
+func unwrap(rid string) ([]string, error) {
```
ishank011, line 48: Let's return the resource ID here itself?

internal/http/services/appprovider/errors.go
```diff
@@ -0,0 +1,78 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package appprovider
```
ishank011, line 19: Why are we duplicating the code? Can we refactor the error handling in the ocmd service to a separate package so that all services can use it?

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -297,16 +297,6 @@ func (fs *Decomposedfs) NewUpload(ctx context.Context, info tusd.FileInfo) (uplo
 		ctx:      ctx,
 	}
 
-	if !info.SizeIsDeferred && info.Size == 0 {
```
ishank011, line 300: I'm not sure if this would work as we skip sending the patch request for zero-size files https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocdav/tus.go#L262

I think this should be covered by the integration tests

wkloucek: Looks like it isn't covered by tests since the CI is green. 
I don't really like the magic skip of zero byte uploads. Therefore i proposed: https://github.com/cs3org/cs3apis/pull/154.

internal/http/services/appprovider/errors.go
wkloucek: OCMD shouldn't expose it's error codes in the first place. They are highly specific for this one particular service (eg. `APIErrorUntrustedService APIErrorCode = "UNTRUSTED_SERVICE"` which maps to `APIErrorUntrustedService: http.StatusForbidden,` ).

I don't see any possibility of adding eg. `appErrorAlreadyExists    appErrorCode = "RESOURCE_ALREADY_EXISTS"` like status code to the OCMD service only to use it in the app provider since this is clearly a violation of concerns.

I agree that we could introduce a common error code mapping but this is out of scope for this PR.

pkg/storage/utils/decomposedfs/upload.go
wkloucek: I removed the skipping of zero byte uploads for now. That's something for a different PR.

pkg/utils/resourceid/owncloud.go
wkloucek: Done :+1:

internal/http/services/appprovider/errors.go
wkloucek: @ishank011 how can we continue on this?

wkloucek (commented):
early upload finish of zero byte files causes a 401 if we actually try to upload the file... needs investigation - back to draft

internal/http/services/appprovider/errors.go
ishank011: Sure, let's address it separately. Seems fine for now

micbar:
@wkloucek fixed by #2055 ?

wkloucek:
> @wkloucek fixed by #2055 ?

No, needs https://github.com/cs3org/reva/pull/2309

glpatcern:
Willy, now this is good to go right? I can merge it if you "undraft" it.

ishank011:
@wkloucek good to merge?

wkloucek:
> @wkloucek good to merge?

Yes, thanks for the ping!

glpatcern (dismissed):

internal/http/services/appprovider/appprovider.go
```diff
@@ -135,34 +126,87 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	// TODO(lopresti) if target is relative, currently the gateway fails to identify a storage provider (?)
-	// and just returns a CODE_INTERNAL error on InitiateFileUpload.
-	// Therefore for now make sure the target is absolute.
-	if target[0] != '/' {
-		target = "/" + target
+	fileRef := &provider.Reference{
+		ResourceId: parentContainer.Info.Id,
+		Path:       utils.MakeRelativePath(filename),
```
ishank011, line 186: Not all drivers support relative references yet. Can we just pass the path as `path.Join(parentContainer.Info.Path, utils.MakeRelativePath(filename)`?

internal/http/services/appprovider/appprovider.go
```diff
@@ -135,34 +126,87 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	filename := r.URL.Query().Get("filename")
+	if filename == "" {
+		writeError(w, r, appErrorInvalidParameter, "missing filename", nil)
+		return
+	}
+
+	dirPart, filePart := path.Split(filename)
+	if dirPart != "" || filePart != filename {
```
ishank011, line 158: Not sure about this. We support requests like `storageID:nodeID!filename` but not `storageID:nodeID!some-dir/filename`. If we say that relative paths are supported, both should be allowed in my opinion

wkloucek: The `/app/new` requests requires two query parameters:
- `parent_container_id`
- `filename`

The code ensures that `filename` is only a filename and does not contain a path segment. I don't see any reason why we should support relative paths in that particular case since we would then need to handle non existent path segments and much more corner cases.  The container referenced by `parent_container_id` can be easily checked if it exists and permissions are sufficient.

I don't see any real world uses cases for these examples:
Example 1:
- `parent_container_id` = `MTI4NGQ...==` => reference to `/home/test/documents`
- `filename` = `../../Document1.odt`

-> one could easily take the reference of `/home` in the first place

Example 2:
- `parent_container_id` = `MTI4NGQ...==` => reference to `/home/test/documents`
- `filename` = `./office/Document2.odt`
-> we would need to check that the folder `/home/test/documents/office` actually exists, which adds extra complexity

internal/http/services/appprovider/appprovider.go
wkloucek: Is changed ;-)

internal/http/services/appprovider/appprovider.go
```diff
@@ -182,8 +182,10 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
 	}
 
 	fileRef := &provider.Reference{
-		ResourceId: parentContainer.Info.Id,
-		Path:       utils.MakeRelativePath(filename),
+		Path: path.Join(parentContainer.Info.Path, utils.MakeRelativePath(filename)),
+		// TODO: switch to a relative reference as soon as all storage providers support this
```
ishank011, line 186: Can we just remove the comments? It doesn't offer any advantages AFAICT

internal/http/services/appprovider/appprovider.go
ishank011: Sure, the web client doesn't need to use this format but I meant for the sake of consistency since these types of requests are supported elsewhere. Adding the extra check sounds okay to me if it means we're consistent across services

internal/http/services/appprovider/appprovider.go
wkloucek: I removed them

internal/http/services/appprovider/appprovider.go
ishank011: I'll merge the PR now and create an issue to keep a track of this part

ishank011 (approved):

PR #2212, opened 2021-10-27
Filter root path according to the agent that makes the request

gmgigi96:


internal/grpc/services/gateway/gateway.go
```diff
@@ -69,6 +69,7 @@ type config struct {
 	HomeMapping         string                            `mapstructure:"home_mapping"`
 	TokenManagers       map[string]map[string]interface{} `mapstructure:"token_managers"`
 	EtagCacheTTL        int                               `mapstructure:"etag_cache_ttl"`
+	HiddenRootFolders   map[string][]string               `mapstructure:"hidden_root_folders"`
```
ishank011, line 72: Have it the other way round, `path -> userAgentList`. Call it `AllowedUserAgents`

internal/grpc/services/gateway/gateway.go
```diff
@@ -116,10 +117,11 @@ func (c *config) init() {
...
 type svc struct {
-	c              *config
-	dataGatewayURL url.URL
-	tokenmgr       token.Manager
-	etagCache      *ttlcache.Cache `mapstructure:"etag_cache"`
+	c                 *config
+	dataGatewayURL    url.URL
+	tokenmgr          token.Manager
+	etagCache         *ttlcache.Cache `mapstructure:"etag_cache"`
+	hiddenRootFolders map[string]map[string]struct{}
```
ishank011, line 124: Let's just store the map as passed by the user. At max, the list will contain 5 elements, searching over those won't lead to any performance drawback

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1648,6 +1650,31 @@ func (s *svc) listSharesFolder(ctx context.Context) (*provider.ListContainerResp
...
+func (s *svc) isFolderHidden(ua, path string) bool {
+	uaSet, ok := s.hiddenRootFolders[ua]
+	if !ok {
+		return false
+	}
+	_, ok = uaSet[path]
+	return ok
+}
+
+func (s *svc) filterProvidersByUserAgent(ctx context.Context, providers []*registry.ProviderInfo) []*registry.ProviderInfo {
```
ishank011, line 1662: Move this logic to the useragent pkg. It should expose only one method `IsUserAgentAllowed(userAgent, allowedUserAgents) bool`

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1657,8 +1684,19 @@ func (s *svc) listContainer(ctx context.Context, req *provider.ListContainerRequ
...
-	if len(providers) == 1 && (utils.IsRelativeReference(req.Ref) || resPath == "" || strings.HasPrefix(resPath, providers[0].ProviderPath)) {
+	p, st := s.getPath(ctx, req.Ref, req.ArbitraryMetadataKeys...)
+	if st.Code != rpc.Code_CODE_OK {
+		return &provider.ListContainerResponse{
+			Status: st,
+		}, nil
+	}
+
+	// check if the path is the root folder
+	if path.Clean(p) == "/" {
```
ishank011, line 1695: Move this code to `listContainerAcrossProviders`. That method is called when there are more than one provider matching a requested path. And there you don't need to call `getPath` as the requests will always be by path. And you can remove the root folder check as well

pkg/ctx/agentctx.go
```diff
@@ -25,16 +25,22 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// UserAgentHeader is the header used for the user agent
+const UserAgentHeader = "x-user-agent"
```
ishank011, line 29: Don't use a separate header. It's logged in other places so we should display the one we're adding

pkg/rgrpc/rgrpc.go
```diff
@@ -292,6 +293,7 @@ func (s *Server) getInterceptors(unprotected []string) ([]grpc.ServerOption, err
 		token.NewUnary(),
 		log.NewUnary(),
 		recovery.NewUnary(),
+		useragent.NewUnary(),
```
ishank011, line 296: Move this above the log unary. I tried to move it above that but I still see the `grpc-go` agent being logged
`2021-11-04 14:00:55.491 DBG ../../internal/grpc/interceptors/log/log.go:69 > unary code=OK end="04/Nov/2021:14:00:55 +0100" from=tcp://127.0.0.1:52341 pid=46490 pkg=rgrpc start="04/Nov/2021:14:00:55 +0100" time_ns=2803304 uri=/cs3.storage.provider.v1beta1.ProviderAPI/ListContainer user-agent=grpc-go/1.26.0`

ishank011 (changes requested):

ishank011:
Why don't we use this [here](https://github.com/cs3org/reva/blob/master/pkg/rgrpc/todo/pool/pool.go#L88)?
https://github.com/grpc/grpc-go/issues/1100#issuecomment-372132051

`server := grpc.Dial(grpc_server, WithUserAgent("user-agent",))`

gmgigi96:
> Why don't we use this [here](https://github.com/cs3org/reva/blob/master/pkg/rgrpc/todo/pool/pool.go#L88)? [grpc/grpc-go#1100 (comment)](https://github.com/grpc/grpc-go/issues/1100#issuecomment-372132051)
> 
> `server := grpc.Dial(grpc_server, WithUserAgent("user-agent",))`

I will leave a comment here to keep track on what we discussed yesterday and other thought I have after playing with user agents in GRPC.

Actually this is the way of specifying a user agent in a grpc call, that internally updates the outcoming metadata in the context when you will call a grpc method, and you can find the user agent in the incoming metadata in the context server-side. But, even if you specify your own user agent, this will be concatenated with the string "grpc-go/<version>". I do not know if this can broke the library we use to parse the user agents, maybe not. But the main problem of using this approach is that for each time you want to get a grpc client for a service you need to specify the user agent that will do the request, if not specified, grpc automatically will set its default ("grpc-go/<version>"). This means that we should modify all the methods in the pool pkg to receive a user agent. But I do not like this approach as we cannot cache anymore the grpc connection for each endpoint address, but we should cache also the user agent. In fact after a dial, there is no way to override it after creating a new connection (I tried also modifying the internal metadata of a context before a call, but apparently grpc reset it with the user agent passed to the Dial method).

So I preferred to define my own user agent token that will be managed internally by reva and not by grpc. With this approach we do not have the cons mentioned above. Of course the user agent log did not work, as it was reading the grpc user-agent metadata, and not the new one. Now it’s fixed and you should see the correct user agent in the log.

ishank011 (approved):

PR #2214, opened 2021-10-27
Enforce permissions in public share apps

C0rby:
A receiver of a read-only public share could still edit files via apps like Collabora.
These changes enforce the share permissions in apps used on publicly shared resources.

### DIFF
--- a/internal/grpc/services/publicstorageprovider/publicstorageprovider.go
+++ b/internal/grpc/services/publicstorageprovider/publicstorageprovider.go
@@ -50,12 +50,14 @@ func init() {
 
 type config struct {
 	MountPath   string `mapstructure:"mount_path"`
+	MountID     string `mapstructure:"mount_id"`
 	GatewayAddr string `mapstructure:"gateway_addr"`
 }
 
 type service struct {
 	conf      *config
 	mountPath string
+	mountID   string
 	gateway   gateway.GatewayAPIClient
 }
 
@@ -88,6 +90,7 @@ func New(m map[string]interface{}, ss *grpc.Server) (rgrpc.Service, error) {
 	}
 
 	mountPath := c.MountPath
+	mountID := c.MountID
 
 	gateway, err := pool.GetGatewayServiceClient(c.GatewayAddr)
 	if err != nil {
@@ -97,6 +100,7 @@ func New(m map[string]interface{}, ss *grpc.Server) (rgrpc.Service, error) {
 	service := &service{
 		conf:      c,
 		mountPath: mountPath,
+		mountID:   mountID,
 		gateway:   gateway,
 	}
 
@@ -142,7 +146,7 @@ func (s *service) translatePublicRefToCS3Ref(ctx context.Context, ref *provider.
 		return nil, "", nil, nil, err
 	}
 
-	originalPath, ls, _, st, err := s.resolveToken(ctx, tkn)
+	ls, shareInfo, st, err := s.resolveToken(ctx, tkn)
 	switch {
 	case err != nil:
 		return nil, "", nil, nil, err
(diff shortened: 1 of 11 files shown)


refs (dismissed):

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -458,9 +462,22 @@ func (s *service) Stat(ctx context.Context, req *provider.StatRequest) (*provide
...
+		nodeID = parts[1]
+	} else if req.Ref.Path != "" {
+		var err error
+		tkn, relativePath, err = s.unwrap(ctx, req.Ref)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	originalPath, ls, ri, st, err := s.resolveToken(ctx, tkn)
```
ishank011, line 483: If `ri` is a file, then `relativePath` should be empty, right? We can save the extra stat call in that case and return early

Similarly, in `resolveToken`, we call `GetPath` and `Stat` but we can just return the path returned by the stat response

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -458,48 +462,95 @@ func (s *service) Stat(ctx context.Context, req *provider.StatRequest) (*provide
...
+	case share.GetPermissions() == nil || !share.GetPermissions().Permissions.Stat:
 		return &provider.StatResponse{
 			Status: status.NewPermissionDenied(ctx, nil, "share does not grant Stat permission"),
 		}, nil
 	}
 
-	p := originalPath
-	if ri.Type == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
-		p = path.Join("/", p, relativePath)
+	if shareInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE {
```
ishank011, line 499: `if shareInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE || (relativePath == "" && nodeID == "")`

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -458,48 +462,95 @@ func (s *service) Stat(ctx context.Context, req *provider.StatRequest) (*provide
...
+	if res.Info != nil {
+		if err := addShare(res.Info, share); err != nil {
+			appctx.GetLogger(ctx).Error().Err(err).Interface("share", share).Interface("info", res.Info).Msg("error when adding share")
 		}
-		statResponse.Info.Path = path.Join(s.mountPath, "/", tkn, relativePath)
-		filterPermissions(statResponse.Info.PermissionSet, ls.GetPermissions().Permissions)
+
+		var sharePath string
+		if shareInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE {
+			sharePath = path.Base(shareInfo.Path)
```
ishank011, line 540: One question, but I might be misunderstanding this. `sharePath` should be empty for file shares right? Because `/$mountID/$token` will resolve to the file itself. If you then append the file name to the path obtained from the token, it won't exist.

C0rby: Yes, you're right. Now I'm wondering why it worked so far. :thinking:

ishank011 (dismissed):

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -458,48 +462,93 @@ func (s *service) Stat(ctx context.Context, req *provider.StatRequest) (*provide
...
+	case share.GetPermissions() == nil || !share.GetPermissions().Permissions.Stat:
 		return &provider.StatResponse{
 			Status: status.NewPermissionDenied(ctx, nil, "share does not grant Stat permission"),
 		}, nil
 	}
 
-	p := originalPath
-	if ri.Type == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
-		p = path.Join("/", p, relativePath)
+	if shareInfo.Type == provider.ResourceType_RESOURCE_TYPE_FILE || (relativePath == "" && nodeID == "") {
```
ishank011, line 499: Sorry one more optimization 😅 
We can also return if `shareInfo.Id. OpaqueId == nodeID`

internal/grpc/services/publicstorageprovider/publicstorageprovider.go
C0rby: Appending the filename to the path is necessary otherwise something breaks. If you don't want to have the filename in there then I need to dig into why it's required. But without it other things like downloading shared files fail.

ishank011: No it's fine by me. I don't understand why it worked previously. Now because of the check on line 499, we return early for file shares. But if we didn't you'd essentially be statting `path-to-parent-dir/file.txt/file.txt` in line 519. Merging for now but please look into it if you have the time. The issue might be on the `web` end.

Issue #1981, opened 2021-08-09
Why require providerId to contain a `:`?


michielbdejong:
https://cs3org.github.io/OCM-API/docs.html#/paths/~1shares/post requires `providerId` to be a string that is unique per provider.

Revad furthermore [requires](https://github.com/cs3org/reva/blob/v1.11.0/internal/grpc/services/ocmcore/ocmcore.go#L111-L122)  that `providerId` is of the form `StorageId:OpaqueId`. Why is this?

When I look at for instance https://github.com/cs3org/reva/blob/v1.11.0/pkg/utils/utils.go#L156-L159 it looks like it might as well have been one unique string instead of two strings with a `:` between them?

Also, since the OCM API spec does not require this, this will make reva incompatible with other OCM implementations.

butonic:
> Identifier to identify the resource at the provider side. This is unique per provider.

reva uses the `StorageId` to route requests to the correct storage provider, which in turn then uses the `OpaqueId` to identify the resource. That was always the case. But for the owncloud ocs and webdav apis we had to glue them together to a single string for legacy reasons. I don't see a conflict with the OCM api, yet. It does not restrict the character set or require a uuid. To other instances the providerid should be an opaque string.

Could you give an example?

michielbdejong:
Sure, it's fine that reva does that internally, and it's also fine when reva is the provider, then it's at liberty to compose the string in that way.

What breaks is when an unsuspecting provider (unrelated to reva) sends a share over OCM to a reva instance. Then a 400 is thrown, from this line:

https://github.com/cs3org/reva/blob/v1.11.0/internal/grpc/services/ocmcore/ocmcore.go#L113

That's the problem I ran into when testing reva's OCM implementation.

glpatcern:
This was fixed at some point, this incorrect requirement is gone

Issue #1962, opened 2021-08-06
How to run revad with TLS enabled


michielbdejong:
Can I run revad with a TLS certificate for grpc and https? If so, how?
I searched for this in the docs and in the code but couldn't find how to do it.

michielbdejong:
While it seems to be impossible to run revad with https directly, it does refer to itself with https for instance here:
https://github.com/cs3org/reva/blob/v1.11.0/internal/http/services/ocmd/config.go#L64

So should revad always be run behind a TLS-offloading proxy, then? Or am I missing something?

michielbdejong:
Answered in https://gitter.im/cs3org/REVA?at=61112c45933ee46b754a5fe4

michielbdejong:
I think it's probably easier to include https support in reva than to tell people (including myself with the ocm-test-suite) to run a separate proxy. I'll see if I can make it work.
These two lines are key:
* [rhttp.New](https://github.com/cs3org/reva/blob/v1.14.0/cmd/revad/runtime/runtime.go#L259)
* [rgrpc.NewServer](https://github.com/cs3org/reva/blob/v1.14.0/cmd/revad/runtime/runtime.go#L249)

michielbdejong:
In [starting the grpc server](https://github.com/cs3org/reva/blob/v1.14.0/pkg/rgrpc/rgrpc.go#L147) and [starting the http server](https://github.com/cs3org/reva/blob/v1.14.0/pkg/rhttp/rhttp.go#L113) we would then probably just use [s.ServeTLS](https://pkg.go.dev/net/http#Server.ServeTLS) instead of 
[s.Serve](https://pkg.go.dev/net/http#Server.Serve).

michielbdejong:
Next error:
```
C2021-10-20 13:08:04.967 DBG ../../reva/internal/grpc/interceptors/auth/auth.go:92 > skipping auth method=/cs3.gateway.v1beta1.GatewayAPI/ListAuthProviders pid=8 pkg=rgrpc
2021-10-20 13:08:04.973 ERR ../../reva/internal/grpc/services/gateway/authregistry.go:45 > gateway error="gateway: error calling ListAuthProviders: rpc error: code = Unavailable desc = connection closed" pid=8 pkg=rgrpc
[...]

Issue #1753, opened 2021-06-01
Support json POST for creating shares?


michielbdejong:
Trying this curl command:
```sh
curl -d'{ "shareWith": "42", "protocol": "42", "meshProvider": "42", "name": "42", "providerId": "42", "owner": "42" }' -H 'content-type: application/json' http://einstein:relativity@localhost:19001/ocm/shares
```
it's unable to extract any of these strings at https://github.com/cs3org/reva/blob/master/internal/http/services/ocmd/shares.go#L63-L64 apparently. 

It does work when you use `application/x-www-form-urlencoded`:
```sh
curl -d'shareWith=42&protocol=42&meshProvider=42&name=42&providerId=42&owner=42' http://einstein:relativity@localhost:19001/ocm/shares
```
(actually there is an 'unauthenticated error then, but at least it finds all 6 post field values).

Issue #1752, opened 2021-06-01
endPoint field in ocm-provider json should be camelCase?


michielbdejong:
On http://localhost:19001/ocm/ocm-provider/ I see:
```json
{
   "enabled": true,
   "apiVersion": "1.0-proposal1",
   "host": "localhost",
   "endpoint": "https://localhost/ocm",
   "provider": "cernbox",
   "resourceTypes": [
      {
         "name": "file",
         "shareTypes": [
            "user"
         ],
         "protocols": {
            "webdav": "/cernbox/ocm_webdav"
         }
      }
   ]
}
```
but my code is looking for `endPoint`, not `endpoint`.

[Question to self:] Is this part of OCM documented somewhere?

michielbdejong:
Looking at https://github.com/cs3org/OCM-API/pull/37 it was suggested as camelCase, but apparently this didn't make it into https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L33.

glpatcern:
This was fixed in one of the last PRs around OCM

PR #2229, opened 2021-11-02
Pass OCM test suite

michielbdejong:
A number of changes in this PR:
* #1752 spell 'endpoint' instead of 'endPoint'
* #1753 support JSON for POST bodies in the OCM API
* #1962 support running the rhttp service with TLS
* #1981 support remote resource identifiers that don't contain a `:`
* support running the ocm service on the domain root

Some of these changes are technically breaking changes (especially the last two) so deploy with care.


michielbdejong:
Closing as stale.

PR #2230, opened 2021-11-03
fix open by default app and expose default app

wkloucek:
needs: https://github.com/cs3org/cs3apis/pull/157

needed because of https://github.com/owncloud/web/pull/5970#issuecomment-958745928

After this PR the `/app/list` endpoint will provide a default app (if configured):
``` json
    {
      "mime_type": "application/vnd.oasis.opendocument.text",
      "ext": "odt",
      "app_providers": [
        {
          "name": "Collabora",
          "icon": "https://www.collaboraoffice.com/wp-content/uploads/2019/01/CP-icon.png"
        }
      ],
      "name": "OpenDocument",
      "description": "OpenDocument text document",
      "allow_creation": true,
      "default_application": "Collabora"
    }
```

### DIFF
--- a/changelog/unreleased/fix-default-app-handling.md
+++ b/changelog/unreleased/fix-default-app-handling.md
@@ -1,0 +1,7 @@
+Bugfix: Fix open by default app and expose default app
+
+We've fixed the open by default app name behaviour which previously only worked, if the default app was configured by the provider address.
+We also now expose the default app on the `/app/list` endpoint to clients.
+
+https://github.com/cs3org/reva/issues/2230
+https://github.com/cs3org/cs3apis/pull/157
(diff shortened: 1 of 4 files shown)


glpatcern (approved):
Looks good, clearly having the default semantically explicit is better than just relying on the sorting of the app providers, which was definitely temporary.

At this point we could actually remove the logic to keep the app providers list ordered.

wkloucek:
@glpatcern I updated the CS3apis, can we get this in?

glpatcern:
(did not merge yet because I'm not code owner - will check with @ishank011 also to update the `CODEOWNERS` rules for those subtrees)

ishank011 (approved):

PR #2231, opened 2021-11-03
Fix share permission on a file in sql (cbox pkg) share driver

gmgigi96:
While creating a share permission with write permission, the cbox/sql share driver converted this to a permission value that corresponds to a viewer role. This PR fix this.


ishank011 (approved):
Nice debugging!

PR #2235, opened 2021-11-04
Specify a list of allowed folders/files to be archived

gmgigi96:
This PR adds a configuration to the archiver service in order to specify a list of folders (as regex) that can be archived.


ishank011 (commented):
Looks good but what kind of regexes are you thinking? Something like `/eos/user/[a-z]/[a-z]./\w.`? Can we come up with something simpler?

labkode:
I propose we keep it like that for time being, it should work. If we want to simplify we can do it later.

labkode (approved):

PR #2239, opened 2021-11-05
cleanup toml config, add ocmd, drop gen subcommand

butonic:
We updated the local and drone configurations, cleanad up the example configs and removed the reva gen subcommand which was generating outdated config.

This is in preparation for OCM tests. They are not included yet, but I want to see if starting ocmd breaks anything in the existing testsuite. Please merge as soon as CI passes.


examples/ocmd/ocmd-server-2.toml
```diff
@@ -87,8 +87,6 @@ driver = "json"
 users = "users.demo.json"
 
 [http]
-enabled_services = ["ocmd"]
```
ishank011, line 90: @butonic can you add these back in `ocmd-server-[1, 2]`?

ishank011 (approved):

PR #2247, opened 2021-11-08
Merge user ACLs from EOS to sys ACLs

ishank011:


labkode:
@ishank011 can you give an explanation for this fix?

ishank011:
@labkode for file shares, we need to set user ACLs as EOS doesn't obey sys ACLs yet. These need to be passed as well when statting or listing resources.

One use case is that when opening a file, we stat it to determine what permissions a user has on it to determine the view mode.

labkode (approved):

Issue #2246, opened 2021-11-08
EOS GRPC doesn't populate the sys ACL field


ishank011:
We need something like
https://github.com/cs3org/reva/blob/80f5ec9b8978bd659459a0db6db6ea9778119169/pkg/eosclient/eosbinary/eosbinary.go#L1089-L1092

ffurano:
Having a look

ffurano:
So, it seems to me that you want the ACL field to contain the content of the xattr field "sys.acl" in addition to the "real" ACLs that are returned by EOS/grpc (if any)
Easy to do, yet since this seems strange to me, could you please confirm that I understood right?

ishank011:
Aren't these the "real" ACLs? Using `file info` you only get this

PR #2252, opened 2021-11-09
eosgrpc and HTTP - Small fixes for ACLs and client certs

ffurano:
Fulfils:
https://github.com/cs3org/reva/issues/2246


ishank011 (commented):
Please also add user ACLs if `sys.eval.useracl` is set. Since EOS doesn't obey sys ACLs for files, we have to set these as user attrs.

https://github.com/cs3org/reva/blob/519aa3d5b09c4c4b60104e73bb6f378424c14608/pkg/eosclient/eosbinary/eosbinary.go#L1098-L1112

labkode (approved):

PR #2253, opened 2021-11-09
Support the file editor role for public links

ishank011:


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -499,6 +499,8 @@ func permissionFromRequest(r *http.Request, h *Handler) (*provider.ResourcePermi
 var ocPublicPermToRole = map[int]string{
 	// Recipients can view and download contents.
 	1: "viewer",
+	// Recipients can view, download and edit single files.
+	3: "file-editor",
```
butonic, line 503: amounts to Read (1) + Write (2): https://github.com/owncloud/core/blob/f685868208be6b4c6fd9104faf18a4c710f1539b/lib/public/Constants.php#L43-L48

butonic (approved):

labkode (approved):

PR #2256, opened 2021-11-10
Return user type in the response of the ocs GET user call

ishank011:


internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
```diff
@@ -65,6 +66,7 @@ type Users struct {
 	Quota       *Quota `json:"quota" xml:"quota"`
 	Email       string `json:"email" xml:"email"`
 	DisplayName string `json:"displayname" xml:"displayname"`
+	UserType    string `json:"usertype" xml:"usertype"`
```
C0rby, line 69: You added `user-type` to the other struct. I would keep the format consistent, so either add the `-` here or remove it from the other struct.

C0rby (changes requested):

internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
ishank011: That's because it's the same for `display-name`, so I thought it's by design. Should I fix both? That'll probably break the tests though

C0rby: Ah I see. Yeah, don't fix the display name in here. We had some problems with that before so I think that should be tackled in another PR. The new field though should be introduced consistently so I would vote for `user-type`.

ishank011: Cool, will do

labkode (approved):

C0rby (approved):

PR #2257, opened 2021-11-10
fix nil pointer during propfinds

C0rby:
Fix a nil pointer which was introduced in a previous PR.
This should resolve the CI issue here: https://github.com/owncloud/ocis/pull/2726


wkloucek (approved):

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -777,8 +777,10 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 						types.WriteString("</oc:share-type>")
 					}
 
-					if _, ok := linkshares[md.Id.OpaqueId]; ok {
-						types.WriteString("<oc:share-type>3</oc:share-type>")
+					if len(linkshares) > 0 {
```
ishank011, line 780: `linkshares` might also be nil?

C0rby: Yes, but the length of a nil map in go is 0.
See https://play.golang.org/p/FLeZQaEK9wF.
Also while making this example for you I noticed that it might not even be the map which is causing our problem but maybe the `md.Id`.
Let me investigate again...

C0rby: Yep, my bad. It wasn't the map which is the problem but the id. A nil map just acts like an empty map.

ishank011 (approved):

internal/http/services/owncloud/ocdav/propfind.go
ishank011: Ah cool, didn't know. I can remove such checks now :D

PR #2263, opened 2021-11-11
Add priority to app providers

gmgigi96:
Before the order of the list returned by the method FindProviders of app providers depended from the order in which the app provider registered themselves.
Now, it is possible to specify a priority for each app provider, and even if an app provider re-register itself (for example after a restart), the order is kept.

This PR also remove all the code used to keep the default app provider always to the top of the list for each mime type, as now the cs3apis support the field DefaultApplication (https://github.com/cs3org/reva/pull/2230 https://github.com/cs3org/cs3apis/pull/157)

### DIFF
--- a/changelog/unreleased/app-registry-priority.md
+++ b/changelog/unreleased/app-registry-priority.md
@@ -1,0 +1,12 @@
+Enhancement: Add priority to app providers
+
+Before the order of the list returned by the method FindProviders
+of app providers depended from the order in which the app provider registered
+themselves.
+Now, it is possible to specify a priority for each app provider, and even if
+an app provider re-register itself (for example after a restart), the order
+is kept.
+
+https://github.com/cs3org/reva/pull/2230
+https://github.com/cs3org/cs3apis/pull/157
+https://github.com/cs3org/reva/pull/2263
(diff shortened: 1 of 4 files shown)


wkloucek:
Code changes look good to me.
But I didn't fully understand the client use case for ordered app lists. How does a client benefit from it? Is is about ordering in the apps menu? (Example from oC Web see screenshot)

image

If that's the case I wonder what happens if all apps have the same priority? Should we then do an alphabetical order, too? Or isn't it up to the client how to sort these apps?

labkode:
> Code changes look good to me. But I didn't fully understand the client use case for ordered app lists. How does a client benefit from it? Is is about ordering in the apps menu? (Example from oC Web see screenshot)
> 
> image
> 
> If that's the case I wonder what happens if all apps have the same priority? Should we then do an alphabetical order, too? Or isn't it up to the client how to sort these apps?

@wkloucek the aim here is that the list presented to the user doesn't change as the service restarts to reduce confusion from the end-user perspective.

Imagine, you click and you have "Open in OnlyOffice" as first entry and then when the service restarts you'll have it at the end.

wkloucek (approved):

labkode (approved):

PR #2285, opened 2021-11-16
Fix ocmd tutorial

### LINKED ISSUE #2284: provider not authorized when following https://reva.link/docs/tutorials/share-tutorial/
Unless it's something I'm doing wrong in my local setup, it seems this broke somewhere between v1.10.0 and master.
Investigating.

michielbdejong:
Reverts part of @butonic's [commit here](https://github.com/cs3org/reva/commit/c35c530fb17661bb7009a33b2678c87617495991#diff-56e04bcb7910dece3dfdbe078825ba79b3ce995f8911fc119c504d4b3a9ee93c)
Reverts part of @butonic's [commit here](https://github.com/cs3org/reva/commit/a4fb51f3155627c0496050f680a4590c35a52526#diff-fe6c2c8565d9bbe5e39cbb10bd6b8efa75801895a6375ef1204a961e3c6e71e2)
Fixes #2284

### DIFF
--- a/examples/ocmd/users.demo.json
+++ b/examples/ocmd/users.demo.json
@@ -2,7 +2,7 @@
 	{
 		"id": {
 			"opaque_id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
-			"idp": "https://cernbox.cern.ch",
+			"idp": "cernbox.cern.ch",
 			"type": 1
 		},
 		"username": "einstein",
@@ -28,7 +28,7 @@
 	{
 		"id": {
 			"opaque_id": "f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c",
-			"idp": "https://cesnet.cz",
+			"idp": "cesnet.cz",
 			"type": 1
 		},
 		"username": "marie",
@@ -54,7 +54,7 @@
 	{
 		"id": {
 			"opaque_id": "932b4540-8d16-481e-8ef4-588e4b6b151c",
-			"idp": "https://example.org",
+			"idp": "example.org",
 			"type": 1
 		},
 		"username": "richard",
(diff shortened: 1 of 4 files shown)


michielbdejong:
@butonic this may not be the desired fix - maybe it's better to leave users.demo.json as you changed it (including the 'https://') and then edit to providers.demo.json to match it?

examples/ocmd/users.demo.json
```diff
@@ -2,7 +2,7 @@
 	{
 		"id": {
 			"opaque_id": "4c510ada-c86b-4815-8820-42cdf82c3d51",
-			"idp": "https://cernbox.cern.ch",
+			"idp": "cernbox.cern.ch",
```
butonic, line 5: The id really should be a url. It may be compared to the oidc provider, which always is a url. Including the protocol.

butonic (commented):
Just commenting why I thought these were wrong. I don't want people to assume the examples are correct. Is this example config used in any tests?

examples/ocmd/users.demo.json
michielbdejong: By 'id' you mean the "idp" field in a UserId object? I agree that [the cs3apis documentation](https://github.com/cs3org/cs3apis/blob/9648574fa9963be02dc3bd2700c0639ee8708492/cs3/identity/user/v1beta1/resources.proto#L36) is undecisive and seems to tend towards making it a URL, but [reva's current providerauthorizer interceptor code](https://github.com/cs3org/reva/blob/33deec960655e886fb399fc32c1524cf97d9dfa1/internal/http/interceptors/providerauthorizer/providerauthorizer.go#L87) clearly requires it to be a domain.

I'll propose a non-breaking change in the providerauthorizer code so that it strips the protocol (if present) from the user's idp string before comparing it to the provider's domain string.

I'll also create a PR on the c3apis repo to make this clearer.

michielbdejong:
Closing this, let's discuss the desired string format in https://github.com/cs3org/cs3apis/pull/159 and then we can adapt the code to that decision.

PR #2289, opened 2021-11-17
Accept old and new userid idp format in provider authorizer interceptor

michielbdejong:


ishank011:
@Daniel-WWU-IT I think the protocols were removed from the IDPs because GOCDB doesn’t work with those. Can you take a look?

michielbdejong:
Providers have a 'domain' field, which by its name is obviously without protocol.
But @butonic recently added (not removed) the protocols in UserID IDP strings in the users.demo.toml files, hence the need for this code change.

Daniel-WWU-IT:
> @Daniel-WWU-IT I think the protocols were removed from the IDPs because GOCDB doesn’t work with those. Can you take a look?

@ishank011 Could you elaborate on what I should be looking at or for? As @michielbdejong already mentioned, the OCM data contains both a domain w/o protocol and a homepage w/ protocol, and both values are stored in the GOCDB separately. So I really don't know what you mean...

michielbdejong:
Will reopen this PR with the branch on my own fork, sorry.

Issue #2292, opened 2021-11-19
downloading an archive with invalid path returns HTTP/500


individual-it:
## Describe the bug

When trying to download a not-existing path using the archiver endpoint the server returns an HTTP 500 error

## Steps to reproduce

1. install and run ocis with `OCIS_INSECURE=true PROXY_ENABLE_BASIC_AUTH=true bin/ocis server`
1. `curl -k -u einstein:relativity 'https://localhost:9200/archiver?path=/home/doesnotexist' --output /tmp/out.tar -v`

## Expected behavior
HTTP/404

## Actual behavior
HTTP/500

## Additional context
```
{"level":"warn","service":"storage","pkg":"rgrpc","time":"2021-11-15T14:53:06+05:45","caller":"/home/artur/www/ocis/vendor/github.com/cs3org/reva/internal/grpc/services/storageprovider/storageprovider.go:647","message":"path not found when statting"}
{"level":"error","service":"storage","pkg":"rhttp","time":"2021-11-15T14:53:06+05:45","message":"error: not found: /home/doesnotexist"}
{"level":"error","service":"storage","pkg":"rhttp","host":"127.0.0.1","method":"GET","uri":"/archiver?path=/home/doesnotexist","url":"/archiver?path=/home/doesnotexist","proto":"HTTP/1.1","status":500,"size":0,"start":"15/Nov/2021:14:53:06 +0545","end":"15/Nov/2021:14:53:06 +0545","time_ns":1050715,"time":"2021-11-15T14:53:06+05:45","message":"http"}
{"level":"info","service":"proxy","proto":"HTTP/1.1","request":"deepthought3/kRSej2PSey-000304","remote-addr":"127.0.0.1:54884","method":"GET","status":500,"path":"/archiver","duration":213.575404,"bytes":0,"time":"2021-11-15T14:53:06+05:45","message":"access-log"}

individual-it:
reopening, see https://github.com/owncloud/ocis/issues/2795

amrita-shrestha:
> reopening, see [owncloud/ocis#2795](https://github.com/owncloud/ocis/issues/2795)

@individual-it Could we close this issue because https://github.com/owncloud/ocis/issues/2768 has been closed in ocis?

[...]

PR #2294, opened 2021-11-19
Fix HTTP status code when path is invalid

gmgigi96:
Before when a path was invalid, the archiver returned a 500 error code.
Now this is fixed and returns a 404 code.

Closes https://github.com/cs3org/reva/issues/2292


labkode (approved):

individual-it:
Im missing unit tests in this PR. The only tests were have for this is in ocis, so the check if it works or has broken anything else will only happen after the next pull off Reva to ocis.
I think it would be good to add unit tests here that fail before and pass after the fix

PR #2297, opened 2021-11-19
Fix public link paths for file shares

ishank011:


ishank011:
@C0rby this takes care of the issue of why tests were failing in https://github.com/cs3org/reva/pull/2214

The download requests go to `/remote.php/dav/public-files/$token/a.txt/a.txt` for file links. Without having the file name in the href twice, the downloaded file is named as the `$token`. But without these fixes, the downloads don't work for me

C0rby (approved):

labkode (approved):

butonic:
the actual problem is that the Content-Disposition Header is wrong:
```
content-disposition | attachment; filename*=UTF-8''gKGaNZfxXMpbVMm; filename="gKGaNZfxXMpbVMm"
```

I struggle to make chrome or edge show me actual download requests. But with firefox this can be seen. I'll comment with a proper fix ...

PR #2305, opened 2021-11-25
Make sure /app/new takes target as absolute path

glpatcern:
A mini-PR to make the `target` parameter absolute (by prepending `/` if missing).

The real issue here is that a relative `target` seems to make the gateway fail to identify a storage provider, and the `InitiateFileUpload` request fails with `CODE_INTERNAL`. And the web frontend happened to send requests with relative paths.

Independently from fixing the web frontend, this patch protects the app provider.

### DIFF
--- a/internal/http/services/appprovider/appprovider.go
+++ b/internal/http/services/appprovider/appprovider.go
@@ -153,6 +153,13 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO(lopresti) if target is relative, currently the gateway fails to identify a storage provider (?)
+	// and just returns a CODE_INTERNAL error on InitiateFileUpload.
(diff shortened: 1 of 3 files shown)


wkloucek:
@ishank011 @glpatcern can we go forward with https://github.com/cs3org/reva/pull/2210? It should also fix this problem

internal/http/services/appprovider/appprovider.go
```diff
@@ -153,6 +153,13 @@ func (s *svc) handleNew(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO(lopresti) if target is relative, currently the gateway fails to identify a storage provider (?)
```
ishank011, line 156: We do support relative references but I don’t know if web can work with those

For now, it’s good enough

ishank011 (approved):

PR #2307, opened 2021-11-26
Accept old and new userid idp format in provider authorizer interceptor

michielbdejong:
Continuation of #2289 -  but with branch on my own fork instead of on the main cs3org repo.


michielbdejong:
Closing as stale.

PR #2310, opened 2021-11-26
fix app provider for editor public links

C0rby:
Fixed opening the app provider in public links with the editor permission.
The app provider failed to open the file in read write mode.

fixes: https://github.com/owncloud/ocis/issues/2803


internal/grpc/services/publicstorageprovider/publicstorageprovider.go
```diff
@@ -108,7 +108,25 @@ func New(m map[string]interface{}, ss *grpc.Server) (rgrpc.Service, error) {
 }
 
 func (s *service) SetArbitraryMetadata(ctx context.Context, req *provider.SetArbitraryMetadataRequest) (*provider.SetArbitraryMetadataResponse, error) {
-	return nil, gstatus.Errorf(codes.Unimplemented, "method not implemented")
+	tkn, relativePath, err := s.unwrap(ctx, req.Ref)
```
ishank011, line 111: Can you call `translatePublicRefToCS3Ref` instead?

C0rby: Good point. I changed that. :+1:

ishank011 (approved):

PR #2314, opened 2021-11-30
OIDC: fallback to "email" if IDP doesn't provide "preferred_username" claim

rhafer:
Some IDPs (e.g. Authelia) don't add the "preferred_username" claim. Fallback to the "email" claim in that case.

See also: https://github.com/owncloud/ocis/issues/2644


wkloucek (approved):
:+1:

labkode (approved):

PR #2315, opened 2021-12-01
Public link propfind

C0rby:
Added a new property "oc:signature-auth" to public link propfinds.
This is a necessary change to be able to support archive downloads in password protected public links.

Together with this PR https://github.com/owncloud/ocis/pull/2831 and some changes in the web client, archive downloads in password protected public links will be possible.


refs (approved):
👍

labkode (approved):

PR #2321, opened 2021-12-03
feat(ocdav meta api): add id to path

fschade:
Added support for fileid to filepath in [ocdav meta api](https://doc.owncloud.com/server/next/developer_manual/webdav_api/meta.html)


butonic:
ther is some related work in https://github.com/cs3org/reva/pull/1604/files

PR #2328, opened 2021-12-06
Rebase master onto edge

refs:
We should catch any attempt from diverging too far from master


refs:
CI is behaving strange because there is no merge commit. Following #2333

refs:
closing in favor of #2333

PR #2329, opened 2021-12-06
 Activate Statcache

kobergj:


changelog/unreleased/activate-statcache.md
```diff
@@ -0,0 +1,6 @@
+Change: activate the statcache
+
+Activates the cache of stat request/responses in the gateway.
+Fixes tests that are failing because of it
```
refs, line 4:
```suggestion

```

internal/grpc/services/gateway/storageprovidercache.go
```diff
@@ -40,6 +40,13 @@ func userKey(ctx context.Context, ref *provider.Reference) string {
 	return u.Id.OpaqueId + "!" + ref.ResourceId.StorageId + "!" + ref.ResourceId.OpaqueId + "!" + ref.Path
 }
 
+// RemoveFromCache removes a reference from the cache
+func RemoveFromCache(cache *ttlcache.Cache, ref *provider.Reference) {
+	// TODO: implement me!
```
refs, line 45:
```suggestion

```

refs (changes requested):

internal/grpc/services/gateway/storageprovidercache.go
```diff
@@ -40,6 +43,20 @@ func userKey(ctx context.Context, ref *provider.Reference) string {
...
 }
 
+// RemoveFromCache removes a reference from the cache
+func RemoveFromCache(cache *ttlcache.Cache, user *userpb.User, res *provider.ResourceId) {
+	remove := user.Id.OpaqueId
+	if res != nil {
+		remove += "!" + res.StorageId + "!" + res.OpaqueId
+	}
+
+	for _, key := range cache.GetKeys() {
```
refs, line 53: O(n)? It might be a niche, but couldn't this be O(1)? 🤔

kobergj: Maybe. But I'm not sure. We need to loop over all keys so we can check for prefixes. Any ideas?

refs (approved):

internal/grpc/services/gateway/storageprovidercache.go
butonic: what he said ;-)

refs: not really, I'd argue about using the OpaqueID as keys so it is a 1:1 match, but I think this is totally out of the scope of this. So you are forced to do this gymnastics around that 😓 because I doubt ttlcache is expressive enough to support queries to it.

kobergj: It isn't unfortunately. Removing keys by opaqueid doesn't work, because I need to have the option to remove all keys of a user, so we anyways loop over all keys.

PR #2309, opened 2021-11-26
Bugfix: Remove early finish for zero byte file uploads

wkloucek:
Bugfix: Remove early finish for zero byte file uploads

We've fixed the upload of zero byte files by removing the
early upload finishing mechanism.

fixes https://github.com/owncloud/ocis/issues/2609

continuation of https://github.com/cs3org/reva/pull/2055

PR #2330, opened 2021-12-06
fix decomposedfs upload

butonic:
The FS.Upload() implementation needs to handle direct uploads that did not initiate a unpload.

fixes ocis spaces registry PR https://github.com/owncloud/ocis/pull/2739

ocis makes direct PUT `/data/spaces/spaceid!spaceid/relative/path/to/file.json`  requests to the metadata storage data provider. No initiate upload call ... we did that to save requests. Might be better to just use TUS with the `creation-with-upload` extension.

For now this allows us to create users with ocis and the accounts service.

### DIFF
--- a/changelog/unreleased/fix-decomposedfs-upload.md
+++ b/changelog/unreleased/fix-decomposedfs-upload.md
@@ -1,0 +1,5 @@
+Bugfix: fix decomposedfs upload
+
+The FS.Upload() implementation needs to handle direct uploads that did not initiate a unpload.
+
+https://github.com/cs3org/reva/pull/2330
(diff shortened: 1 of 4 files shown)


IljaN (approved):

wkloucek:
This is completely against the CS3api spec... It will create hidden complexity and produce bugs like https://github.com/owncloud/ocis/issues/2609. That's why we removed zero-byte-uploads and the functionality you just added again in https://github.com/cs3org/reva/pull/2309.

Can we please revert this and solve it properly?

CC @C0rby

butonic:
I agree the different upload mechanisms currently are confusing:
- we have the InitiateFileUpload call which returns a "simple" or "tus" protocol endpoint
- both protocols take an upload id, but
  - the eos driver uses a path as the upload token
  - the ocis metadata service bypasses all this as it used the "simple" protocol with a path

We should fix the metadata service to make the proper InitiateFileUpload + PUT request. While this will create more requests we can improve that later, similar to the [`creation-with-upload` extension of TUS](https://tus.io/protocols/resumable-upload.html#creation-with-upload).

PR #2336, opened 2021-12-07
handle sending all permissions when creating public links

butonic:
For backwards compatability we now reduce permissions to readonly when a create public link carries all permissions.

Adresses https://github.com/owncloud/ocis/issues/1269#issuecomment-646278608

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
@@ -423,6 +423,13 @@ func (h *Handler) removePublicShare(w http.ResponseWriter, r *http.Request, shar
 }
 
 func ocPublicPermToCs3(permKey int, h *Handler) (*provider.ResourcePermissions, error) {
+
+	if permKey == 31 {
+		// apiSharePublicLink1/createPublicLinkShare.feature:109
+		// Scenario Outline: Trying to create a new public link share of a file with edit permissions only grants read access using the public WebDAV API
+		permKey = 1
+	}
+
 	// TODO refactor this ocPublicPermToRole[permKey] check into a conversions.NewPublicSharePermissions?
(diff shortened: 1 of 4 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -423,6 +423,13 @@ func (h *Handler) removePublicShare(w http.ResponseWriter, r *http.Request, shar
 }
 
 func ocPublicPermToCs3(permKey int, h *Handler) (*provider.ResourcePermissions, error) {
+
+	if permKey == 31 {
```
C0rby, line 427: Could we use the constants from the `conversions` package here instead of "magic numbers"?

C0rby (changes requested):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -423,6 +423,13 @@ func (h *Handler) removePublicShare(w http.ResponseWriter, r *http.Request, shar
...
+		// apiSharePublicLink1/createPublicLinkShare.feature:109
+		// Scenario Outline: Trying to create a new public link share of a file with edit permissions only grants read access using the public WebDAV API
```
refs, line 429: could we instead of having this comment here, link https://github.com/owncloud/ocis/issues/1269 in the changelog? This way we can trace the bug and the fix :D

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
refs: 👍 there is `conversions.PermissionAll` and `conversions.PermissionRead`

C0rby (approved):

PR #2338, opened 2021-12-08
remove unused user agent filter

butonic:
I removed the unused `allowed_user_agents` config option. It isnot used anywhere in the code, but three places had the config option. Dropping it before it spreads further. We can bring it back in a clean way later.


refs (approved):

ishank011:
@butonic this is used in the master branch. Please add back the code to the edge branch as this is a required functionality. Context: https://github.com/cs3org/reva/pull/2212

PR #2340, opened 2021-12-08
Space registry multiple spaces per provider

butonic:
The spaces registry can now use multiple space configurations to allow personal and project spaces on the same provider

### DIFF
--- a/pkg/storage/registry/spaces/spaces.go
+++ b/pkg/storage/registry/spaces/spaces.go
@@ -117,10 +117,12 @@ func (c *config) init() {
 	for _, provider := range c.Providers {
 		for _, space := range provider.Spaces {
 
+			if space.MountPoint == "" {
+				space.MountPoint = "/"
+			}
+
 			// if the path template is not explicitly set use the mount point as path template
-			// do not use the mount point if it is a regex (starting with ^ - TODO add test)
-			if space.PathTemplate == "" && strings.HasPrefix(space.MountPoint, "/") {
-				// TODO err if the mount point is a regex
+			if space.PathTemplate == "" {
 				space.PathTemplate = space.MountPoint
 			}
 
@@ -311,6 +313,8 @@ func (r *registry) findProvidersForResource(ctx context.Context, id string) []*r
 		}
 
 		switch len(spaces) {
+		case 0:
+			// nothing to do, will continue with next provider
 		case 1:
 			space := spaces[0]
 			for spaceType, sc := range provider.Spaces {
@@ -331,7 +335,7 @@ func (r *registry) findProvidersForResource(ctx context.Context, id string) []*r
 					return []*registrypb.ProviderInfo{p}
 				}
 			}
-		case 2:
+		default:
 			// there should not be multiple spaces with the same id per provider
 			appctx.GetLogger(ctx).Error().Err(err).Interface("provider", provider).Interface("spaces", spaces).Msg("multiple spaces returned, ignoring")
 		}
--- a/pkg/storage/registry/spaces/spaces.go
+++ b/pkg/storage/registry/spaces/spaces.go
@@ -476,7 +476,7 @@ func (r *registry) findStorageSpaceOnProvider(ctx context.Context, addr string, 
 	if err != nil {
 		return nil, err
 	}
-	if res.Status.Code != rpc.Code_CODE_OK {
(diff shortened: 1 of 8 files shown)


pkg/storage/registry/spaces/spaces.go
```diff
@@ -52,17 +52,31 @@ func init() {
...
-	Aliases           map[string]string `mapstructure:"aliases"`
-	AllowedUserAgents []string          `mapstructure:"allowed_user_agents"`
-	PathTemplate      string            `mapstructure:"path_template"`
-	template          *template.Template
+type spaceConfig struct {
+	// MountPoint determines where a space is mounted. Can be a regex
+	// It is used to determine which storage provider is responsible when only a path is given in the request
+	MountPoint string `mapstructure:"mount_point"`
+	// PathTemplate is used to build the path of an individual space. Layouts can access {{.Space...}} and {{.CurrentUser...}}
+	PathTemplate string `mapstructure:"path_template"`
```
kobergj, line 60: Just for clarification: Will `MountPoint` always be a prefix of `PathTemplate`?

pkg/storage/registry/spaces/spaces.go
```diff
@@ -98,27 +103,35 @@ func (c *config) init() {
...
-			rule.PathTemplate = rule.MountPath
-		}
+	// cleanup space paths
+	for _, provider := range c.Providers {
+		for _, space := range provider.Spaces {
 
-		// cleanup path template
-		rule.PathTemplate = filepath.Clean(rule.PathTemplate)
+			// if the path template is not explicitly set use the mountpath as path template
+			if space.PathTemplate == "" && strings.HasPrefix(space.MountPoint, "/") {
```
kobergj, line 121: 1) Should we error if both `space.PathTemplate` and `space.MountPoint` are empty strings?

2) Why does `space.MountPoint` need to start with a `"/"`? Isn't that added automatically later?

pkg/storage/registry/spaces/spaces.go
```diff
@@ -172,43 +184,43 @@ type registry struct {
...
+			spacePath := ""
+			var err error
+			if space.SpaceType != "" && spaceType != space.SpaceType {
 				continue
 			}
-			match, err := regexp.MatchString(rule.MountPath, mountPath)
-			if err != nil {
-				continue
+			if space.Owner != nil {
+				spacePath, err = sc.SpacePath(nil, space)
```
kobergj, line 203: Shouldn't we pass `space.Owner` instead `nil` here? We explicitly checked it is not nil?

pkg/storage/registry/spaces/spaces.go
```diff
@@ -269,52 +281,56 @@ func (r *registry) ListProviders(ctx context.Context, filters map[string]string)
...
 			continue
 		}
 
 		if len(spaces) > 0 {
-			space := spaces[0] // there shouldn't be multiple
-			providerPath, err := rule.ProviderPath(currentUser, space)
-			if err != nil {
-				appctx.GetLogger(ctx).Error().Err(err).Interface("rule", rule).Interface("space", space).Msg("failed to execute template, continuing")
-				continue
+			space := spaces[0] // there should not be multiple per provider
```
kobergj, line 313: `"should"`. Do we want to check that there are not multiple so we can error?

kobergj (changes requested):

pkg/storage/registry/spaces/spaces.go
butonic: AFAICT yes. But we cannot just use one config option because the mount_point might be a regex.

pkg/storage/registry/spaces/spaces.go
butonic: 1. No, mount point and path template should default to `/`
2. The check makes sure it is not a regex, starting with `^` ... yes, that only works with a ton of assumptions needs better test coverage to explain how regexes are used

pkg/storage/registry/spaces/spaces.go
butonic: the first parameter is the current user, however for personal spaces the current user may not be the owner of a space. The latter is part of the space struct we pass as the second parameter to `SpacePath`.

pkg/storage/registry/spaces/spaces.go
kobergj: I don't quite understand 2. If it is a regex, why can't it start with `"/"`?

kobergj (approved):

PR #2341, opened 2021-12-08
Use permissions API in decomposedfs

C0rby:
This PR is part of the introduction of the permissions API to the CS3 APIs.
It demonstrates the usage of the permissions API for the use case `ListStorageSpaces`. A user with the permission `list-all-spaces` should be able to see all spaces.

### DIFF
--- a/internal/grpc/services/gateway/gateway.go
+++ b/internal/grpc/services/gateway/gateway.go
@@ -71,6 +71,7 @@ type config struct {
 	EtagCacheTTL        int                               `mapstructure:"etag_cache_ttl"`
 	AllowedUserAgents   map[string][]string               `mapstructure:"allowed_user_agents"` // map[path][]user-agent
 	CreateHomeCacheTTL  int                               `mapstructure:"create_home_cache_ttl"`
+	PermissionsEndpoint string                            `mapstructure:"permissionssvc"`
 }
 
 // sets defaults
--- a/internal/grpc/services/gateway/gateway.go
+++ b/internal/grpc/services/gateway/gateway.go
@@ -57,6 +57,7 @@ type config struct {
 	GroupProviderEndpoint         string `mapstructure:"groupprovidersvc"`
 	DataTxEndpoint                string `mapstructure:"datatx"`
 	DataGatewayEndpoint           string `mapstructure:"datagateway"`
+	PermissionsEndpoint           string `mapstructure:"permissionssvc"`
 	CommitShareToStorageGrant     bool   `mapstructure:"commit_share_to_storage_grant"`
 	CommitShareToStorageRef       bool   `mapstructure:"commit_share_to_storage_ref"`
 	DisableHomeCreationOnLogin    bool   `mapstructure:"disable_home_creation_on_login"`
@@ -71,7 +72,6 @@ type config struct {
 	EtagCacheTTL        int                               `mapstructure:"etag_cache_ttl"`
 	AllowedUserAgents   map[string][]string               `mapstructure:"allowed_user_agents"` // map[path][]user-agent
 	CreateHomeCacheTTL  int                               `mapstructure:"create_home_cache_ttl"`
-	PermissionsEndpoint string                            `mapstructure:"permissionssvc"`
 }
 
 // sets defaults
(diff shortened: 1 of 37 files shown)


refs:
CS3 Api counterpart: https://github.com/cs3org/cs3apis/pull/161

labkode:
@C0rby I've merged the cs3apis. Can we have a "not implemented" implementation for the rest of drivers? Thanks

C0rby:
I had to add a "no-op" permission service for the CI since our real implementation is in the oCIS repo.
If we want to test scenarios which rely on certain permissions we should test that in the oCIS CI anyways so IMO there is no problem here.

C0rby:
@ishank011, could you review this please?

internal/grpc/services/gateway/gateway.go
```diff
@@ -71,6 +71,7 @@ type config struct {
 	EtagCacheTTL        int                               `mapstructure:"etag_cache_ttl"`
 	AllowedUserAgents   map[string][]string               `mapstructure:"allowed_user_agents"` // map[path][]user-agent
 	CreateHomeCacheTTL  int                               `mapstructure:"create_home_cache_ttl"`
+	PermissionsEndpoint string                            `mapstructure:"permissionssvc"`
```
ishank011, line 74: Group this along with the other services' endpoints?

pkg/permission/manager/ocisci/ocisci.go
```diff
@@ -0,0 +1,43 @@
...
+package ocisci
+
+import (
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/pkg/permission"
+	"github.com/cs3org/reva/pkg/permission/manager/registry"
+)
+
+func init() {
+	registry.Register("ocisci", New)
```
ishank011, line 28: Maybe rename this as demo?

pkg/permission/manager/ocisci/ocisci.go
```diff
@@ -0,0 +1,43 @@
...
+	return manager{}, nil
+}
+
+type manager struct {
+}
+
+func (m manager) CheckPermission(permission string, subject string, ref *provider.Reference) bool {
+	// We can currently return false all the time.
+	// Once we beginn testing roles we need to somehow check the roles of the users here
+	return false
```
ishank011, line 42: I'd rather return true by default. You can get rid of the `checkRes.Status.Code == v1beta11.Code_CODE_OK` check in the `ListStorageSpaces` in decomposedfs then as well

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -200,6 +202,28 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
 		return spaces, nil
 	}
 
+	client, err := pool.GetGatewayServiceClient(fs.o.GatewayAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	checkRes, err := client.CheckPermission(ctx, &permissionsv1beta1.CheckPermissionRequest{
```
ishank011, line 210: Can we move this check to `gateway.ListStorageSpaces`? Then you won't need to duplicate it in every driver

pkg/permission/manager/ocisci/ocisci.go
C0rby: Not really. The response contains the status `ok` or `permission denied` depending of if the user has the permission.
In hindsight I think this is bad API design and there should be a dedicated `result` field since the status is more concerning the transport.
But I'd change that afterwards to not delay this PR further.

pkg/storage/utils/decomposedfs/spaces.go
C0rby: I agree that this would be nicer but it's not really possible since we need the permission information in the decomposedfs in line https://github.com/cs3org/reva/pull/2341/files/cc9def031675c4301ca9256cbf6fbfc8555fd71e#diff-d308a0558595437a537e77e29346472e66ab3bc31c213fcd2dd9b5f64c02aa91R384.

Sure this could be passed down via an `Opaque` field but that wouldn't really be secure since everybody could just put that field into the Opaque map.

C0rby:
@ishank011 could you have another look.

ishank011 (approved):

PR #2343, opened 2021-12-08
allow multiple space type filters on decomposedfs

butonic:
The decomposedfs driver now evaluates multiple space type filters when listing storage spaces.


kobergj (approved):

individual-it:
is that a thing that can be tested in unit tests?

PR #2345, opened 2021-12-09
Switch LDAP test to use entryUUID as unique id for groups

rhafer:
The entryUUID attribute is generated by the LDAP server. By using this
me can make sure that a group gets a different IDs with every
test. So that we can avoid setting DELETE_USER_DATA_CMD to delete
the USER data after each test in the long run.

Also fix the user- and group-providers to use the correct attribute
mapping when searching for groups by id (was using CN in some places)


kobergj (approved):
:+1:

PR #2348, opened 2021-12-09
fix-archiver

butonic:
The archiver can now handle the spaces protocol. The ocis testsuite caught this one in https://drone.owncloud.com/owncloud/ocis/8266/21/6

### DIFF
--- a/changelog/unreleased/fix-archiver.md
+++ b/changelog/unreleased/fix-archiver.md
@@ -1,0 +1,5 @@
+Bugfix: make archiver handle spaces protocol
+
+The archiver can now handle the spaces protocol
+
+https://github.com/cs3org/reva/pull/2348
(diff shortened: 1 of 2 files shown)


micbar:
Can we add an api test?

butonic:
@micbar not right now, I am tryding to make the ocis PR green ... I tried fo 30min to write api tists for this. I'll have to mock even an http client and pass that in ... something I can do while waiting for the CI ...

kobergj (approved):

individual-it:
@butonic are the tests in an other PR?
and wouldn't that an easy thing to test with unit tests?

butonic:
I discovered this in the ocis test suite. yes it is easy to cover in reva with a unit test, which I will write during ci runs ... but ... as it turns out ... far more is broken ... need to dig and will try to cover this as I go.

PR #2351, opened 2021-12-10
Fix Statcache issue

kobergj:
Statcache is not working correctly. Some tests fail in local but are green on ci.

Goals:

1) Make tests red on ci
2) Make tests green on ci

Note: The order is important

### DIFF
--- a/internal/grpc/services/gateway/storageprovidercache.go
+++ b/internal/grpc/services/gateway/storageprovidercache.go
@@ -47,17 +47,25 @@ func userKey(ctx context.Context, ref *provider.Reference) string {
 func RemoveFromCache(cache *ttlcache.Cache, user *userpb.User, res *provider.ResourceId) {
 	uid := "uid:" + user.Id.OpaqueId
 	sid := ""
+	oid := ""
 	if res != nil {
-		sid = "sid:" + res.StorageId + "!oid:" + res.OpaqueId
+		sid = "sid:" + res.StorageId
+		oid = "oid:" + res.OpaqueId
 	}
 
-	for _, key := range cache.GetKeys() {
+	keys := cache.GetKeys()
+	for _, key := range keys {
 		if strings.Contains(key, uid) {
 			_ = cache.Remove(key)
 			continue
 		}
 
 		if sid != "" && strings.Contains(key, sid) {
+			_ = cache.Remove(key)
+			continue
+		}
+
+		if oid != "" && strings.Contains(key, oid) {
 			_ = cache.Remove(key)
 			continue
 		}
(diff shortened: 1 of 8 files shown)


kobergj:
@rhafer this should fix the (locally) failing test in ci and on local. Please double check!

internal/grpc/services/gateway/storageprovidercache.go
```diff
@@ -86,6 +101,11 @@ func (c *cachedAPIClient) Stat(ctx context.Context, in *provider.StatRequest, op
 		return resp, nil
 	case key == "":
 		return resp, nil
+	case strings.Contains(key, "sid:a0ca6a90-a365-4782-871e-d44447bbc668"):
```
rhafer, line 104: Shouldn't this ID be defined as a constant in some central place? Having the hardcoded ID here (and elsewhere) makes the code a) harder to understand and b) makes it hard to keep stuff in sync in the code.

rhafer (changes requested):
> this should fix the (locally) failing test in ci and on local. Please double check!

Yeah, this works locally here now was well. Thanks. 

This looks good to me overall. Though I would really like to avoid adding the hardcoded spaceid here. Is there a way we can move that into a constant?

internal/grpc/services/gateway/storageprovidercache.go
kobergj: Good point. The question is: Where would that constant live? For now it is created within the sharestorageprovider, see here: https://github.com/cs3org/reva/blob/edge/internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go#L352

Ofc it is no problem to move it to a constant in the gateway package, but if we want to use the same constant in sharesstorageprovider and gateway we need a common package that both can import...

rhafer: Good question ;-). How about pkg/utils? It's not really nice, but it  should work...

refs: There was another magic string like this I recently found. If we start setting default id's without a formal configuration then things will get confusing real fast. This is the one I refer to:

https://github.com/cs3org/reva/blob/edge/internal/grpc/services/publicstorageprovider/publicstorageprovider.go#L53

kobergj: @refs @rhafer I moved both to utils package and import them from there. Should be easy to add configuration if we see fit. For now if you really need to change the IDs you can overwrite them via build flags

refs: 🙆‍♂️ Thanks! :D

rhafer (approved):

PR #2352, opened 2021-12-10
Log traces only when these are available

refs:
Prevent from logging `traceid=0000000000000000`. Now logs would appear as:

image

when traces are enabled, then the output looks like:

image

as god intended.

### DIFF
--- a/internal/grpc/interceptors/appctx/appctx.go
+++ b/internal/grpc/interceptors/appctx/appctx.go
@@ -81,11 +81,8 @@ func (ss *wrappedServerStream) Context() context.Context {
 }
 
 func getTraceIDFromSpan(span trace.Span) string {
-	traceID := ""
-	if span.SpanContext().TraceID() == [16]byte{} {
-		traceID = ""
-	} else {
-		traceID = span.SpanContext().TraceID().String()
+	if span.SpanContext().TraceID() != [16]byte{} {
+		return span.SpanContext().TraceID().String()
 	}
-	return traceID
+	return ""
 }
--- a/internal/grpc/interceptors/appctx/appctx.go
+++ b/internal/grpc/interceptors/appctx/appctx.go
@@ -37,7 +37,7 @@ func NewUnary(log zerolog.Logger) grpc.UnaryServerInterceptor {
 			ctx, span = rtrace.Provider.Tracer("grpc").Start(ctx, "grpc unary")
 		}
 
-		sub := log.With().Str("traceid", span.SpanContext().TraceID().String()).Logger()
+		sub := log.With().Str("traceid", getTraceIDFromSpan(span)).Logger()
 		ctx = appctx.WithLogger(ctx, &sub)
 		res, err := handler(ctx, req)
 		return res, err
@@ -57,7 +57,7 @@ func NewStream(log zerolog.Logger) grpc.StreamServerInterceptor {
 			ctx, span = rtrace.Provider.Tracer("grpc").Start(ctx, "grpc stream")
 		}
 
-		sub := log.With().Str("traceid", span.SpanContext().TraceID().String()).Logger()
+		sub := log.With().Str("traceid", getTraceIDFromSpan(span)).Logger()
 		ctx = appctx.WithLogger(ctx, &sub)
 
 		wrapped := newWrappedServerStream(ctx, ss)
@@ -78,4 +78,14 @@ type wrappedServerStream struct {
 
 func (ss *wrappedServerStream) Context() context.Context {
 	return ss.newCtx
+}
+
+func getTraceIDFromSpan(span trace.Span) string {
+	traceID := ""
+	if span.SpanContext().TraceID() == [16]byte{} {
+		traceID = ""
+	} else {
+		traceID = span.SpanContext().TraceID().String()
+	}
+	return traceID
 }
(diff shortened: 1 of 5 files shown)


C0rby (approved):

internal/grpc/interceptors/appctx/appctx.go
```diff
@@ -79,3 +79,13 @@ type wrappedServerStream struct {
 func (ss *wrappedServerStream) Context() context.Context {
 	return ss.newCtx
 }
+
+func getTraceIDFromSpan(span trace.Span) string {
+	traceID := ""
+	if span.SpanContext().TraceID() == [16]byte{} {
+		traceID = ""
```
kobergj, line 86: Why do overwrite an empty string with an empty string?

internal/http/interceptors/appctx/appctx.go
```diff
@@ -49,9 +49,19 @@ func handler(log zerolog.Logger, h http.Handler) http.Handler {
...
 		ctx = appctx.WithLogger(ctx, &sub)
 		r = r.WithContext(ctx)
 		h.ServeHTTP(w, r)
 	})
 }
+
+func getTraceIDFromSpan(span trace.Span) string {
+	traceID := ""
+	if span.SpanContext().TraceID() == [16]byte{} {
+		traceID = ""
```
kobergj, line 62: Same here. Why not just do
```
if span.SpanContext().TraceID() != [16]byte{} {
```

internal/http/interceptors/appctx/appctx.go
```diff
@@ -49,9 +49,19 @@ func handler(log zerolog.Logger, h http.Handler) http.Handler {
...
+		sub := log.With().Str("traceid", getTraceIDFromSpan(span)).Logger()
 		ctx = appctx.WithLogger(ctx, &sub)
 		r = r.WithContext(ctx)
 		h.ServeHTTP(w, r)
 	})
 }
+
+func getTraceIDFromSpan(span trace.Span) string {
+	traceID := ""
+	if span.SpanContext().TraceID() == [16]byte{} {
```
kobergj, line 61: Same here. Why not just do
```
if span.SpanContext().TraceID() != [16]byte{} {
```

kobergj (changes requested):

pkg/rgrpc/status/status.go
```diff
@@ -204,5 +204,9 @@ func NewErrorFromCode(code rpc.Code, pkgname string) error {
 // internal function to attach the trace to a context
 func getTrace(ctx context.Context) string {
 	span := trace.SpanFromContext(ctx)
+	if span.SpanContext().TraceID() == [16]byte{} {
```
kobergj: Maybe save `span.SpanContext().TraceID()` in its own var? It is used twice and that would make reading it simpler...

internal/grpc/interceptors/appctx/appctx.go
```diff
@@ -79,3 +79,10 @@ type wrappedServerStream struct {
 func (ss *wrappedServerStream) Context() context.Context {
 	return ss.newCtx
 }
+
+func getTraceIDFromSpan(span trace.Span) string {
+	if span.SpanContext().TraceID() != [16]byte{} {
+		return span.SpanContext().TraceID().String()
```
kobergj, line 85: I'd suggest to do 
```go
if b := span.SpanContext().TraceID(); b != [16]byte{} {
     return b.String()
}
```
(in all three places). But it's not critical.

refs:
@kobergj thanks for the extra pair of eyes! I applied your suggestions, could you have another look? 🚀

kobergj (approved):
Looks awesome now :+1:

labkode:
@refs originally the traceid was not taken from the tracing framework, it was always visible as it was generated from a Reva middeware. With the change to the tracing framework we have now the possibility to configure a tracing target to send the traces remotely. 

However we should split two functionalities:
1) trace id in the log is always useful and should be available independently if the remote tracing collection is enabled or not.
2) send traces to remote collector for trace analysis.

With this in mind I'm not in favour of dropping the traceid rathen to fix the code to make sure it is always there

labkode (changes requested):
Traceid should always be present in the logs independently if trace collector is enabled or not.

refs:
> Traceid should always be present in the logs independently if trace collector is enabled or not.

Hm, I see your argument of always having traces regardless of these are reported or not. I will update the code to reflect that. I completely misjudged the benefits of having the traces always available 👍

ishank011:
@refs I'd propose enabling tracing by default if it has not been explicitly disabled

```diff
diff --git a/cmd/revad/runtime/runtime.go b/cmd/revad/runtime/runtime.go
index 22d55eb3..9783092c 100644
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -311,6 +311,15 @@ func parseCoreConfOrDie(v interface{}) *coreConf {
                os.Exit(1)
        }

+       // tracing defaults to enabled if not explicitly configured
+       if v == nil {
+               c.TracingEnabled = true
+               c.TracingEndpoint = "localhost:6831"
+       } else if _, ok := v.(map[string]interface{})["tracing_enabled"]; !ok {
+               c.TracingEnabled = true
+               c.TracingEndpoint = "localhost:6831"
+       }
+
        return c
 }
```

refs:
That's a way to do it. I'm currently testing what would happen if there is no sidecar agent and we have traces. I have seen it fail trying to connect to the agent's address. Once I got it working I'll update here. But out of the box seems like a good compromise, if it works 👯

refs:
update: the exporter fails, and crashes if missconfigured. I will figure a way to preserve the traceid for the logs regardless of enabled tracing

ishank011:
Superseded by https://github.com/cs3org/reva/pull/2515

PR #2353, opened 2021-12-10
reduce logging output

C0rby:
Reduced log output. Some errors or warnings were logged multiple times or even unnecesarily. 

Now, there are probably a lot of places where logging is missing but these can be added when we touch those spots again.

### DIFF
--- a/changelog/unreleased/reduce-log-output.md
+++ b/changelog/unreleased/reduce-log-output.md
@@ -1,0 +1,6 @@
+Change: Reduce log output
+
+Reduced log output. Some errors or warnings were logged multiple times or even unnecesarily. 
+
+
+https://github.com/cs3org/reva/pull/2353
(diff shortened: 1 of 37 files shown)


C0rby:
Similarly, should both the services and the gateway log stuff or only one of both? If the latter who should log? I'd say the service itself would make sense.

C0rby:
I added error logging to the storage provider. It probably would make sense to have more logging in other places too but I propose to add those when we are working at those places.

refs:
I find this approach better rather than having a wall of text. Adding logging when needed (if needed) is more beneficial 👍

C0rby:
There are more things that can be done but I'm going to leave it at that to not make this PR any longer.

refs (approved):

rhafer (approved):

kobergj (approved):

PR #2367, opened 2021-12-13
[tests-only] Remove DELETE_USER_DATA_CMD from tests

rhafer:
With recent changes in the test configs for user and group providers we
no longer need to manually remove the user data after each testcase.
User and groups will get a new unique id with every test run.


rhafer:
Hm, seems this is slowing the tests down massively (IIUC we ran into an 1hour limit). I guess we need to figure out what is slowing things down so much when we leave the old files behind, before we can merge this.

cc @kobergj @butonic

PR #2372, opened 2021-12-13
[tests-only] Refactor gateway caching

kobergj:
Move all of gateway caching to same place for easier maintenance


internal/grpc/services/gateway/storageprovidercache.go
```diff
@@ -22,30 +22,70 @@ import (
...
 	}
-	u, ok := ctxpkg.ContextGetUser(ctx)
-	if !ok {
-		return ""
+
+	return "uid" + user.Id.OpaqueId + "!sid:" + ref.ResourceId.StorageId + "!oid:" + ref.ResourceId.OpaqueId + "!path:" + ref.Path
+}
+
+// Caches holds all caches used by the gateway
+type Caches struct {
```
refs, line 48: I have a conflict with this type. Could we define `Caches` as: `type Caches [3]*ttlcache.Cache`? or an enum, in case accessing by index loses meaning, and accessing it can be done like:


```go
const (
	stat = iota
	home
	provider
)

type Caches [3]*ttlcache.Cache

// only access
func main() {
	c := Caches{}
	statCache := c[stat]
	homeCache := c[home]
	providerCache := c[provider]
}
```

internal/grpc/services/gateway/authprovider.go
```diff
@@ -75,6 +75,7 @@ func (s *svc) Authenticate(ctx context.Context, req *gateway.AuthenticateRequest
 			Status: res.Status,
 		}, nil
 	case res.Status.Code != rpc.Code_CODE_OK:
+		_ = status.NewErrorFromCode(res.Status.Code, "gateway")
```
C0rby, line 78: I think you can remove this line if you are not going to use the error.

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1437,13 +1447,7 @@ func (s *svc) findProviders(ctx context.Context, ref *provider.Reference) ([]*re
...
 	case ref == nil:
 		return nil, errtypes.BadRequest("missing reference")
-	case ref.ResourceId != nil: // can we use the provider cache?
-		// only the StorageId is used to look up the provider. the opaqueid can only be a share and as such part of a storage
-		if value, exists := s.providerCache.Get(ref.ResourceId.StorageId); exists == nil {
-			if providers, ok := value.([]*registry.ProviderInfo); ok {
-				return providers, nil
-			}
-		}
+	case ref.ResourceId != nil:
```
C0rby, line 1450: Is this case intended to do nothing? If so could you add a comment?

C0rby (dismissed):

internal/grpc/services/gateway/authprovider.go
kobergj: :+1:

internal/grpc/services/gateway/storageprovider.go
kobergj: Exactly. No Action needed. I added a comment

butonic (approved):

PR #2373, opened 2021-12-13
Decomposedfs lookup outermost root

butonic:
When stating a resource by id the decomposedfs will now also return the outermost space root. This allows the gateway to properly reconstruct global paths by combining the space mountpoint and the relative path.
For the owner the root will be the space root, For share recipiens the root will be the outermost shared resource. For now, the root is returned as an opaque property.


butonic:
Closed in favor of making the gateway dumb again by moving the space lookup logic to clients, when the ocdav handler is one of them.

PR #2378, opened 2021-12-15
[tests-only] add .drone.env to CODEOWNERS as it is part of the test files

phil-davis:
`.drone.env` contains the branch and commit id pointing to where the API tests are found in `owncloud/core`

It often needs to be modified by the same people/groups who modify tests. This PR makes that addition to CODEOWNERS.


phil-davis:
@ishank011 @labkode is this OK?
It means we won't have to ping you for reviews when we just update the core commit id for tests (and adjust the expected-failures files)

ishank011 (approved):

PR #2380, opened 2021-12-15
[tests-only]Remove test from expected to fail and bump commit id

SwikritiT:
This PR removes the test scenario from the expected to fail file in accordance with the fixes made by PR owncloud/core#39598 and bumps the commit id as well.

PR #2381, opened 2021-12-15
[tests-only] Remove test from expected to fail and bump commit id

SwikritiT:
This PR removes the test scenario from the expected to fail file in accordance with the fixes made by PR owncloud/core#39598 and bumps the commit id as well.

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,3 +1,3 @@
 # The test runner source for API tests
-CORE_COMMITID=2987544bfd6f6823df68f6ea392ba41c1ea4f07c
+CORE_COMMITID=38ee68e6358443e980ba5e7036cf1fb554443814
 CORE_BRANCH=acceptance-test-changes-waiting-2021-11
--- a/tests/acceptance/expected-failures-on-OCIS-storage.md
+++ b/tests/acceptance/expected-failures-on-OCIS-storage.md
@@ -1284,5 +1284,5 @@ _ocs: api compatibility, return correct status code_
 -   [apiWebdavOperations/downloadFile.feature:169](https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdavOperations/downloadFile.feature#L169)
 -   [apiWebdavOperations/downloadFile.feature:170](https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdavOperations/downloadFile.feature#L170)
 
-### [send PUT requests to another user's webDav endpoints as normal user](https://github.com/owncloud/ocis/issues/2759)
+### [send PUT requests to another user's webDav endpoints as normal user](https://github.com/owncloud/ocis/issues/2893)
 -   [apiAuthWebDav/webDavPUTAuth.feature:40](https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiAuthWebDav/webDavPUTAuth.feature#L40)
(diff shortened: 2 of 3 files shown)


phil-davis (approved):

phil-davis:
https://drone.cernbox.cern.ch/cs3org/reva/4730/8/6
```
  Scenario: send PUT requests to another user's webDav endpoints as normal user                             # /drone/src/tmp/testrunner/tests/acceptance/features/apiAuthWebDav/webDavPUTAuth.feature:40
    When user "Brian" requests these endpoints with "PUT" including body "doesnotmatter" about user "Alice" # OCSContext::userSendsRequestToTheseEndpointsWithBody()
      | endpoint                                       |
      | /remote.php/dav/files/%username%/textfile1.txt |
      | /remote.php/dav/files/%username%/PARENT        |
    Then the HTTP status code of responses on all endpoints should be "403"                                 # FeatureContext::theHTTPStatusCodeOfResponsesOnAllEndpointsShouldBe()
      Responses did not return expected http status code
      Failed asserting that 409 is identical to 403.
    When user "Brian" requests these endpoints with "PUT" including body "doesnotmatter" about user "Alice" # OCSContext::userSendsRequestToTheseEndpointsWithBody()
      | endpoint                                           |
      | /remote.php/dav/files/%username%/PARENT/parent.txt |
    Then the HTTP status code of responses on all endpoints should be "403"                                 # FeatureContext::theHTTPStatusCodeOfResponsesOnAllEndpointsShouldBe()
```

That is strange - PR #2380 to `master` branch has passed. But in `edge` branch it returns 409 even for the file and folder in the root.

On oCIS it returned 403 - all was good.

Needs investigation.

wkloucek:
> On oCIS it returned 403 - all was good.

oCIS `master` still runs on REVA `master`, therefore this looks like a regression on `edge`.

I think there is no way around having two different expected failure files for REVA `master` / `edge`

phil-davis:
> I think there is no way around having two different expected failure files for REVA master / edge

The same expected-failures file will have different contents in `master` and `edge` branches.

@wkloucek It would be nice to understand where this happened. It is not a problem on oCIS, so I guess that the difference is something that is waiting to be pulled from `edge` into oCIS?

wkloucek:
> > I think there is no way around having two different expected failure files for REVA master / edge
> 
> The same expected-failures file will have different contents in `master` and `edge` branches.
> 
> @wkloucek It would be nice to understand where this happened. It is not a problem on oCIS, so I guess that the difference is something that is waiting to be pulled from `edge` into oCIS?

@butonic @kobergj @aduffeck might know, since they are working on `edge`

kobergj:
@phil-davis we are still working on including `edge` branch into ocis. As of now it is still running on `master`.

`edge`  contains the spaces registry, which is a huge change compared to `master`. It could be that this somehow leads to a different statuscode returned. If it is not critical I would prefer having different content in the expected failure file on `edge` and `master` for now.

phil-davis:
> If it is not critical I would prefer having different content in the expected failure file on edge and master for now.

OK, then should I raise an issue about this "regression/difference" and, in expected-failures, link to the issue?

That would give the issue visibility.

But where to raise the issue? This is not a problem in `cs3org/reva` `master` so it seems a bit "unfair" to raise the issue there. And it is not an issue in `owncloud/ocis`

kobergj:
I see the problem. But we are working with highest prio on switching to `edge` branch for ocis. After this is completed it will be an issue for `owncloud/ocis` so we can raise the issue there. (Even if it is not 100% correct I would have no problem with having the issue raised before the switch is completed)

wkloucek:
> But where to raise the issue?

We should still do it in oCIS. Basically there are three options:
- issue is on expected failures on `edge` and `master`
- issue is on expected failures on `edge`
- issue is on expected failures on `master`

Depending on what branch oCIS uses these expected failures would also apply to oCIS.

So ideally we would track in the ticket where it does apply. But that's nothing for you and your team. I would say just continue creating tickets as of now. Developers can comment where it does apply

phil-davis (approved):

phil-davis:
I raised issue https://github.com/owncloud/ocis/issues/2893 and linked it in expected-failures.

PR #2394, opened 2021-12-21
Make gateway dumb again

kobergj:
Remove the logic that takes care of converting absolute path requests to relative ones. Gateway now expects to get relative request always


kobergj:
Closed in favor of this https://github.com/cs3org/reva/pull/2437

PR #2395, opened 2021-12-21
Fix building on FreeBSD

wkloucek:
rebased version of https://github.com/cs3org/reva/pull/1499, closes #1499


ishank011:
@wkloucek the issue is that with the go-prompt update, taking user input doesn't work, at least on mac systems. https://github.com/cs3org/reva/blob/master/cmd/reva/login.go#L99

PR #2419, opened 2022-01-06
Spaces fixes

butonic:
The decomposedfs now also checks the permissons on the root node of a space. This makes spaces of type 'project' show up for user that have received a share for the space.


C0rby (approved):
Just two small nitpicks. :+1:

PR #2433, opened 2022-01-11
fix shares provider filter

butonic:
The shares storage provider now correctly filters space types

related https://github.com/owncloud/ocis/pull/2946


kobergj (approved):
Awesome fix! I would have done it exactly the same!

PR #2436, opened 2022-01-12
Adjust "groupfilter" to be able to search by member name

rhafer:
Previously the input for the LDAP Groupfilter to lookup all groups a
specific user is member of was the userpb.UserId part of the User
object. I.e. it assumed we could run a single LDAP query to get all
groups a user is member of by specifying the userid. However most
LDAP Servers store the GroupMembership by either username (e.g. in
memberUID Attribute) or by the user's DN (e.g. in member/uniqueMember).

The GetUserGroups method was already updated recently to do a two-staged
lookup (first lookup the user's name by Id then search the Groups by
username). This change just removes the userpb.UserId template processing
from the GroupFilter and replaces it with a single string (the
username) to get rid of the annoying `{{.}}` template values in the
config.

In the future we should add a config switch to also allow lookups by
member DN.


refs (approved):

butonic (approved):
There really was no reason to use an interface in the signature, anymore. We can safely go with a plain string. While we could introduce a new config var and try to stay backwards compatible, I prefer to break now as otherwise misconfiguration is more error prone.

PR #2437, opened 2022-01-13
Make gateway dumb again

kobergj:
This PR will remove all logic from gateway except forwarding requests to other services


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -573,37 +570,48 @@ func (s *svc) InitiateFileUpload(ctx context.Context, req *provider.InitiateFile
...
+	req.ResourceId = ref.ResourceId
+	return c.GetPath(ctx, req)
 
-	res, err := c.GetPath(ctx, req)
-	if err != nil {
-		return &provider.GetPathResponse{
-			Status: status.NewStatusFromErrType(ctx, "gateway could not call GetPath", err),
-		}, nil
-	}
+	/*
```
C0rby, line 583: You can delete the commented out code.

internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -458,7 +458,7 @@ func (p *Handler) getResourceInfos(ctx context.Context, w http.ResponseWriter, r
 					// add path to resource
 					res.Infos[i].Path = filepath.Join(info.Path, res.Infos[i].Path)
 					if spacesPropfind {
-						res.Infos[i].Path = utils.MakeRelativePath(res.Infos[i].Path)
+						res.Infos[i].Path = utils.MakeRelativePath(filepath.Join(info.Path, res.Infos[i].Path))
```
C0rby, line 461: Isn't this reintroducing a bug?
The `Join` is already happening in line 459.

C0rby (changes requested):

internal/http/services/owncloud/ocdav/propfind/propfind.go
kobergj: Good catch :+1: Removed it.

butonic (changes requested):
filepath -> path, then we can merge

butonic (approved):
❤️

C0rby (approved):

PR #2438, opened 2022-01-13
CreateStorageSpace allows already existing spaces.

dragotin:
The response contains all neccessary data as if it was created
newly.


dragotin:
Not solving...

PR #2443, opened 2022-01-14
fix app provider on public links

wkloucek:


wkloucek:
changes added to https://github.com/cs3org/reva/pull/2625

PR #2444, opened 2022-01-14
Implementation of CS3 Lock API for EOS storage driver

gmgigi96:


glpatcern (changes requested):
A few changes here and there, let's discuss them later

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -537,22 +543,248 @@ func (fs *eosfs) UnsetArbitraryMetadata(ctx context.Context, ref *provider.Refer
...
+	}
+	if l1.AppName != "" || l2.AppName != "" {
+		same = l1.AppName == l2.AppName
+	}
+	return same
 }
 
 // Unlock removes an existing lock from the given reference
 func (fs *eosfs) Unlock(ctx context.Context, ref *provider.Reference) error {
-	return errtypes.NotSupported("unimplemented")
```
butonic, line 555: AFAICT we should change the signature to 
```
func (fs *eosfs) Unlock(ctx context.Context, ref *provider.Reference, lock *provider.Lock) error {
```
as we have to send a matching lockid to unlock a resource

gmgigi96: Thanks, that's true. I will change the code

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -537,22 +543,257 @@ func (fs *eosfs) UnsetArbitraryMetadata(ctx context.Context, ref *provider.Refer
...
+		Type: UserAttr,
+		Key:  LockKeyAttr,
+		Val:  encodedLock,
+	}
+
+	return fs.c.SetAttr(ctx, auth, attr, false, false, path)
+}
+
+func sameHolder(l1, l2 *provider.Lock) bool {
+	same := true
```
glpatcern, line 745:
```suggestion
	same := false
```

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -537,22 +543,257 @@ func (fs *eosfs) UnsetArbitraryMetadata(ctx context.Context, ref *provider.Refer
...
+		}
+	}
+
+	user, err := getUser(ctx)
+	if err != nil {
+		return errors.Wrap(err, "eosfs: error getting user")
+	}
+
+	// check if the user hold the old lock
+	if oldLock.User != nil && !utils.UserEqual(oldLock.User, user.Id) {
```
glpatcern, line 699: Given that you have the `sameHolder()` check just below, this is redundant actually

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1843,10 +2084,10 @@ func (fs *eosfs) convert(ctx context.Context, eosFileInfo *eosclient.FileInfo) (
 		}
 	}
 
-	// filter 'sys' attrs
+	// filter 'sys' attrs and key attr
```
glpatcern, line 2087:
```suggestion
	// filter 'sys' attrs and the reserved lock key attr
```

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -487,7 +493,7 @@ func (fs *eosfs) SetArbitraryMetadata(ctx context.Context, ref *provider.Referen
```
glpatcern, line 493: We need to add a check to reject `k == LockKeyAttr`

glpatcern (changes requested):
Looks good, a couple of minor changes follow...

pkg/storage/utils/eosfs/eosfs.go
gmgigi96: Why false? Because the user and appname in the lock are optional, and must be checked if not empty, if they are either empty or nil, they have the same "holder" (i.e. no one)

glpatcern: ah right, forget that then! Let me double check that we enforce somewhere that either the user or the appname are set on `SetLock`...

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -537,22 +547,300 @@ func (fs *eosfs) UnsetArbitraryMetadata(ctx context.Context, ref *provider.Refer
...
+	if err != nil {
+		return errors.Wrap(err, "eosfs: cannot check if user has write access on resource")
+	}
+	if !has {
+		return errtypes.BadRequest("user has not write access on resource")
+	}
+
+	// the user in the lock could differ from the user in the context
+	// in that case, also the user in the lock MUST have the write permission
+	if !utils.UserEqual(user.Id, l.User) {
```
glpatcern, line 626:
```suggestion
	if l.User != nil && !utils.UserEqual(user.Id, l.User) {
```

glpatcern (approved):
We tested this patch with WOPI (cs3org/wopiserver#51) and it all works. More tests are needed on the WOPI side but ~~this PR can be merged now~~.

Edit: following recent discussions, we're going to mimic more closely the behavior of the decomposedfs implementation, in particular in terms of checking for expired locks, and we'll ease the interoperability with EOS' `flock` support.

glpatcern:
Adding a reference to https://github.com/cs3org/reva/pull/2582 for the lock implementation: we may want to look at it in the future to address the known race conditions.

glpatcern:
Going forward, the eos driver implementation will be based on 3 xattrs:

`sys.reva.lock.payload` = a base64-encoded JSON with the lock payload but excluding `type` and `expiration`
`sys.reva.lock.expiration` = the Unix time when the lock expires
`sys.reva.lock.type` = an integer representing the lock type according to https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.LockType

The assumption is that EOS will check `sys.reva.lock.expiration` and `sys.reva.lock.type` when handling `flock` FUSE operations.

We're going to target the lazy deletion of expired locks similarly to the decomposedFS driver, but for the time being we don't address the race conditions due to rm-after-check or set-after-check cases (to be done using some EOS primitives later on). Also, for now we focus on `WRITE` locks only in order to support the WOPI use cases.

glpatcern (approved):
We have tested this now and it all works fine. Some caveats to be addressed in future PRs:
* Race conditions exist in some cases, and once EOS provides an endpoint to `flock` a file they will be fixed.
* Attributes are currently stored on the version folder, but this can be reverted once the propagation of xattrs across versions is implemented in EOS (work in progress). This affects _all_ xattrs, not just the locks.

Given that, the PR can be merged.

Once we reconcile with the edge branch, we could factorize out some common logic out of the `decomposedfs` and `eosfs` storage providers.

labkode (approved):

PR #2455, opened 2022-01-19
Modifications in the localfs storage driver so that it can use a dCache nfs4.1 mount as file system

rontrompert:
Modifications:
pkg/storage/utils/localfs/localfs.go and pkg/storage/fs/localhome have been modified so that DataDirectory and Shadow are now configurable. 
pkg/storage/utils/localfs/uploads.go has been modified so that the uploaded file can be moved to a directory on a different file system. os.Rename cannot do that.

Configuration:
In the appropriate toml file "data_directory" and "shadow" can be defined in the [grpc.services.storageprovider.drivers.localhome] and [http.services.dataprovider.drivers.localhome] section.

example:
[grpc.services.storageprovider.drivers.localhome]
data_directory = "/dCache/data"
shadow = "/dCache/.shadow"

[http.services.dataprovider.drivers.localhome]
data_directory = "/dCache/data"
shadow = "/dCache/.shadow"

with a root as /var/tmp/reva, create a directory /var/tmp/reva/dCache and mount dCache on this directory. data_diretcory and shadow should be both on the dCache file system to make deleting and restoring files fast. The uploads diretcory should be on the local file system of the node since files are appended there and dCache cannot handle that.


pkg/storage/utils/localfs/upload.go
```diff
@@ -329,10 +329,24 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) error {
 		}
 	}
 
-	err := os.Rename(upload.binPath, np)
+	//err := os.Rename(upload.binPath, np)
```
kobergj, line 332: Could you remove the commented code?

pkg/storage/utils/localfs/upload.go
```diff
@@ -329,10 +329,24 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) error {
...
+	outputFile, err := os.Create(np)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(outputFile, inputFile)
+	if err != nil {
+		return err
+	}
+	inputFile.Close()
+	outputFile.Close()
```
kobergj, line 349: Can we `defer` the `.Close()` calls? Otherwise if `os.Create(np)` fails the `inputFile` will not be closed

kobergj (changes requested):
@rontrompert  Thanks for contributing! Would you mind adding a changelog item to this PR? We need it for releases and the ci-pipeline will not run without it.

Note that this could have an performance impact with large files (copy vs rename). But I am fine tackling this if it becomes a problem.

pkg/storage/utils/localfs/upload.go
rontrompert: I will remove it.

kobergj (commented):
Looks good now :+1: 

CI is still red, I guess because there is a new line between `Enhancement` and `Modifications in the localfs storage driver so that it can use a dCache nfs4.1 mount as file system`. Linter seems to want them on one line

@rontrompert could you fix that? Then we can merge...

pkg/storage/utils/localfs/upload.go
```diff
@@ -329,10 +329,23 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) error {
...
+	//and Uploads directory to be on diferent file systems.
+	inputFile, err := os.Open(upload.binPath)
 	if err != nil {
 		return err
 	}
+	outputFile, err := os.Create(np)
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(outputFile, inputFile)
```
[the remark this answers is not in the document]
butonic: why don't you spread multiple storage spaces over multiple nfs shares? then all uploads can use the atomic rename.

PR #2459, opened 2022-01-20
prevent purging of enabled spaces

C0rby:
Without these changes it is possible to purge a space with one request.


kobergj (commented):
Looks good. But one would still be able to delete and purge a drive within the same call, right?

If so, is that we want @micbar

kobergj:
Skip my comment. Stupid me was remembering wrong...

kobergj (approved):

PR #2482, opened 2022-01-27
[tests-only] [full-ci] Add end-of-line to expected-failures files

phil-davis:
`apiWebdavProperties1/createFileFolderWhenSharesExist.feature:60` is actually passing, but is not always reported as an "unexpected pass" because it is at the very end of the expected-failures files, and there is no newline character at the end.

For now, do not put a scenario reference on the last line. Tomorrow, I will think about making the bash script smarter in core.

Various discussion of this "last line" problem: https://stackoverflow.com/questions/12916352/shell-script-read-missing-last-line


phil-davis:
@ishank011 this makes CI pass. `createFileFolderWhenSharesExist.feature:60` does always pass, and should never have been in expected-failures.

wkloucek (approved):

PR #2488, opened 2022-01-28
Cephfs: support keyrings with IDs

thmour:
Support keyrings with IDs. We can't expose an admin keyring to a webserver, only keyrings that edit a specified subpath in CephFS.


thmour:
@ishank011 we need this for the hpc use-case

ishank011 (dismissed):

ishank011 (approved):

PR #2494, opened 2022-01-31
eos: Use sys ACLs for file permissions

gmgigi96:


ishank011 (commented):
The change also needs to be reflected when reading the ACLs

gmgigi96:
> The change also needs to be reflected when reading the ACLs

@ishank011 thanks for the review. I did it. Let me know if it's ok now :)

ishank011:
Will merge once we update our MGMs

ishank011 (approved):

PR #2502, opened 2022-02-03
use registry

wkloucek:


wkloucek:
closed in favor of another approach like eg. https://github.com/cs3org/reva/pull/2665

PR #2506, opened 2022-02-04
fix propfind listing for files

C0rby:
When doing a propfind for a file the result contained the files twice.
See  https://github.com/owncloud/ocis/issues/3066


kobergj (approved):
Just remarks, not critical

Issue #2480, opened 2022-01-27
Overwrite must be forbidden on public links with drop zone


labkode:
If two people upload the same file they will overwrite each other.
The expected behaviour is to to make each upload unique.
In CERNBox we prepend a random UUID.

<img width="1440" alt="Screenshot 2022-01-27 at 14 35 18" src="https://user-images.githubusercontent.com/5175912/151369707-e8fd5304-c5d5-4d20-89c1-01fa61317835.png">

pmaier1:
I'd like to advocate against using UUIDs here as they appear very technical and regular users might feel uncomfortable. Instead I'd propose to *append* a number in parentheses "`(2)`". This is the behavior we have in the classic ownCloud. /cc @tbsbdr

ishank011:
I think appending the timestamp might be even better? That'll help the user differentiate between the various files

pmaier1:
> I think appending the timestamp might be even better? That'll help the user differentiate between the various files

Agree, that could be valuable as well. Dropbox behaves like this, for example.

labkode:
@pmaier1 make sense!

phil-davis:
What is happening here? We are happy to adjust tests. There has been plenty of discussion in PR #2510 - we just need a decision on the naming format to move forward.

PR #2510, opened 2022-02-07
Append a timestamp to filename when a collision is detected during upload

gmgigi96:
Closes https://github.com/cs3org/reva/issues/2480


labkode:
@gmgigi96 this only need to trigger inside a public link with "Uploader" role only, I think the current implementation applies to all writes. @ishank011 please confirm as well my assumption.

pkg/utils/utils.go
```diff
@@ -352,3 +352,11 @@ func GetViewMode(viewMode string) gateway.OpenInAppRequest_ViewMode {
 		return gateway.OpenInAppRequest_VIEW_MODE_INVALID
 	}
 }
+
+// AppendNowToPath appends the now timestamp just before the extension (if present)
+func AppendNowToPath(p string) string {
+	now := time.Now().Format("2006-01-02 15:04:05")
+	ext := path.Ext(p)
+	baseFile := strings.TrimSuffix(p, ext)
+	return fmt.Sprintf("%s (%s)%s", baseFile, now, ext)
```
glpatcern, line 361: To ease use without escaping chars, I'd just do time in ISO format `2006-01-02T15:04:05` and the format string like `%s.%s%s`.

phil-davis:
Note: https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiSharePublicLink2/uploadToPublicLinkShare.feature#L24
```
  Scenario: Uploading same file to a public upload-only share multiple times via new API
    # The new API does the autorename automatically in upload-only folders
    Given user "Alice" has created a public link share with settings
      | path        | FOLDER |
      | permissions | create |
    When the public uploads file "test.txt" with content "test" using the new public WebDAV API
    And the public uploads file "test.txt" with content "test2" using the new public WebDAV API
    Then the HTTP status code should be "201"
    And the following headers should match these regular expressions
      | ETag | /^"[a-f0-9:\.]{1,32}"$/ |
    And the content of file "/FOLDER/test.txt" for user "Alice" should be "test"
    And the content of file "/FOLDER/test (2).txt" for user "Alice" should be "test2"
```

And https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiSharePublicLink2/uploadToPublicLinkShare.feature#L268 has a similar scenario.

Those scenarios define and explicit file-naming system that is used on oC10 - `test (2).txt`

The requirement here in reva and oCIS is different. I will adjust the core API  test suite so that it covers the different requirement/expectation for reva-oCIS. The file name format looks like it will be:

`test (2022-01-29 15:04:05).txt`

with the date in the ISO format YYYY-MM-DD.

So the test can check that the file name matches that pattern.

Will there be any date-time localization possible? US-style MM-DD-YYYY appearing? Or are we making the format fixed as above?
("asking for a friend" to double-check that the requirement is "well defined")

pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -207,6 +208,8 @@ func (c *Client) executeXRDCopy(ctx context.Context, cmdArgs []string) (string,
 				err = nil
 			case int(syscall.ENOENT):
 				err = errtypes.NotFound(errBuf.String())
+			case 54:
```
glpatcern, line 211: This is all good but unfortunately xrootd may return `54` for several errors as far as I know (including on conditions that could be retried). We could add a log that includes the full `errBuf` to start with...

glpatcern (commented):
A couple of changes, one being more an observation/request for discussion ;-)

pkg/utils/utils.go
glpatcern: Of course this folds with @phil-davis ' comment...

gmgigi96:
@labkode @phil-davis indeed here I'm changing the file name for all the uploads, I will fix it considering the "Uploader" role only.

@phil-davis for the format I think we could agree on a fixed one, to keep things as simple as possible, and maybe the one proposed by @glpatcern is a good format, so `file.2006-01-02T15:04:05.txt`. What do you think?

phil-davis:
> @phil-davis for the format I think we could agree on a fixed one, to keep things as simple as possible, and maybe the one proposed by @glpatcern is a good format, so file.2006-01-02T15:04:05.txt. What do you think?

I am not so sure about ending up with 2 "." dots in the filename plus extension. That often confuses "ordinary users" like my mum. But the space character is also annoying to me (for nerds using the command line, or scripts it means that the file name has to be quoted or escaped or something special done with it).

 @micbar @butonic @dragotin @pmaier1or anyone else. Comments on the "timestamp" format please.

butonic:
um ... `:` is a reserved character on windows and cannot be used in directory or filensames ...

phil-davis:
> windows explorer appends (n+1) when a file already exists / a conflict occurs

yes, mum already gets confused. So I'm not sure that we will be able to "unconfuse" her when I give her an oCIS-reva-based cloud storage anyway ;)

gmgigi96:
@butonic maybe the UNIX time is confusing not only for mum, what about 2018_02_03_07_30_00?

butonic:
> @butonic maybe the UNIX time is confusing not only for mum, what about 2018_02_03_07_30_00?

I don't think having the date in the filename adds any value:
- the mtime of the file already contains the time when the conflict occured
- the longer the appended string the earlier will we run into max filename length issues

I recommend using ` (n+1)` ...

phil-davis:
> I recommend using `(n+1)`

Note: see also the discussion in and around issue comment https://github.com/cs3org/reva/issues/2480#issuecomment-1024087521

"somebody"/"somebodies" will need to come up with an agreed requirement.

pmaier1:
I'm fine with both, `(n+1)` and using the timestamp. I think timestamp has slightly better usability. See Dropbox example screenshot.
Screenshot from 2022-02-07 12-52-48

butonic:
lol ... so they use `.` to separate the time components

ishank011:
Move the checks to the ocdav service, for the PUT and POST handlers

ishank011 (commented):
Just check if stat returned ok here https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocdav/put.go#L147. If it did, keep updating the appended timestamp to the path you pass to InitiateFileUpload

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -149,6 +151,24 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+		user, ok := ctxpkg.ContextGetUser(ctx)
+		if !ok {
+			log.Error().Msg("error getting user from context")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if hasUploaderRole(user) {
+			ref.Path = utils.AppendNowToPath(ref.Path)
+			log.Debug().Msgf("user %s has uploader role, changed file destination to %s", user.Username, ref.Path)
```
butonic, line 168: please never use Msgf with structured loggers. Instead use the proper key value calls
```suggestion
			log.Debug().Str("user", user.Username).Interface("path", ref.Path).Msg("user has uploader role, changed file name")
```

internal/http/services/owncloud/ocdav/tus.go
```diff
@@ -138,6 +139,24 @@ func (s *svc) handleTusPost(ctx context.Context, w http.ResponseWriter, r *http.
...
+		user, ok := ctxpkg.ContextGetUser(ctx)
+		if !ok {
+			log.Error().Msg("error getting user from context")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if hasUploaderRole(user) {
+			ref.Path = utils.AppendNowToPath(ref.Path)
+			log.Debug().Msgf("user %s has uploader role, changed file destination to %s", user.Username, ref.Path)
```
butonic, line 156:
```suggestion
			log.Debug().Str("user", user.Username).Str("path", ref.Path).Msg("user has uploader role, changed file name")
```

butonic:
This actually violates the PUT semantics, because we are not creating the resource requested in PUT. It cannot be retrieved with a subsequent GET request. IMO this should be fixed on the client side.

gmgigi96:
> This actually violates the PUT semantics, because we are not creating the resource requested in PUT. It cannot be retrieved with a subsequent GET request. IMO this should be fixed on the client side.

This is true, but here we are only modifying the behavior in a drop zone (where an user has only `uploader` pemission), so it makes no sense do a GET request as it will fail anyway as the user does not have the read permission.

> What happens when a file with the appended timestamp already exists, because two uploaders upload in the same second?

Indeed that's a problem, one of the twos files will be overridden.

phil-davis:
> Indeed that's a problem, one of the twos files will be overridden.

And we have this kind of problem in oC10 trashbin and versions. Those use 1-second timestamps and there is "trouble" if more than one action happens per second.

And anyway, someone might have uploaded an actual files called `file.txt` and `file - 2022-02-14 16:30:40.txt` a week ago, just to be an annoying person. And at exactly 2022-02-14 16:30:40 someone uploads `file.txt` again. To be sure, there needs to be a code path that, if the target filename with date-time-stamp already exists, then it has a system to generate another unique name - add `(2)` or `(3)`... to the end after the timestamp?

tbsbdr:
I also favour the `file (1).txt` approach, mainly because of 3 reasons:
1. **Feasible:** Robust against same-seconds-problem
2. **Known:** My mom has seen the parenthesis-schema already and has a clue why this happened
3. **Simple:** Cognitive costs for reading the filename are comparatively low, because its short (compared to our other approaches)

Personally I don't like parenthesis and spaces in names, but I can handle it - no dealbreaker.

Any objections against the (n+1) approach?

PR #2512, opened 2022-02-07
Consolidate all metadata Get's and Set's to central functions.

dragotin:
In the decomposed FS, access to xattr was spread all over. This
patch consolidates that to use either the Node.SetMetadata() or
xattrs.Set(). This allows us to hook in indexing for example.

This does not yet handle Remove, List and others. Subject of another patch.

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -110,14 +110,13 @@ func New(id, parentID, name string, blobsize int64, blobID string, owner *userpb
 func (n *Node) ChangeOwner(new *userpb.UserId) (err error) {
 	nodePath := n.InternalPath()
 	n.owner = new
-	if err = xattr.Set(nodePath, xattrs.OwnerIDAttr, []byte(new.OpaqueId)); err != nil {
-		return errors.Wrap(err, "Decomposedfs: could not reset owner id attribute")
-	}
-	if err = xattr.Set(nodePath, xattrs.OwnerIDPAttr, []byte(new.Idp)); err != nil {
-		return errors.Wrap(err, "Decomposedfs: could not reset owner idp attribute")
-	}
-	if err = xattr.Set(nodePath, xattrs.OwnerTypeAttr, []byte(utils.UserTypeToString(new.Type))); err != nil {
-		return errors.Wrap(err, "Decomposedfs: could not reset owner idp attribute")
+
+	var attribs = map[string]string{xattrs.OwnerIDAttr: new.OpaqueId,
+		xattrs.OwnerIDPAttr:  new.Idp,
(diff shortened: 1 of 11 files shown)


pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -127,3 +128,55 @@ func CopyMetadata(s, t string, filter func(attributeName string) bool) error {
 	}
 	return nil
 }
+
+// The primitive function to set the extended attribute
```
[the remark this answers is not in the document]
dragotin: @aduffeck FYI

dragotin:
Get, Set and All also give us a central point to deal with locking, indexing, plausability checks and logging.

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -127,47 +126,42 @@ func (n *Node) ChangeOwner(new *userpb.UserId) (err error) {
...
+}
+
+// WriteAllNodeMetadata writes the Node metadata to disk
+func (n *Node) WriteAllNodeMetadata(owner *userpb.UserId) (err error) {
+	attribs := make(map[string]string)
+
+	attribs[xattrs.ParentidAttr] = n.ParentID
+	attribs[xattrs.NameAttr] = n.Name
+	attribs[xattrs.BlobIDAttr] = n.BlobID
+	attribs[xattrs.BlobsizeAttr] = fmt.Sprintf("%d", n.Blobsize)
```
C0rby, line 151:
```suggestion
	attribs[xattrs.BlobsizeAttr] = strconv.FormatInt(n.Blobsize, 10)
```

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -127,3 +128,57 @@ func CopyMetadata(s, t string, filter func(attributeName string) bool) error {
...
+		name := attrNames[i]
+		val, err := xattr.Get(filePath, name)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to read extended attrib")
+		}
+		attribs[name] = string(val)
+	}
+
+	return attribs, nil
+}
```
C0rby, line 184:
```suggestion
func All(filePath string) (map[string]string, error) {
	attrNames, err := xattr.List(filePath)
	if err != nil {
		return nil, errors.Wrap(err, "xattrs: Can not list extended attributes")
	}

       attribs := make(map[string]string, len(attrNames))
	for _, name := range attrNames {
		val, err := xattr.Get(filePath, name)
		if err != nil {
			return nil, errors.Wrap(err, "Failed to read extended attrib")
		}
		attribs[name] = string(val)
	}

	return attribs, nil
}
```

C0rby (changes requested):

C0rby (approved):

PR #2513, opened 2022-02-07
Format adler checksum header as expected by the sync client

ishank011:


ishank011:
@phil-davis The tests fail with
```
Expected: webDav checksum should be ... ADLER32:1048035a but got ... Adler32:1048035a
```

But the desktop client expects the header to be `Adler32` (https://github.com/owncloud/client/blob/15fc9d017fcdcc4cc95728c16a2dd171d0395b85/src/common/checksums.h#L38-L42)

And reports the error `The checksum header contained an unknown checksum type 'ADLER32'`

TheOneRing:
There was an issue in desktop clients < 2.7.6 fixed in https://github.com/owncloud/client/pull/8371 .
The issue was first discovered in https://github.com/owncloud/client/issues/7999 .

butonic:
@ishank011 I think we should stick to the oc10 Uppercase for all hashing algorithms. The desktop client 2.7.6 is case insensitive.

ishank011:
@butonic We won't be able to support older clients with ocis then. @SamuAlfageme I'm not sure what the upgrade timeline looks like from our end?

If oc10 always sends the header in uppercase, how do clients older than 2.7.6 work with it? I guess you don't run tests with Adler?

phil-davis:
We can easily change the acceptance test expectations. If all known current clients (desktop, Android, iOS etc) that do any processing of `ADLER32`/`Adler32` can handle any upper/lower case combinations, and if there are some (or one) older clients version(s) that really look for `Adler32` then a change from `ADLER32` to `Adler32` might be good/useful.

IMO, we should do the same in oC10 so that all implementations respond with the same `Adler32` - and we will need to check various client versions to make sure that nothing breaks for a reasonable range of existing clients.

labkode:
> We can easily change the acceptance test expectations. If all known current clients (desktop, Android, iOS etc) that do any processing of `ADLER32`/`Adler32` can handle any upper/lower case combinations, and if there are some (or one) older clients version(s) that really look for `Adler32` then a change from `ADLER32` to `Adler32` might be good/useful.
> 
> IMO, we should do the same in oC10 so that all implementations respond with the same `Adler32` - and we will need to check various client versions to make sure that nothing breaks for a reasonable range of existing clients.

This is the key point.  Clients process this attribute as case insensitive. Artificial tests (the ones failing) do not honour this and change the semantics to case sensitive, meaning that they are not testing anything useful. 

Conclusion: the tests needs to be adapted to be case insensitive to ensure client behaviours are properly taken into account.

@phil-davis can we adapt the tests please to be case insensitive please?

PR #2514, opened 2022-02-07
Reuse ocs role objects in other drivers

ishank011:


gmgigi96:
This should fix https://github.com/cs3org/reva/issues/2521

pkg/auth/scope/scope.go
```diff
@@ -56,5 +56,5 @@ func VerifyScope(ctx context.Context, scopeMap map[string]*authpb.Scope, resourc
 }
 
 func hasRoleEditor(scope authpb.Scope) bool {
-	return scope.Role == authpb.Role_ROLE_EDITOR
+	return scope.Role == authpb.Role_ROLE_OWNER || scope.Role == authpb.Role_ROLE_EDITOR || scope.Role == authpb.Role_ROLE_UPLOADER
```
gmgigi96, line 59: I think it's better create two functions similar to this, one for checking the editor role and the other for uploader, and than while checking the publicshareScope for the `InitiateFileUploadRequest` check the hasRoleUploader also.

This also require that this function https://github.com/cs3org/reva/blob/master/internal/grpc/interceptors/auth/scope.go#L218 is aware of the uploader role (as is aware of the editor role)

internal/grpc/interceptors/auth/scope.go
```diff
@@ -52,7 +52,7 @@ func expandAndVerifyScope(ctx context.Context, req interface{}, tokenScope map[s
 
 	hasEditorRole := false
 	for _, v := range tokenScope {
-		if v.Role == authpb.Role_ROLE_EDITOR {
+		if v.Role == authpb.Role_ROLE_OWNER || v.Role == authpb.Role_ROLE_EDITOR || v.Role == authpb.Role_ROLE_UPLOADER {
```
gmgigi96, line 55: See the comment below

gmgigi96 (changes requested):
Hi @ishank011, I left some comments
Also you could fix this https://github.com/cs3org/reva/blob/master/internal/http/services/owncloud/ocs/conversions/role.go#L245 with the RoleUploader

labkode (approved):

PR #2519, opened 2022-02-08
remove creation of .space folder

micbar:
# Description

We do not develop the feature around the `.space` folder further.


dragotin (approved):
I like the idea to just use the existing Metadata functions to store the space attributes. Ok.

PR #2527, opened 2022-02-10
Add space props

micbar:
# Description

We need to store more space attributes in the storage.

This implements

- [x] Space Short Description
- [x] Space Image ID
- [x] Space Readme ID

... in the DecomposedFS

### DIFF
--- a/pkg/storage/utils/decomposedfs/spaces.go
+++ b/pkg/storage/utils/decomposedfs/spaces.go
@@ -387,9 +387,7 @@ func (fs *Decomposedfs) UpdateStorageSpace(ctx context.Context, req *provider.Up
 	}
 	space.Owner = u
 
-	space.SpaceType = filepath.Base(filepath.Dir(matches[0]))
-
-	metadata := make(map[string]string)
+	metadata := make(map[string]string, 5)
 	if space.Name != "" {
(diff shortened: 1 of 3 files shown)


butonic (approved):

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -387,16 +387,32 @@ func (fs *Decomposedfs) UpdateStorageSpace(ctx context.Context, req *provider.Up
 	}
 	space.Owner = u
 
+	space.SpaceType = filepath.Base(filepath.Dir(matches[0]))
+
+	metadata := make(map[string]string)
```
dragotin, line 392: Nitpick: Shouldn't you add the size to the make here? It is static and known... I think that is more efficient.

dragotin (approved):
nice one!

pkg/storage/utils/decomposedfs/spaces.go
micbar: Yes. Of course. The thing is, we don't know how many props the Client will set in the request, none of them is mandatory.

butonic (approved):

butonic (approved):

PR #2528, opened 2022-02-10
xattrs: use unix flocks when writing multiple xattrs

butonic:
checking the flock implementation options...

- [x] x/sys/unix ... current state of this PR ... ties us to unix
- [x] https://github.com/gofrs/flock ... uses a dedicated file ... I still like this ... but it seems unmaintaind
- [ ] wait for official lockfile: [Proposal: make the internal lockedfile package public](https://go.googlesource.com/proposal/+/master/design/33974-add-public-lockedfile-pkg.md)
- [ ] use https://github.com/rogpeppe/go-internal/blob/master/lockedfile/lockedfile.go until then? because it seems maintained.

### DIFF
--- a/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
+++ b/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
@@ -294,5 +294,5 @@ func All(filePath string) (attribs map[string]string, err error) {
 		err = errors.New("Failed to read all xattrs")
 	}
 
-	return attribs, nil
+	return attribs, err
 }
--- a/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
+++ b/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
@@ -19,6 +19,7 @@
 package xattrs
 
 import (
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


butonic:
https://github.com/gofrs/flock was the first implementation we looked at, but stumbled over the fact that it is intendet to write a lock *file*, not folders. Furthermore, we discovered the proposal to expose the existing locking implementation in the internal `golang/go/src/cmd/go/internal/lockedfile` package. But all threads lead to file locks, so I implemented a unix based lock.

butonic:
we will use https://github.com/rogpeppe/go-internal/blob/master/lockedfile/lockedfile.go to write a `{node}.flock` file when setting multiple extended attributes, OR when creating a application level log (`{node}.lock`).

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -178,20 +253,45 @@ func GetInt64(filePath, key string) (int64, error) {
...
+			attribs[name] = string(val)
 		}
-		attribs[name] = string(val)
+	}
+
+	if xerrs > 0 {
+		err = errors.New("Failed to read all xattrs")
 	}
 
 	return attribs, nil
```
C0rby, line 297: You shouldn't return nil here. If you do that then the `err` in line 270 will always be nil.

dragotin (approved):
Awesome stuff!

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -142,21 +191,48 @@ func Set(filePath string, key string, val string) error {
...
-			return err
+		if xerr = xattr.Set(filePath, key, []byte(val)); xerr != nil {
+			// log
+			xerrs++
 		}
 	}
+	if xerrs > 0 {
+		err = errors.Wrap(xerr, "Failed to set all xattrs")
+	}
 	return nil
```
C0rby, line 229: This also needs to be changed
```suggestion
	return err
```

PR #2530, opened 2022-02-11
Events for master

kobergj:
Brings events to reva!

Includes examples in pkg/events/example, Try them out!
Includes possibility to start a nats (stan) server

Will have no further effects as no services are configured to use eventing for now.


kobergj:
@labkode as requested here is the events PR for master. This is the version as accepted by ocis team, but I'm happy to implement any changes that you see fit!

labkode (approved):
@kobergj I think is a good first step there is no point into having two split branches for this feature so you can always push to master and bring the changes to edge.

This is very promising, many use-cases will be unblock with this work, good job!

PR #2533, opened 2022-02-13
Use space description on creation

micbar:
# Description

1) Make it possible to send an opaque "description" value in the CreateStorageSpaces Request.

2) Fixed a bug in the sharing roles. CoOnwners are now Managers.

### DIFF
--- a/pkg/storage/utils/decomposedfs/spaces.go
+++ b/pkg/storage/utils/decomposedfs/spaces.go
@@ -67,6 +67,13 @@ func (fs *Decomposedfs) CreateStorageSpace(ctx context.Context, req *provider.Cr
 			spaceID = string(e.Value)
 		}
 	}
+	// allow sending a space description
+	var description string
+	if req.Opaque != nil && req.Opaque.Map != nil {
+		if e, ok := req.Opaque.Map["description"]; ok && e.Decoder == "plain" {
+			description = string(e.Value)
+		}
+	}
 	// TODO enforce a uuid?
 	// TODO clarify if we want to enforce a single personal storage space or if we want to allow sending the spaceid
 	if req.Type == "personal" {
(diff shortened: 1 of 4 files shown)


butonic:
well, IMO we should rename Coowners, as ... they no longer exist, right?

micbar:
ok, but for me not in the scope of this PR

That would make merging master into edge more difficult.

butonic (approved):

butonic (approved):

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -157,6 +151,17 @@ func (fs *Decomposedfs) CreateStorageSpace(ctx context.Context, req *provider.Cr
 		return nil, err
```
dragotin, line 151: When you return here, the upper call to `SetMultiple()` has changed the nodes metadata.

Maybe it makes sense to let `SetMultiple` return a map of the previous values which can be used in a defer?

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -157,6 +151,17 @@ func (fs *Decomposedfs) CreateStorageSpace(ctx context.Context, req *provider.Cr
...
+	if err != nil {
+		return nil, err
+	}
+
+	resp := &provider.CreateStorageSpaceResponse{
+		Status: &v1beta11.Status{
+			Code: v1beta11.Code_CODE_OK,
+		},
+		StorageSpace: space,
+	}
```
dragotin, line 164: Oh, the response is created here.... That reminds me on the discussion that we had about "from where do we report errors?" - Maybe instead of just returning `nil, err` in the code flow before, we should return a proper response object with an error code set? (I haven't really checked the code if that would make sense here, you probably know).

dragotin (approved):
Comments...

pkg/storage/utils/decomposedfs/spaces.go
micbar: This is the storage driver, the errors are catched by the storage provider which converts them into proper responses with error codes. (application level errors)

pkg/storage/utils/decomposedfs/spaces.go
micbar: i need to really read the attributes from the storage again. This is done in https://github.com/cs3org/reva/pull/2533/files#diff-d308a0558595437a537e77e29346472e66ab3bc31c213fcd2dd9b5f64c02aa91R154

micbar: so the space has been successfully created already and the metadata has been set. If the AddGrant fails we should not delete the space again IMO.

pkg/storage/utils/decomposedfs/spaces.go
micbar: good point. Currently the CS3Apis are constructed in that way. The storageprovider is a Wrapper around the storageDrivers, which implement an internal interface.

PR #2539, opened 2022-02-14
[tests-only] Update composer.json

micbar:
# Description

Fixes a naming collision. The go modules try to use the vendor folder from the test php deps.

Solution: Change the name

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
   "name": "cs3org/reva",
   "config" : {
     "platform": {
-      "php": "7.2"
+      "php": "7.4"
     },
     "vendor-dir": "./vendor-php"
   },
--- a/composer.json
+++ b/composer.json
@@ -2,8 +2,9 @@
   "name": "cs3org/reva",
   "config" : {
     "platform": {
-      "php": "7.4"
-    }
+      "php": "7.2"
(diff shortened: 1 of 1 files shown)


phil-davis (dismissed):

phil-davis:
Note: going down from PHP 7.4 to min 7.2 is OK. But in practice, I think that the test code has has PHP 7.3 and/or 7.4 parameter and return type declarations etc added to it. So the test code might not actually run with PHP 7.2.

micbar:
I don't want to go down, i just took the composer.json from the edge branch and  modified it.

micbar:
fixed

phil-davis:
> I don't want to go down, i just took the composer.json from the edge branch and modified it.

I am surprised. master gets merged into edge every few weeks, so `composer.json` should get updated in edge. If that is not happening, then I might make a PR to the edge branch to apply the PHP 7.4 change.

micbar:
@ishank011 @labkode this should be merged to master to prevent collisions with the `vendor` folders from php deps and golang deps.

wkloucek:
@ishank011 @labkode can we please have this merged?

ishank011 (approved):

PR #2540, opened 2022-02-14
Decomposedfs xattr errors

butonic:
The xattrs package now uses the xattr.ENOATTR instead of os.ENODATA or os.ENOATTR to check attribute existence.


aduffeck:
LGTM

C0rby (approved):

PR #2547, opened 2022-02-15
implement if-match check in storageprovider

C0rby:
Implement a check for the if-match value in InitiateFileUpload to prevent overwrites of newer versions.

Implements this feature: https://github.com/owncloud/ocis/issues/3174

### DIFF
--- a/internal/grpc/services/storageprovider/storageprovider.go
+++ b/internal/grpc/services/storageprovider/storageprovider.go
@@ -291,6 +291,29 @@ func (s *service) InitiateFileUpload(ctx context.Context, req *provider.Initiate
 		}, nil
 	}
 
+	if ifMatch := req.GetIfMatch(); ifMatch != "" {
+		sRes, err := s.Stat(ctx, &provider.StatRequest{Ref: req.Ref})
+		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -291,14 +291,38 @@ func (s *service) InitiateFileUpload(ctx context.Context, req *provider.Initiate
...
+	ifMatch := req.GetIfMatch()
+	if ifMatch != "" {
+		sRes, err := s.Stat(ctx, &provider.StatRequest{Ref: req.Ref})
+		if err != nil {
+			return nil, err
+		}
+
+		switch sRes.Status.Code {
+		case rpc.Code_CODE_OK:
+			if sRes.Info.Etag != req.GetIfMatch() {
```
kobergj, line 304: Any reason why we are calling `req.GetIfMatch()` again instead of using `ifMatch` variable?

C0rby: good catch! :+1:

butonic (approved):

PR #2550, opened 2022-02-16
[tests-only] test update-getPersonalSpaceIdForUser

phil-davis:
core PR https://github.com/owncloud/core/pull/39793 changes test method `getPersonalSpaceIdForUser` so that it can work with 2 different space id formats.

This PR is to test that the core code does correctly process the current base64-encoded space-id.

Running with master branch.


phil-davis:
passed - good

PR #2552, opened 2022-02-16
 [tests-only] Test update-getPersonalSpaceIdForUser

phil-davis:
This is code from PR #2542 "remove base64 encoding of ids" and running the `getPersonalSpaceIdForUser` test code from core PR https://github.com/owncloud/core/pull/39793 that is core branch `update-getPersonalSpaceIdForUser`

This PR is just to confirm that the changed test code works.

PR #2553, opened 2022-02-16
[tests-only] Break test fix getPersonalSpaceIdForUser

phil-davis:
This is PR #2552 but with the idDelimiter string changed from "!" to "#". This should break some tests. I want to confirm that stuff fails.


phil-davis:
Example: https://drone.cernbox.cern.ch/cs3org/reva/5716/9/6
various scenarios fail because they cannot detect the personal space id, which is what I expect from this test.

PR #2574, opened 2022-02-21
use NATS JetStream

wkloucek:
NATS Streaming is depreciated in favor of NATS JetStream.

Please also see changes in https://github.com/owncloud/ocis/pull/3192


internal/grpc/interceptors/eventsmiddleware/events.go
```diff
@@ -67,6 +67,8 @@ func NewUnary(m map[string]interface{}) (grpc.UnaryServerInterceptor, int, error
 		if ev != nil {
 			if err := events.Publish(publisher, ev); err != nil {
 				log.Error(err)
+			} else {
+				log.Debug("event emitted just fine")
```
kobergj, line 71: If we need this log, please also add an identifier which event was emitted fine. Otherwise the log is mostly useless.

internal/grpc/interceptors/eventsmiddleware/events.go
```diff
@@ -93,6 +95,7 @@ func publisherFromConfig(m map[string]interface{}) (events.Publisher, error) {
 	case "nats":
 		address := m["address"].(string)
 		cid := m["clusterID"].(string)
-		return server.NewNatsStream(nats.Address(address), nats.ClusterID(cid))
+		pub, err := server.NewNatsStream(natsjs.Address(address), natsjs.ClusterID(cid))
+		return pub, err
```
kobergj, line 99: What is the point in this change?

pkg/events/example/example.go
```diff
@@ -53,15 +55,22 @@ func main() {
 
 // Server generates a nats server
 func Server() {
-	err := server.RunNatsServer()
+	err := server.RunNatsServer(
+		server.ClusterID("test-cluster"),
+		server.Host("127.0.0.1"),
+		server.Port(4222),
```
kobergj, line 61: use `9233` as port for nats please

pkg/events/example/example.go
```diff
@@ -53,15 +55,22 @@ func main() {
...
 	if err != nil {
 		log.Fatal(err)
 	}
 }
 
 // Client builds a nats client
 func Client() events.Stream {
-	c, err := server.NewNatsStream(nats.Address("127.0.0.1:9233"), nats.ClusterID("test-cluster"))
+	c, err := server.NewNatsStream(
+		natsjs.Address("127.0.0.1:4222"),
```
kobergj, line 71: same here port `9233`

pkg/events/server/nats.go
```diff
@@ -21,32 +21,43 @@ package server
...
 
 	for _, o := range opts {
-		o(&natsOpts, stanOpts)
+		o(options)
 	}
-	_, err := stanServer.RunServerWithOpts(stanOpts, &natsOpts)
-	return err
+
+	server, err := nserver.NewServer(
+		options,
```
kobergj, line 40: can we have this in one line? I don't see the point in having this streched over multiple lines

kobergj (changes requested):
Found some smaller ones...

pkg/events/server/nats.go
```diff
@@ -21,32 +21,41 @@ package server
...
-	return err
+
+	server, err := nserver.NewServer(options)
+	if err != nil {
+		return err
+	}
+
+	c := &nserver.JetStreamConfig{}
+
+	server.EnableJetStream(c)
```
kobergj, line 46: Linter is angry that we don't check the error
```suggestion
	_ = server.EnableJetStream(c)
```

kobergj (approved):

butonic:
we should update https://github.com/owncloud/ocis/pull/3192

PR #2575, opened 2022-02-21
[tests-only] Don't error if lock is not found

kobergj:
Is supposed to fix failing pipelines like this: https://drone.owncloud.com/owncloud/ocis/9533/25/4

### DIFF
--- a/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
+++ b/pkg/storage/utils/decomposedfs/xattrs/xattrs.go
@@ -130,7 +130,7 @@ func CopyMetadata(src, target string, filter func(attributeName string) bool) (e
 		rerr := filelocks.ReleaseLock(writeLock)
 
 		// if err is non nil we do not overwrite that
-		if err == nil {
+		if err == nil && !isNotFoundError(rerr) {
 			err = rerr
 		}
 	}()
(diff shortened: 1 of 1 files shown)


pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -130,7 +130,7 @@ func CopyMetadata(src, target string, filter func(attributeName string) bool) (e
 		rerr := filelocks.ReleaseLock(writeLock)
 
 		// if err is non nil we do not overwrite that
-		if err == nil {
+		if err == nil && !isNotFoundError(rerr) {
```
C0rby, line 133:
```suggestion
		if err == nil && !errors.Is(rerr, fs.ErrNotExist) {
```

C0rby (changes requested):

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
kobergj: Much better :+1:

dragotin:
Oh no, please don't. This is hiding a real problem, which is: There are two processes trying to create a file. There was a bug in the locking code that failed to detect that and wait a moment until the first one has released.

That was fixed in 59a37c326a5b2ccdcbfb95a5174665017b7ea0bb . Please double check if you have that commit, and if you have, we need to investigate...

dragotin (changes requested):
Please dont merge. See comment.

C0rby:
Right, I think we should do an error check before `ReleaseLock`. If the lock couldn't be set then we should try to release it. But still return the original error.

PR #2581, opened 2022-02-23
set status values to match what is currently in oCIS frontend.go

phil-davis:
https://github.com/owncloud/ocis/blob/master/storage/pkg/command/frontend.go currently has hard-coded values in capabilities core status that get returned in a capabilities response.

reva has 10.0.9.5 still in some places, but other places have 10.0.11.5

Make the hardcoded values for version, product name etc. the same - then at least they will be consistent for now.

There has been other mention of implementing some proper advertising of a product name and version of reva and oCIS using semver etc. That is a different issue, and can be sorted out after this PR gets the hard-coded values consistent.

### DIFF
--- a/changelog/unreleased/change-status-values.md
+++ b/changelog/unreleased/change-status-values.md
@@ -1,0 +1,5 @@
+Change: Update hard-coded status values
+
+The hard-coded version and product values have been updated to be consistent in all places in the code.
+
+https://github.com/cs3org/reva/pull/2581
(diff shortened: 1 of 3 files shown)


phil-davis:
If this is approved then I will make a similar PR to master, so that this is a least consistent.

micbar (approved):

C0rby (approved):

PR #2585, opened 2022-02-23
Add metadata storage and indexer

aduffeck:
This PR adds the metadata storage layer and the indexer from ocis-pkg. The indexer has also been extended to allow for indexing nested fields and supporting index functions that are necessary for indexing polymorphic fields.


pkg/storage/utils/metadata/cs3.go
```diff
@@ -0,0 +1,379 @@
...
+		return err
+	}
+
+	md, _ := metadata.FromOutgoingContext(ctx)
+	req.Header.Add(ctxpkg.TokenHeader, md.Get(ctxpkg.TokenHeader)[0])
+	resp, err := cs3.dataGatewayClient.Do(req)
+	if err != nil {
+		return err
+	}
+	if err = resp.Body.Close(); err != nil {
```
kobergj, line 163: Again unnecessary `if` case

pkg/storage/utils/metadata/cs3.go
```diff
@@ -0,0 +1,379 @@
...
+	return nil
+}
+
+// CreateSymlink creates a symlink
+func (cs3 *CS3) CreateSymlink(ctx context.Context, oldname, newname string) error {
+	if _, err := cs3.ResolveSymlink(ctx, newname); err == nil {
+		return os.ErrExist
+	}
+
+	err := cs3.SimpleUpload(ctx, newname, []byte(oldname))
```
kobergj, line 335:
```suggestion
	return cs3.SimpleUpload(ctx, newname, []byte(oldname))
```

pkg/storage/utils/sync/mutex.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// RUnlock undoes a single RLock call.
+func (m *NamedRWMutex) RUnlock(name string) {
+	m.loadOrStore(name).RUnlock()
+}
+
+func (m *NamedRWMutex) loadOrStore(name string) *sync.RWMutex {
+	pmu := m.pool.Get()
+	mmu, loaded := m.mus.LoadOrStore(name, pmu)
```
kobergj, line 61: What happens if `name` is not in the `mus` map? Wouldn't `mmu` be `nil` then? This would lead to a nil pointer reference

pkg/storage/utils/indexer/index/non_unique.go
```diff
@@ -0,0 +1,240 @@
...
+	var matches = make([]string, 0)
+	paths, err := idx.storage.ReadDir(context.Background(), path.Join("/", idx.indexRootDir, v))
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, p := range paths {
+		matches = append(matches, path.Base(p))
+	}
```
C0rby, line 98:
```suggestion
	paths, err := idx.storage.ReadDir(context.Background(), path.Join("/", idx.indexRootDir, v))

	if err != nil {
		return nil, err
	}
       matches = make([]string, 0, len(paths))
	for _, p := range paths {
		matches = append(matches, path.Base(p))
	}
```

pkg/storage/utils/indexer/indexer.go
```diff
@@ -0,0 +1,432 @@
...
+					root.Children[1].Token.Value,
+				},
+			})
+			for _, child := range root.Children {
+				if err := buildTreeFromOdataQuery(child, tree.left); err != nil {
+					return err
+				}
+			}
+		default:
+			return fmt.Errorf("operator not supported")
```
C0rby, line 428:
```suggestion
			return errors.New("operator not supported")
```

pkg/storage/utils/metadata/disk.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// ReadDir returns the resource infos in a given directory
+func (disk *Disk) ReadDir(_ context.Context, p string) ([]string, error) {
+	infos, err := ioutil.ReadDir(disk.targetPath(p))
+	if err != nil {
+		return nil, err
+	}
+
+	entries := []string{}
```
C0rby, line 72:
```suggestion
	entries := make([]string, 0, len(infos))
```

C0rby (commented):
Just some small nitpicks

pkg/storage/utils/indexer/helper.go
```diff
@@ -0,0 +1,15 @@
+package indexer
+
+// dedup removes duplicate values in given slice
+func dedup(s []string) []string {
+	for i := 0; i < len(s); i++ {
```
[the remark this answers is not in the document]
C0rby: You just have a small typo it should be: `exists[ss]`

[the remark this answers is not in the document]
C0rby: I find your proposal more readable but @aduffeck's implementation seems to be faster and doesn't do any allocations.

pkg/storage/utils/indexer/indexer.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	return results, nil
+}
+
+// t is used to infer the indexed field names. When building an index search query, field names have to respect Golang
+// conventions and be in PascalCase. For a better overview on this contemplate reading the reflection package under the
+// indexer directory. Traversal of the tree happens in a pre-order fashion.
+// TODO implement logic for `and` operators.
+func (i *Indexer) resolveTree(t interface{}, tree *queryTree, partials *[]string) error {
```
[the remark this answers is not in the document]
aduffeck: I had to try it because I wasn't sure myself, but when the slice is given by value you can alter the elements being pointed at by both the original slice and the copy, but you can't alter the original slice itself, i.e. adding/removing elements is not possible. See https://go.dev/play/p/Mk4IkD0dKz9

pkg/storage/utils/sync/mutex.go
aduffeck: `m.mus.LoadOrStore` returns the existing value if there is one with that key already or stores and returns the given one if not. So I don't think there's a case where `mmu` could be nil.

pkg/storage/utils/indexer/helper.go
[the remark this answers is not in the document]
aduffeck: I ran some benchmarks and @kobergj's proposal is actually quite a bit faster, deduing a 1000 element slice takes ~0.1ms now, down from ~1.3ms with the original implementation. O(n²) is bad ;)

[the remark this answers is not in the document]
C0rby: True my slice wasn't as big. With bigger slices the difference will become bigger.

kobergj (approved):
LGTM :+1:

PR #2588, opened 2022-02-24
cephfs: fix Dockerfile.revad-ceph to use the right base image

SamuAlfageme:
- In Aug2021 https://hub.docker.com/r/ceph/daemon-base was moved
  to quay.ceph.io and the builds for this image were failing for
  some weeks after January.


glpatcern (approved):

ishank011 (approved):
The lint issue is because of the timeout config change. Should be fixed after merging

PR #2589, opened 2022-02-24
replace deprecated linter modules

C0rby:
Replaced the deprecated linter modules with the recommended ones.


kobergj (approved):
Nice :+1:

PR #2595, opened 2022-02-28
Add integration test for groupprovider 

rhafer:
Provide a similar set of tests as for the userprovider. The LDAP tests
are still mostly failing because of issues in the LDAP driver for the
groupprovider. (They are skipped by default anyway)


micbar (approved):
LGTM

PR #2608, opened 2022-03-03
trace: honour tracing_service_name config directive

labkode:
The already available `tracing_service_name` directive was not been honoured as the `reva` value was hardcoded.


labkode:
@cs3org/owncloud-team ping!

ishank011 (approved):

PR #2613, opened 2022-03-03
Externalize custom mime types configuration for storage providers

glpatcern:
As discussed with @ishank011 this small PR moves the custom mime types configuration to a JSON file, so that it can be reused when many storage providers are deployed at the same time.


examples/nextcloud-integration/custom-mime-types-demo.json
```diff
@@ -0,0 +1,10 @@
+[
+    {
+        "file_ext": ".zmd",
```
ishank011, line 3: Can we have a string->string map here? We can directly read the json into a map object and not have to do the parsing.

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -61,7 +63,12 @@ type config struct {
 	DataServerURL    string                            `mapstructure:"data_server_url" docs:"http://localhost/data;The URL for the data server."`
 	ExposeDataServer bool                              `mapstructure:"expose_data_server" docs:"false;Whether to expose data server."` // if true the client will be able to upload/download directly to it
 	AvailableXS      map[string]uint32                 `mapstructure:"available_checksums" docs:"nil;List of available checksums."`
-	MimeTypes        map[string]string                 `mapstructure:"mimetypes" docs:"nil;List of supported mime types and corresponding file extensions."`
+	CustomMimeTypes  string                            `mapstructure:"custom_mimetypes" docs:"nil;An optional mapping file with the list of supported custom file extensions and corresponding mime types."`
```
ishank011, line 66: Let's call it `CustomMimeTypesJson`

PR #2620, opened 2022-03-07
improve grant handling for space members

micbar:
# Description

Improve the grant handling

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -701,7 +701,16 @@ func (s *svc) removeSpaceShare(ctx context.Context, ref *provider.ResourceId, gr
 	if err != nil {
 		return nil, errors.Wrap(err, "gateway: error getting grant to remove from storage")
 	}
-	removeGrantStatus, err := s.removeGrant(ctx, ref, grantee, listGrantRes.Grants[0].Permissions)
+	var permissions *provider.ResourcePermissions
+	for _, g := range listGrantRes.Grants {
+		if isEqualGrantee(g.Grantee, grantee) {
+			permissions = g.Permissions
+		}
+	}
+	if permissions == nil {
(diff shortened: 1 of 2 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -698,10 +698,19 @@ func (s *svc) removeShare(ctx context.Context, req *collaboration.RemoveShareReq
 
 func (s *svc) removeSpaceShare(ctx context.Context, ref *provider.ResourceId, grantee *provider.Grantee) (*collaboration.RemoveShareResponse, error) {
 	listGrantRes, err := s.listGrants(ctx, ref)
+	var permissions *provider.ResourcePermissions
```
kobergj, line 701: can you move this line down? It doesn't make sense to have `permissions` created here before the `err` from `listGrants` is checked...

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -733,3 +742,23 @@ func shareIsSpaceRoot(key *collaboration.ShareKey) bool {
 	}
 	return refIsSpaceRoot(key.ResourceId)
 }
+
+func isEqualGrantee(a, b provider.Grantee) bool {
```
kobergj, line 746: Would it make sense to have a and b being `*provider.Grantee`? We would avoid having to dereference them in line `706`.

kobergj (changes requested):
LG :+1: Just some minors

internal/grpc/services/gateway/usershareprovider.go
micbar: done

internal/grpc/services/gateway/usershareprovider.go
micbar: done

kobergj (approved):

PR #2622, opened 2022-03-07
implement listing shares in spaces

C0rby:
Instead of the query parameter `path` one can now send `space_ref` to
list shares in spaces. The URL would look like this:
`https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/shares?reshares=true&space_ref=ae96eb17-1b37-4bfc-8ed6-9c044723a43f/elmo.gif`
The difference to the 'old' URL is the parameter `space_ref` instead of the `path` parameter.
`path` is always assumed to be relative to the users home and `space_ref` is the combined spaceid and the path relative to the space root.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -42,6 +42,10 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+const (
+	spaceIDDelimiter = "!"
+)
+
 var (
 	matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
 	matchAllCap   = regexp.MustCompile("([a-z0-9])([A-Z])")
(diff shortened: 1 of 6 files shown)


tests/oc-integration-tests/local/ldap-users.toml
```diff
@@ -34,7 +34,7 @@ cn="cn"
 driver = "ldap"
 
 [grpc.services.userprovider.drivers.ldap]
-hostname="ldap"
+hostname="localhost"
```
wkloucek, line 37: that one shouldn't change?

pkg/utils/utils.go
```diff
@@ -357,6 +363,35 @@ func ParseStorageSpaceReference(sRef string) (provider.Reference, error) {
 	}, nil
 }
 
+// FormatStorageSpaceReference will format a storage space reference into a string representation.
+// If ref or ref.ResourceId are nil an empty string will be returned.
```
kobergj, line 367: Not quite true. It errors in that case.

kobergj (changes requested):

pkg/utils/utils.go
C0rby: Oh, true. Forgot to update the comment.

kobergj (approved):

PR #2623, opened 2022-03-07
add driveAlias

micbar:
# Description

Implements the base features for space aliases as a space property in the decomposedFS

## Detailed description

https://github.com/owncloud/ocis/pull/3283#issue-1161675086

### DIFF
--- a/pkg/storage/utils/decomposedfs/spaces.go
+++ b/pkg/storage/utils/decomposedfs/spaces.go
@@ -105,7 +105,7 @@ func (fs *Decomposedfs) CreateStorageSpace(ctx context.Context, req *provider.Cr
 		return nil, err
 	}
 
-	metadata := make(map[string]string, 3)
+	metadata := make(map[string]string, 6)
 
 	// always enable propagation on the storage space root
 	// mark the space root node as the end of propagation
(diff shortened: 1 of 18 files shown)


micbar:
I discovered a possible problem in https://github.com/cs3org/reva/blob/40f7f008a1f7f476e3f10c6d422ead61c4512fd6/pkg/storage/utils/decomposedfs/spaces.go#L223

We cannot assume that the length is as we expect it to be. I found that during unit testing.
`../../spaces/4c/510ada-c86b-4815-8820-42cdf82c3d51/nodes/4c/51/0a/da/-c86b-4815-8820-42cdf82c3d51` could be
`../../spaces/us/erid/nodes/4c/51/0a/da/-c86b-4815-8820-42cdf82c3d51`

1) Create a user with a userid which is not UUIDv4
2) Log in, we will create a personal space with the userid as spaceID
3) the string will be shorter as we expect it

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -62,22 +62,18 @@ func (fs *Decomposedfs) CreateStorageSpace(ctx context.Context, req *provider.Cr
...
-		}
-	}
+	description := utils.ReadPlainFromOpaque(req.Opaque, "description")
+	// allow sending a spaceAlias
+	alias := utils.ReadPlainFromOpaque(req.Opaque, "spaceAlias")
 	// TODO enforce a uuid?
 	// TODO clarify if we want to enforce a single personal storage space or if we want to allow sending the spaceid
 	if req.Type == spaceTypePersonal {
 		spaceID = req.GetOwner().GetId().GetOpaqueId()
+		alias = spaceTypePersonal + "/" + req.GetOwner().GetUsername()
```
kobergj, line 76: This means that personal spaces will always have an `alias` and I cannot change this as a user. 

Is this intended behaviour?

kobergj (commented):
Just a small question

pkg/storage/utils/decomposedfs/spaces.go
micbar: yes. We intend to resolve it in the url always as `https://cloud.owncloud.com/files/spaces/personal/einstein/relative/path/to/file.txt` appended by the ID which will result in the stable and user-friendly URL like `https://cloud.owncloud.com/files/spaces/personal/einstein/relative/path/to/file.txt?id=95b4c6f6-d8f2-4976-b0a3-1b016bdfcb7e`

micbar: But this can be changed. We only set it during the creation. When updating the space, this can be changed also.

It does not need to be unique.

kobergj (approved):

pkg/storage/utils/decomposedfs/spaces.go
micbar: But let us discuss the defaults.

micbar:
@kobergj I added configurable template strings.

kobergj (approved):
Looks still good to me :+1:

micbar:
general approach needs to be discussed

@dragotin FYI

micbar:
general approach needs to be discussed

@dragotin FYI

PR #2625, opened 2022-03-09
Locking fixes

wkloucek:
- expire locks
- pass lockid from the InitiateUploadRequest via metadata to the FinishUpload, so that we don't need to provide that info the the datagateway
- enable lock modification for app based locks independently of the user
- implement locking for public storage provider

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/locks.go
+++ b/pkg/storage/utils/decomposedfs/node/locks.go
@@ -129,7 +129,7 @@ func (n Node) ReadLock(ctx context.Context) (*provider.Lock, error) {
 			return nil, errors.Wrap(err, "Decomposedfs: could not remove expired lock file")
 		}
 		// we successfully deleted the expired lock
-		return nil, nil
+		return nil, errtypes.NotFound("no lock found")
 	}
 
 	return lock, nil
--- a/pkg/storage/utils/decomposedfs/node/locks.go
+++ b/pkg/storage/utils/decomposedfs/node/locks.go
@@ -123,14 +123,15 @@ func (n Node) ReadLock(ctx context.Context) (*provider.Lock, error) {
 	}
 
 	// lock already expired
-	if time.Now().After(time.Unix(int64(lock.Expiration.Seconds), int64(lock.Expiration.Nanos))) {
+	if lock.Expiration != nil && time.Now().After(time.Unix(int64(lock.Expiration.Seconds), int64(lock.Expiration.Nanos))) {
 		if err = os.Remove(f.Name()); err != nil {
 			return nil, errors.Wrap(err, "Decomposedfs: could not remove expired lock file")
 		}
+		// we successfully deleted the expired lock
 		return nil, nil
 	}
 
-	return lock, err
+	return lock, nil
 }
 
 // RefreshLock refreshes the node's lock
--- a/pkg/storage/utils/decomposedfs/node/locks.go
+++ b/pkg/storage/utils/decomposedfs/node/locks.go
@@ -318,24 +318,31 @@ func isLockModificationAllowed(ctx context.Context, oldLock *provider.Lock, newL
 		return true, nil
 	}
 
-	var allowed bool = true
+	var forbidden bool = true
 	var err error
 
-	if oldLock.AppName != "" && oldLock.AppName != newLock.AppName {
-		allowed = false
-		err = errtypes.PermissionDenied("cannot change holder (app name) when refreshing a lock")
+	if oldLock.AppName != "" || newLock.AppName != "" {
+		if oldLock.AppName != newLock.AppName {
+			err = errtypes.PermissionDenied("cannot change holder (app name) when refreshing a lock")
+		} else {
+			forbidden = forbidden && true
+		}
 	}
 
-	if oldLock.User != nil && !utils.UserEqual(oldLock.User, newLock.GetUser()) {
-		allowed = false
-		err = errtypes.PermissionDenied("cannot change holder (user) when refreshing a lock")
+	if oldLock.User != nil || newLock.GetUser() != nil {
+		if !utils.UserEqual(oldLock.User, newLock.GetUser()) {
+			err = errtypes.PermissionDenied("cannot change holder (user) when refreshing a lock")
+		} else {
+			forbidden = forbidden && true
+		}
+
+		u := ctxpkg.ContextMustGetUser(ctx)
+		if !utils.UserEqual(oldLock.User, u.Id) {
+			err = errtypes.PermissionDenied("mismatching holder")
+		} else {
+			forbidden = forbidden && true
+		}
 	}
 
-	u := ctxpkg.ContextMustGetUser(ctx)
-	if oldLock.User != nil && !utils.UserEqual(oldLock.User, u.Id) {
-		allowed = false
-		err = errtypes.PermissionDenied("mismatching holder")
-	}
-
-	return allowed, err
+	return !forbidden, err
 }
--- a/pkg/storage/utils/decomposedfs/node/locks.go
+++ b/pkg/storage/utils/decomposedfs/node/locks.go
@@ -32,6 +32,7 @@ import (
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/filelocks"
+	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -121,7 +121,17 @@ func (n Node) ReadLock(ctx context.Context) (*provider.Lock, error) {
...
 	}
-	return lock, err
+
+	// lock already expired
+	if lock.Expiration != nil && time.Now().After(time.Unix(int64(lock.Expiration.Seconds), int64(lock.Expiration.Nanos))) {
+		if err = os.Remove(f.Name()); err != nil {
+			return nil, errors.Wrap(err, "Decomposedfs: could not remove expired lock file")
+		}
+		// we successfully deleted the expired lock
+		return nil, nil
```
kobergj, line 131: Is this correct? When I call `ReadLock` I would expect to get an `error` or a working lock.

kobergj (changes requested):

pkg/storage/utils/decomposedfs/node/locks.go
dragotin: Confusing... This method only reads an existing Application-Lock file and returns the content. If the lock is expired, it is indeed correct to return nothing.

What you mean is `RefreshLock` or `SetLock`.

The method documentation should be more verbose on that.

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -165,15 +175,6 @@ func (n *Node) RefreshLock(ctx context.Context, lock *provider.Lock) error {
...
 
-	u := ctxpkg.ContextMustGetUser(ctx)
-	if !utils.UserEqual(oldLock.User, u.Id) {
-		return errtypes.PermissionDenied("cannot refresh lock of another holder")
-	}
-
-	if !utils.UserEqual(oldLock.User, lock.GetUser()) {
-		return errtypes.PermissionDenied("cannot change holder when refreshing a lock")
-	}
-
```
dragotin, line 176: This is IMHO controversal: It means a lock can be removed by anybody, right? And we need that in the context of user changes through the wopi server if I understood correctly.

Can't we get an additional flag that indicates if we allow that?

dragotin (approved):

pkg/storage/utils/decomposedfs/node/locks.go
kobergj: Mmh. I still don't get it. What would it return if the requested lock does not exist?

wkloucek: An expired lock doesn't even exist from an outside perspective. We just don't have a collector for expired locks. Therefore we delete it lazily when someone calls `ReadLock`

pkg/storage/utils/decomposedfs/node/locks.go
wkloucek: You're right... We need to change our behaviour dependent on the lock type: https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.LockType

WOPI needs to use a shared lock.

glpatcern: IIRC the concept was that a lock can be refreshed either by the same `user` or by the same `appname`. Cf. the WIP for EOS: https://github.com/cs3org/reva/pull/2444/files#diff-65d95a652cc692a947b89561eaea871da49ceb5eb0b8d53ce741ae5f996b90cbR783 (which should likely be factored out).

WOPI needs to use write locks, not shared locks, otherwise we defeat the purpose of preventing sync/web clients from writing an open file.

pkg/storage/utils/decomposedfs/node/locks.go
kobergj: Yes I got that. But in that case it should return `not found` imho, because `An expired lock doesn't even exist from an outside perspective.`

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -121,7 +122,17 @@ func (n Node) ReadLock(ctx context.Context) (*provider.Lock, error) {
 		appctx.GetLogger(ctx).Error().Err(err).Msg("Decomposedfs: could not decode lock file, ignoring")
 		return nil, errors.Wrap(err, "Decomposedfs: could not read lock file")
 	}
-	return lock, err
+
+	// lock already expired
+	if lock.Expiration != nil && time.Now().After(time.Unix(int64(lock.Expiration.Seconds), int64(lock.Expiration.Nanos))) {
+		if err = os.Remove(f.Name()); err != nil {
+			return nil, errors.Wrap(err, "Decomposedfs: could not remove expired lock file")
```
dragotin, line 129: I think the `return` is not correct here. It is not nice if the file lock can not be removed physically but in the context of the application, that is no error condition in regards of the lock from the callers perspective. 

So imho this should only place a warning in the logfile, and fall through to the _no lock found_ reply below.

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -307,3 +312,24 @@ func (n *Node) hasLocks(ctx context.Context) bool {
...
+	if oldLock.AppName != "" && oldLock.AppName != newLock.AppName {
+		allowed = false
+		err = errtypes.PermissionDenied("cannot change holder (app name) when refreshing a lock")
+	}
+
+	if oldLock.User != nil && !utils.UserEqual(oldLock.User, newLock.GetUser()) {
+		allowed = false
+		err = errtypes.PermissionDenied("cannot change holder (user) when refreshing a lock")
+	}
+
```
dragotin, line 333: Not sure if the error logic is correct here: If the oldLock.AppName is empty, the lock modification is allowed, same for user. Is that intended?

dragotin (approved):
Fine with me.

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -307,3 +312,24 @@ func (n *Node) hasLocks(ctx context.Context) bool {
 	_, err := os.Stat(n.LockFilePath()) // FIXME better error checking
 	return err == nil
 }
+
+func isLockModificationAllowed(oldLock *provider.Lock, newLock *provider.Lock) (bool, error) {
+	if oldLock.Type == provider.LockType_LOCK_TYPE_SHARED {
+		return true, nil
+	}
+
+	var allowed bool = true
```
kobergj, line 321: Can we reverse the logic and default this to `false`? 
It's good practice to have permissions default to `false` because in case of a logic error in the function (maybe through later refactoring) the possibility of unauthenticated access is minimized.

pkg/storage/utils/decomposedfs/node/locks.go
dragotin: I think that is fine.

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -307,3 +312,37 @@ func (n *Node) hasLocks(ctx context.Context) bool {
...
+	if oldLock.Type == provider.LockType_LOCK_TYPE_SHARED {
+		return true, nil
+	}
+
+	var forbidden bool = true
+	var err error
+
+	if oldLock.AppName != "" || newLock.AppName != "" {
+		if oldLock.AppName != newLock.AppName {
+			err = errtypes.PermissionDenied("cannot change holder (app name) when refreshing a lock")
```
kobergj, line 326: Does it make sense to continue after we encountered this error? Imho we could just `return false, err`

pkg/storage/utils/decomposedfs/node/locks.go
```diff
@@ -307,3 +312,37 @@ func (n *Node) hasLocks(ctx context.Context) bool {
...
+		} else {
+			forbidden = forbidden && true
```
kobergj, line 328: I don't understand this logic. With `forbidden` being defaulted to `true` wouldn't this be `true && true` which would be `true`?

So my question is: How is `forbidden` supposed to become `false`?

kobergj (changes requested):

pkg/storage/utils/decomposedfs/node/locks.go
wkloucek: If the user is not present in both locks, we allow modification. Same applies for the app name.

pkg/storage/utils/decomposedfs/node/locks.go
wkloucek: done :+1:

pkg/storage/utils/decomposedfs/node/locks.go
wkloucek: done :+1:

kobergj (commented):
LGTM :+1: But lock tests are red...

wkloucek:
> LGTM +1 But lock tests are red...

I fixed them in https://github.com/cs3org/reva/pull/2625/commits/520f58b4f53dde011c5b25331ee6a5133cb87865

kobergj (approved):

PR #2626, opened 2022-03-09
Cs3 share manager

aduffeck:
This PR introduces a new share manager which uses the new metadata storage backend.


pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+type ReceivedShareMetadata struct {
+	State      collaboration.ShareState `json:"state"`
+	MountPoint *provider.Reference      `json:"mountpoint"`
+}
+
+func init() {
+	registry.Register("cs3", NewDefault)
+}
+
+type config struct {
```
kobergj, line 82: Can we make this public? Having an exported overview of the configuration for this might be beneficial

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+func New(s Storage, indexer Indexer) (*Manager, error) {
+	return &Manager{
+		storage:     s,
+		indexer:     indexer,
+		initialized: false,
+	}, nil
+}
+
+func (m *Manager) initialize() error {
+	if m.initialized {
```
kobergj, line 117: We can have an concurrency issue here. Imagine two routines go into this function at the same the time. `m.initialized` would be false for both, so both start the initialization process. This can lead to unforeseeable consequences

My best approach was to `Lock` if not initialized and then check for `initialized` again. But there might be better ones.

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+}
+
+// Share creates a new share
+func (m *Manager) Share(ctx context.Context, md *provider.ResourceInfo, g *collaboration.ShareGrant) (*collaboration.Share, error) {
+	if err := m.initialize(); err != nil {
+		return nil, err
+	}
+	user := ctxpkg.ContextMustGetUser(ctx)
+	now := time.Now().UnixNano()
+	ts := &typespb.Timestamp{
```
kobergj, line 156: maybe use `utils.TSNow()` instead?

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	metadataPath := path.Join("metadata", share.Id.OpaqueId)
+	err = m.storage.MakeDirIfNotExist(ctx, metadataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = m.indexer.Add(share)
+	if err != nil {
```
kobergj, line 192: `err` check is unnecessary. Just `return share, err`

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+		return err
+	}
+
+	fn := path.Join("shares", ref.GetId().OpaqueId)
+	err = m.storage.Delete(ctx, fn)
+	if err != nil {
+		return err
+	}
+
+	return m.indexer.Delete(share)
```
kobergj, line 231: Are we sure we want to error if the indexer fails to delete but the storage was successful? What should the client do with that information?

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+}
+
+// ListShares returns the shares created by the user. If md is provided is not nil,
```
kobergj, line 234: Comment unclear. What is `md`?

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	share.Permissions = p
+
+	data, err := json.Marshal(share)
+	if err != nil {
+		return nil, err
+	}
+
+	fn := path.Join("shares", share.Id.OpaqueId)
+	err = m.storage.SimpleUpload(ctx, fn, data)
+	if err != nil {
```
kobergj, line 281: Again unnecessary `if` statement

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	for _, id := range receivedIds {
+		share, err := m.getShareByID(ctx, id)
+		if err != nil {
+			return nil, err
+		}
+		metadata, err := m.downloadMetadata(ctx, share)
+		if err != nil {
+			if _, ok := err.(errtypes.NotFound); ok {
```
kobergj, line 334: Can we reverse the logic here?
```golang
if !ok {
    return nil, err
}
```
Then we can get rid of the else statement which reduces cognitive complexity.

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+
+	metadata, err := m.downloadMetadata(ctx, share)
+	if err != nil {
+		if _, ok := err.(errtypes.NotFound); ok {
+			// use default values if the grantee didn't configure anything yet
+			metadata = ReceivedShareMetadata{
+				State: collaboration.ShareState_SHARE_STATE_PENDING,
+			}
+		} else {
```
kobergj, line 370: Same here. The `else` statement makes the function unnecessary complicated.

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	fn, err := metadataFilename(rshare.Share, user)
+	if err != nil {
+		return nil, err
+	}
+	err = m.storage.SimpleUpload(ctx, fn, data)
+	if err != nil {
```
kobergj, line 424: `err` check also here unnecessary

pkg/storage/utils/indexer/errors/errors.go
```diff
@@ -50,8 +47,5 @@ func (e *NotFoundErr) Error() string {
 	return fmt.Sprintf("%s with %s=%s not found", e.TypeName, e.IndexBy.String(), e.Value)
 }
 
-// IsNotFoundErr checks whether an error is of type IsNotFoundErr.
-func IsNotFoundErr(e error) bool {
-	_, ok := e.(*NotFoundErr)
-	return ok
-}
+// IsNotFound implements the IsNotFound interface.
+func (e *NotFoundErr) IsNotFound() {}
```
kobergj, line 51: I like :+1: That's much simpler than before.

kobergj (changes requested):
Found some smaller ones. Looks good in general!

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,564 @@
...
+	"github.com/mitchellh/mapstructure"
+	"github.com/pkg/errors"
+	"google.golang.org/genproto/protobuf/field_mask"
+)
+
+//go:generate mockery -name Storage
+//go:generate mockery -name Indexer
+
+// Storage is the interface to the metadata storage backend
+type Storage interface {
```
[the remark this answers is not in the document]
aduffeck: Yes, you are right. I'm currently refactoring this as part of the publicshare manager pull request here: https://github.com/aduffeck/reva/commit/f1b31ef53f9d7156b88b415e2f23a788c33fb52b 
So I hope it's fine to leave it in for this pr, it will be gone soon ;)

pkg/share/manager/cs3/cs3.go
aduffeck: Hm, maybe. But currently the struct is only used for extracting the configuration values from the generic interface which is being passed as part of the registry contract. The values are only passed on to the storage and not even stored in the manager itself.
Personally I would rather keep it private and expose it when it's actually being used.

pkg/share/manager/cs3/cs3.go
aduffeck: My idea was that running the initializing steps multiple times should be fine because they should be idempotent. I also *think* that running them in parallel shouldn't do any harm, but I agree it's better to err on the safe side here. I'll add some locking to it.

pkg/share/manager/cs3/cs3.go
aduffeck: But if the share couldn't be persisted in the index we don't want the caller to use the share object as that could lead to inconsistencies. I think it's safer to not give that option in the first place, the share hasn't actually be (fully) created.

pkg/share/manager/cs3/cs3.go
aduffeck: See my comment above.

pkg/share/manager/cs3/cs3.go
kobergj: We return a non-nil error. That should be enough indication that the share is not fully initialized and you shouldn't use it at all. I think it is common understanding in go that you do not use a value if the corresponding error is not nil. 

So in my opinion we can safely assume that the `share` value is not used if `err != nil`

aduffeck: You are probably right and it'll never be a problem, my personal preference would still have been to keep it a tad cleaner at the cost of an additional `if`. I've also seen both ways being used in go projects. It's fine though, I'll change it and push a commit in a sec.

pkg/share/manager/cs3/cs3.go
```diff
@@ -0,0 +1,558 @@
...
+		Ctime:       ts,
+		Mtime:       ts,
+	}
+
+	data, err := json.Marshal(share)
+	if err != nil {
+		return nil, err
+	}
+
+	fn := path.Join("shares", share.Id.OpaqueId)
```
dragotin, line 183: I saw a few places where we hardcode path segments like this. I wonder if we should not have a central place where we define these instead of hardcoding.

Maybe not really part of this PR..

aduffeck: @dragotin I just pushed a commit which changes the remaining places to also make use of `shareFilename()` which encapsulates the knowledge about the directoy layout.
Not sure if that's what you meant, but I don't think we need to move this to a central place inside the repository as the layout is only relevant to the share manager itself.

kobergj (approved):

PR #2628, opened 2022-03-10
Webdav trash-bin API for spaces

C0rby:
Added the webdav trash-bin API for spaces. It works the same way as the 'normal' API.
The spaces trashbin is reachable at '/remote.php/dav/spaces/trash-bin/{spaceid}`.

### DIFF
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -27,7 +27,6 @@ import (
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
-	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/errors"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/net"
@@ -37,6 +36,10 @@ import (
 	"github.com/cs3org/reva/v2/pkg/rhttp/router"
 	"github.com/cs3org/reva/v2/pkg/utils"
 	"google.golang.org/grpc/metadata"
+)
+
+const (
+	_trashbinPath = "trash-bin"
 )
 
 type tokenStatInfoKey struct{}
@@ -172,7 +175,7 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 			base := path.Join(ctx.Value(net.CtxKeyBaseURI).(string), "spaces")
 			ctx := context.WithValue(ctx, net.CtxKeyBaseURI, base)
 			r = r.WithContext(ctx)
-			h.SpacesHandler.Handler(s).ServeHTTP(w, r)
+			h.SpacesHandler.Handler(s, h.TrashbinHandler).ServeHTTP(w, r)
 		case "public-files":
 			base := path.Join(ctx.Value(net.CtxKeyBaseURI).(string), "public-files")
 			ctx = context.WithValue(ctx, net.CtxKeyBaseURI, base)
@@ -203,9 +206,9 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 			case err != nil:
 				w.WriteHeader(http.StatusInternalServerError)
 				return
-			case res.Status.Code == rpcv1beta1.Code_CODE_PERMISSION_DENIED:
+			case res.Status.Code == rpc.Code_CODE_PERMISSION_DENIED:
 				fallthrough
-			case res.Status.Code == rpcv1beta1.Code_CODE_UNAUTHENTICATED:
+			case res.Status.Code == rpc.Code_CODE_UNAUTHENTICATED:
 				w.WriteHeader(http.StatusUnauthorized)
 				if hasValidBasicAuthHeader {
 					b, err := errors.Marshal(http.StatusUnauthorized, "Username or password was incorrect", "")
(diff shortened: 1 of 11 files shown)


JanAckermann:
:heart_eyes:

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -39,6 +38,10 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+const (
+	_trashbinPath = "trash-bin"
```
kobergj, line 42: Curious: could you explain (or point me to where is explained) when to use underscores as constant prefixes and when not?

internal/http/services/owncloud/ocdav/move.go
```diff
@@ -99,7 +100,9 @@ func (s *svc) handleSpacesMove(w http.ResponseWriter, r *http.Request, srcSpaceI
 	ctx, span := rtrace.Provider.Tracer("ocdav").Start(r.Context(), "spaces_move")
 	defer span.End()
 
-	dst, err := extractDestination(r)
+	dh := r.Header.Get(net.HeaderDestination)
+	baseURI := r.Context().Value(net.CtxKeyBaseURI).(string)
+	dst, err := net.ParseDestination(baseURI, dh)
```
kobergj, line 105: Is there a reason why `extractDestination` was removed? I mean the three lines above could also be in that function, then no code changes would be needed here.

internal/http/services/owncloud/ocdav/net/net.go
```diff
@@ -110,6 +118,41 @@ func ParseDepth(s string) (Depth, error) {
 	case DepthInfinity.String():
 		return DepthInfinity, nil
 	default:
-		return "", fmt.Errorf("invalid depth: %s", s)
+		return "", errors.Wrapf(ErrInvalidHeaderValue, "invalid depth: %s", s)
+	}
+}
+
+// ParseOverwrite parses the overwrite header value defined in https://datatracker.ietf.org/doc/html/rfc4918#section-10.6
+// Valid values are "T" and "F". An empty string will be parse to true.
```
kobergj, line 126: Not quite true. `"f"` and `"t"` are also valid. 

I wonder if we should use `strconv.ParseBool` (then `true` would also be valid)?

internal/http/services/owncloud/ocdav/spaces.go
```diff
@@ -135,3 +144,56 @@ func (h *SpacesHandler) Handler(s *svc) http.Handler {
...
+		log.Debug().Str("key", key).Str("path", r.URL.Path).Str("dst", dst).Msg("spaces restore")
+
+		dstRef := ref
+		dstRef.Path = utils.MakeRelativePath(dst)
+
+		trashbinHandler.restore(w, r, s, ref, dstRef, key, r.URL.Path)
+	case http.MethodDelete:
+		trashbinHandler.delete(w, r, s, ref, key, r.URL.Path)
+	default:
+		http.Error(w, "501 Not implemented", http.StatusNotImplemented)
```
kobergj, line 197: Maybe `405` would be better?

internal/http/services/owncloud/ocdav/spaces.go
```diff
@@ -135,3 +144,56 @@ func (h *SpacesHandler) Handler(s *svc) http.Handler {
...
+
+	var key string
+	key, r.URL.Path = router.ShiftPath(r.URL.Path)
+
+	switch r.Method {
+	case MethodPropfind:
+		trashbinHandler.listTrashbin(w, r, s, ref, path.Join(_trashbinPath, spaceID), key, r.URL.Path)
+	case MethodMove:
+		if key == "" {
+			http.Error(w, "501 Not implemented", http.StatusNotImplemented)
```
kobergj, line 175: Curious: Why `501`? Isn't it more like a `Bad Request`?

internal/http/services/owncloud/ocdav/trashbin.go
```diff
@@ -290,11 +276,11 @@ func (h *TrashbinHandler) listTrashbin(w http.ResponseWriter, r *http.Request, s
 	}
 
 	// TODO when using space based requests we should be able to get rid of this path unprefixing
-	for i := range items {
-		items[i].Ref.Path = strings.TrimPrefix(items[i].Ref.Path, basePath)
-	}
+	// for i := range items {
+	// 	items[i].Ref.Path = strings.TrimPrefix(items[i].Ref.Path, basePath)
+	// }
```
kobergj, line 281: No longer needed? Can we remove it then?

kobergj (changes requested):

internal/http/services/owncloud/ocdav/net/net.go
C0rby: > I wonder if we should use strconv.ParseBool (then true would also be valid)?
No, I would stick to the defined values from the RFC, I'll change the code to not include the upper casing.

internal/http/services/owncloud/ocdav/move.go
C0rby: `extractDestination` was moved to `net.ParseDestination`. It just doesn't get the value from the header itself because this way the parse function is more generic and you can have different baseURIs.

internal/http/services/owncloud/ocdav/dav.go
C0rby: I like to prefix unexported package constants with an underscore. That way you immediately see that it's not a local variable. And you don't accidentally shadow those variables.

internal/http/services/owncloud/ocdav/spaces.go
C0rby: Maybe but it was already there and is probably checked by the integration tests so I won't change it.

internal/http/services/owncloud/ocdav/spaces.go
C0rby: Maybe but it was already there and is probably checked by the integration tests so I won't change it.

internal/http/services/owncloud/ocdav/dav.go
C0rby: I like the uber style guide for example: https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_

kobergj: Cool. Thanky!

kobergj (approved):

PR #2636, opened 2022-03-15
Delay reconnect log

kobergj:
When services are connecting to nats server but don't succeed they print a message like 
```
can't connect to nats (stan) server, retrying in 676.400829ms
```

This is confusing for the end user as it is not an error if this message only appears once. For avoiding the confusion we will only log this when the reconnect is bigger than a second already

### DIFF
--- a/pkg/events/server/nats.go
+++ b/pkg/events/server/nats.go
@@ -19,11 +19,11 @@
 package server
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/asim/go-micro/plugins/events/nats/v4"
 	"github.com/cenkalti/backoff"
+	"github.com/cs3org/reva/v2/pkg/logger"
 	"go-micro.dev/v4/events"
 
 	stanServer "github.com/nats-io/nats-streaming-server/server"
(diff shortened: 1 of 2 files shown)


pkg/events/server/nats.go
```diff
@@ -46,10 +47,11 @@ func NewNatsStream(opts ...nats.Option) (events.Stream, error) {
 	b := backoff.NewExponentialBackOff()
 	var stream events.Stream
 	o := func() error {
+		n := b.NextBackOff()
 		s, err := nats.NewStream(opts...)
-		if err != nil {
+		if err != nil && n > time.Second {
 			// TODO: should we get the standard logger here? if yes: How?
-			fmt.Printf("can't connect to nats (stan) server, retrying in %s\n", b.NextBackOff())
+			fmt.Printf("can't connect to nats (stan) server, retrying in %s\n", n)
```
rhafer, line 54: In order to address the above TODO item couldn't we just use 

```
logger.New().Error().Err(err).Msg("can't connect to nats (stan) server, retrying")
```

here? (https://github.com/cs3org/reva/blob/master/pkg/logger/logger.go#L48)
It's not exactly nice, but would at least be a bit cleaner than just doing fmt.Printf here.

kobergj: Not perfect. But much better than before :+1:

rhafer (approved):

PR #2638, opened 2022-03-15
[tests-only] [full-ci] Bump core commit id 20220315

phil-davis:
Issue https://github.com/owncloud/QA/issues/731

I added the new "product" data item into the capabilities, version and status responses. That will keep those compatible with the new expected responses.


individual-it (approved):

individual-it (approved):

phil-davis:
@kulmann I added the "product" data items to the various capabilities version, core, status responses here - hardcoded in the same way as the other existing values. I also enhanced the test code in core so that can request a capabilities response without relying on the "admin" account. The test scenarios for capabilities (status and version sections) and for the status.php end-point now pass here in reva.

So you can make the version, product etc responses more flexible (e.g. to report a "real" version number...) and there are test scenarios that should keep passing. The tests check that values in the capabilities response match with values in the status.php response, that major/minor/micro version numbers are actual numbers, and that the version string correctly starts with the numbers from major.minor.micro. So if new flexible code changes version numbers, product name etc, then the tests will keep passing as long as all the endpoints report consistent values.

PR #2643, opened 2022-03-17
emit linkaccessfailed event when share is nil

butonic:
The code no longer panics when a link access failed event has no share.


kobergj (approved):
:+1: Good find, Thanks!

PR #2644, opened 2022-03-17
Implement a publicshare manager using the metadata storage backend

aduffeck:
This pr adds a `cs3` publicshare manager which uses the cs3 metadata storage backend for persisting the share information.


pkg/publicshare/manager/cs3/cs3.go
```diff
@@ -0,0 +1,401 @@
...
+	return &Manager{
+		storage:          storage,
+		indexer:          indexer,
+		passwordHashCost: passwordHashCost,
+		initialized:      false,
+	}, nil
+}
+
+func (m *Manager) initialize() error {
+	if m.initialized {
```
kobergj, line 102: Would it make sense to lock? Then we do not need to make unnecessary calls to `Init`, `MakeDirIfNotExists` and `AddIndex`

pkg/publicshare/manager/cs3/cs3.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// CreatePublicShare creates a new public share
+func (m *Manager) CreatePublicShare(ctx context.Context, u *user.User, ri *provider.ResourceInfo, g *link.Grant) (*link.PublicShare, error) {
+	if err := m.initialize(); err != nil {
+		return nil, err
+	}
+
+	id := &link.PublicShareId{
+		OpaqueId: utils.RandString(15),
```
kobergj, line 135: Why can't this be a `uuid`?

pkg/publicshare/manager/cs3/cs3.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	tkn := utils.RandString(15)
+	now := time.Now().UnixNano()
+
+	displayName := tkn
+	if ri.ArbitraryMetadata != nil {
+		metadataName, ok := ri.ArbitraryMetadata.Metadata["name"]
+		if !ok {
```
kobergj, line 144: `if !ok`? So we would overwrite `displayName` with an empty string?

pkg/publicshare/manager/cs3/cs3.go
```diff
@@ -0,0 +1,401 @@
...
+			displayName = metadataName
+		}
+	}
+
+	var passwordProtected bool
+	password := g.Password
+	if len(password) > 0 {
```
kobergj, line 151: Use `password != ""` instead `len(password) > 0`

It makes it clearer that we are working with a string in this case

pkg/publicshare/manager/cs3/cs3.go
```diff
@@ -0,0 +1,401 @@
...
+		return nil, errtypes.BadRequest("neither id nor token given")
+	}
+}
+
+func (m *Manager) getByID(ctx context.Context, id string) (*PublicShareWithPassword, error) {
+	tokens, err := m.indexer.FindBy(&link.PublicShare{}, "Id.OpaqueId", id)
+	if err != nil {
+		return nil, err
+	}
+	if len(tokens) == 0 {
```
kobergj, line 264: Should we consider the case `len(tokens) > 1`? I could imagine debugging will be hard in that case.

kobergj (changes requested):

pkg/publicshare/manager/cs3/cs3.go
aduffeck: Hm, the index being used is a "unique" index so it'll never return more than one tokens. And *if* there were more than one they would (hopefully!) be identical in which case it would be fine to just pick one. I would vote for not bothering in this case :)

pkg/publicshare/manager/cs3/cs3.go
aduffeck: good catch!

pkg/publicshare/manager/cs3/cs3.go
aduffeck: Hm, I'm not sure, I just made it consistent with what the `json` manager does. If you think uuids would be better we should maybe think about changing all managers at once.

pkg/publicshare/manager/cs3/cs3.go
kobergj: Ok. Fine for me :+1:

pkg/publicshare/manager/cs3/cs3.go
kobergj: Yes fine. Just wanted to know if you have insights :smile:

kobergj (approved):

PR #2657, opened 2022-03-21
add a spaceid to the propfind response

C0rby:
Added the spaceid to propfind responses so that clients have the necessary data to send subsequent requests.
See: https://github.com/owncloud/ocis/issues/3345

### DIFF
--- a/internal/http/services/owncloud/ocdav/trashbin.go
+++ b/internal/http/services/owncloud/ocdav/trashbin.go
@@ -191,7 +191,7 @@ func (h *TrashbinHandler) listTrashbin(w http.ResponseWriter, r *http.Request, s
 	}
 
 	if depth == net.DepthZero {
-		propRes, err := h.formatTrashPropfind(ctx, s, refBase, nil, nil)
+		propRes, err := h.formatTrashPropfind(ctx, s, ref.ResourceId.StorageId, refBase, nil, nil)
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocdav/trashbin.go
```diff
@@ -355,29 +355,31 @@ func (h *TrashbinHandler) itemToPropResponse(ctx context.Context, s *svc, refBas
 
 	t := utils.TSToTime(item.DeletionTime).UTC()
 	dTime := t.Format(time.RFC1123Z)
+	size := strconv.FormatUint(item.Size, 10)
 
 	// when allprops has been requested
 	if pf.Allprop != nil {
 		// return all known properties
-		response.Propstat = append(response.Propstat, propfind.PropstatXML{
+		propstatOK := propfind.PropstatXML{
```
micbar, line 363: :+1:

micbar (approved):
LGTM

kulmann:
What's the reasoning for naming the property `spaceid` instead of `storageid` though? 🤔

C0rby:
:man_shrugging: I implemented what was discussed in the linked issue. The storages in the backend are also called spaces so spaceid is technically correct. Storageid would be more generic though. I don't have a preference for either.

PR #2664, opened 2022-03-22
Grant id in mountpoint

micbar:
# Description

We distinguish between an mountpoint and grants for shares.

### DIFF
--- a/internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
+++ b/internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
@@ -954,13 +954,28 @@ func (s *service) fetchShares(ctx context.Context) ([]*collaboration.ReceivedSha
 
 func findEarliestShare(receivedShares []*collaboration.ReceivedShare, shareMd map[string]utils.ShareMetadata) (earliestShare *collaboration.Share, atLeastOneAccepted bool) {
 	for _, rs := range receivedShares {
+		var hasCurrentMd bool
+		var hasEarliestMd bool
+
 		current := rs.Share
 		if rs.State == collaboration.ShareState_SHARE_STATE_ACCEPTED {
 			atLeastOneAccepted = true
 		}
+
+		// We cannot assume that every share has metadata
+		if current.Id != nil {
+			_, hasCurrentMd = shareMd[current.Id.OpaqueId]
+		}
+		if earliestShare != nil && earliestShare.Id != nil {
+			_, hasEarliestMd = shareMd[earliestShare.Id.OpaqueId]
+		}
+
 		switch {
 		case earliestShare == nil:
 			earliestShare = current
+		// ignore if one of the shares has no metadata
+		case !hasEarliestMd || !hasCurrentMd:
+			continue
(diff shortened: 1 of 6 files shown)


internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
```diff
@@ -687,13 +689,14 @@ func (s *service) Stat(ctx context.Context, req *provider.StatRequest) (*provide
 					OpaqueId:  utils.ShareStorageProviderID,
 				},
 				Type:          provider.ResourceType_RESOURCE_TYPE_CONTAINER,
-				Mtime:         &typesv1beta1.Timestamp{},
+				Mtime:         resourceResp.Info.Mtime,
```
kobergj, line 692: Just to be sure I got it right: We want the `Mtime` of the resource behind the earliest share, NOT the earliest `Mtime` of all resources behind all shares, right?

internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
```diff
@@ -939,3 +942,45 @@ func (s *service) rejectReceivedShare(ctx context.Context, receivedShare *collab
...
+		if entry, ok := lsRes.Opaque.Map["etags"]; ok {
+			// If we can't get the etags thats fine, just continue.
+			_ = json.Unmarshal(entry.Value, &shareEtags)
+		}
+	}
+	return receivedShares, shareEtags, nil
+}
+
+func findEarliestShare(receivedShares []*collaboration.ReceivedShare) (earliestShare *collaboration.Share, atLeastOneAccepted bool) {
+	atLeastOneAccepted = false
```
kobergj, line 969: This line is unnecessary, you have already defined `atLeastOneAccepted` through the named return parameter

kobergj (changes requested):

internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
micbar: that is tricky. Need to think about it again. I want the mtime of the earliest share.

kobergj (approved):

internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
micbar: i changed the implementation. @C0rby FJYI

kobergj (approved):

pkg/utils/utils.go
```diff
@@ -66,6 +66,12 @@ var (
 	errInvalidSpaceReference = errors.New("invalid storage space reference")
 )
 
+// ShareMetadata contains Metadata for a share
+type ShareMetadata struct {
```
C0rby, line 70: Could you move this to the `share` package and then rename it to just `Metadata`?

C0rby (approved):
Just one small thing. Apart from that it looks fine.

pkg/utils/utils.go
micbar: :+1:

PR #2665, opened 2022-03-22
Micro ocdav

butonic:
This PR adds a new `pkg/micro` package that exposes the internal ocdav http handler as a go micro service. For testing `pkg/micro/ocdav/cmd/main.go` can start the service standalone.
- [x] add auth, ctx, log, middlewares
- [x] add tracing middleware
- [ ] add more config options


kobergj:
@wkloucek could you check tracing commit?

pkg/micro/ocdav/service.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+
+	// log
+	lm := revaLogMiddleware.New()
+
+	// tracing
+	tm := func(h http.Handler) http.Handler { return h }
+	if sopts.TracingEnabled {
+		tm = traceHandler("ocdav", sopts.TracingCollector, sopts.TracingEndpoint)
```
butonic, line 149: While this differs from the way rhttp (it wraps all middlewares), I think adding one trace middleware is the right approach. If other middlewares should be traced they should add their own span.

micbar (approved):
Code looks good IMO. It only adds a new way of starting the ocdav service whithout changing existing behavior.

PR #2668, opened 2022-03-24
Minor cleanup

butonic:
- The `chunk_folder` config option is unused
- Prevent a panic when looking up spaces

### DIFF
--- a/internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
+++ b/internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
@@ -105,7 +105,13 @@ func LookUpStorageSpacesForPathWithChildren(ctx context.Context, client gateway.
 	}
 
 	lSSRes, err := client.ListStorageSpaces(ctx, lSSReq)
-	if err != nil || lSSRes.Status.Code != rpc.Code_CODE_OK {
+	if err != nil {
+		return nil, nil, err
+	}
+	if lSSRes == nil {
+		return nil, &rpc.Status{Code: rpc.Code_CODE_INTERNAL, Message: "no error but result was nil"}, nil
+	}
+	if lSSRes.Status.Code != rpc.Code_CODE_OK {
 		return nil, lSSRes.Status, err
 	}
 
--- a/internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
+++ b/internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
@@ -111,7 +111,7 @@ func LookUpStorageSpacesForPathWithChildren(ctx context.Context, client gateway.
 	if lSSRes == nil {
 		return nil, &rpc.Status{Code: rpc.Code_CODE_INTERNAL, Message: "no error but result was nil"}, nil
 	}
-	if lSSRes.Status.Code != rpc.Code_CODE_OK {
+	if lSSRes.Status.GetCode() != rpc.Code_CODE_OK {
 		return nil, lSSRes.Status, err
 	}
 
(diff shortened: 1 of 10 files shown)


internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
```diff
@@ -105,7 +105,13 @@ func LookUpStorageSpacesForPathWithChildren(ctx context.Context, client gateway.
 	}
 
 	lSSRes, err := client.ListStorageSpaces(ctx, lSSReq)
-	if err != nil || lSSRes.Status.Code != rpc.Code_CODE_OK {
+	if err != nil {
+		return nil, nil, err
+	}
+	if lSSRes == nil {
+		return nil, &rpc.Status{Code: rpc.Code_CODE_INTERNAL, Message: "no error but result was nil"}, nil
```
kobergj, line 112: This should NEVER happen. If we need to check it, please log it with highest priority so we are aware of it and can fix it.

kobergj (changes requested):

internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
butonic: when `err != nil` then `lSSRes.Status` will nill because `lSSRes` might be nil ;-)

butonic: it will all go away when we have a service registry ... for it is good enough to return the errors properly.

C0rby: > when `err != nil` then `lSSRes.Status` will nill because `lSSRes` might be nil ;-)

But if `err != nil` then `lSSRes.Status.Code != rpc.Code_CODE_OK` would never be executed.
If `err == nil` and `lSSRes == nil` too then something in `ListStorageSpaces` is wrong.

kobergj: That's exactly what I meant. In my opinion if a handler returns an `err == nil` then the result MUST never be nil so I can check the status code.

aduffeck: I agree that `lssRes` should never be nil when `err == nil` but the `Status` field could still be nil which would make the `lssRes.Status.Code` access fail, , maybe that's what @butonic was seeing?
We can't expect `lssRes.Status` to never be nil in the no-error case I think, so how about changing the check to `if lssRes.Status == nil`?

kobergj: Actually not needed. `lssRes.Status.GetCode` will return `rpc.Code_CODE_INVALID` even if `lssRes.Status` is `nil`.

aduffeck: That's even better :) :+1:

butonic: @kobergj could you suggest a code change?

internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
```diff
@@ -105,7 +105,13 @@ func LookUpStorageSpacesForPathWithChildren(ctx context.Context, client gateway.
...
+	}
+	if lSSRes.Status.Code != rpc.Code_CODE_OK {
```
kobergj, line 114:
```suggestion
	if lSSRes.Status.GetCode() != rpc.Code_CODE_OK {
```

internal/http/services/owncloud/ocdav/spacelookup/spacelookup.go
kobergj: @C0rby @aduffeck I removed the line. Can some of you approve this? My approve did go away because I was pushing...

aduffeck (approved):

aduffeck:
lgtm

kobergj (approved):

PR #2684, opened 2022-03-28
remove ioutil and use io or os appropriately

nilsocket:
package `ioutil` is deprecated from `1.16` and same functionality is implemented in across packages `os` and `io`.

This commit replaces `ioutil` with `io` and `os` packages.
Simple find and replace.

### DIFF
--- a/cmd/reva/common.go
+++ b/cmd/reva/common.go
@@ -21,7 +21,7 @@ package main
 import (
 	"bufio"
 	"encoding/json"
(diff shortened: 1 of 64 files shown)


labkode:
@nilsocket can you please create a changelog? Thanks for cleaning the code.

nilsocket:
Will resubmit, their were some changes which need to be made for `os.ReadDir()`

PR #2687, opened 2022-03-28
[full-ci] Allow no permissions on Links

kobergj:
Adds an endpoint to ocs service that exposes information  about link tokens.
Usage:

curling the `unprotected` endpoint returns standard information
```
curl --insecure -X GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/unprotected/jqDCLhaiyIvXZTp

<TokenInfo>
  <token>jqDCLhaiyIvXZTp</token>
  <linkurl>/s/jqDCLhaiyIvXZTp</linkurl>
  <passwordprotected>false</passwordprotected>
  <storageid>f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c</storageid>
  <opaqueid>c6378f93-0e2d-4f77-812b-a1b2b3ee04ae</opaqueid>
  <path/>
  <spacePath/>
  <spaceAlias/>
  <spaceURL/>
</TokenInfo>
```

curling a password protected link only returns minimal information
```
curl --insecure -X GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/unprotected/jqDCLhaiyIvXZTp

<TokenInfo>
  <token>jqDCLhaiyIvXZTp</token>
  <linkurl>/s/jqDCLhaiyIvXZTp</linkurl>
  <passwordprotected>true</passwordprotected>
  <storageid/>
  <opaqueid/>
  <path/>
  <spacePath/>
  <spaceAlias/>
  <spaceURL/>
</TokenInfo>
```

curling the `protected` endpoint returns full information if user has native access
```
curl --insecure -X GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/protected/jqDCLhaiyIvXZTp -u marie:radioactivity

<TokenInfo>
  <token>jqDCLhaiyIvXZTp</token>
  <linkurl>/s/jqDCLhaiyIvXZTp</linkurl>
  <passwordprotected>false</passwordprotected>
  <storageid>f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c</storageid>
  <opaqueid>c6378f93-0e2d-4f77-812b-a1b2b3ee04ae</opaqueid>
  <path/>
  <spacePath>/users/f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c</spacePath>
  <spaceAlias>personal/marie</spaceAlias>
  <spaceURL>personal/marie/c6378f93-0e2d-4f77-812b-a1b2b3ee04ae</spaceURL>
</TokenInfo>
```

curling the `protected` endpoint returns same information as `unprotected` endpoint if user does not have native access
```
curl --insecure -X GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/protected/jqDCLhaiyIvXZTp -u einstein:relativity

<TokenInfo>
  <token>jqDCLhaiyIvXZTp</token>
  <linkurl>/s/jqDCLhaiyIvXZTp</linkurl>
  <passwordprotected>false</passwordprotected>
  <storageid>f7fbf8c8-139b-4376-b307-cf0a8c2d0d9c</storageid>
  <opaqueid>c6378f93-0e2d-4f77-812b-a1b2b3ee04ae</opaqueid>
  <path/>
  <spacePath/>
  <spaceAlias/>
  <spaceURL/>
</TokenInfo>
```

curling the `protected` endpoint without authentication returns `401 Unauthorized`
```
curl --insecure -X GET https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/tokeninfo/protected/jqDCLhaiyIvXZTp

HTTP/1.1 401 Unauthorized
```

### DIFF
--- a/internal/http/services/owncloud/ocdav/token.go
+++ b/internal/http/services/owncloud/ocdav/token.go
@@ -14,12 +14,12 @@ import (
 	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
-	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
-	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/net"
+	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/spacelookup"
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/v2/pkg/rhttp/router"
+	"github.com/cs3org/reva/v2/pkg/utils"
 	"google.golang.org/grpc/metadata"
 )
 
@@ -29,7 +29,6 @@ type TokenHandler struct{}
 // Handler handles http requests
 func (t TokenHandler) Handler(s *svc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		fmt.Println("TOKEN Handler")
 		if r.Method != http.MethodGet {
 			w.WriteHeader(http.StatusMethodNotAllowed)
 			return
@@ -55,13 +54,20 @@ func (t TokenHandler) Handler(s *svc) http.Handler {
 
 // TokenInfo contains information about the token
 type TokenInfo struct {
+	// for all callers
 	Token             string `xml:"token"`
 	LinkURL           string `xml:"linkurl"`
 	PasswordProtected bool   `xml:"passwordprotected"`
 
+	// if not password protected
 	StorageID string `xml:"storageid"`
 	OpaqueID  string `xml:"opaqueid"`
 	Path      string `xml:"path"`
+
+	// if native access
+	SpacePath  string `xml:"spacePath"`
+	SpaceAlias string `xml:"spaceAlias"`
+	SpaceURL   string `xml:"spaceURL"`
 }
 
 func (s *svc) handleGetToken(w http.ResponseWriter, r *http.Request, tkn string, c gateway.GatewayAPIClient, protected bool) {
@@ -88,17 +94,14 @@ func (s *svc) handleGetToken(w http.ResponseWriter, r *http.Request, tkn string,
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
-		ref := &provider.Reference{
-			ResourceId: &provider.ResourceId{
-				StorageId: t.StorageID,
-				OpaqueId:  t.OpaqueID,
-			},
-			Path: "",
+		space, status, err := spacelookup.LookUpStorageSpaceByID(ctx, c, t.StorageID)
+		// add info only if user is able to stat
+		if err == nil && status.Code == rpc.Code_CODE_OK {
+			t.SpacePath = utils.ReadPlainFromOpaque(space.Opaque, "path")
+			t.SpaceAlias = utils.ReadPlainFromOpaque(space.Opaque, "spaceAlias")
+			t.SpaceURL = path.Join(t.SpaceAlias, t.OpaqueID, t.Path)
 		}
-		res, err := c.Stat(ctx, &provider.StatRequest{
-			Ref: ref,
-		})
-		fmt.Println("FILE STAT", res, err)
+
 	}
 
 	b, err := xml.Marshal(t)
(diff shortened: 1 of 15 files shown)


internal/http/services/owncloud/ocdav/token.go
```diff
@@ -0,0 +1,171 @@
...
+	t.StorageID = ls.ResourceId.GetStorageId()
+	t.OpaqueID = ls.ResourceId.GetOpaqueId()
+
+	return t, nil
+}
+
+func getInfoForToken(tkn string, q url.Values, c gateway.GatewayAPIClient) (owner *user.User, token string, passwordProtected bool, err error) {
+	ctx := context.Background()
+
+	sig := q.Get("signature")
```
C0rby, line 147: Will there ever be a signature in the URL? Also what about password authentication via basic auth?

kobergj: I just used same logic as with public links. Don't know if this is really needed.

Regarding second question: Even if there was a password authentication that still doesn't mean the user has access to the link. We use the link to determine the resourceID (and some other information) of the underlying resource in all cases.
I didn't see the point in trying to use the users access token, since he can just use the link token to get that information.

C0rby: > Even if there was a password authentication that still doesn't mean the user has access to the link.

I don't mean user authentication but a public link can be password protected.

Let's assume this file tree:
/
-+ Folder1 <- Shared with public link. Token: foobar42, password: secret
--+ Cat.jpg
-+ Folder2

Now if I would like to PROPFIND the share I would have to send:
```
curl -u public:secret -X PROPFIND https://localhost:9200/remote.php/dav/public-files/foobar42

The signature is also only present for password protected public links. It is used when clients want to download a file with a standard html link.
Then the link would look like this: `https://localhost:9200/remote.php/dav/public-files/foobar42/Cat.jpg?signature=<signature>&expiration=<timestamp>`

That is why I think you should rather handle the link password instead of the signature.

Comment (kobergj) [reply internal/http/services/owncloud/ocdav/token.go]: Ah. Now I got it. Thanks for explanation :+1: 
Please check again, we are using `BaicAuth(...)` now.

**Comment by butonic** [comment]
why does it make sense to have two endpoints? when should clients use `/dav/tokeninfo/unprotected/{token}` and when  `/dav/tokeninfo/protected/{token}`? why is a single `/dav/tokeninfo/{token}` endpoint not sufficient?

**Comment by kobergj** [comment]
@butonic Clients should use the `protected` endpoint when they have an authenticated user already. The `unprotected` endpoint is used when there is no user logged in yet.

Of course this could also be done in one endpoint. But then we would need to duplicate authentication logic from the auth interceptor. By using different endpoints the auth middleware can take care of authentication, and future refactorings can be done in auth middleware only.

IMO it was the simplest and stablest approach to use two endpoints and reuse existing code.

File: internal/http/services/owncloud/ocdav/token.go [hunk 2]
```diff
@@ -0,0 +1,181 @@
...
+	// if not password protected
+	StorageID string `xml:"storageid"`
+	OpaqueID  string `xml:"opaqueid"`
+	Path      string `xml:"path"`
+
+	// if native access
+	SpacePath  string `xml:"spacePath"`
+	SpaceAlias string `xml:"spaceAlias"`
+	SpaceURL   string `xml:"spaceURL"`
+}
```
butonic, line 88: hm, sometimes camelcase, sometimes all lowercase. Looking at the existing propfind, try sticking to all lowercase.

butonic: to lowercase `TokenInfo` you can add
```
type TokenInfo struct {
	XMLName  xml.Name  `xml:"tokeninfo"`
```

butonic (changes requested):

internal/http/services/owncloud/ocdav/token.go
kobergj: structs in ocs didn't have that file and are not xml specific, so I didn't add it yet. Recomment if I should.

kobergj: I used camelCase for json and alllowercase for xml. Does that sound correct?

kobergj:
@butonic I moved the handler to ocs. Please recheck!

internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -154,6 +154,24 @@ type ShareeData struct {
...
+
+	// if not password protected
+	StorageID string `json:"storageID" xml:"storageid"`
+	OpaqueID  string `json:"opaqueID" xml:"opaqueid"`
+	Path      string `json:"path" xml:"path"`
+
+	// if native access
+	SpacePath  string `json:"spacePath" xml:"spacepath"`
+	SpaceAlias string `json:"spaceAlias" xml:"spacealias"`
+	SpaceURL   string `json:"spaceURL" xml:"spaceurl"`
```
butonic, line 172: the ocs api uses [snake_case](https://en.wikipedia.org/wiki/Snake_case) check the structs above.

butonic (changes requested):

internal/http/services/owncloud/ocs/conversions/main.go
kobergj: Yeah, and the ones below use camelCase :laughing: But ok, then snake_case it is...

butonic: oh, did I mention that the ocs api has its shortcomings? ;-)

PR #2697, opened 2022-03-30
Public shares improvements

C0rby:
Enhanced the json and cs3 public share manager so that it lists shares in spaces for all members.


butonic (approved):
👍

PR #2708, opened 2022-03-31
 LDAP: rework configuration and move to a common module

rhafer:
This reworks the LDAP configuration of the user- and groups-providers.
As the LDAP groups provider needs to be aware of users (e.g. for
looking up groups members) and the LDAP users provider needs to be aware
of groups (e.g. for looking up groupmembership of a user) this commit
basically unifies the configuration of both providers.
    
Additionally the LDAP configuration is reworked to no longer rely on
templating LDAP filters in the configuration with can be error prone and
confusing.
    
This commit also move all code executing LDAP queries into the common
utlis/ldap module for being able to share more code between both
providers.

Instead of open a new connection with every single requests. We now use
a long-lived connection per service with a simple mechanism to
automatically reconnect in case of failures.


pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,569 @@
...
+		IDIsOctetString: false,
+		Username:        "cn",
+		Mail:            "mail",
+		DisplayName:     "displayName",
+		UIDNumber:       "uidNumber",
+		GIDNumber:       "gidNumber",
+	},
+}
+
+// Default userConfig (Active Directory)
```
kobergj, line 104: typo `groupConfig`

pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,569 @@
...
+			i.User.Schema.Mail,
+			i.User.Schema.Username,
+		},
+		nil,
+	)
+	log.Debug().Str("backend", "ldap").Str("basedn", dn).Str("filter", filter).Int("scope", i.User.scopeVal).Msg("LDAP Search")
+	res, err := lc.Search(searchRequest)
+
+	if err != nil {
+		return nil, errtypes.NotFound(dn)
```
kobergj, line 221: Any `err` means `NotFound`? Do we need to log the real error for debugging?

pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,569 @@
...
+	if err != nil {
+		var errmsg string
+		if lerr, ok := err.(*ldap.Error); ok {
+			if lerr.ResultCode == ldap.LDAPResultSizeLimitExceeded {
+				errmsg = fmt.Sprintf("too many results searching for group '%s'", filter)
+				log.Debug().Str("backend", "ldap").Err(lerr).
+					Str("groupfilter", filter).Msg("too many results searching for group")
+			}
+		}
+		return nil, errtypes.NotFound(errmsg)
```
kobergj, line 336: Same as above. Are all errors considered `NotFound` or do we need to log the real error for debugging?

pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,569 @@
...
+		i.Group.BaseDN,
+		ldap.ScopeWholeSubtree, ldap.NeverDerefAliases, 0, 0, false,
+		i.getGroupFindFilter(query),
+		[]string{i.Group.Schema.ID, i.Group.Schema.Groupname, i.Group.Schema.Mail, i.Group.Schema.DisplayName, i.Group.Schema.GIDNumber},
+		nil,
+	)
+
+	sr, err := lc.Search(searchRequest)
+	if err != nil {
+		return nil, err
```
kobergj, line 358: In other cases `lc.Search` errors were replaced by `NotFound`. Not here?

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,341 @@
...
+// Search implements the ldap.Client interface
+func (c *ConnWithReconnect) Search(sr *ldap.SearchRequest) (*ldap.SearchResult, error) {
+	conn, err := c.getConnection()
+
+	if err != nil {
+		return nil, err
+	}
+
+	var res *ldap.SearchResult
+	for try := 0; try <= c.retries; try++ {
```
kobergj, line 86: 1) Can we move this loop to a dedicated function? It is used by all Methods so generates lots of duplication.

2) Should we do a more sophisticated retry logic? For example `backoff.Exponential` proved valuable in my experience,

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,341 @@
...
+			if l != nil && l == resConn {
+				c.logger.Debug().Msgf("closing connection %v", &l)
+				l.Close()
+			}
+			if l == resConn || l == nil {
+				c.logger.Debug().Msg("reconnecting to LDAP")
+				l, err = c.ldapConnect(config)
+			} else {
+				c.logger.Debug().Msg("already reconnected")
+			}
```
kobergj, line 233: Uh, this `if` collection is making my head burn :smile: How about refactoring it into a switch?
```golang
switch {
case l == nil:
    l, err = c.ldapConnect(config)
case l != resConn:
    continue
default:
   l.Close()
   l, err = c.ldapConnect(config)
}
```

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,341 @@
...
+	var err error
+	var l *ldap.Conn
+	if config.TLSConfig != nil {
+		l, err = ldap.DialURL(config.URI, ldap.DialWithTLSConfig(config.TLSConfig))
+	} else {
+		l, err = ldap.DialURL(config.URI)
+	}
+
+	if err != nil {
+		c.logger.Debug().Err(err).Msg("could not get ldap Connection")
```
kobergj, line 251: `return nil, err` after this line. Then we can omit the `else` case

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// ModifyWithResult implements the ldap.Client interface
+func (c *ConnWithReconnect) ModifyWithResult(m *ldap.ModifyRequest) (*ldap.ModifyResult, error) {
+	conn, err := c.getConnection()
+	if err != nil {
+		return nil, err
+	}
+
+	return conn.ModifyWithResult(m)
```
kobergj, line 325: Curious: Would this work for all Methods? So let's say I want to Implement `SetTimeout` Method. Would I just do:
```golang
conn, err := c.getConnection()
if err != nil {
	return
}

return conn.SetTimeout(t)
```

kobergj (changes requested):

pkg/utils/ldap.go
```diff
@@ -21,30 +21,62 @@ package utils
...
-func GetLDAPConnection(c *LDAPConn) (*ldap.Conn, error) {
-	tlsconfig := &tls.Config{InsecureSkipVerify: c.Insecure}
+// GetLDAPClientWithReconnect initializes a long-lived LDAP connection that
+// automatically reconnects on connection errors. It allows to set TLS options
+// e.g. to add trusted Certificates or disable Certificate verification
+func GetLDAPClientWithReconnect(c *LDAPConn) (ldap.Client, error) {
+	var tlsConf *tls.Config
+	if c.Insecure {
+		tlsConf = &tls.Config{
+			//nolint:gosec // We need the ability to run with "insecure" (dev/testing)
```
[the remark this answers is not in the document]
rhafer: Yeah, it's not really pretty. The alternative would be to use "real" certificates everywhere. The `insecure` flag is something we use in many other place throughout reva (and ocis). I am just happy that we meanwhile killed the places where `insecure` was hardcoded without the possibility to disable it. 

Adding a log message seem like a good idea though. (It a bit difficult in that place though, as the services don't seem to have their logging fully setup). I'll see what I can get.

pkg/utils/ldap/reconnect.go
rhafer: No. This one slipped through from an earlier experiment. It's lacking the retry. But for now we don't even need the `ModifyWithResult` call.

pkg/utils/ldap/reconnect.go
rhafer: done

pkg/utils/ldap/reconnect.go
rhafer: I added a separate retry function now.  But I'd rather move a more sophisticated retry mechanism to a later PR if that's ok

pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,574 @@
...
+	)
+
+	log.Debug().Str("backend", "ldap").Str("basedn", i.Group.BaseDN).Str("filter", filter).Int("scope", i.Group.scopeVal).Msg("LDAP Search")
+	sr, err := lc.Search(searchRequest)
+	if err != nil {
+		log.Debug().Str("backend", "ldap").Err(err).Str("filter", filter).Msg("Error looking up group memberships")
+		return []string{}, err
+	}
+
+	groups := []string{}
```
kobergj, line 280: Better to use make for allocating slices, especially when the capacity is know: https://github.com/uber-go/guide/blob/master/style.md#specifying-slice-capacity

pkg/utils/ldap/identity.go
```diff
@@ -0,0 +1,574 @@
...
+		filter,
+		[]string{i.Group.Schema.ID},
+		nil,
+	)
+
+	log.Debug().Str("backend", "ldap").Str("basedn", i.Group.BaseDN).Str("filter", filter).Int("scope", i.Group.scopeVal).Msg("LDAP Search")
+	sr, err := lc.Search(searchRequest)
+	if err != nil {
+		log.Debug().Str("backend", "ldap").Err(err).Str("filter", filter).Msg("Error looking up group memberships")
+		return []string{}, err
```
kobergj, line 277: Just curious, no change needed: Is there a special reason you prefer `[]string{}` over `nil`?

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,317 @@
...
+		return err
+	}
+
+	for try := 0; try <= retries; try++ {
+		if try > 0 {
+			c.logger.Debug().Msgf("retrying attempt %d", try)
+			conn, err = c.reconnect(conn)
+			if err != nil {
+				// reconnection failed stop this attempt
+				return err
```
kobergj, line 90: That will make it stop retrying if the 2nd reconnect resulted in an error, not? That isn't what we want, is it?

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,317 @@
...
+	BindPassword string
+	TLSConfig    *tls.Config
+}
+
+// NewLDAPWithReconnect Returns a new ConnWithReconnect initialized from config
+func NewLDAPWithReconnect(config Config) *ConnWithReconnect {
+	conn := ConnWithReconnect{
+		conn:    make(chan ldapConnection),
+		reset:   make(chan *ldap.Conn),
+		retries: 1,
```
kobergj, line 64: Shouldn't this be configurable? At least in a dedicated `var` so we can override it on build time?

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,317 @@
...
+	go conn.ldapAutoConnect(config)
+	return &conn
+}
+
+// SetLogger sets the logger for the current instance
+func (c *ConnWithReconnect) SetLogger(logger *zerolog.Logger) {
+	c.logger = logger
+}
+
+func (c *ConnWithReconnect) retry(retries int, fn func(c ldap.Client) error) error {
```
kobergj, line 77: Imho it doesn't make sense to have `retry` take the number of retries as a parameter. `retry` is a method of `ConnWithReconnect` which has a field `retries`. I would expect `c.retry` to use this value.

pkg/utils/ldap/reconnect.go
```diff
@@ -0,0 +1,317 @@
...
+
+	retryErr := c.retry(c.retries, func(c ldap.Client) error {
+		res, err = c.Search(sr)
+		return err
+	})
+
+	if retryErr != nil {
+		return nil, ldap.NewError(ldap.ErrorNetwork, errMaxRetries)
+	}
+	return res, err
```
kobergj, line 118: Wouldn't `err` be always `nil`? If it is not, it would be returned as `retryErr` and the function would never come here.

Same counts for all methods that use the `retry` method. (For sake of simplicity I won't comment them all...)

kobergj (changes requested):
Unfortunately I still have some questions, sorry...

pkg/utils/ldap/identity.go
rhafer: No special reason. Probably mostly related to me still being a go noob in some areas.

pkg/utils/ldap/reconnect.go
rhafer: Yes. For the moment that is ok. I am most interested in covering the case where we need to recover after the server closed the connection (e.g. because it was restarted or some idle timeout kicked in).

We can always add more sophisticated reconnection login later. (waiting with exponential backoff and so forth)

pkg/utils/ldap/reconnect.go
rhafer: Hm, true.

pkg/utils/ldap/reconnect.go
rhafer: Good catch!

pkg/utils/ldap/reconnect.go
kobergj: Ah, I see. That also explains the hardcoded `retries: 1` in the comment below. 

Ok then, scratch both of them. We'll do it when we improve the reconnection logic.

kobergj (approved):
:+1:

wkloucek (approved):

PR #2709, opened 2022-04-01
Decomposed FS precondition failed locking

wkloucek:
We've fixed the return code from permission denied to precondition failed if a
user tries to lock an already locked file.

fixes https://github.com/owncloud/ocis/issues/3434


kobergj (approved):

wkloucek:
@kobergj I had to change the logic to ensure the right return code when a file was already locked. Therefore I again requested a review

kobergj (approved):
Tricky solution :smile: Bit works for me!

PR #2712, opened 2022-04-04
fix: update eos-revad dockerfile

jimil749:
Updates the dockerfile for `revad-eos` to not break.

The compressed size of the image is 2.76GB. 

/cc @SamuAlfageme 

### DIFF
--- a/Dockerfile.revad-eos
+++ b/Dockerfile.revad-eos
@@ -35,9 +35,6 @@ RUN make build-revad-docker && cp /go/src/github/cs3org/reva/cmd/revad/revad /us
 RUN mkdir -p /etc/revad/ && echo "" > /etc/revad/revad.toml
 RUN mkdir -p /usr/local/bin
 
-# COPY --from=builder /go/bin/revad /usr/bin/revad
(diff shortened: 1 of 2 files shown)


jimil749:
Just built and checked the compressed size of the image locally: `2.59GB`
https://hub.docker.com/repository/docker/jimil749/revad-eos-test-2/tags?page=1&ordering=last_updated

SamuAlfageme:
@jimil749 could you give a go the `RUN rm -rf ` the following dirs: 

- [ ] `/go`
- [ ] `/var/lib/rpm/Packages`
- [ ] `/var/log`

... and report back on the size?

jimil749:
I added the following line:

```
RUN rm -rf /go && rm -rf /var/lib/rpm/Packages && rm -rf /var/log
```

in the Dockerfile but the size appears to be the same around ~`2.59GB`. Moreover, I am not able to push the image to dockerhub. I get `unexpected EOF` error

SamuAlfageme:
Let's merge for the time being and fix the size in a follow-up PR

PR #2713, opened 2022-04-04
fix permissions in propfind reponses

C0rby:
Project space members would receive the 'shared' permissions/flag in their propfind responses. This would then look like as if they received the resources through a share which isn't correct. This PR fixes this.

See: https://github.com/owncloud/ocis/issues/3373


micbar (approved):
LGTM

PR #2714, opened 2022-04-05
build: simplify

labkode:


dolmen:
@labkode The second commit should be in a separate pull request because it is not dependant.

dolmen:
I approve the first commit (simplify the Makefile) becasue applying `goimports` must be done by the developer BEFORE committing.

The second commit should not be here.

glpatcern:
(Obsoleted, the first commit was already in)

PR #2715, opened 2022-04-05
Quicklinks

kobergj:
Introduces quicklinks.

Creating a Link with the flag `"quicklink=true"` will create a Quicklink. If another create link request with "quicklink=true" comes in, the server will return the existing link instead of creating a new one.

- Since that functionality is not part of the cs3api we needed to update it.
- Besides `"true"` usual go convertible string values are supported for the quicklink flag: ` 1, t, T, TRUE, true, True` see https://golang.google.cn/pkg/strconv/#ParseBool

Curl example:
```bash
curl --insecure -X POST "https://localhost:9200/ocs/v1.php/apps/files_sharing/api/v1/shares" -u admin:admin -d 'shareType=3&path=/rhino.png&permissions=15&name=BLUBB&quicklink=true'

<ocs>
  <meta>
    <status>ok</status>
    <statuscode>100</statuscode>
    <message>OK</message>
  </meta>
  <data>
    <id>exmHTRXGQCwSOyL</id>
    <share_type>3</share_type>
    <uid_owner>admin</uid_owner>
    <displayname_owner>Admin</displayname_owner>
    <additional_info_owner>admin@example.org</additional_info_owner>
    <permissions>3</permissions>
    <stime>1649167505</stime>
    <parent/>
    <expiration/>
    <token>avEAzCCwPcfNuer</token>
    <uid_file_owner>admin</uid_file_owner>
    <displayname_file_owner>Admin</displayname_file_owner>
    <additional_info_file_owner>admin@example.org</additional_info_file_owner>
    <state>0</state>
    <path>/rhino.png</path>
    <item_type>file</item_type>
    <mimetype>image/png</mimetype>
    <storage_id>shared::/rhino.png</storage_id>
    <storage>0</storage>
    <item_source>ddc2004c-0977-11eb-9d3f-a793888cd0f8!db6f0bc2-dc85-4b5e-9834-621dd76d1e68</item_source>
    <file_source>ddc2004c-0977-11eb-9d3f-a793888cd0f8!db6f0bc2-dc85-4b5e-9834-621dd76d1e68</file_source>
    <file_parent/>
    <file_target>/rhino.png</file_target>
    <share_with_additional_info/>
    <mail_send>0</mail_send>
    <name>BLUBB</name>
    <url>https://localhost:9200/s/avEAzCCwPcfNuer</url>
    <quicklink>true</quicklink>
  </data>
</ocs>

```

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
@@ -38,10 +38,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-// QuicklinkName is the reserved name for a quicklink.
-// Creating (or updating) a link with (to) the same name will be blocked by the server
-const QuicklinkName = "Quicklink"
-
 func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request, statInfo *provider.ResourceInfo) (*link.PublicShare, *ocsError) {
 	ctx := r.Context()
 	log := appctx.GetLogger(ctx)
@@ -64,18 +60,8 @@ func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request, 
 		}
 	}
 
-	// check reserved names
-	linkname := r.FormValue("name")
-	quick, _ := strconv.ParseBool(r.FormValue("quicklink")) // no need to check the error - defaults to zero value!
-	if !quick && linkname == QuicklinkName {
-		return nil, &ocsError{
-			Code:    response.MetaBadRequest.StatusCode,
-			Message: fmt.Sprintf("not allowed to use `%s` as name", linkname),
-		}
-
-	}
-
 	// check if a quicklink should be created
+	quick, _ := strconv.ParseBool(r.FormValue("quicklink")) // no need to check the error - defaults to zero value!
 	if quick {
 		f := []*link.ListPublicSharesRequest_Filter{publicshare.ResourceIDFilter(statInfo.Id)}
 		req := link.ListPublicSharesRequest{Filters: f}
@@ -95,12 +81,10 @@ func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request, 
 		}
 
 		for _, l := range res.GetShare() {
-			if l.DisplayName == QuicklinkName {
+			if l.Quicklink {
 				return l, nil
 			}
 		}
-
-		linkname = QuicklinkName
 	}
 
 	newPermissions, err := permissionFromRequest(r, h)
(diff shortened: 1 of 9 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -37,6 +37,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// QuicklinkName is the reserved name for a quicklink.
+// Creating (or updating) a link with (to) the same name will be blocked by the server
+var QuicklinkName = "Quicklink"
```
C0rby, line 42: You should make a constant out of that.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -59,6 +63,54 @@ func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request,
...
+		return nil, &ocsError{
+			Code:    response.MetaBadRequest.StatusCode,
+			Message: fmt.Sprintf("not allowed to use `%s` as name", linkname),
+		}
+
+	}
+
+	// check if a quicklink should be created
+	if quick {
+		_, f, err := h.addFilters(w, r, &provider.Reference{ResourceId: statInfo.Id})
```
C0rby, line 79: Since you already have the resource id and you are only interested in the public link filters, you can just do this: 
```golang
linkFilters := []*link.ListPublicSharesRequest_Filter{publicshare.ResourceIDFilter(info.Id)}
```

Then you don't do an extra stat request in `addFilters`

C0rby (changes requested):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
kobergj: Good point :+1: will do so

Any feedback on the name itself? `"Internal Quicklink"` could make it less likely for users to accidentally run into the bad request. Or we just use a fixed uuid?
cc @micbar

C0rby: I guess the name needs to be decided by PM.
Also not important for this ticket but in the future the name probable needs to be translatable. :see_no_evil:

C0rby: Is that not the name shown by web?

kobergj: No, in the mockups I saw the quicklink doesn't show a name at all

C0rby (approved):
Looks good. :+1:

PR #2717, opened 2022-04-05
add events definitions for users and groups

C0rby:
This is the preparation for the user and group events for auditing.


rhafer (approved):
lgtm

PR #2744, opened 2022-04-17
`RFC3339` as layout for parsing time 

### LINKED ISSUE #2322: Use colon `:` for timezone layout when parsing dates
The `ParseTimestamp()` method currently uses `2006-01-02T15:04:05Z0700` as layout for parsing dates. This is somewhat inconsistent and suboptimal for oC Web, we'd rather expect `2006-01-02T15:04:05Z07:00` as the layout already uses delimiters like `-` and `:`.

Is there a chance to align that? Otherwise we would need to implement a custom format in oC Web as well.

https://github.com/cs3org/reva/blob/d270f61dc546e894b1cb88863536652b2120a5f1/internal/http/services/owncloud/ocs/conversions/main.go#L294

amal-thundiyil:
The `time.Parse()` method currently uses `2006-01-02T15:04:05Z0700` as layout for parsing dates. This is somewhat inconsistent with oC Web. This PR changes `time.Parse()` to use the `RFC3339` layout for parsing dates.

Fixes #2322

### DIFF
--- a/changelog/unreleased/rfc3339-fix.md
+++ b/changelog/unreleased/rfc3339-fix.md
@@ -1,0 +1,6 @@
+Bugfix: Use RFC3339 for parsing dates
+
+We have used the RFC3339 format for parsing dates to be consistent with oC Web.
+
+https://github.com/cs3org/reva/issues/2322
+https://github.com/cs3org/reva/pull/2744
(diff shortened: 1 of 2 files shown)


labkode (approved):

micbar:
where was that changed in web?

PR #2759, opened 2022-04-20
Made uid/gid claims parsing more robust in OIDC auth provider

glpatcern:


pkg/auth/manager/oidc/oidc.go
```diff
@@ -192,6 +198,16 @@ func (am *mgr) Authenticate(ctx context.Context, clientID, clientSecret string)
 	if claims["email"] == nil {
 		return nil, nil, fmt.Errorf("no \"email\" attribute found in userinfo: maybe the client did not request the oidc \"email\"-scope")
 	}
+	if uid, ok := claims[am.c.UIDClaim].(float64); ok {
+		claims[am.c.UIDClaim] = int64(uid)
+	} else {
+		return nil, nil, fmt.Errorf("malformed or missing uid claim in userinfo: '%v'", claims[am.c.UIDClaim])
```
ishank011, line 204: UID and GID are not essential for users, we shouldn't raise errors here

gmgigi96: I think we could leave the code as before, just type assert to `float64` instead of `int64`

glpatcern: Interesting, but in case the `uid` was missing the previous logic defaulted to `0` and the storage provider would typically fail...

Anyway I can surely drop this check (provided we have a non-zero default? or silently leaving 0?) but we do need to go for the double type assertion. The point is that the additional code checking for user mapping does produce on `int64` uids/gids.

glpatcern: Answering myself, yes definitely a token may come without uid/gid, after all this is how ESCAPE tokens are handled and mapped. But the question becomes shall we log a warning (if not fail?) if after the whole processing the uid/gid is still 0?

gmgigi96 (changes requested):

ishank011 (approved):

PR #2774, opened 2022-04-22
[tests-only] Merge master to edge

ishank011:


ishank011:
@butonic @micbar @wkloucek @C0rby this is green! PLEASE merge it before any other PRs

labkode (approved):

phil-davis (approved):
There are no changes to expected-failures, or to the drone CI pipelines. So the large amount of code, docs, example... changes here have not introduced any obvious regressions (and have not fixed anything either :)

Someone else can review more detail (if that is possible with 10k lines of diff). But from my point-of-view it may as well be merged and then we move on.,..

micbar:
We need to check for regressions with the full ocis CI.

PR in https://github.com/owncloud/ocis/pull/3570

micbar (changes requested):
We have some regressions with public links. Maybe due to middleware changes.

micbar:
We need to investigate https://drone.owncloud.com/owncloud/ocis/10843/22/6

internal/http/services/owncloud/ocs/conversions/main.go
```diff
@@ -302,7 +302,7 @@ func timestampToExpiration(t *types.Timestamp) string {
 
 // ParseTimestamp tries to parses the ocs expiry into a CS3 Timestamp
 func ParseTimestamp(timestampString string) (*types.Timestamp, error) {
-	parsedTime, err := time.Parse("2006-01-02T15:04:05Z0700", timestampString)
+	parsedTime, err := time.Parse(time.RFC3339, timestampString)
```
micbar, line 305: This change breaks the current test suite. ownCloud Web is sending an expire date string like `expireDate: 2022-04-29T23:59:59+0200`

micbar: It was introduced in https://github.com/cs3org/reva/pull/2744

BUT afaict it was not yet changed in web. @amal-thundiyil Can you clarify? ownCloud Web sends it in the old format. This breaks the tests in oCIS.

micbar (changes requested):
The issue is the date format change for expire dates. It was only done in the backend, but not yet in ownCloud web. I suggest to revert this one line to make the tests pass. Then do a sychronized change of backend and frontend at the same time.

internal/http/services/owncloud/ocs/conversions/main.go
micbar:
```suggestion
	parsedTime, err := time.Parse("2006-01-02T15:04:05Z0700", timestampString)
```

micbar: We also need to check the behavior with oc10 backend. Web needs to be compatible with both backends.

micbar:
Apart from that, awesome job! 🎉

internal/http/services/owncloud/ocs/conversions/main.go
ishank011: Thanks @micbar! I've pushed the fix

micbar:
I restarted the full-CI in oCIS. Should become green now. I will approve then.

@ishank011 @labkode BUT: One last thing, we should not "Squash-merge" this PR, I would like to merge all the commits as they are. Can you change the github config to make that possible?

ishank011:
@micbar makes sense. I've enabled thay option

micbar (approved):
oCIS Full test suite is green

Issue #2216, opened 2021-10-28
Some "insecure" options are hardcoded to true


wkloucek:
## Describe the bug
Some REVA "insecure" configurations default to true.

## Expected behavior
"insecure" options (most of the time it's SSL cert validation check skip) must be configurable and default to false.

## Actual behavior
some "insecure" options default to true, eg.

- https://github.com/cs3org/reva/blob/a133b602f6259335f1a799434d15b712d960800d/cmd/reva/main.go#L114
- https://github.com/cs3org/reva/blob/8485d4046dbc4f8b4395d82eea78129b928833bb/cmd/reva/download.go#L108
- https://github.com/cs3org/reva/blob/a8c61401b662d8e09175416c0556da8ef3ba8ed6/pkg/cbox/utils/tokenmanagement.go#L67

GRPC connections use also the insecure option (https://grpc.io/docs/guides/auth/#go)
- https://github.com/cs3org/reva/blob/63cd96849ff22995752d393c357854dcd503d0b5/pkg/rgrpc/todo/pool/pool.go#L87-L90

PR #2775, opened 2022-04-23
fix: make hard-coded HTTP "insecure" options configurable

amal-thundiyil:
re #2216

### DIFF
--- a/cmd/reva/download.go
+++ b/cmd/reva/download.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"time"
 
 	"github.com/cheggaaa/pb"
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
@@ -50,7 +49,7 @@ func downloadCommand() *command {
 		remote := cmd.Args()[0]
 		local := cmd.Args()[1]
 
-		client, err := getClient()
+		gatewayClient, err := getClient()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


cmd/reva/main.go
```diff
@@ -114,10 +114,8 @@ func main() {
 	}
 
 	client = rhttp.GetHTTPClient(
-		// TODO make insecure configurable
-		rhttp.Insecure(true),
-		// TODO make timeout configurable
-		rhttp.Timeout(time.Duration(24*int64(time.Hour))),
+		rhttp.Insecure(insecure),
```
ishank011, line 117: This `insecure` flag denotes if the GRPC connection to the gateway should work without TLS or not. I'm not sure if we should use the same one for the data gateway service.

ishank011 (commented):
Hi @amal-thundiyil, I left a couple of comments

cmd/reva/download.go
```diff
@@ -104,10 +104,8 @@ func downloadCommand() *command {
 			httpReq.Header.Set(datagateway.TokenTransportHeader, p.Token)
 			httpClient := rhttp.GetHTTPClient(
```
ishank011, line 105: Let's use the `client` declared in the `main.go` file instead of a new one.

ishank011 (commented):
Hi @amal-thundiyil. One minor comment, we can merge then

cmd/reva/download.go
```diff
@@ -38,7 +37,7 @@ import (
 	"github.com/studio-b12/gowebdav"
 )
 
-func downloadCommand() *command {
+func downloadCommand(httpClient *http.Client) *command {
```
ishank011, line 40: `client` is a global variable, you don't need to pass it as a function argument

amal-thundiyil: Ah, sorry. Fixed it.

PR #2781, opened 2022-04-25
in memory user provider

butonic:
We added an in memory implementation for the user provider that reads the users from the mapstructure passed in.

We will use it to configure a second minimal gateway with a single system user for the metadata storage to keep it completely independent of the normal users.

See https://github.com/owncloud/ocis/issues/3343


pkg/auth/manager/memory/memory.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Credentials holds a pair of secret and userid
+type Credentials struct {
+	ID           *user.UserId    `mapstructure:"id" json:"id"`
+	Username     string          `mapstructure:"username" json:"username"`
+	Mail         string          `mapstructure:"mail" json:"mail"`
+	MailVerified bool            `mapstructure:"mail_verified" json:"mail_verified"`
+	DisplayName  string          `mapstructure:"display_name" json:"display_name"`
+	Secret       string          `mapstructure:"secret" json:"secret"`
```
rhafer, line 51: IMO, it would be nice to uses a password hash here. In order to avoid having to add cleartext password in the config in too many places. OTOH the service that will use this authprovider will need to have the cleartext configured anyways. :man_shrugging:

pkg/auth/manager/memory/memory.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2018-2021 CERN
```
rhafer, line 1: We're in 2022 already since a while :smile:

rhafer (commented):
Looks fine to me overall. I just wonder, didn't we agree to just use the a separate instance of the machine auth provider for auth. That way we would just need the new userprovider. I might be mixing things up though.

pkg/auth/manager/memory/memory.go
C0rby: I also think that this should better be a password hash. 

> OTOH the service that will use this authprovider will need to have the cleartext configured anyways. man_shrugging

Yeah, but it is actually the responsibility of the admin to store sensitive values in a secure way like a vault store or something like that. Or at least that would be the nice way.

pkg/auth/manager/memory/memory.go
butonic: yep, tooling problem

PR #2785, opened 2022-04-27
implement workaround for chi.RegisterMethod

C0rby:
Implemented a workaround for `chi.RegisterMethod` because of a concurrent map read write issue.
This needs to be fixed upstream in go-chi. Basically this map needs to support concurrent access: https://github.com/go-chi/chi/blob/c97bc988430d623a14f50b7019fb40529036a35a/tree.go#L34


butonic (approved):
interesting hack ;-)

rhafer (approved):

PR #2790, opened 2022-04-28
Fake ProviderIDs

kobergj:
Adds providerids to the storageproviders. These have no function so far, but are needed for the static registry


micbar (approved):
Awesome!

PR #2794, opened 2022-04-28
[tests-only] bump core commitid

micbar:


butonic:
we need to update the expected failures ...

micbar:
@phil-davis I see a large change in the whole gherkin. Was that intended? I yes, then we need help ;-)

phil-davis:
@micbar yes, we are improving API test checks, so there are lots of line number changes. I will get it passing...

phil-davis:
see #2796

PR #2799, opened 2022-04-29
add flag to unrestrict spaces listing

C0rby:
Listing spaces now only returns all spaces when the user has the permissions and it was explicitly requested.
The default will only return the spaces the current user has access to.

This is necessary to be able to limit the list in the graph `/me/drives` API.


micbar (approved):
LGTM

PR #2813, opened 2022-05-02
Add custom mimetypes config to the app provider

glpatcern:
Same as for the storage provider, this adds the ability to configure custom mimetypes to ext mappings for the WOPI app provider.

If it needs to be moved "up" at the appprovider level, it can be done at a later stage. For now this only affects the WOPI driver as it is extension-based, whereas the appprovider is mimetype-based in general.


micbar:
@glpatcern Do you need that in master? Or can you merge to edge?

glpatcern:
@micbar I need it in master but I guess it's for both anyway!

ishank011 (approved):
@micbar can someone from @cs3org/owncloud-team create a master to edge merge next week?

PR #2816, opened 2022-05-02
centralize space id util functions

C0rby:
Merged the utility functions regarding space ids or references into one package. Also updated the functions to support the newly added provider id in the spaces id.


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -1116,15 +1116,6 @@ func (s *service) RemoveGrant(ctx context.Context, req *provider.RemoveGrantRequ
...
 
-	ctx = ctxpkg.ContextSetLockID(ctx, req.LockId)
-
-	// check targetType is valid
-	if req.Grant.Grantee.Type == provider.GranteeType_GRANTEE_TYPE_INVALID {
-		return &provider.RemoveGrantResponse{
-			Status: status.NewInvalid(ctx, "grantee type is invalid"),
-		}, nil
-	}
-
```
butonic, line 1127: so ... this is no longer needed?

pkg/storagespace/storagespace.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// FormatReference will format a storage space reference into a string representation.
+// If ref or ref.ResourceId are nil an error will be returned.
+// The function doesn't check if all values are set.
+// The resulting format can be:
+//
+// "storage_id!opaque_id"
+// "storage_id!opaque_id/path"
+// "storage_id/path"
```
butonic, line 127: the `opaque_id` might contain `/`, eg for the localfs, which uses the path ins the fileid. we may have to change localfs to use another form of id ... it already uses sqlite. might as well use an identifier for fileid to path lookup.

I currently only see ̀ FormatReference` used in the share cache ... if so make this code a private function there. I fear people might expect to be able to parse `FormatReference` back, which might not always be the case.

butonic (dismissed):

internal/grpc/services/storageprovider/storageprovider.go
C0rby: Argh, no I didn't intend to delete this. This probably happened during rebase. I had some conflicts and must have accidentally removed this.

pkg/storagespace/storagespace.go
C0rby: I think we have more places in the code where we manually build a storage space reference. Usually like `path.Join(storageid + opaqueid, relPath)`. But I think it would be beneficial to have that logic in a function so that changing the format of the reference string is easy. That is what this function should be for. If we don't need/want it fine. :shrug:

pkg/storagespace/storagespace.go
```diff
@@ -0,0 +1,151 @@
...
+	if len(parts) == 2 {
+		return parts[0], parts[1]
+	}
+	return "", parts[0]
+}
+
+// FormatResourceID converts a ResourceId into the string format.
+// The result format will look like:
+// <storageid>!<opaqueid>
+func FormatResourceID(sid provider.ResourceId) string {
```
kobergj, line 75: Could this take `*provider.ResourceId` to avoid having to dereference the `sid` all the time? We only ever work with `*ResourceId` usually

pkg/storagespace/storagespace.go
```diff
@@ -0,0 +1,151 @@
...
+	ErrInvalidSpaceID = errors.New("invalid storage space id")
+)
+
+// SplitID splits a storage space ID into a provider ID and a node ID.
+// The accepted formats results of the storage space ID and respective results
+// are:
+// <providerid>$<spaceid>!<nodeid> 	-> <providerid>$<spaceid>, <nodeid>
+// <providerid>$<spaceid>			-> <providerid>$<spaceid>, <spaceid>
+// <spaceid>						-> <spaceid>, <spaceid>
+func SplitID(ssid string) (storageid, nodeid string, err error) {
```
kobergj, line 48: This func is pretty complicated
1) I need to read the comment to use it
2) It can unwrap `storageid!opaqueid` but that is not obvious

Imho it would be better to have a func that only does `storageid!opaqueid` -> `storageid, opaqueid` It would be much cleaner what this function is supposed to do.

pkg/storagespace/storagespace.go
butonic: Ok, that makes sense. Maybe add this intention to the doc string?

pkg/storagespace/storagespace.go
C0rby: I really dislike using pointers everywhere. IMO the fact that the ResourceId is a pointer almost everywhere else is a bad design so I'd like to slowly reduce that but if you insist I can change that to use pointers.

pkg/storagespace/storagespace.go
C0rby: The thing is that we have implicitly defined the behavior that if there is no node ID we will take the space ID as the node ID.
Sure we can make this function dumb and just split at "!" but then we have to implement that logic with using the space ID as the node ID multiple times in different places.

pkg/storagespace/storagespace.go
kobergj: I understand. But the question is: Is there a realistic chance that cs3apis change from using pointers to absolute values? If not we will have do dereferencing and referencing all the time. This will lead to a worse performance and ultimately to more bugs.

So long story short: I don't care too much if it is a pointer or a struct, but I would prefer to have it consistent. Afais cs3api won't change at any time to use structs instead of pointers. But if you tell me they will, I will believe you...

pkg/storagespace/storagespace.go
kobergj: This behaviour doesn't sound correct to me in all places. If the nodeID is the spaceID then we are querying the space root. But this should not be default behaviour. A client should clarify that it wants to query the space root and not hope that the server repairs the request for him.

pkg/storagespace/storagespace.go
C0rby: > Is there a realistic chance that cs3apis change from using pointers to absolute values?

I see your point. Probably not, at least not in the short-term.

> If not we will have do dereferencing and referencing all the time. This will lead to a worse performance and ultimately to more bugs.

Performance wise it should be a net zero. Yes, copying the memory for stack allocations do have some costs but the compiler can then better optimize those function call. Regarding bugs, I don't see how that would lead to bugs. Sure it's annoying to have to de-reference at many places but if you forget that your compiler will tell you so that shouldn't be too much of a problem. More an annoyance, I agree.

Just to write down  the result of our ~~offline~~ off-GitHub discussion:
We decided to leave it like that and try to reduce the use of pointers generally. It would also make sense to define a coding style guide we can refer to. Maybe something like: https://github.com/uber-go/guide/blob/master/style.md
But that is not part of this pull request.

pkg/storagespace/storagespace.go
C0rby: In our ~~offline~~ off-GitHub discussion we clarified that this is the current implemented behavior. I.e. it already is like this in edge. We have also documented this and therefor will leave it as it is.

pkg/storagespace/storagespace.go
C0rby: > Maybe add this intention to the doc string?

I don't think we need a comment saying "Use this instead of building your own." Or do you mean something else?

kobergj (approved):

PR #2822, opened 2022-05-03
[tests-only] [full-ci] Test fix trashbin propfind

phil-davis:


internal/http/services/owncloud/ocdav/trashbin.go
```diff
@@ -295,7 +295,7 @@ func (h *TrashbinHandler) formatTrashPropfind(ctx context.Context, s *svc, space
 	responses := make([]*propfind.ResponseXML, 0, len(items)+1)
 	// add trashbin dir . entry
 	responses = append(responses, &propfind.ResponseXML{
-		Href: net.EncodePath(ctx.Value(net.CtxKeyBaseURI).(string) + "/"), // url encode response.Href TODO
+		Href: net.EncodePath(path.Join(ctx.Value(net.CtxKeyBaseURI).(string), refBase) + "/"), // url encode response.Href TODO
```
phil-davis, line 298: Note to future readers. The `+ "/"` is the same functionality as code in `itemToPropResponse`:
```
	if item.Type == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
		ref += "/"
	}
```

The response for containers (collections) has a `/` on the end.

In the here, we are constructing the href for the root of the trash-bin - that is always a container (collection). So there is no need for and `if` condition.

phil-davis:
Need to look at the output of new scenario apiTrashbin/trashbinDelete.feature:327 - that might give some clue about what is going wrong with the trashbin href processing.

phil-davis:
The fails in the last CI run are all due to other core test suite changed suite names and/or feature files that moved. That is OK - unrelated to the purpose of this test PR.

PR #2828, opened 2022-05-06
Adjust permissions for deleting spaces

kobergj:
- Do not allow viewers and editors to delete a space
- Block deleting spaces via dav service (graph should be used)

### DIFF
--- a/internal/http/services/owncloud/ocdav/delete.go
+++ b/internal/http/services/owncloud/ocdav/delete.go
@@ -67,15 +67,6 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
 
 	req := &provider.DeleteRequest{Ref: ref}
 
-	// do not allow deleting spaces via dav endpoint - use graph one instead
-	// we get a relative reference coming from the space root
-	// so if the path is "empty" we a referencing the space
-	if ref.Path == "." {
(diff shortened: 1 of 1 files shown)


micbar (approved):
LGTM

micbar (changes requested):
I think the logic for the space root is not good enough.

internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -67,6 +67,15 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
 
 	req := &provider.DeleteRequest{Ref: ref}
 
+	// do not allow deleting spaces via dav endpoint - use graph one instead
+	// we get a relative reference coming from the space root
+	// so if the path is "empty" we a referencing the space
+	if ref.Path == "." {
```
micbar, line 73: That could also be node further down in the tree. 

IMO a better check would be if storageID and opaqueID are the same.

kobergj: That is only working for decomposedfs. Eos for example is having different IDs for space and opaque. We need are more sophisticated way to determine a space root from other items.

For now it is fine though, the `ref` is created via `makeRelativeReference` which uses the space root and a path from there.

micbar (approved):

PR #2829, opened 2022-05-06
Add a provider id to mountpoint spaces as well

aduffeck:


micbar (changes requested):
@aduffeck Thanks! This is needed for the future. Please do not merge until i have the go from the web team. Could be that we need to pull reva again and avoid having this changes already in the beta1

butonic:
@kulmann @micbar does that mean we can merge this now?

PR #2842, opened 2022-05-09
SDK: Fix download action

Daniel-WWU-IT:
The download action was no longer working in the SDK (used by our testing probes); this PR fixes the underlying issue.

### DIFF
--- a/changelog/unreleased/sdk-download-fix.md
+++ b/changelog/unreleased/sdk-download-fix.md
@@ -1,0 +1,5 @@
+Bugfix: SDK - Fix download action
+
+The download action was no longer working in the SDK (used by our testing probes); this PR fixes the underlying issue.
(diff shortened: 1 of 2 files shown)


micbar:
@Daniel-WWU-IT Please check if you really need this change. oCIS and ownCloud Web are using ID based requests. They use the code from reva `edge` branch.

Daniel-WWU-IT:
> @Daniel-WWU-IT Please check if you really need this change. oCIS and ownCloud Web are using ID based requests. They use the code from reva `edge` branch.

This change is needed and only affects the SDK - which is used by our testing probes (has nothing to do with code checks). The change shouldn't have anything to do with ownCloud or any other integration stuff.

labkode (approved):

Issue #2843, opened 2022-05-09
Spaces registry: path templates don't obey mountpoints


ishank011:

butonic:
umm `/users/[l-z]` does not make sense for a user with username `Alice`, does it? If you mean that a storageprovider with the `mount_point` `/users/[a-k]` **should** return a space path `/users/Alice`, then I agree.

ishank011:
> umm `/users/[l-z]` does not make sense for a user with username `Alice`, does it?

Exactly what I'm saying. If `alice` sends a `ListStorageProviders` request with a `resourceID` filter corresponding to a `personal` storage provider mounted at `/users/[l-z]`, it should return a not found error. Currently it just frames the path depending on the user sending the request irrespective of whether the path would actually exist on the provider or not.

PR #2844, opened 2022-05-09
Make spaces registry path templates obey the mountpoints

ishank011:
Fixes https://github.com/cs3org/reva/issues/2843

### DIFF
--- a/pkg/storage/registry/spaces/spaces.go
+++ b/pkg/storage/registry/spaces/spaces.go
@@ -199,11 +199,7 @@ func (r *registry) GetProvider(ctx context.Context, space *providerpb.StorageSpa
 				if err != nil {
 					continue
 				}
-				match, err := regexp.MatchString(sc.MountPoint, spacePath)
-				if err != nil {
-					continue
-				}
-				if !match {
+				if match, err := regexp.MatchString(sc.MountPoint, spacePath); err != nil || !match {
 					continue
 				}
 			}
@@ -348,15 +344,21 @@ func (r *registry) findProvidersForFilter(ctx context.Context, filters []*provid
 					appctx.GetLogger(ctx).Error().Err(err).Interface("provider", provider).Interface("space", space).Msg("failed to execute template, continuing")
 					continue
 				}
+				if match, err := regexp.MatchString(sc.MountPoint, spacePath); err != nil || !match {
+					continue
+				}
+
 				setPath(space, spacePath)
 				validSpaces = append(validSpaces, space)
 			}
 
-			if err := setSpaces(p, validSpaces); err != nil {
-				appctx.GetLogger(ctx).Debug().Err(err).Interface("provider", provider).Interface("spaces", validSpaces).Msg("marshaling spaces failed, continuing")
-				continue
+			if len(validSpaces) > 0 {
+				if err := setSpaces(p, validSpaces); err != nil {
+					appctx.GetLogger(ctx).Debug().Err(err).Interface("provider", provider).Interface("spaces", validSpaces).Msg("marshaling spaces failed, continuing")
+					continue
+				}
+				providerInfos = append(providerInfos, p)
 			}
-			providerInfos = append(providerInfos, p)
 		}
 	}
 	return providerInfos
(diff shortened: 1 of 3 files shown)


pkg/storage/registry/spaces/spaces_test.go
```diff
@@ -333,24 +333,24 @@ var _ = Describe("Spaces", func() {
 
 			It("filters by id", func() {
 				filters := map[string]string{
-					"storage_id": "barspace",
-					"opaque_id":  "barspace",
+					"storage_id": "foospace",
+					"opaque_id":  "foospace",
```
butonic, line 337: hm, `filters by id` should only test if filtering by id works. changing from `barspace` to `foospace` only affects the tests because a different provider is used. And the username lowercasing is the last change. How is this testing the `mount_point` logic you changed? Should it?

Could you update a more specific test of it it does not exist add one that covers your changes?

That being said, I noticed that the test sets up two `personal` spaces, both owned by alice. That should not be the case: either a filter by type `personal` should list all personal spaces and each is owned by a different user, or a filter by owner returns the storage provider that is responsible for that specific space.

We'll look into that.

pkg/storage/registry/spaces/spaces.go
```diff
@@ -348,15 +344,21 @@ func (r *registry) findProvidersForFilter(ctx context.Context, filters []*provid
...
-				appctx.GetLogger(ctx).Debug().Err(err).Interface("provider", provider).Interface("spaces", validSpaces).Msg("marshaling spaces failed, continuing")
-				continue
+			if len(validSpaces) > 0 {
+				if err := setSpaces(p, validSpaces); err != nil {
+					appctx.GetLogger(ctx).Debug().Err(err).Interface("provider", provider).Interface("spaces", validSpaces).Msg("marshaling spaces failed, continuing")
+					continue
+				}
+				providerInfos = append(providerInfos, p)
 			}
-			providerInfos = append(providerInfos, p)
```
butonic, line 359: There is a code path that tries to fetch only provdier info. Let me see if I can dig that up ...

pkg/storage/registry/spaces/spaces_test.go
ishank011: What I mean to say is a storage provider with the config:
```
"mount_point":   "/users/[l-z]"
```
should not return a space with path `/users/Alice`

To make the test pass, I just changed the resource ID to one of a provider capable of returning such a path. I can add one more test where sending a request with a `barspace` filter should return empty.

Regarding the capitalization, again, with a `mount_point: /users/[a-k]`, you shouldn't return `/users/Alice`

pkg/storage/registry/spaces/spaces.go
ishank011: Doesn't matter. Mount points should always be obeyed.

Assume I have two providers mounted at `/users/[a-k]` and `/users/[l-z]`. When `alice` wants to list personal spaces, she should just get one space located on the first provider. At the moment, she also gets `/users/Alice` mounted on the second one

Can you please look into making the tests pass?

pkg/storage/registry/spaces/spaces_test.go
butonic: The `mount_point` is used for path based access in a sharded directory. It should determine which storage provider should be queried for a given _path_. When querying by id or with any other filter every storage provider needs to be queried (unless we can take into account the storageproviderid, eg. for an id based request).

The list of spaces returend by a storage provider should never be filtered by `mount_point`:
- either the path determined the storage provider, based on the mount_point or
- the id determined the provider and the path is built by the `path_template`

ishank011: But that's just inconsistent. The path built by `path_template` should always be of the form dictated by `mount_point`.

Assume that the directory `/users/[username]` exists in both the sharded providers but according to our sharding config, we only provide a user access to one. Both of this would be returned when the user queries personal spaces.
```
[
  {"localhost:11001": {"/users/alice", ...}},
  {"localhost:11002": {"/users/alice", ...}},
]
```

Config:
```
/users/[a-k]: "localhost:11001",
/users/[l-z]: "localhost:11002",
```

ishank011: So we can add a log, not return this response to the user

ishank011: This would also hamper the case when you want to balance load among users based on their usernames, even though the reva daemons might be connecting to the same storage backend.

butonic:
@ishank011 I set this to draft. Please let us know your thoughts after evaluating current web whch uses the /dav/spaces endpoint and only makes relative requests

PR #2851, opened 2022-05-10
[tests-only] Fix flocks by ignoring not found errors

kobergj:
cc @butonic @dragotin 

"A" solution to https://github.com/owncloud/ocis/issues/3757 Better ones very much welcome


pkg/storage/utils/filelocks/filelocks.go
```diff
@@ -140,8 +141,11 @@ func ReleaseLock(lock *flock.Flock) error {
 
 	err = lock.Unlock()
 	if err == nil {
-		if !lock.Locked() {
+		if !lock.Locked() && !lock.RLocked() {
 			err = os.Remove(n)
+			if err != nil && strings.Contains(err.Error(), "no such file or directory") {
```
butonic, line 146: try `os.IsNotExist()`

kobergj: :+1: Thanks!

C0rby: Actually it should be `errors.Is(err, fs.ErrNotExist)` now because `os.IsNotExist()` is deprecated but this is fine for now.

kobergj: Fixed it quickly, pipeline needed restart anyways...

PR #2852, opened 2022-05-10
fix cbox dep on reva 1.18

butonic:
The cbox package no longer depends on reva 1.18.


pkg/cbox/utils/conversions.go
```diff
@@ -28,7 +28,7 @@ import (
 	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	typespb "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
-	"github.com/cs3org/reva/internal/http/services/owncloud/ocs/conversions"
+	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocs/conversions"
```
butonic, line 31: @ishank011 I think this just slipped in ...

kobergj (approved):

PR #2863, opened 2022-05-12
don't include versions in quota check

C0rby:
Fixed the quota check to not count the quota of previous versions.

This will fix the situation where the quota is exceeded and the user wants to create a new version of an existing file.
See this issue for more information: https://github.com/owncloud/ocis/issues/2829

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -1108,20 +1108,27 @@ func IsSpaceRoot(r *Node) bool {
 }
 
 // CheckQuota checks if both disk space and available quota are sufficient
-var CheckQuota = func(spaceRoot *Node, fileSize uint64) (quotaSufficient bool, err error) {
+// Overwrite must be set to true if the new file replaces the old file e.g.
+// when creating a new file version. In such a case the function will
+// reduce the used bytes by the old file size and then add the new size.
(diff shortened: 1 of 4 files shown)


pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -1108,20 +1108,27 @@ func IsSpaceRoot(r *Node) bool {
...
 		return false, errtypes.InsufficientStorage("disk full")
 	}
 	quotaByte, _ := xattrs.Get(spaceRoot.InternalPath(), xattrs.QuotaAttr)
-	var total uint64
 	if quotaByte == "" {
 		// if quota is not set, it means unlimited
 		return true, nil
 	}
-	total, _ = strconv.ParseUint(quotaByte, 10, 64)
+	total, _ := strconv.ParseUint(quotaByte, 10, 64)
```
dragotin, line 1125: When reading the code, I found `total` a bit of a missleading name. Maybe change to `quotaByte` and the String of quotaByte to `quotaByteStr` ?

dragotin (approved):
LGTM. Nice code style :)

pkg/storage/utils/decomposedfs/node/node.go
C0rby: :+1:

PR #2864, opened 2022-05-12
Share jail fixes

micbar:
# Bugfix: Fix Grant Space IDs 

The opaqueID for a grant space was incorrectly overwritten with the root space id.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,11 @@ test: off
 test-integration: build-ci
 	cd tests/integration && go test -race ./...
 
+# This is needed for osx because this os does not support static linking
+# Use the build target without the static flag
+test-integration-osx: build
+	cd tests/integration && go test -race ./...
+
 litmus-test-old: build
 	cd tests/oc-integration-tests/local && ../../../cmd/revad/revad -c frontend.toml &
 	cd tests/oc-integration-tests/local && ../../../cmd/revad/revad -c gateway.toml &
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -56,6 +56,11 @@ test: off
 test-integration: build-ci
 	cd tests/integration && go test -race ./...
 
+# This is needed for osx because this os does not support static linking
+# Use the build target without the static flag
+test-integration-osx: build
+	cd tests/integration && go test -race ./...
```
fschade, line 62: i would prefer to keep `test-integration` intact and detect the os instead 

```bash
if [[ $OSTYPE == 'darwin'* ]]; then
  ...
fi
```

micbar: as this is used in CI, i would like to keep it separate.

fschade (approved):
LGTM 🚀

pkg/storage/utils/decomposedfs/testhelpers/helpers.go
```diff
@@ -252,10 +257,14 @@ func (t *TestEnv) CreateTestStorageSpace(typ string, quota *providerv1beta1.Quot
...
+	res, err := storagespace.ParseID(id)
+	if err != nil {
+		return nil
+	}
 	return &providerv1beta1.Reference{
 		ResourceId: &providerv1beta1.ResourceId{
-			StorageId: id,
-			OpaqueId:  id,
+			StorageId: res.StorageId,
+			OpaqueId:  res.OpaqueId,
```
butonic, line 267: `ResourceId: res`

butonic (approved):

PR #2869, opened 2022-05-13
fix creating new files with wopi

C0rby:
Fixed a bug in the appprovider which prevented creating new files.
 
 See https://github.com/owncloud/ocis/issues/3505 for more infos.


butonic:
❤️

PR #2870, opened 2022-05-13
Update quota calculation

butonic:
We now render the `free` and `definition` quota properties, taking into account the remaining bytes reported from the storage space and calculating `relative` only when possible.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
+++ b/internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
@@ -19,14 +19,16 @@
 package users
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strconv"
 
-	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
-	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
-	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
-	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
-	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/errors"
+	cs3gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
+	cs3identity "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	cs3rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	cs3storage "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocs/config"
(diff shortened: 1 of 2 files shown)


kobergj (dismissed):

internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
```diff
@@ -96,98 +103,128 @@ type Groups struct {
...
 	}
-	if res.Status.Code != rpc.Code_CODE_OK {
-		errors.HandleErrorStatus(sublog, w, res.Status)
+	if res.Status.Code != cs3rpc.Code_CODE_OK {
 		return
 	}
 
 	if len(res.StorageSpaces) == 0 {
 		sublog.Error().Err(err).Msg("list spaces returned empty list")
-		w.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
butonic: no, if something goes wrong, I want to gracefully degrade the response. The more important bit is the user information. if the quota is not available clients should just hide quota information.

internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
```diff
@@ -69,19 +72,24 @@ func (h *Handler) GetGroups(w http.ResponseWriter, r *http.Request) {
...
+type User struct {
+	Enabled           string `json:"enabled" xml:"enabled"`
+	UserType          string `json:"user-type" xml:"user-type"`
+	UserID            string `json:"id" xml:"id"` // UserID is mapped to the preferred_name attribute in accounts
+	DisplayName       string `json:"display-name" xml:"display-name"`
+	LegacyDisplayName string `json:"displayname" xml:"displayname"`
+	Email             string `json:"email" xml:"email"`
+	Quota             *Quota `json:"quota,omitempty" xml:"quota,omitempty"`
+	UIDNumber         int64  `json:"uidnumber" xml:"uidnumber"`
+	GIDNumber         int64  `json:"gidnumber" xml:"gidnumber"`
```
rhafer, line 92: I think `UIDNumber` and `GIDNumber` should also be optional here. Currently the CS3 API returns them always (with some hardcoded defaults) but in the long run we should change that IMO.

rhafer (changes requested):
Looks ok to me. Just some minor nitpicking.

And tests would be great :speak_no_evil:

internal/http/services/owncloud/ocs/handlers/cloud/users/users.go
```diff
@@ -96,98 +103,128 @@ type Groups struct {
...
 
+	d := &User{
+		Enabled:     "true", // TODO include in response only when admin?
+		DisplayName: user.DisplayName,
+		Email:       user.Mail,
+		UserType:    conversions.UserTypeString(user.Id.Type),
+		Quota:       &Quota{},
+	}
+	// TODO how do we fill lastlogin of a user when another user (with the necessary permissions) looks up the user?
+	// TODO someone needs to fill last-login
```
kobergj, line 140: :rofl: love the comment...

kobergj (approved):

PR #2882, opened 2022-05-18
File uploaded event in dataprovider

aduffeck:
This pr fixes a problem with the `FileUploaded` event which has been emitted too early until now.

The previous code emitted `FileUploaded` after the `InitiateFileUpload` call already, which is too early because the file hasn't actually been uploaded yet at that point and isn't available to be statted yet, for example.

The fix is to emit the event from the dataprovider only when the upload has actually finished.

*Note*: This PR only implements this functionality for the decomposedfs storage driver, other drivers might still need to be adapted.

### DIFF
--- a/internal/http/services/dataprovider/dataprovider.go
+++ b/internal/http/services/dataprovider/dataprovider.go
@@ -23,11 +23,14 @@ import (
 	"net/http"
 
 	"github.com/cs3org/reva/v2/pkg/appctx"
+	"github.com/cs3org/reva/v2/pkg/events"
+	"github.com/cs3org/reva/v2/pkg/events/server"
 	datatxregistry "github.com/cs3org/reva/v2/pkg/rhttp/datatx/manager/registry"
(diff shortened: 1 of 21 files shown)


butonic (approved):
works in combination with https://github.com/owncloud/ocis/pull/3821

internal/http/services/dataprovider/dataprovider.go
```diff
@@ -75,7 +80,16 @@ func New(m map[string]interface{}, log *zerolog.Logger) (global.Service, error)
 		return nil, err
 	}
 
-	dataTXs, err := getDataTXs(conf, fs)
+	var publisher events.Publisher
+
+	if conf.NatsAddress != "" && conf.NatsClusterID != "" {
```
kobergj, line 85: Should It log if it is not configured to use nats? Otherwise admins might not be aware...

aduffeck: Agreed :+1:

Issue #2881, opened 2022-05-17
Edge performance issues


ishank011:
When calling the `GET /ocs/v1.php/apps/files_sharing/api/v1/shares?include_tags=false&state=all&shared_with_me=true` endpoint, each received share is statted 23 times (twice via the `Stat` call, the rest of the times via the `ListStorageSpaces` call), since `ListStorageSpaces` is called every time any type of operation needs to be performed.

Enabling the storageprovider cache in the gateway reduces this to 12 stats for subsequent calls. For an actual heterogenous, distributed system, this makes the platform unusable.

The performance is further worsened by two calls to list all the received shares for EVERY received share in the shares storage provider.

Note that these numbers have been obtained after the change to route requests by mount IDs was merged.

cc @labkode @butonic @micbar @dragotin

ishank011:
One suggestion would be to not call ListStorageSpaces for every lookup. Just verify that the provider ID or mountpoint matches, return that and assume that it’ll handle the actual storage request.

I tried to implement something related to this but since generating the space path requires attributes of the space a reference belongs to, I couldn’t. The gateway storage provider constructs a relative reference using this path and further uses that in the request. IMO this should be delegated to the storage provider itself.

butonic:
cc @kobergj @aduffeck ^

butonic:
We have to differentiate to types of requests as they follow different lookup semantics:
1. relative requests that have a ResourceID and a Path set
2. Path based requests

For 1. we added https://github.com/cs3org/reva/pull/2792 which should use the providerids that are provided in ocis since: https://github.com/owncloud/ocis/pull/3619. ocdav does a LookUpStorageSpaceByID call which contains an ID filter. The gateway storageprovider splits the spaceid into a providerid and a nodeid. The space registries ListProvider then makes a findProvidersForResource call which currently then makes a ListStorageSpaces call on the storageprovider with the matching providerid or a provider without an id set.

I think we mixed responsibilities. An the reva storage registry may not need to do an actual ListStorageSpaces call on the provider as the reva storage registry is only responsible for routing. To route a PROPFIND or MKCOL request the ocdav service does not need to know the etag or mtime of the storage space. The gateway certainly does not need it for routing.

I may have failed to clarify that the graph service, which lists the spaces available to a user really is the only service that needs to be aware of etags and mtimes...

🤔 

[...]

PR #2885, opened 2022-05-20
move stat out of usershareprovider

butonic:
When moving the stat to the shgaresstorageprovider we can at least omit stating unaccepted shares. Should reduce the overall amount of stats when listing shares.

First step related to https://github.com/cs3org/reva/issues/2881 and https://github.com/owncloud/ocis/issues/3839

This optimization can be pushed further to the usershareprovider and the backends by adding a State Filter to CS3 https://github.com/owncloud/ocis/issues/3843


micbar (approved):
LGTM

PR #2888, opened 2022-05-23
Introduce spaces field mask

butonic:
We now use a field mask to select which properties to retrieve when looking up storage spaces. This allows the gateway to only ask for `root` when trying to forward id or path based requests. It allows the spaces storage registry to omit calling ListStorageSpaces on the storage provider.

Related: https://github.com/cs3org/reva/issues/2881


butonic:
there are problems in the ocis test suite I need to check before continuing: https://drone.owncloud.com/owncloud/ocis/12029/37/7

aduffeck (approved):

PR #2896, opened 2022-05-25
stat only after put

butonic:
OCDAV no longer makes a Stat call before making an InitiateFileUpload request. The storageprovider will call GetMD to check if a file exists before initiating the upload and return `"exists"="true"` the opaque data of the InitiateUploadResponse. Storage drivers now should return an Already Exists error when an upload tries to overwrite a directory. 

There is a race condition when two requests try to PUT a file at the same time, both might return a 201 created, but one of them actually creates a new revision. I tried avoiding that by making `FS.Upload()` return a `created` flag. Then the  simple and spaces dataproviders could return 201 / 204 for PUT requests and the ocdav PUT could use that response code. However, an indirection to a new resource eg. when storing on an S3 bucket would always lead to a 201, even for existing files ... and returning a created flag would require the implementation to be able to report if the file existed (the s3 client used currently does not reveal if a file was overwritten or created).

Furthermore, it would be nice to report back more metadata, eg the etag, mtime and fileid. Currently, ocdav still does an additional stat request to do that.

The tus driver for the dataprovider does not need to differentiate between 201 vs 200/204 for PUT requests as it uses PATCH requests returning 204 No Content or POST requests with 201 created on upload URLs which are ephemeral anyway.

related: https://github.com/cs3org/reva/issues/2881

### DIFF
--- a/changelog/unreleased/stat-after-put.md
+++ b/changelog/unreleased/stat-after-put.md
@@ -1,0 +1,5 @@
+Enhancement: stat only after PUT
+
+OCDAV no longer makes a Stat call before making an InitiateFileUpload request. The storageprovider will call GetMD to check if a file exists before initiating the upload and return `"exists"="true"` the opaque data of the InitiateUploadResponse. Storage drivers now should return an Already Exists error when an upload tries to overwrite a directory. 
+
+https://github.com/cs3org/reva/pull/2896
--- a/internal/grpc/services/storageprovider/storageprovider.go
+++ b/internal/grpc/services/storageprovider/storageprovider.go
@@ -324,28 +324,29 @@ func (s *service) InitiateFileUpload(ctx context.Context, req *provider.Initiate
 	}
 
 	metadata := map[string]string{}
-	ifMatch := req.GetIfMatch()
-	if ifMatch != "" {
-		sRes, err := s.Stat(ctx, &provider.StatRequest{Ref: req.Ref})
-		if err != nil {
-			return nil, err
+	info, err := s.storage.GetMD(ctx, req.GetRef(), nil)
+	switch err.(type) {
+	case nil:
+		if info.Type == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
+			return &provider.InitiateFileUploadResponse{
+				Status: status.NewAlreadyExists(ctx, errors.New("cannot overwrite directiory"), "cannot overwrite directiory"),
+			}, nil
 		}
-
-		switch sRes.Status.Code {
-		case rpc.Code_CODE_OK:
-			if sRes.Info.Etag != ifMatch {
+		ifMatch := req.GetIfMatch()
+		if ifMatch != "" {
+			if info.Etag != ifMatch {
 				return &provider.InitiateFileUploadResponse{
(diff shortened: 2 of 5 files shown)


butonic:
from https://drone.cernbox.cern.ch/cs3org/reva/7258/12/6
```
@files_sharing-app-required @issue-ocis-reva-196 @skipOnOcV10 | 407s
-- | --
4663 | Scenario Outline: Modifying a shared file should return correct checksum in the propfind using new DAV path                                                                 # /drone/src/tmp/testrunner/tests/acceptance/features/apiMain/checksums.feature:215 | 407s
4664 | Given the administrator has set the default folder for received shares to "Shares"                                                                                        # OccContext::theAdministratorHasSetTheDefaultFolderForReceivedSharesTo() | 407s
4665 | And auto-accept shares has been disabled                                                                                                                                  # FeatureContext::autoAcceptSharesHasBeenDisabled() | 407s
4666 | And using <dav_version> DAV path                                                                                                                                          # FeatureContext::usingOldOrNewDavPath() | 407s
4667 | And user "Brian" has been created with default attributes and without skeleton files                                                                                      # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles() | 407s
4668 | And user "Alice" has uploaded file "filesForUpload/textfile.txt" to "/myChecksumFile.txt" with checksum "MD5:d70b40f177b14b470d1756a3c12b963a"                            # ChecksumContext::userHasUploadedFileToWithChecksumUsingTheAPI() | 407s
4669 | And user "Alice" has shared file "/myChecksumFile.txt" with user "Brian"                                                                                                  # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi() | 407s
4670 | And user "Brian" has accepted share "/myChecksumFile.txt" offered by user "Alice"                                                                                         # FeatureContext::userHasReactedToShareOfferedBy() | 407s
4671 | When user "Brian" uploads file with checksum "SHA1:ce5582148c6f0c1282335b87df5ed4be4b781399" and content "Some Text" to "/Shares/myChecksumFile.txt" using the WebDAV API # FeatureContext::userUploadsAFileWithChecksumAndContentTo() | 407s
4672 | Then the HTTP status code should be "204"                                                                                                                                 # FeatureContext::thenTheHTTPStatusCodeShouldBe() | 407s
4673 | And as user "Alice" the webdav checksum of "/myChecksumFile.txt" via propfind should match "<checksum>"                                                                   # ChecksumContext::theWebdavChecksumOfViaPropfindShouldMatch() | 407s
4674 |   | 407s
4675 | Examples: | 407s
4676 | \| dav_version \| checksum                                                                                            \| | 407s
4677 | \| new         \| SHA1:ce5582148c6f0c1282335b87df5ed4be4b781399 MD5:56e57920c3c8c727bfe7a5288cdf61c4 ADLER32:1048035a \| | 407s
4678 | HTTP status code 201 is not the expected value 204 | 407s
4679 | Failed asserting that 201 matches expected '204'. | 407s
4680 | \| spaces      \| SHA1:3ee962b839762adb0ad8ba6023a4690be478de6f MD5:d70b40f177b14b470d1756a3c12b963a ADLER32:8ae90960 \| | 408s
4681 | HTTP status code 404 is not the expected value 204 | 408s
4682 | Failed asserting that 404 matches expected '204'.

```

PR #2899, opened 2022-05-27
fix meta requests with no ID

micbar:
# Description

Bugfix for https://github.com/owncloud/ocis/issues/3879


micbar:
@kobergj @butonic I tried to handle it in the existing method, was cleaner to write a new handler func. More duplication but cleaner.

kobergj (approved):
Fine for me. A little copying is better than a little dependency :smile:

PR #2903, opened 2022-05-30
user owncloudsql: use correct user id

butonic:
fixes owncloudsql backed users having the wrong userid when a ownclouduuid property is configured


kobergj (approved):
CI looks flaky, Code looks good,

PR #2909, opened 2022-05-31
decomposedfs: actually check GetPath permission

butonic:
After fixing the meta endpoint and introducing the fieldmask the GetPath call is made directly to the storageprovider. The decomposedfs now checks if the current user actually has the permission to get the path. Before the two previous PRs this was covered by the list storage spaces call which used a stat request and the stat permission. 

I assume we broke edge beclause the two PRs got merged because each on its own did not trigger the bug ... 🤷‍♂️


C0rby (approved):
Awesome!

PR #2910, opened 2022-06-01
[tests-only][full-ci]Update expected to fail file edge

SwikritiT:
Part of :
- https://github.com/owncloud/QA/issues/744
- https://github.com/owncloud/web/issues/7011


phil-davis (approved):
Looks reasonable. It will be good to get this merged now (and similar changes for master #2917 ). And then we can continue going through expected-failures and looking at diffs between reva master, reva edge and oCIS, and document carefully what is the current real issue with each test that fails.

PR #2918, opened 2022-06-02
fix propfinds with depth 0

C0rby:
Fixed the response for propfinds with depth 0. The response now doesn't contain the shares jail anymore.

Fixes: https://github.com/owncloud/ocis/issues/3704

@butonic, can you please double check that this is indeed the fix.


kobergj (approved):
LGTM :+1:

butonic:
This approach omits stating the children and uses the virtual share jail as is. But the virtual share jail does not have a proper etag and mtime ...

well, /dav/spaces/{sharejailid} now need to see all mountpoints as well, which means the share jail now has to calculate etag end mtime for the root anyway. So this might actually be the fix ... but for depth 0 we still need to aggregate the mtimes, size and use the most recent etag...

C0rby:
> So this might actually be the fix ... but for depth 0 we still need to aggregate the mtimes, size and use the most recent etag...

Isn't this happening here? https://github.com/cs3org/reva/pull/2918/files#diff-fa4130be45a2f0a202cc47c8f7ab33722e36ceccb5ba67e881b772e077a59251L448-L462

C0rby:
Oh wait, I still need to test this with subfolders. I only tested it with the root folder.

C0rby:
> Oh wait, I still need to test this with subfolders. I only tested it with the root folder.

This still worked.

PR #2928, opened 2022-06-07
fix mkcol response code

butonic:
We now return the correct response code when an mkcol fails.

found in https://drone.owncloud.com/owncloud/ocis/12322/37/6


dragonchaser (approved):
lgtm

dragonchaser (approved):

rhafer (approved):

PR #2932, opened 2022-06-08
Stat mount points only for accepted shares, configure updating existing shares

ishank011:


ishank011:
Sonarcloud seems to have some sort of problem with me :/

ishank011:
@butonic can you please review and merge this one and https://github.com/cs3org/reva/pull/2929?

PR #2945, opened 2022-06-09
add an executant to all events

C0rby:
Added the executant field to all events.


kobergj (approved):
:+1: Awesome. Thanks for doing this!

PR #2962, opened 2022-06-14
Tracing improvements

rhafer:
Create separate TracerProviders per service to improve tracing when
running multiple services in a single process (like e.g. ocis does).
Previously all service were referring to the same package-level
TracerProvider (which, among other things, caused the services to
overwrite each others services names).

We still create a "global" TracerProvider to cover the cases where
we don't have access to the per service Provider yet. This will be
improved in a follow up change.

### DIFF
--- a/changelog/unreleased/per-service-trace.md
+++ b/changelog/unreleased/per-service-trace.md
@@ -1,0 +1,7 @@
+Enhancement: Per service TracerProvider
+
+To improve tracing we create separate TracerProviders per service now.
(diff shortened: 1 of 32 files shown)


rhafer:
I must admit I am not really sure about the whole change (especially the second commit in the PR). I am not a fan of attaching this kind of stuff to the context. I think ideally we would instantiate a tracer per "sub"-service at startup and utilized that. (That would however require quite a bit of rework I fear)

butonic (approved):

PR #2963, opened 2022-06-14
[tests-only] Upload post processing

kobergj:
Introduces post processing of items after upload

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind/propfind.go
@@ -708,6 +708,14 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 		Propstat: []PropstatXML{},
 	}
 
+	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
+		response.Propstat = append(response.Propstat, PropstatXML{
+			Status: "HTTP/1.1 425 TOO EARLY", // TODO: use proper status code
+			Prop:   []prop.PropertyXML{},
+		})
+		return &response, nil
+	}
+
 	var ls *link.PublicShare
 
 	// -1 indicates uncalculated
--- a/internal/http/services/owncloud/ocdav/propfind/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind/propfind.go
@@ -708,14 +708,6 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 		Propstat: []PropstatXML{},
 	}
 
-	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
-		response.Propstat = append(response.Propstat, PropstatXML{
-			Status: "HTTP/1.1 425 TOO EARLY", // TODO: use proper status code
-			Prop:   []prop.PropertyXML{},
-		})
-		return &response, nil
-	}
-
 	var ls *link.PublicShare
 
 	// -1 indicates uncalculated
@@ -1223,6 +1215,14 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 				}
 			}
 		}
+	}
+
+	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
+		response.Propstat = append(response.Propstat, PropstatXML{
+			Status: "HTTP/1.1 425 TOO EARLY", // TODO: use proper status code
+			Prop:   propstatOK.Prop,
+		})
+		return &response, nil
 	}
 
 	if len(propstatOK.Prop) > 0 {
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -643,7 +643,6 @@ func (n *Node) AsResourceInfo(ctx context.Context, rp *provider.ResourcePermissi
 
 	if n.IsProcessing() {
 		ri.Opaque = utils.AppendPlainToOpaque(ri.Opaque, "status", "processing")
-		return ri, nil
 	}
 
 	if nodeType == provider.ResourceType_RESOURCE_TYPE_CONTAINER {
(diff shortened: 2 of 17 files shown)


pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -641,6 +641,11 @@ func (n *Node) AsResourceInfo(ctx context.Context, rp *provider.ResourcePermissi
 		ParentId:      parentID,
 	}
 
+	if n.IsProcessed() {
+		ri.Opaque = utils.AppendPlainToOpaque(ri.Opaque, "status", "processing")
```
C0rby, line 645: Shouldn't the value be 'processed'?

C0rby (changes requested):

butonic:
related: https://github.com/owncloud/ocis/issues/214#issuecomment-630684815

kobergj:
@butonic should we rename the steps? Not sure if that was a call to action or just an extra info :smile:

internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -708,6 +708,14 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 		Propstat: []PropstatXML{},
 	}
 
+	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
+		response.Propstat = append(response.Propstat, PropstatXML{
+			Status: "HTTP/1.1 425 TOO EARLY", // TODO: use proper status code
```
butonic, line 713: Yeah, 425 is about replay attacks and the `Early-Data` header.

IMO [202 Accepted] would be the correct Status for a resource that is still being uploaded:
> The *202 (Accepted)* status code indicates that the request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon, as it might be disallowed when processing actually takes place. There is no facility in HTTP for re-sending a status code from an asynchronous operation.

But existing desktop clients might try to download files with a 200 status ... @TheOneRing what would desktop clients do with status 202? ignore the file? would be great for now...

pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -0,0 +1,446 @@
...
+	if err = os.Truncate(upload.binPath, 0); err != nil {
+		sublog.Err(err).
+			Msg("Decomposedfs: could not truncate")
+		return err
+	}
+	if err = os.Rename(upload.binPath, targetPath); err != nil {
+		sublog.Error().Err(err).Msg("Decomposedfs: could not rename")
+		return err
+	}
+	// the rename dropped the "processing" status - we need to set it again
```
butonic, line 329: hm, we could copy the metadata from the node to the upload.binPath file, so the rename atomically has all properties set. Copying the metadata from a previous version as in lines 334-350 is brittle ... and might not even work when versioning is somehow disabled. Not necessarily a task for this PR. Please open an issue if you want to tackle this separately.

pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -0,0 +1,446 @@
...
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/xattrs"
+	"github.com/cs3org/reva/v2/pkg/utils"
+	"github.com/cs3org/reva/v2/pkg/utils/postprocessing"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
+	tusd "github.com/tus/tusd/pkg/handler"
+)
+
+// Tree is used to manage a tree hierarchy
+type Tree interface {
```
butonic, line 52: can't you use the `Tree` interface from decomposedfs.go?

butonic:
@dragotin @kobergj I was trying to find an issue where we decide how to implement this, eg the status to return to clients for uploads that are in progress. If and how to return the current progress in the propfind (or only via graph) etc. It is not an ADR either ... did you discuss and design something off the record?

My find-related-issue-fu only came up with https://github.com/owncloud/ocis/issues/214#issuecomment-630684815, which is why I linked it here. Maybe we can clarify there what clients can expect from the server and how we want to handle this?

pkg/storage/utils/decomposedfs/upload/upload.go
kobergj: I'd love to. But `decomposedfs` imports `upload` pkg. So if I import `decomposedfs` in `upload` it won't build any more.

internal/http/services/owncloud/ocdav/propfind/propfind.go
TheOneRing: I'm pretty sure to guess what a 423 will cause.
For everything indicating ok I'd expect a download to be triggered. If you need more info I'd need to dig deeper.

kobergj: As discussed in cross platform meeting we will stick with the 425 status code for now. We can revisit that decision later.

pkg/storage/utils/decomposedfs/upload/upload.go
kobergj: We could use the tree interface from here in decomposedfs. But I don't see the benefit in that...

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -1124,6 +1129,22 @@ func (n *Node) FindStorageSpaceRoot() error {
...
+
+// UnmarkProcessing removes the processing flag from the node
+func (n *Node) UnmarkProcessing() error {
+	return n.RemoveMetadata("user.ocis.nodestatus")
+}
+
+// IsProcessing returns true if the node is currently being processed
+func (n *Node) IsProcessing() bool {
+	v, err := n.GetMetadata("user.ocis.nodestatus")
+	return err == nil && v == "processing"
```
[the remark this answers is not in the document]
kobergj: :+1:

pkg/storage/utils/decomposedfs/upload/upload.go
kobergj: Did it quickly. Should work imo

butonic (approved):

butonic:
cool!

PR #2973, opened 2022-06-15
Fix target uri for webdav prefix for datatx

redblom:
When a webdav prefix is used it appears in both host and name parameter of the target uri for data transfer. This PR fixes that.
The resulting target uri should be similar to:
`http://...token...@reva.eu/prefix/?name=remote.php/webdav/home/...`

### DIFF
--- a/internal/grpc/services/gateway/webdavstorageprovider.go
+++ b/internal/grpc/services/gateway/webdavstorageprovider.go
@@ -238,6 +238,7 @@ func (s *svc) extractEndpointInfo(ctx context.Context, targetURL string) (*webda
 }
 
 func (s *svc) getWebdavEndpoint(ctx context.Context, domain string) (string, error) {
+	fmt.Printf("getWebdavEndpoint( domain = %v )\n", domain)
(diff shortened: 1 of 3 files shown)


internal/grpc/services/gateway/webdavstorageprovider.go
```diff
@@ -238,6 +238,7 @@ func (s *svc) extractEndpointInfo(ctx context.Context, targetURL string) (*webda
 }
 
 func (s *svc) getWebdavEndpoint(ctx context.Context, domain string) (string, error) {
+	fmt.Printf("getWebdavEndpoint( domain = %v )\n", domain)
```
wkloucek, line 241: please make sure to use the logger with an appropriate log level or remove this (development/debug?) statement

michielbdejong:
Apart from @redblom's manual testing I also quickly tested it with the OCM test suite and didn't encounter any problems. 👍

glpatcern:
Excellent, thanks @michielbdejong :+1:

PR #2974, opened 2022-06-15
return early on ocdav move requests with body

C0rby:
Added a check if a ocdav move request contains a body. If it does a 415 415 (Unsupported Media Type) will be returned.

https://datatracker.ietf.org/doc/html/rfc4918#section-8.4


wkloucek (approved):

phil-davis:
Note: I added a core issue to make a couple of API test scenarios for this https://github.com/owncloud/core/issues/40144
We might like to make oC10 behave to-spec for this in the future. And we should also check that COPY and DELETE requests in reva/oCIS have this 415 behavior. The new test will come in a commit-id update in the next week or so, that will clearly show if COPY and DELETE are behaving correctly or not.

PR #2977, opened 2022-06-16
ocdav: skip space lookup on spaces propfind

butonic:
we don't need to look up the space for spaces based requests. we already have it

fixes https://github.com/owncloud/ocis/issues/1277
fixes https://github.com/owncloud/ocis/issues/2144
fixes https://github.com/owncloud/ocis/issues/3073

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind/propfind.go
@@ -298,8 +298,10 @@ func (p *Handler) HandleSpacesPropfind(w http.ResponseWriter, r *http.Request, s
 			if sRes.Status.Code != rpc.Code_CODE_OK {
 				// return not found error so we dont leak existence of a space
 				status = http.StatusNotFound
-				m = "Resource not found" // mimic the oc10 error message
 			}
+		}
+		if status == http.StatusNotFound {
+			m = "Resource not found" // mimic the oc10 error message
 		}
 		w.WriteHeader(status)
 		b, err := errors.Marshal(status, m, "")
--- a/internal/http/services/owncloud/ocdav/propfind/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind/propfind.go
@@ -46,6 +46,7 @@ import (
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/publicshare"
+	rstatus "github.com/cs3org/reva/v2/pkg/rgrpc/status"
 	"github.com/cs3org/reva/v2/pkg/rhttp/router"
 	"github.com/cs3org/reva/v2/pkg/storagespace"
 	"github.com/cs3org/reva/v2/pkg/utils"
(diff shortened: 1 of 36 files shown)


internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -255,35 +248,122 @@ func (p *Handler) HandleSpacesPropfind(w http.ResponseWriter, r *http.Request, s
...
+		status := rstatus.HTTPStatusFromCode(res.Status.Code)
+		if res.Status.Code == rpc.Code_CODE_ABORTED {
+			// aborted is used for etag an lock mismatches, which translates to 412
+			// in case a real Conflict response is needed, the calling code needs to send the header
+			status = http.StatusPreconditionFailed
+		}
+		m := res.Status.Message
+		if res.Status.Code == rpc.Code_CODE_PERMISSION_DENIED {
+			// check if user has access to resource
+			sRes, err := client.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{ResourceId: ref.GetResourceId()}})
```
C0rby, line 291: What is the difference between this stat and the original one above?

butonic: The original stat might return an error, to check if we need to hide that error this stat checks if we are allowed to stat the space (well only the resource id, ignoring any path). If this stat returns OK we have read access and can return other errors like PrecondtitionFailed when an etag does not match.

The more I think about it, the more I think we should move 'hiding the existence of spaces' int the storage drivers. an id based lookup has to find the space first. if a user has no access to it we can always return 404 and stop any further request processing.

But that is a task for a separate PR.

internal/http/services/owncloud/ocdav/report.go
```diff
@@ -110,7 +110,7 @@ func (s *svc) doFilterFiles(w http.ResponseWriter, r *http.Request, ff *reportFi
 			infos = append(infos, statRes.Info)
 		}
 
-		responsesXML, err := propfind.MultistatusResponse(ctx, &propfind.XML{Prop: ff.Prop}, infos, s.c.PublicURL, namespace, "", nil)
+		responsesXML, err := propfind.MultistatusResponse(ctx, &propfind.XML{Prop: ff.Prop}, infos, s.c.PublicURL, namespace /*"",*/, nil)
```
C0rby, line 113: This comment can also be removed, I guess.

internal/http/services/owncloud/ocdav/versions.go
```diff
@@ -189,7 +189,7 @@ func (h *VersionsHandler) doListVersions(w http.ResponseWriter, r *http.Request,
 		infos = append(infos, vi)
 	}
 
-	propRes, err := propfind.MultistatusResponse(ctx, &pf, infos, s.c.PublicURL, "", "", nil)
+	propRes, err := propfind.MultistatusResponse(ctx, &pf, infos, s.c.PublicURL, "" /*"",*/, nil)
```
C0rby, line 192: Same

internal/http/services/owncloud/ocdav/propfind/propfind.go
butonic: I have to admit this is really confusing. The testsuite does not even cover all requests and we actually need to do this error code handling for every request type...

I'll move this to the driver. That will also allow eos to return 403 forbidden and decomposedfs to return 404 not found. And we can even configure this kind of hiding per space.

C0rby (approved):

PR #2986, opened 2022-06-17
Use email as display name for external users opening WOPI apps

glpatcern:
This mini PR sets the `username` passed over to the wopiserver also for external/federated accounts. This is then used as a `displayName` in WOPI.


labkode:
@glpatcern changelog missing :)

glpatcern:
> @glpatcern changelog missing :)

Yep was there but with the wrong prefix :)

labkode (approved):

PR #2991, opened 2022-06-20
make sharesstorageprovider get accepted share

butonic:
The sharesstorageprovider now gets an accepted share instead of filtering all shares.


C0rby (approved):
Looks good! :+1:

PR #2998, opened 2022-06-22
Fix 0 byte uploads

aduffeck:
This PR fixes a problem with 0-byte uploads by using `TouchFile` instead of going through TUS (decomposedfs and owncloudsql storage drivers only for now).

Fixes https://github.com/owncloud/ocis/issues/3154


C0rby:
There are still many failures in the CI. The code looks good but I don't want to approve as long as the CI fails.

C0rby (approved):
Cool. :+1:

PR #3009, opened 2022-06-24
Prevent recursive copy/move operations on the API level

aduffeck:
Fixes https://github.com/owncloud/ocis/issues/3992

### DIFF
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -27,8 +27,10 @@ import (
 
 	"github.com/ReneKroon/ttlcache/v2"
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
+	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/net"
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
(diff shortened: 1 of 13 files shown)


internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -384,3 +393,100 @@ func addAccessHeaders(w http.ResponseWriter, r *http.Request) {
...
+	granteeCtx = metadata.AppendToOutgoingContext(granteeCtx, ctxpkg.TokenHeader, authRes.Token)
+	return granteeCtx, nil
+}
+
+func (s *svc) sspReferenceIsChildOf(ctx context.Context, client gatewayv1beta1.GatewayAPIClient, child, parent *provider.Reference) (bool, error) {
+	parentStatRes, err := client.Stat(ctx, &provider.StatRequest{Ref: parent})
+	if err != nil {
+		return false, err
+	}
+	parentAuthCtx, err := authContextForUser(client, parentStatRes.Info.Owner, s.c.MachineAuthAPIKey)
```
butonic, line 421: When the MachineAuthAPIKey is not set, can we return a `NotSupported` error? Then referenceIsChildOf or the move.go implementation can return a good error ur just return a forbidden error

butonic (changes requested):
I see now why you were unhappy. Doing a GetPath request as the owner requires ocdav to know the machine auth api key ...

what would happen if we send the GetPath as the current user?

In theory the storage driver should give us a path relative to the same root when calling GetPath with two different references. Each call should look up the highest resourceid the user has access to and we should be able to compare the path. We can also check thet the references resourceid that we get in GetPath is the same ... dang ... GetPath has no resource id ... fck

I remember discussing having to extend GetPath with the space root for exactly this reason ... maybe in https://github.com/cs3org/reva/pull/1721#issue-897520929

Then again, we extended `ResourceInfo` with `Name` and could use the filedmask to query `Path` and `Space` in a Stat request instead of a GetPath request ... 🤔

butonic (approved):

phil-davis:
@aduffeck this is good and fixes the loop.

Now the related test fails because the API request returns HTTP status 400 in this case. But oC10 returns 409 in this case.

Is it easy to make another PR that adjusts the return status to 409?

See my updated issue https://github.com/owncloud/ocis/issues/3023 and PR #3030 that enables the test again.

Issue #2967, opened 2022-06-14
CERN: Deny of ACLS


labkode:
See attached presentation from September 2021:
https://new.cernbox.cern.ch/s/qliRAxgPsMNmPOx

labkode:
This functionality is already implemented in the server and OCIS Web:
https://github.com/cernbox/reva/commit/5d4054628b5a43d9b7bf0877367ea10400a3cdc2
https://github.com/cernbox/web/commit/1d75877919e862fdecfdff3bae366ede3951ba1a

Testing of this functionality in the sync client has been positive.

micbar:
@labkode See #3011 as a first attempt to make it work on the edge branch.

labkode:
Work in progress, waiting for PM approval for GA.

pmaier1:
UI proposal: https://github.com/owncloud/web/issues/7180

micbar:
Done. This is part of the ocis 2.0.0 release. The feature is disabled by default.

Enable it `FRONTEND_OCS_ENABLE_DENIALS=true`

PR #3011, opened 2022-06-24
Expose capability to deny access in OCS API

micbar:
# Description

This adds #2967 to the edge branch.

@labkode @dragotin @kulmann Let us try to make this work. We need to test and estimate the impact.

@kobergj Your input is welcome 😄


internal/http/services/owncloud/ocs/conversions/role.go
```diff
@@ -154,6 +165,15 @@ func NewUnknownRole() *Role {
 	}
 }
 
+// NewDeniedRole creates a fully denied role
+func NewDeniedRole() *Role {
+	return &Role{
+		Name:                   RoleDenied,
+		cS3ResourcePermissions: &provider.ResourcePermissions{},
```
micbar, line 172: @butonic @kobergj we need to check how the decomposedFS works with these permissions. The idea is that this node and all childs should not have any permissions for the grantee. Maybe we need to implement it.

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -39,21 +39,30 @@ const (
 	PermissionDelete
 	// PermissionShare grants share permissions on a resource
 	PermissionShare
+	// PermissionDeny grants permissions to deny access on a resource
+	// The recipient of the resource will then have PermissionNone.
+	PermissionDeny
+	// PermissionNone grants no permissions on a resource
+	PermissionNone
+	// PermissionMax is to be used within value range checks
+	PermissionMax Permissions = (1 << (iota - 1)) - 1
```
kobergj, line 48: `PermissionMax` makes no sense. It would include `PermissionNone` which would take away all other permissions

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -39,21 +39,30 @@ const (
...
 	// PermissionAll grants all permissions on a resource
-	PermissionAll Permissions = (1 << (iota - 1)) - 1
+	PermissionAll = PermissionMax - PermissionNone
```
kobergj, line 50: Since we anyway add new permissions now we could also include a `PermissionManager`. I worked around it in my PR but a permission for it would be better.

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -39,21 +39,30 @@ const (
...
+	// PermissionMin is to be used within value range checks
+	PermissionMin = PermissionRead
```
kobergj, line 52: The naming is confusing. The minimal permission is `PermissionNone`. I would opt removing this completely

internal/http/services/owncloud/ocs/conversions/role.go
kobergj: Should work imo. Decomposedfs checks for specific `ResourcePermissions` before doing anything. In this case `ResourcePermissions` would be all false.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -463,6 +463,24 @@ func (h *Handler) extractPermissions(reqRole string, reqPermissions string, ri *
...
+	if permissions == conversions.PermissionNone {
+		if !ri.PermissionSet.DenyGrant {
+			return nil, nil, &ocsError{
+				Code:    http.StatusNotFound,
+				Message: "Cannot set the requested share permissions: no deny grant on resource",
+				Error:   errors.New("Cannot set the requested share permissions: no deny grant on resource"),
+			}
+		}
+		existingPermissions := conversions.RoleFromResourcePermissions(ri.PermissionSet).OCSPermissions()
+		if permissions == conversions.PermissionInvalid || !existingPermissions.Contain(permissions) {
```
kobergj, line 476: Does this actually work? I mean `permissions == conversions.PermissionsNone` as stated in line 467. So

1) `permissions` can never be equal to `conversions.PermissionInvalid`
2) `existingPermissions` (which is `conversions.PermissionsAll` in best case) can NEVER contain `conversions.PermissionNone`.

So we would always run into the error below

kobergj (changes requested):
Some questions/remarks

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
micbar: pushed a fix

internal/http/services/owncloud/ocs/conversions/permissions.go
micbar: This is not about the minimal permissions, as far as I understand @labkode and @ishank011 it is the upper maximum for the bitmask which is used to validate possible input values. Same for `PermissionMax`

internal/http/services/owncloud/ocs/conversions/permissions.go
micbar: See my comment below, it is used for validation purposes. Maybe you have a problem with the naming.

Should it be `PermissionMaxInput` ?

internal/http/services/owncloud/ocs/conversions/permissions.go
micbar: suggest to use `PermissionMinInput`

internal/http/services/owncloud/ocs/conversions/role.go
micbar: Ok, so that works for the whole tree below that node? Let me test it.

pkg/storage/utils/ace/ace.go
```diff
@@ -365,6 +368,9 @@ func getACEPerm(set *provider.ResourcePermissions) string {
 	if set.RemoveGrant || set.UpdateGrant {
 		b.WriteString("o")
 	}
+	if set.DenyGrant {
+		b.WriteString("O")
```
kobergj, line 372: Why `"O"`? We are using linux acls here: https://linux.die.net/man/5/nfs4_acl There is no capital `"O"` flag afais.

kobergj (changes requested):

internal/http/services/owncloud/ocs/conversions/permissions.go
kobergj: Ok ja. Much better now. :+1:

internal/http/services/owncloud/ocs/conversions/permissions.go
kobergj: Yes, that is much clearer :+1:

pkg/storage/utils/ace/ace.go
micbar: small `o` is "change owner"

There is nothing there for "revoke acess" so i picked capital "O". Open for suggestions.

pascalwengerter (commented):
Just skimmed the code quickly but seems like the actual capability (as in "entry in capabilities.go so clients know whether they can use negative ACLs") is missing 💩

micbar:
@pascalwengerter this is configurable now. It needs no capability because the web client reads the webdav permissions. It only makes the denials available if it sees a `Z` permission.

micbar:
@dragotin @kobergj For the backend: The ace implementation which mimics nfsv4 acls has no letter to indicate a denial. I added capital `O`

## from the manpage

https://linux.die.net/man/5/nfs4_acl

```
ACE PERMISSIONS:

There are a variety of different ACE permissions (13 for files, 14 for directories), each represented by a single character. An ACE should have one or more of the following permissions specified:
r
read-data (files) / list-directory (directories)

w

write-data (files) / create-file (directories)

a

append-data (files) / create-subdirectory (directories)

x

execute (files) / change-directory (directories)

d

delete - delete the file/directory. Some servers will allow a delete to occur if either this permission is set in the file/directory or if the delete-child permission is set in its parent direcory.

D

delete-child - remove a file or subdirectory from within the given directory (directories only)

t

read-attributes - read the attributes of the file/directory.

T

write-attributes - write the attributes of the file/directory.

n

read-named-attributes - read the named attributes of the file/directory.

N

write-named-attributes - write the named attributes of the file/directory.

c

read-ACL - read the file/directory NFSv4 ACL.

C

write-ACL - write the file/directory NFSv4 ACL.

o

write-owner - change ownership of the file/directory.

y

synchronize - allow clients to use synchronous I/O with the server.
```

## Warning

```
Deny ACEs should be avoided whenever possible. Although they are a valid part of NFSv4 ACLs, Deny ACEs can be confusing and complicated. This stems primarily from the fact that, unlike POSIX ACLs and CIFS ACLs, the ordering of ACEs within NFSv4 ACLs affects how they are evaluated.

First, it is important to note that (despite some unfortunate ambiguity in RFC3530) NFSv4 ACLs are "default-deny" in practice. That is, if a permission is not explicitly granted, it is denied.

In general, when a principal is attempting to perform an action over NFSv4 which requires one or more permissions, an access check is performed. The NFSv4 ACL (assuming one is present) is evaluated ACE-by-ACE until every one of those permissions has been addressed, or until the end of the ACL is reached. If every requisite permission was granted by Allow ACEs and was not forbidden by Deny ACEs (see next paragraph), the action is allowed to proceed. Otherwise, the action is forbidden.

Note that each requisite permission is only addressed once -- that is, after a permission has been explicitly Allowed or Denied once during an access check, any subsequent ACEs in the ACL which affect that permission are no longer considered. This often introduces problematic ordering issues when Deny ACEs are present.

Additionally, in some cases Group-Deny ACEs can be difficult (if not impossible) to enforce, since a server might not know about all of a given principal's memberships in remote groups, e.g.

Because NFSv4 ACLs are "default-deny", the use of Deny ACEs can (and should) be avoided entirely in most cases.
```

## Conclusion

We should keep the deny feature at this point where we only allow "full denials". Going beyond that should never be considered.

kobergj (changes requested):
Some comments on dav part. Rest looks good imo :+1:

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -46,6 +46,8 @@ const (
 	PermissionMaxInput = PermissionAll
 	// PermissionMinInput is to be used within value range checks
 	PermissionMinInput = PermissionRead
+	// PermissionsNone is to be used to deny access on a resource
+	PermissionsNone = 64
```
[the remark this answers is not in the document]
micbar: @kobergj Unfortunately we cannot change it. The existing permissions should not change. 31 is the `PermissionsAll` representation. Further than that, this is a forward port from master and is already running on Cernbox.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -445,14 +447,32 @@ func (h *Handler) extractPermissions(reqRole string, reqPermissions string, ri *
...
 			Code:    http.StatusNotFound,
 			Message: "Cannot set the requested share permissions",
 			Error:   errors.New("cannot set the requested share permissions"),
 		}
 	}
 
+	if role.Name == conversions.RoleDenied {
+		switch {
+		case !h.deniable:
+			return nil, nil, &ocsError{
```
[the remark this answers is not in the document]
micbar: This is a switch to disable the feature even if we have the CanDeny permission. This feature is currently disabled by default and not supported. So we need a config to make the request impossible.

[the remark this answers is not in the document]
micbar: Changed the wording.

internal/http/services/owncloud/ocs/conversions/permissions.go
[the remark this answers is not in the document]
kobergj: Ok. But please note that if we do so, we cannot add another permission any more. `PermissionAll` must then forever stay 31 as it cannot increase without including `PermissionsNone`

internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -46,6 +46,8 @@ const (
 	PermissionMaxInput = PermissionAll
```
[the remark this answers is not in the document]
kobergj: I'm talking about `MaxInput`. Afaik it is used to sanitized requests. One must set only values lower than this. If this is set to `31` (`PermissionAll`) then one cannot request to set `64` (`PermissionNone`)

internal/http/services/owncloud/ocs/conversions/role.go
```diff
@@ -120,12 +124,19 @@ func (r *Role) WebDAVPermissions(isDir, isShared, isMountpoint, isPublic bool) s
 	if isDir && r.ocsPermissions.Contain(PermissionCreate) {
 		fmt.Fprintf(&b, "CK")
 	}
+
+	if r.CS3ResourcePermissions().DenyGrant && deniable {
```
kobergj, line 128: Why is it needed to check for `deniable` here? One should block requests containing the `DenyGrant` in the webdavhandler, possibly returning a BadRequest. We should NOT magically sanitize requests like we do here.

kobergj (changes requested):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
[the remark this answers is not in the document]
micbar: solved

internal/http/services/owncloud/ocs/conversions/role.go
micbar: this is the ocs part where we can block all deny requests by the config.

micbar:
@kobergj @kulmann I added the capability.

pkg/storage/utils/ace/ace.go
```diff
@@ -187,8 +188,34 @@ type ACE struct {
...
+			ListFileVersions:     true,
+			ListRecycle:          true,
+			Move:                 true,
+			RemoveGrant:          true,
+			PurgeRecycle:         true,
+			RestoreFileVersion:   true,
+			RestoreRecycleItem:   true,
+			Stat:                 true,
+			UpdateGrant:          true,
+			DenyGrant:            true,
```
kobergj, line 214: Why do we need to set those? Afais they are not used at all?

pkg/storage/utils/decomposedfs/grants.go
```diff
@@ -36,8 +36,40 @@ import (
...
 
 // DenyGrant denies access to a resource.
-func (fs *Decomposedfs) DenyGrant(ctx context.Context, ref *provider.Reference, g *provider.Grantee) error {
-	return errtypes.NotSupported("decomposedfs: not supported")
+func (fs *Decomposedfs) DenyGrant(ctx context.Context, ref *provider.Reference, grantee *provider.Grantee) error {
+	log := appctx.GetLogger(ctx)
+
+	log.Debug().Interface("ref", ref).Interface("grantee", grantee).Msg("DenyGrant()")
+
+	// f
```
kobergj, line 44: Is this comment intended?

kobergj (changes requested):
Only two small questions open. Rest looks good to go :+1:

pkg/storage/utils/decomposedfs/grants.go
micbar: 🙈

pkg/storage/utils/ace/ace.go
micbar: ok, not needed because we only support full denials

kobergj (approved):

PR #3023, opened 2022-07-01
Fix public link resharing

kobergj:
See here https://github.com/owncloud/ocis/issues/3881 or here https://github.com/owncloud/ocis/issues/4061

Basically privilege escalation was possible through public links and resharing


butonic:
hm, shouldn't the public link share manager prevent elevating permissions? With this approach you are just shadowing a broken implementation, aren't you?

kobergj:
@butonic I sticked to same implementation as for normal shares. Permissions are checked within the ocs handler. As of now the share manager doesn't care about permissions.

butonic:
YOLO 😞

butonic (approved):

PR #3034, opened 2022-07-06
[tests-only] [full-ci] Test with core branch ocis-4095

phil-davis:
to check that the test code in core PR https://github.com/owncloud/core/pull/40180 does not cause any regression in tests.


phil-davis:
No longer needed.

PR #3035, opened 2022-07-06
prevent cross space move

butonic:
decomposedfs now prevents moving across space boundaries

fixes https://github.com/owncloud/ocis/issues/4114

### DIFF
--- a/changelog/unreleased/prevent-cross-storage-move.md
+++ b/changelog/unreleased/prevent-cross-storage-move.md
@@ -1,0 +1,5 @@
+Bugfix: Prevent cross space move
+
+decomposedfs now prevents moving across space boundaries
+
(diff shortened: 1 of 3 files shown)


micbar (approved):

labkode:
Maybe you should implement the move as "copy/remove" rather than forbidden this feature? 
Is copying still possible?

micbar:
> Maybe you should implement the move as "copy/remove" rather than forbidden this feature? Is copying still possible?

Yes. COPY & DELETE works fine

butonic:
@labkode the move is only forbidden for decomposedfs. It is technically possible, but requires moving all children, from one space to another, which is why I decided to forbid it for now. Other drivers are not affected.

PR #3042, opened 2022-07-07
[tests-only] Event based postprocessing

kobergj:
We decided to put more effort into postprocessing and schedule it via events. This is ofcourse the superior solution as load can be shared among service instances now


pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -126,21 +131,108 @@ func NewDefault(m map[string]interface{}, bs tree.Blobstore) (storage.FS, error)
...
+			default:
+				log.Error().Str("action", ev.Action).Str("uploadID", ev.UploadID).Msg("unknown postprocessing outcome - aborting")
+				fallthrough
+			case "abort":
+				failed = true
+				keepUpload = true
+			case "continue":
+				if err := upload.Finalize(up); err != nil {
+					log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("could not finalize upload")
+					keepUpload = true // should we keep the upload when assembling failed?
```
fschade, line 204: we should answer the question before we merge

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -126,21 +131,108 @@ func NewDefault(m map[string]interface{}, bs tree.Blobstore) (storage.FS, error)
...
+			case "delete":
+				failed = true
+			}
+
+			upload.Cleanup(up, failed, keepUpload)
+
+			if err := events.Publish(fs.stream, events.UploadReady{UploadID: ev.UploadID}); err != nil {
+				log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("Failed to publish UploadReady event")
+			}
+		case events.VirusscanFinished:
```
fschade, line 216: not too happy that the reva postprocessing knows about the virus service, fine for now but would be nice to have some more generic solutions.

fschade (approved):
unfortunately still not able to approve, but at least i've added some comments.

Anyway, LGTM 🚀

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj: Yes. I was thinking that since we reported the client we have all bytes, we cannot delete the data without reason. But I'm fine changing it.

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj: Yes that would be better. But I'm not sure how a more generic event should look like, so I opt keeping it until we have more events and can deduct a reasonable generic event

kobergj: @butonic 
Not entirely. Reva needs to update the node with the postprocessing result. E.g. last virusscan date. This information is important for the user. It could do this in two ways:
- Wait for the `PostprocessingFinished` event and update data for all steps at once
- Update the node when each postprocessing step is finished
The second option is imo superior as clients could potentially show the status of the postprocessing

That being said, as @fschade mentioned above already, it would be better to have a generic event like `PostprocessingStepFinished` on which reva could update the node. But as I explained above I would wait for that until we have more postprocessing steps and then design this event according to common values

kulmann (approved):
Giving a 👍 from @fschade

fschade:
@kulmann thank you 😗

PR #3043, opened 2022-07-07
indexer: introduce LookupCtx

butonic:
The index interface now has a new LookupCtx that can look up multiple values so we can more efficiently look up multiple shares by id. It also takes a context so we can pass on the trace context to the CS3 backend

### DIFF
--- a/pkg/storage/utils/indexer/index/autoincrement.go
+++ b/pkg/storage/utils/indexer/index/autoincrement.go
@@ -75,17 +75,53 @@ func (idx *Autoincrement) Init() error {
 
 // Lookup exact lookup by value.
 func (idx *Autoincrement) Lookup(v string) ([]string, error) {
-	searchPath := path.Join(idx.indexRootDir, v)
-	oldname, err := idx.storage.ResolveSymlink(context.Background(), searchPath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			err = &idxerrs.NotFoundErr{TypeName: idx.typeName, IndexBy: idx.indexBy, Value: v}
-		}
+	return idx.LookupCtx(context.Background(), v)
+}
 
-		return nil, err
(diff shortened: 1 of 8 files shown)


pkg/storage/utils/indexer/indexer.go
```diff
@@ -166,9 +166,14 @@ func (i *StorageIndexer) FindBy(t interface{}, queryFields ...Field) ([]string,
...
-			for _, idx := range fields.IndicesByField[strcase.ToCamel(field.Name)] {
-				res, err := idx.Lookup(field.Value)
+		for fieldName, queryFields := range groupFieldsByName(queryFields) {
+			idxes := fields.IndicesByField[strcase.ToCamel(fieldName)]
+			values := make([]string, 0, len(queryFields))
+			for _, f := range queryFields {
+				values = append(values, f.Value)
+			}
+			for _, idx := range idxes {
+				res, err := idx.LookupCtx(context.Background(), values...)
```
butonic, line 176: here we no longer make one Lookup call per queryField but instead aggregate by field name and then use the new LookupCtx to fetch all results at once.

I introduced the context because we want to be able to pass the trace context into the index as well ...

pkg/storage/utils/indexer/index/autoincrement.go
```diff
@@ -75,17 +75,53 @@ func (idx *Autoincrement) Init() error {
...
+		allValues, err = idx.storage.ReadDir(context.Background(), path.Join("/", idx.indexRootDir))
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		allValues = values
+	}
+
+	var matches = []string{}
+	for _, av := range allValues {
```
C0rby, line 95: It could be more efficient to turn allValues into a `map[string]struct{}` to have faster lookups.

C0rby (approved):

PR #3060, opened 2022-07-11
[cbox-commit-5] Format adler checksum header as expected by the sync client

labkode:


butonic (approved):
@phil-davis we should make the testsuite case insensitive for the checksum hash algorithm

micbar:
@butonic @rhafer please also check if we can forward port this to the edge branch

PR #3064, opened 2022-07-11
[cbox-commit-1] Expose capability to deny access in OCS API

labkode:


internal/http/services/owncloud/ocs/conversions/permissions.go
```diff
@@ -38,21 +38,30 @@ const (
...
+	PermissionDeny
+	// PermissionNone grants no permissions on a resource
+	PermissionNone
+	// PermissionMax is to be used within value range checks
+	PermissionMax Permissions = (1 << (iota - 1)) - 1
 	// PermissionAll grants all permissions on a resource
-	PermissionAll Permissions = (1 << (iota - 1)) - 1
+	PermissionAll = PermissionMax - PermissionNone
+	// PermissionMin is to be used within value range checks
+	PermissionMin = PermissionRead
```
micbar, line 51: We should not change the bitmask. We should use roles instead. The permissions bitmask is considered deprecated and will not work in spaces.

pkg/share/share.go
```diff
@@ -118,9 +117,7 @@ func MatchesFilter(share *collaboration.Share, filter *collaboration.Filter) boo
 	case collaboration.Filter_TYPE_GRANTEE_TYPE:
 		return share.Grantee.Type == filter.GetGranteeType()
 	case collaboration.Filter_TYPE_EXCLUDE_DENIALS:
-		// This filter type is used to filter out "denial shares". These are currently implemented by having the permission "0".
-		// I.e. if the permission is 0 we don't want to show it.
-		return int(conversions.RoleFromResourcePermissions(share.Permissions.Permissions).OCSPermissions()) != 0
+		return share.Permissions.Permissions.DenyGrant
```
micbar, line 120: You should use permissionsNone here because you want to filter out the denials not the DenyGrant

micbar (commented):
See comments.

internal/http/services/owncloud/ocs/conversions/permissions.go
labkode: That's fine, this PR is not expected to work with spaces, only for the master branch.

micbar: Ok but the testsuite will never pass on this. The permissions bitmask ends with `31` everything above that will be considered as "invalid permission"

labkode: Yeah we found that earlier, I think we will add those tests in the expected failures files for time being as this will (are) fixed in the edge branch.

PR #3065, opened 2022-07-11
[tests-only] Allow plugin bamarni/composer-bin-plugin

SwikritiT:
Part of: https://github.com/owncloud/QA/issues/749


phil-davis (approved):
App;roved if CI passes

PR #3070, opened 2022-07-12
Allow http service to expose prefixes containing slashes (/)

gmgigi96:
This PR allows to expose services under URLs like `api/v1/mystuff` rather than just on `api`


gmgigi96:
@labkode I will fix the code to pass the tests

labkode:
@gmgigi96 the thing is I don't know why they failed ...

gmgigi96:
> @gmgigi96 the thing is I don't know why they failed ...

I think I got it. Before the handler was called with the URL containing only the tail part of the original URL request, and this is what the different http services are expecting.

labkode (approved):

PR #3071, opened 2022-07-13
make CS3 sharing drivers parse legacy resource id

butonic:
The CS3 public and user sharing drivers will now correct a resource id that is missing a spaceid when it can split the storageid.


micbar (approved):
LGTM

PR #3075, opened 2022-07-14
Check move destination perms

aduffeck:
We now properly check the permissions on the target of move operations.

Fixes https://github.com/owncloud/ocis/issues/4192


micbar (approved):
LGTM

PR #3079, opened 2022-07-15
Allow empty permissions on links again

kobergj:
Empty permissions are needed for infamous alias links. Through some PR that functionailty was dropped. Now it is back again


pascalwengerter:
Works like a charm 👌🏽

butonic (approved):

PR #3080, opened 2022-07-15
make dataprovider send headers

butonic:
Instead of ocdav doing an additional Stat request we now rely on the dataprovider to return the necessary metadata information as headers. As a bonus it prevents race conditions between the Stat and InitiateFileDownload calls.

related: https://github.com/owncloud/ocis/issues/4153

### DIFF
--- a/changelog/unreleased/tune-download.md
+++ b/changelog/unreleased/tune-download.md
@@ -1,0 +1,5 @@
+Bugfix: Make dataproviders return more headers
+
+Instead of ocdav doing an additional Stat request we now rely on the dataprovider to return the necessary metadata information as headers.
+
+https://github.com/owncloud/reva/issues/3080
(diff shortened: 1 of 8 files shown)


labkode:
@butonic what about encapsulating all this information into a JSON structure and pass that payload into one single header that can be decoded in a simpler way compared to have the information spread across multiple headers?

butonic:
@labkode most of the headers are standard headers: ContentType, ContentDisposition, ETag and LastModified.

Which leaves the OC prefixed headers: `OC-ETag` (which in contrast to `ETag` remains the same, regardless of the content encoding, so clients can use it to sync), `OC-FileId` and `OC-Checksum`.

- `OC-Checksum` should be deprecated in favor of [Digest](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Digest)

- For the other headers maybe sth like the TUS `Upload-Metadata` could be used. It could be `Download-Metadata`, `CS3-Metadata` or individual `CS3-{property}` headers.

The question next question then becomes how can clients request what metadata to return, eg by providing a field mask in the get request ... again, as a header.

micbar (approved):

butonic:
`OC-FileId` should be deprecated and split into `CS3-StorageID`, `CS3-SpaceID` and `CS3-OpaqueID`,
`OC-ETag` should be deprecated in favor of a `CS3-ETag`

related https://github.com/owncloud/ocis/issues/3782

PR #3084, opened 2022-07-18
Fix list space permissions

butonic:


pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -282,7 +282,7 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
 		return nil, errtypes.PermissionDenied(fmt.Sprintf("user %s is not allowed to list spaces of other users", ctxpkg.ContextMustGetUser(ctx).GetId().GetOpaqueId()))
```
dragotin, line 282: Don't you want to check for `userId != spaceIDAny` here rather than for `spaceTypeAny`?

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -282,7 +282,7 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
 	}
 
-	checkNodePermissions := !canListAllSpaces && !unrestricted
+	checkNodePermissions := !canListAllSpaces || !unrestricted
```
dragotin, line 285: The flag `unrestricted` is ambiguous for me. It is not documented at all in the function. You can only guess what should be unrestricted. Yes, that is possible, but not nice.

dragotin (commented):
Just two little nitpicks.

pkg/storage/utils/decomposedfs/spaces.go
butonic: true
1. we may have to touch this again anyway when adding a user filter, see https://github.com/owncloud/ocis/pull/4072
2. the permissions need to be reworked anyway, as there are better ways than passing an explicit unrestricted flag over the api.

so, I'll not touch this in this PR as we just want to push out the ocis/reva release

pkg/storage/utils/decomposedfs/spaces.go
butonic: true, we should probably introduce a new `anyID` const instead of three https://github.com/cs3org/reva/pull/3084/files/9e683197d0f3bd17af4941ac2732a6f46ec77f04#diff-d308a0558595437a537e77e29346472e66ab3bc31c213fcd2dd9b5f64c02aa91R247-R249

not part of this pr, though

dragonchaser (approved):

micbar (approved):

PR #3091, opened 2022-07-20
non personal spaces need no owner

micbar:
# Description

Project spaces should not have a real user as an owner.

That fixes the bug when the space creator leaves the space and is no longer granted access to the space.

@butonic @kobergj @dragotin 

## Solution approach

Public links will become the new unified links. Therefore they will have a longer lifecycle than public links.
Inside a space, we need a stable way to access a public link regardless of the users which can join or leave the space at any time.

This PR assigns an owner with the spaceID as UserID and type `invalid` as a space owner. In this way, the public link scoped token authentication can work in the same way like inside of a personal space.

@C0rby JFYI

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -298,6 +298,9 @@ func UserTypeMap(accountType string) userpb.UserType {
 		t = userpb.UserType_USER_TYPE_FEDERATED
 	case "lightweight":
 		t = userpb.UserType_USER_TYPE_LIGHTWEIGHT
+	// FIXME new user type
+	case "spaceowner":
+		t = 8
 	}
 	return t
 }
@@ -320,6 +323,9 @@ func UserTypeToString(accountType userpb.UserType) string {
 		t = "federated"
 	case userpb.UserType_USER_TYPE_LIGHTWEIGHT:
 		t = "lightweight"
+	// FIXME new user type
+	case 8:
+		t = "spaceowner"
 	}
 	return t
 }
--- a/changelog/unreleased/space-owner.md
+++ b/changelog/unreleased/space-owner.md
@@ -1,0 +1,5 @@
+Bugfix: Project spaces need no real owner
+
+Make it possible to use a non existing user as a space owner.
+
+https://github.com/cs3org/reva/pull/3091
--- a/internal/grpc/services/gateway/publicshareprovider.go
+++ b/internal/grpc/services/gateway/publicshareprovider.go
@@ -21,6 +21,7 @@ package gateway
 import (
 	"context"
 
+	userprovider "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	"github.com/cs3org/reva/v2/pkg/appctx"
@@ -138,6 +139,13 @@ func (s *svc) UpdatePublicShare(ctx context.Context, req *link.UpdatePublicShare
 	if err != nil {
 		return nil, errors.Wrap(err, "error updating share")
 	}
-	s.cache.RemoveStat(ctxpkg.ContextMustGetUser(ctx), res.Share.ResourceId)
+	s.cache.RemoveStat(
+		&userprovider.User{
+			Id: &userprovider.UserId{
+				OpaqueId: res.Share.Owner.GetOpaqueId(),
+			},
+		},
+		res.Share.ResourceId,
+	)
 	return res, nil
 }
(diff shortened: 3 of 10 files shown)


pkg/utils/utils.go
```diff
@@ -59,6 +59,15 @@ var (
 	PublicStorageProviderID = "7993447f-687f-490d-875c-ac95e89a62a4"
 	// PublicStorageSpaceID is the space id used by the sharestorageprovider
 	PublicStorageSpaceID = "7993447f-687f-490d-875c-ac95e89a62a4"
+	// VirtualSpaceOwner is the non-existing user which is needed for public share impersonation
```
butonic, line 62:
```suggestion
	// VirtualSpaceOwner is the non-existing user, used for public share impersonation
```

pkg/utils/utils.go
```diff
@@ -59,6 +59,15 @@ var (
...
+	VirtualSpaceOwner = userpb.User{
+		Id: &userpb.UserId{
+			OpaqueId: "dbdf733d-427f-4ac3-8093-2cba5c307e4c",
```
butonic, line 65: should we use something more speaking like `virtual-public-link-user-id`?

butonic (approved):

pkg/utils/utils.go
micbar: Ok, we can do this safely because it is not a providerID?

I was just following the pattern for the provider IDs

pkg/utils/utils.go
```diff
@@ -59,6 +59,15 @@ var (
...
+			Type:     userpb.UserType_USER_TYPE_INVALID,
+		},
+		DisplayName: "Virtual Space Owner",
+		Username:    "virtualowner",
```
butonic, line 69:
```suggestion
		DisplayName: "No Owner",
		Username:    "nobody",
```
Hm, should we use the nobody/nogroup analogy from unix? https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/usernames.html

butonic:
I like the concept, however, naming is hard. I don't think this is a 'virtual' user. Conceptually, it represents *no* user, aka *nobody* in unix ... well ... in unix nobody should never 'own' any files or folders ...

pkg/utils/utils.go
kobergj: Mmmh. We should stick to uuids in case we add some validation later imo. Can we just move it to a `var VirtualPublinLinkUID`? Then we can use this everywhere in the code.

pkg/utils/utils.go
phil-davis, line 69: This Username value can be returned by `mustGetIdentifiers()` so callers will see it somewhere.

Whatever its string value is ("virtualowner", "nobody", "somethingelse"...), what will happen if there is a real user with the same value for Username?

Is it OK if the Username attribute is the same for multiple users? (And the `OpaqueId`, of course, is different - unique for every user)

pkg/utils/utils.go
micbar: you need an extra var? what about `utils.VirtualSpaceOwner.Id.GetOpaqueId()`

pkg/utils/utils.go
micbar: That is the same challenge for all other use cases with users, not only for this one 😄

butonic: for the ocs api I think oc10 already blocks `public` as a username when accessing via public link: https://github.com/owncloud/core/blob/eebd754771a99bc44820eb67a6d57cea238e6b86/apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php#L148

so we could use 'public', however that does not reflect the concept of a project space not having an owner. 

Not that I think of it ... won't this be a security risk if any public link will become the owner of all project spaces?

butonic: ok, nevermind. the virtual user is not used to implement public link access, yet...

butonic: internally all permission checks are done one user id, not on username. The ocs api thinks users are identified by username and expectst them to be uniqe. Simplest case: searching for users.

We can move to userids by either extending ocs or (my preferred choice) flesh out libregraph ...

butonic (approved):

pkg/utils/utils.go
phil-davis: For reference, oC10 has "special" prohibited usernames:
https://github.com/owncloud/core/blob/master/lib/public/User/Constants.php

The oC10 username-namespace is polluted with all sorts of things.

micbar:
After checking the public link auth, this needs another approach.

pkg/utils/utils.go
kobergj: Yes. Sorry. That's obviously good enough :+1:

micbar:
@butonic @kobergj @dragotin 
I changed the approach. Public links need to be bound to the user who created them. There is no other solution currently without weakening the security. 

1) Virtual owner --> very bad idea. Especially if it is the same for every space
2) Random Non existing owner --> needs a list for lookup, introduces more state, will have a clean solution only with the guests feature

## Solution approach

Public links are short term collaboration. Most of them have a very short lifecycle. It is reasonable to bind them to the link creator and delete them when the creator leaves the space.

butonic (approved):

butonic:
While I am not happy with using an invalid user type, we can update the CS3 api with a new `USER_TYPE_SPACE_OWNER` ... hm maybe we should then use `8` (the next free int, which would be assigned to `USER_TYPE_SPACE_OWNER`) before merging this?

micbar:
@butonic done

PR #3095, opened 2022-07-21
Cleanup uploads

aduffeck:
This PR makes storage providers aware of upload expiration and adds an interface for FS which support listing and purging expired uploads. It also implements said interface for decomposedfs.

This is required for https://github.com/owncloud/ocis/issues/2622

### DIFF
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -509,7 +509,7 @@ func (s *svc) InitiateFileDownload(ctx context.Context, req *provider.InitiateFi
 
 			// TODO(labkode): calculate signature of the whole request? we only sign the URI now. Maybe worth https://tools.ietf.org/html/draft-cavage-http-signatures-11
 			target := u.String()
-			token, err := s.sign(ctx, target, s.c.TransferExpires)
+			token, err := s.sign(ctx, target, time.Now().UTC().Add(time.Duration(s.c.TransferExpires)*time.Second).Unix())
 			if err != nil {
 				return &gateway.InitiateFileDownloadResponse{
 					Status: status.NewStatusFromErrType(ctx, "error creating signature for download", err),
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -81,13 +81,12 @@ type transferClaims struct {
 	Target string `json:"target"`
 }
 
-func (s *svc) sign(_ context.Context, target string) (string, error) {
+func (s *svc) sign(_ context.Context, target string, expiresAt int64) (string, error) {
 	// Tus sends a separate request to the datagateway service for every chunk.
 	// For large files, this can take a long time, so we extend the expiration
-	ttl := time.Duration(s.c.TransferExpires) * time.Second
 	claims := transferClaims{
 		StandardClaims: jwt.StandardClaims{
-			ExpiresAt: time.Now().Add(ttl).Unix(),
+			ExpiresAt: expiresAt,
 			Audience:  "reva",
 			IssuedAt:  time.Now().Unix(),
 		},
@@ -510,7 +509,7 @@ func (s *svc) InitiateFileDownload(ctx context.Context, req *provider.InitiateFi
 
 			// TODO(labkode): calculate signature of the whole request? we only sign the URI now. Maybe worth https://tools.ietf.org/html/draft-cavage-http-signatures-11
 			target := u.String()
-			token, err := s.sign(ctx, target)
+			token, err := s.sign(ctx, target, s.c.TransferExpires)
 			if err != nil {
 				return &gateway.InitiateFileDownloadResponse{
 					Status: status.NewStatusFromErrType(ctx, "error creating signature for download", err),
@@ -572,7 +571,12 @@ func (s *svc) InitiateFileUpload(ctx context.Context, req *provider.InitiateFile
 
 			// TODO(labkode): calculate signature of the whole request? we only sign the URI now. Maybe worth https://tools.ietf.org/html/draft-cavage-http-signatures-11
 			target := u.String()
-			token, err := s.sign(ctx, target)
+			ttl := time.Duration(s.c.TransferExpires) * time.Second
+			expiresAt := time.Now().Add(ttl).Unix()
+			if storageRes.Protocols[p].Expiration != nil {
+				expiresAt = utils.TSToTime(storageRes.Protocols[p].Expiration).Unix()
+			}
+			token, err := s.sign(ctx, target, expiresAt)
 			if err != nil {
 				return &gateway.InitiateFileUploadResponse{
 					Status: status.NewStatusFromErrType(ctx, "error creating signature for upload", err),
--- a/internal/grpc/services/storageprovider/storageprovider.go
+++ b/internal/grpc/services/storageprovider/storageprovider.go
@@ -28,6 +28,7 @@ import (
 	"path"
 	"sort"
 	"strconv"
+	"time"
 
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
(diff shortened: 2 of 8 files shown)


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -510,7 +509,7 @@ func (s *svc) InitiateFileDownload(ctx context.Context, req *provider.InitiateFi
 
 			// TODO(labkode): calculate signature of the whole request? we only sign the URI now. Maybe worth https://tools.ietf.org/html/draft-cavage-http-signatures-11
 			target := u.String()
-			token, err := s.sign(ctx, target)
+			token, err := s.sign(ctx, target, s.c.TransferExpires)
```
kobergj, line 512: That seems wrong. `s.c.TransferExpires` is an amount of seconds, while expiresAt requires a timestamp.

Speaking of which: Why did we make expiresAt an paramter to `sign()` instead of just calculating it there?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -572,7 +571,12 @@ func (s *svc) InitiateFileUpload(ctx context.Context, req *provider.InitiateFile
 
 			// TODO(labkode): calculate signature of the whole request? we only sign the URI now. Maybe worth https://tools.ietf.org/html/draft-cavage-http-signatures-11
 			target := u.String()
-			token, err := s.sign(ctx, target)
+			ttl := time.Duration(s.c.TransferExpires) * time.Second
+			expiresAt := time.Now().Add(ttl).Unix()
+			if storageRes.Protocols[p].Expiration != nil {
+				expiresAt = utils.TSToTime(storageRes.Protocols[p].Expiration).Unix()
```
kobergj, line 577: Mmh. Why would the protocol expiration overwrite our upload expiration? How are they related? And who wins if they have different values?

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -65,6 +66,7 @@ type config struct {
 	AvailableXS         map[string]uint32                 `mapstructure:"available_checksums" docs:"nil;List of available checksums."`
 	CustomMimeTypesJSON string                            `mapstructure:"custom_mimetypes_json" docs:"nil;An optional mapping file with the list of supported custom file extensions and corresponding mime types."`
 	MountID             string                            `mapstructure:"mount_id"`
+	UploadExpiration    int64                             `mapstructure:"upload_expiration" docs:"0;Duration for how long uploads will be valid."`
```
kobergj, line 69: Do we need a new config value? Didn't we just want to use the `TransferExpires` value?

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -383,12 +397,58 @@ func (fs *Decomposedfs) GetUpload(ctx context.Context, id string) (tusd.Upload,
...
+func (fs *Decomposedfs) ListUploads() ([]tusd.FileInfo, error) {
+	infos, err := filepath.Glob(filepath.Join(fs.o.Root, "uploads", "*.info"))
+	if err != nil {
+		return nil, err
+	}
+	list := make([]tusd.FileInfo, len(infos))
+	idRegexp := regexp.MustCompile(".*/([^/]+).info")
+	for i, info := range infos {
+		match := idRegexp.FindStringSubmatch(info)
+		upload, err := fs.readInfo(match[1])
```
kobergj, line 416: Could we run into a panic here? `index out of range`?

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -383,12 +397,58 @@ func (fs *Decomposedfs) GetUpload(ctx context.Context, id string) (tusd.Upload,
...
+		if err != nil {
+			return nil, err
+		}
+		list[i] = upload
+	}
+	return list, nil
+}
+
+// PurgeExpiredUploads scans the fs for expired downloads and removes any leftovers
+func (fs *Decomposedfs) PurgeExpiredUploads(purgedChan chan<- tusd.FileInfo) error {
```
kobergj, line 426: Imho that should only purge an upload not list the uploads too. That is already done by `ListUploads`. One should call `ListUploads` then `PurgeUpload` when the upload expired. There is no need for `PurgeExpiredUploads` to copy the same logic as `ListUploads`

kobergj (changes requested):

internal/grpc/services/gateway/storageprovider.go
aduffeck: Fixed. It became a parameter because `expiresAt` can - and in fact generally will be - set to something different than the gateway fallback expiration duration.

internal/grpc/services/gateway/storageprovider.go
aduffeck: The expiration can now be configured per storage and this is how the expiration is propagated to the gateway. The expiration setting in the gateway is merely a fallback for storage providers that do not set their own expiration.

internal/grpc/services/storageprovider/storageprovider.go
aduffeck: The default values are identical and they probably won't be changed by many people but since the setting **is** specific to the storage I would rather we make it possible to actually configure it.

pkg/storage/utils/decomposedfs/upload.go
aduffeck: I can't see how since the list is a glob of `uploads/*.infos` and the regexp always matches those results. I can add a safety check though if you prefer, it shouldn't hurt.

pkg/storage/utils/decomposedfs/upload.go
aduffeck: Hm, `PurgeExpiredUploads` iterates over all uploads just like `ListUploads`, but it doesn't really copy its logic.
`ListUploads` returns a list of all the uploads it finds while `PurgeExpiredUploads` looks for uploads that are expired an deletes those. Does that make sense?

aduffeck: Looking at the code again I will DRY it up some by extracting the list-compiling code into a separate function.

pkg/storage/utils/decomposedfs/upload.go
kobergj: No you are probably right. The only thing I could imagine was a faulty upload with empty ID leading to `uploads/.info` which wouldn't be found by the regexp. But I'll leave assessing ciritically to you, I'm fine with leaving as is.

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -383,12 +397,60 @@ func (fs *Decomposedfs) GetUpload(ctx context.Context, id string) (tusd.Upload,
...
+
+	for _, info := range infos {
+		expires, err := strconv.Atoi(info.MetaData["expires"])
+		if err != nil {
+			continue
+		}
+		if int64(expires) < time.Now().Unix() {
+			purgedChan <- info
+			os.Remove(info.Storage["BinPath"])
+			os.Remove(filepath.Join(fs.o.Root, "uploads", info.ID+".info"))
```
kobergj, line 448: Should we log errrors here? We might want to inform "someone" when there is a problem removing files from the disk..

kobergj (approved):
:+1: Looks good. Just one small comment which is not critical.

I might need your help when I merge this into experimental though as it will cause merge conflicts

pkg/storage/utils/decomposedfs/upload.go
aduffeck: :+1: sounds reasonable, I'll add logging for those cases and merge. And sure, let me know when you need help with merging.

aduffeck: I took a look and getting a logging facility in there would be quite involved. I think it's fine to just error out if something goes wrong instead. Errors during deletion aren't expected anyway and should be investigated if they happen. Let me know if that doesn't work for you.

PR #3096, opened 2022-07-21
add userid matching

dragonchaser:
We fixed the user filter to display the users drives properly and allow admins to list other users drives.

### DIFF
--- a/pkg/storage/utils/decomposedfs/spaces.go
+++ b/pkg/storage/utils/decomposedfs/spaces.go
@@ -244,9 +244,9 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
 	// we would not need /nodes/root if access always happened via spaceid+relative path
 
 	var (
-		spaceID = spaceIDAny
-		nodeID  = spaceIDAny
-		userID  = spaceIDAny
+		spaceID         = spaceIDAny
+		nodeID          = spaceIDAny
+		requestedUserID = spaceIDAny
 	)
 
 	spaceTypes := map[string]struct{}{}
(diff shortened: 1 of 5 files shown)


pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -427,6 +427,36 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
+	case canListAllSpaces && requestedUserID == spaceTypeAny:
+		return true
+	case canListAllSpaces && authenticatedUserID != requestedUserID:
+		// as admin you have to be able to see other users spaces
+		return false
+	case canListAllSpaces, unrestricted:
+		// standard case, requesting user needs to check space permissions
+		return false
+	}
+	return true
```
kobergj, line 443: The func is unclear to me. But maybe that is just because of naming?
- What exactly is the function supposed to do? I mean what does it mean if this function returns `true`?
- The parameter `unrestricted` is unclear to me. What is unrestricted? And why does it result in a `false`?
- `requestedUserID == spaceTypeAny` makes no sense to me. How should a userID and a SpaceType match?

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -427,6 +427,36 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
+}
+
+func (fs *Decomposedfs) canListSpacesOfRequestedUser(ctx context.Context, requestedUserID string) bool {
+	authenticatedUserID := ctxpkg.ContextMustGetUser(ctx).GetId().GetOpaqueId()
+	switch {
+	case requestedUserID == spaceTypeAny:
```
kobergj, line 449: Makes no sense imho. A userID and a spaceType can't match. Maybe rename `spaceTypeAny`?

kobergj (changes requested):
Need some clarification. Rest looks good :+1:

kobergj:
Sorry. Github wrote email I should review. Didn't realize it's on draft. Ignore comments if they are still worked on.

pkg/storage/utils/decomposedfs/spaces.go
dragonchaser: the parameter `unrestricted` ins inherited from `ListStorageSpaces` above.

dragonchaser: if it returns true, then permissions on the nodes that are being red from the decomposedfs need to be checked, e.g. 
- user einstein wants to read spaces of user marie => does not have the permission => spaces won't be visible to that user.
- user einstein wants to read its own spaces => has permission => own spaces will be listed

dragonchaser:
> Sorry. Github wrote email I should review. Didn't realize it's on draft. Ignore comments if they are still worked on.

no problem, input during dev is also valuable :)

kobergj (approved):
LGTM

PR #3098, opened 2022-07-22
use Form instead of Query to also support `application/x-www-form-urlencoded` for the app provider http endpoint

wkloucek:
Enhancement: app provider http endpoint uses Form instead of Query

We've improved the http endpoint now uses the Form instead of Query
to also support `application/x-www-form-urlencoded` parameters on
 the app provider http endpoint.

The support for query parameter stays the same. See https://pkg.go.dev/net/http#Request.ParseForm. https://pkg.go.dev/net/url#URL.Query only parses the parameters from the URL


glpatcern (approved):
Nice! Could you please push this to master as well?

PR #3101, opened 2022-07-25
use Form instead of Query to also support `application/x-www-form-urlencoded` for the app provider http endpoint

wkloucek:
port to master of #3098


micbar (approved):

glpatcern (approved):
Thanks!

PR #3106, opened 2022-07-27
[tests-only]Bump core commit id for tests

SwikritiT:
Part of: https://github.com/owncloud/QA/issues/753


phil-davis:
The PHP code dependencies used for acceptance tests are fetched into `vendor-php` these days. That was done because "something else" (I don't remember exactly) also used a directory called `vendor`

I added `vendor-php` to `.gitignore` so that I (and other people) don't accidentally commit it after doing some local testing.

phil-davis (approved):

PR #3110, opened 2022-07-29
fix indexes and list all spaces

micbar:
# Fix the indexes on the DecomposedFS

The spacetype index was not updated during space creation.

@kobergj @C0rby Open for suggestions / improvements

## To-Do

- [x] Check RemoveGrant()

The user-id-index needed a change when removing a space grant

### DIFF
--- a/internal/grpc/services/storageprovider/storageprovider.go
+++ b/internal/grpc/services/storageprovider/storageprovider.go
@@ -980,11 +980,15 @@ func (s *service) AddGrant(ctx context.Context, req *provider.AddGrantRequest) (
 
 	// TODO: update CS3 APIs
 	// FIXME these should be part of the AddGrantRequest object
-	if req.Opaque != nil {
-		_, spacegrant := req.Opaque.Map["spacegrant"]
-		if spacegrant {
-			ctx = context.WithValue(ctx, utils.SpaceGrant, struct{}{})
-		}
+	// https://github.com/owncloud/ocis/issues/4312
+	if utils.ExistsInOpaque(req.Opaque, "spacegrant") {
+		ctx = context.WithValue(
+			ctx,
+			utils.SpaceGrant,
+			struct{ SpaceType string }{
+				SpaceType: utils.ReadPlainFromOpaque(req.Opaque, "spacetype"),
+			},
+		)
 	}
 
 	// check grantee type is valid
@@ -1031,6 +1035,13 @@ func (s *service) RemoveGrant(ctx context.Context, req *provider.RemoveGrantRequ
 		return &provider.RemoveGrantResponse{
 			Status: status.NewInvalid(ctx, "grantee type is invalid"),
 		}, nil
+	}
+
+	// TODO: update CS3 APIs
+	// FIXME these should be part of the RemoveGrantRequest object
+	// https://github.com/owncloud/ocis/issues/4312
+	if utils.ExistsInOpaque(req.Opaque, "spacegrant") {
+		ctx = context.WithValue(ctx, utils.SpaceGrant, struct{}{})
 	}
 
 	err := s.storage.RemoveGrant(ctx, req.Ref, req.Grant)
--- a/pkg/storage/utils/decomposedfs/grants.go
+++ b/pkg/storage/utils/decomposedfs/grants.go
@@ -20,6 +20,7 @@ package decomposedfs
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -156,6 +157,8 @@ func (fs *Decomposedfs) RemoveGrant(ctx context.Context, ref *provider.Reference
 		return err
 	}
 
+	spaceGrant := ctx.Value(utils.SpaceGrant)
+
 	var attr string
 	if g.Grantee.Type == provider.GranteeType_GRANTEE_TYPE_GROUP {
 		attr = xattrs.GrantGroupAcePrefix + g.Grantee.GetGroupId().OpaqueId
@@ -164,7 +167,16 @@ func (fs *Decomposedfs) RemoveGrant(ctx context.Context, ref *provider.Reference
 	}
 
 	if err = xattrs.Remove(node.InternalPath(), attr); err != nil {
-		return
+		return err
+	}
+
+	if spaceGrant != nil {
+		// remove from user index
+		// TODO we need an index for groups
+		userIDPath := filepath.Join(fs.o.Root, "indexes", "by-user-id", g.Grantee.GetUserId().OpaqueId, node.SpaceID)
+		if err := os.Remove(userIDPath); err != nil {
+			return err
+		}
 	}
 
 	return fs.tp.Propagate(ctx, node)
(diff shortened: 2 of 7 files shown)


pkg/storage/utils/decomposedfs/grants.go
```diff
@@ -252,6 +252,11 @@ func (fs *Decomposedfs) storeGrant(ctx context.Context, n *node.Node, g *provide
 		if err != nil {
 			return err
 		}
+	} else {
```
micbar, line 255: @butonic I needed to fix the indexes. Please check if i got that right.

micbar: this code will not work for arbitrary space types

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -435,13 +435,7 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
 func (fs *Decomposedfs) MustCheckNodePermissions(ctx context.Context, requestedUserID string, unrestricted bool) bool {
 	canListAllSpaces := fs.canListAllSpaces(ctx)
-	switch {
-	case canListAllSpaces && !unrestricted:
-		return true
-	case (canListAllSpaces && requestedUserID == userIDAny):
-		// admins should not see any other spaces when they request their own, without settings filters
-		return true
-	case canListAllSpaces, unrestricted:
+	if canListAllSpaces && unrestricted {
```
aduffeck, line 438: Is this supposed to be a simplification or a logic change? It seems to be slightly different, e.g. `!canListAllSpaces` / `unrestricted` will return `true` now but returned `false` before.

pkg/storage/utils/decomposedfs/grants.go
```diff
@@ -240,18 +252,30 @@ func (fs *Decomposedfs) storeGrant(ctx context.Context, n *node.Node, g *provide
...
+	// this is not a grant on a space root we are just adding a share
+	if spaceGrant == nil {
+		spaceType = spaceTypeShare
+	}
+	// this is a grant to a space root, the receiver needs the space type to update the indexes
+	if sg, ok := spaceGrant.(struct{ SpaceType string }); ok && sg.SpaceType != "" {
+		spaceType = sg.SpaceType
+	}
+
+	// set the grant
```
aduffeck, line 266: At this point `spaceType` could be an empty string if both conditions failed and looking at the `updateIndexes` function it doesn't seem to be a valid case. Do we have to bail out here in that case?

pkg/storage/utils/decomposedfs/spaces.go
micbar: That is a change. IMO that is the correct behavior. The case `!canListAllSpaces` / `unrestricted` should return `true``

`!canListAllSpaces` means the user has not the permission from the role and we need to check the grants.

pkg/storage/utils/decomposedfs/grants.go
micbar: If we have no space type then the updateIndexes can only create the `by-user-id` index. 
So IMO it can handle empty strings for the spaceTyoe

micbar: https://github.com/micbar/reva/blob/5ae2aa5da34f2092a36ccef4703142fd2a1e5599/pkg/storage/utils/decomposedfs/spaces.go#L654

pkg/storage/utils/decomposedfs/spaces.go
kobergj: This is the second time I am reading a PR on this and I still haven't understood what `unrestricted` means. Could you please add a comment so I understand next time?

Because from my naive understanding: If something is unrestricted I should not need to check node permissions for it.

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -1033,6 +1039,15 @@ func (s *service) RemoveGrant(ctx context.Context, req *provider.RemoveGrantRequ
 		}, nil
 	}
 
+	// TODO: update CS3 APIs
+	// FIXME these should be part of the RemoveGrantRequest object
+	if req.Opaque != nil {
+		_, spacegrant := req.Opaque.Map["spacegrant"]
+		if spacegrant {
```
kobergj, line 1046: Maybe use `utils.ExistsInOpaque` instead?

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -983,7 +983,13 @@ func (s *service) AddGrant(ctx context.Context, req *provider.AddGrantRequest) (
 	if req.Opaque != nil {
 		_, spacegrant := req.Opaque.Map["spacegrant"]
 		if spacegrant {
-			ctx = context.WithValue(ctx, utils.SpaceGrant, struct{}{})
+			ctx = context.WithValue(
```
kobergj, line 986: Maybe use `utils.ExistsInOpaque` instead?

kobergj (changes requested):

dragonchaser (approved):

pkg/storage/utils/decomposedfs/spaces.go
micbar: will do. `unrestricted` means the user is trying to list all drives (calling `/drives` instead of `/me/drives`), but this should only be possible if we have `canListAllSpaces`

internal/grpc/services/storageprovider/storageprovider.go
micbar: good idea

internal/grpc/services/storageprovider/storageprovider.go
micbar: good idea

pkg/storage/utils/decomposedfs/spaces.go
micbar: done

pkg/storage/utils/decomposedfs/grants.go
```diff
@@ -164,7 +167,16 @@ func (fs *Decomposedfs) RemoveGrant(ctx context.Context, ref *provider.Reference
 	}
 
 	if err = xattrs.Remove(node.InternalPath(), attr); err != nil {
-		return
+		return err
+	}
+
+	if spaceGrant != nil {
+		// remove from user index
+		// TODO we need an index for groups
```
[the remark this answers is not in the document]
micbar: done

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -434,14 +434,13 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
 
 // MustCheckNodePermissions checks if permission checks are needed to be performed when user requests spaces
 func (fs *Decomposedfs) MustCheckNodePermissions(ctx context.Context, requestedUserID string, unrestricted bool) bool {
```
kobergj, line 436: Thanks for updating the comment. It is now clear to me. But that makes `requestedUserID` obsolete, as it isn't used any more, right?

kobergj (approved):
Cleanup action would be a plus, but it is not critical.

pkg/storage/utils/decomposedfs/spaces.go
micbar: yes, let me cleanup.

PR #3114, opened 2022-08-02
Use more efficient ReadDir method, save one Stat call in a loop.

dragotin:


dragonchaser (approved):
lgtm

dragotin:
Further investigation on failing tests reveal the following: `ReadDir` does not follow symlinks, but uses `lstat` internally. That breaks the code as it is, which would be fixed again by a subsequent `stat()` call as before. But that means that we have exactly the same amount of stats as before and do not win anything (but knowledge ;-)
What really would help is a specific `readir` based implementation that does only one stat to follow the symlinks. That would half the amount of stat calls, which would be a good optimization.

PR #3116, opened 2022-08-02
Add cback driver for backup browsing

KyleFearne:
This PR add the cback storage driver implementation that allows the user to browse their backup


pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	client := &http.Client{}
+
+	resp, err := client.Do(req)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode == 404 {
```
gmgigi96, line 116: for all these if cases, use a switch statement, and instead of using directly the numbers, use the const values defined in the http library (404 -> http.StatusNotFound)

Before check for a of status (status >= 200 && status < 300).

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil, errtypes.BadRequest("cback")
+	}
+
+	return resp.Body, nil
+}
+
+func (fs *cback) listSnapshots(userName string, backupID int) ([]snapshotResponse, error) {
+
+	url := fs.conf.APIURL + "/backups/" + strconv.Itoa(backupID) + "/snapshots"
+	requestType := "GET"
```
gmgigi96, line 138: no need for this variable, you can use it directly into the function call. Better if you use http.MethodGet

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer responseData.Close()
+
+	/*Unmarshalling the JSON response into the Response struct*/
+	responseObject := []snapshotResponse{}
+	json.NewDecoder(responseData).Decode(&responseObject)
```
gmgigi96, line 149: When you decode a json, check the error

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	return responseObject, nil
+}
+
+func (fs *cback) matchBackups(userName, pathInput string) (*backUpResponse, error) {
+
+	url := fs.conf.APIURL + "/backups/"
+	requestType := "GET"
```
gmgigi96, line 157: same as above

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer responseData.Close()
+
+	/*Unmarshalling the JSON response into the Response struct*/
+	responseObject := []backUpResponse{}
+	json.NewDecoder(responseData).Decode(&responseObject)
```
gmgigi96, line 168: same as above

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+		if strings.HasPrefix(pathInput, responseObject[i].Source) {
+			substr := strings.TrimPrefix(pathInput, responseObject[i].Source)
+			substr = strings.TrimLeft(substr, "/")
+			responseObject[i].Substring = substr
+			return &responseObject[i], nil
+		}
+	}
+
+	return nil, nil
```
gmgigi96, line 200: please document what means a nil response

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer responseData.Close()
+
+	/*Unmarshalling the JSON response into the Response struct*/
+	responseObject := []contents{}
+	json.NewDecoder(responseData).Decode(&responseObject)
```
gmgigi96, line 251: same as above

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	resp := make([]fsReturn, len(responseObject))
```
gmgigi96, line 253: change this to make([]*fsReturn, 0, len(responseObject))
Note the pointer

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+		if err != nil {
+			return nil, err
+		}
+		/*fmt.Printf("\nName is: %v\n", responseObject[i].Name)
+		fmt.Printf("Type is: %d\n", m)
+		fmt.Printf("Time is: %v", uint64(responseObject[i].Mtime))
+		fmt.Printf("\n")*/
+
+		resp[i].Size = response.Size
```
gmgigi96, line 267: create an fsReturn struct on the flight and append it to the resp list

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+		return 0, err
+	}
+
+	epoch := tm.Unix()
+	return epoch, nil
+}
+
+func (fs *cback) pathFinder(userName, path string) ([]string, error) {
+	url := fs.conf.APIURL + "/backups/"
+	requestType := "GET"
```
gmgigi96, line 290: same as above

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	defer responseData.Close()
+
+	/*Unmarshalling the JSON response into the Response struct*/
+	responseObject := []backUpResponse{}
+	json.NewDecoder(responseData).Decode(&responseObject)
```
gmgigi96, line 302: same as above

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+	returnString := make([]string, len(responseObject))
```
gmgigi96, line 304: change it to returnString := make([]string, 0, len(responseObject))

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+func (fs *cback) pathTrimmer(snapshotList []snapshotResponse, resp *backUpResponse) (string, string) {
+
+	var ssID, searchPath string
+	for _, snapshot := range snapshotList {
+
+		if snapshot.ID == resp.Substring {
+			ssID = resp.Substring
+			searchPath = resp.Source
+			break
```
gmgigi96, line 337: return instead of break, is not clear otherwise the flow

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+
+		} else if strings.HasPrefix(resp.Substring, snapshot.ID) {
+			searchPath = strings.TrimPrefix(resp.Substring, snapshot.ID)
+			searchPath = resp.Source + searchPath
+			ssID = snapshot.ID
+			break
```
gmgigi96, line 343: same as last comment

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+	// returns the storage.FS interface
+	return &cback{conf: c}, nil
+
+}
+
+func (fs *cback) GetMD(ctx context.Context, ref *provider.Reference, mdKeys []string) (*provider.ResourceInfo, error) {
+	var ssID, searchPath string
+
+	user, _ := ctxpkg.ContextGetUser(ctx)
+	UID, _ := ctxpkg.ContextGetUserID(ctx)
```
gmgigi96, line 48: no need for this. You can get the userID from the user struct

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+
+	resp, err := fs.matchBackups(user.Username, ref.Path)
+
+	if err != nil {
+		fmt.Print(err)
```
gmgigi96, line 53: remove this print

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+		return nil, err
+	}
+
+	snapshotList, err := fs.listSnapshots(user.Username, resp.ID)
+
+	if err != nil {
+		fmt.Print(err)
```
gmgigi96, line 60: remove this print

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+
+		return ri, nil
+
+	}
+
+	ret, err := fs.statResource(resp.ID, ssID, user.Username, searchPath, resp.Source)
+
+	if err != nil {
+		fmt.Print(err)
```
gmgigi96, line 99: remove this print

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+	return ri, nil
+}
+
+func (fs *cback) ListFolder(ctx context.Context, ref *provider.Reference, mdKeys []string) ([]*provider.ResourceInfo, error) {
+
+	var path string = ref.GetPath()
+	var ssID, searchPath string
+
+	user, _ := ctxpkg.ContextGetUser(ctx)
+	UID, _ := ctxpkg.ContextGetUserID(ctx)
```
gmgigi96, line 140: same as a previous comment

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+
+	resp, err := fs.matchBackups(user.Username, path)
+
+	if err != nil {
+		fmt.Print(err)
```
gmgigi96, line 145: remove this print

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+
+			files[index] = &f
+		}
+
+		return files, nil
+
+	}
+
+	//If match in path, therefore print the Snapshot IDs
+	files := make([]*provider.ResourceInfo, len(snapshotList))
```
gmgigi96, line 252: change to make([]*provider.ResourceInfo, 0, len(snapshotList))

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+			Etag:          "",
+			Owner:         UID,
+			PermissionSet: &permID,
+			Id:            &ident,
+			MimeType:      mime.Detect(true, ref.Path+"/"+snapshot.ID),
+			Size:          0,
+			Mtime:         &setTime,
+			Type:          provider.ResourceType_RESOURCE_TYPE_CONTAINER,
+		}
+		files[index] = &f
```
gmgigi96, line 284: append the resource info struct to the slice instead of accessing it by index

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+
+	// returns the storage.FS interface
+	return &cback{conf: c}, nil
+
+}
+
+func (fs *cback) GetMD(ctx context.Context, ref *provider.Reference, mdKeys []string) (*provider.ResourceInfo, error) {
+	var ssID, searchPath string
+
+	user, _ := ctxpkg.ContextGetUser(ctx)
```
gmgigi96, line 47: the second parameter returned is a bool indicating if the user is in the context. Either you check it and return a proper error, or you use the method ContextMustGetUser

pkg/storage/fs/cback/cback.go
```diff
@@ -0,0 +1,473 @@
...
+			responseData, err := fs.getRequest(user.Username, url, requestType, nil)
+
+			if err != nil {
+				return nil, err
+			}
+
+			return responseData, nil
+		}
+
+		err = errors.New("can only download files")
```
gmgigi96, line 333: change this error to bad request

pkg/storage/fs/cback/options.go
```diff
@@ -0,0 +1,7 @@
+package cback
+
+// Options for the CBACK module
+type Options struct {
+	ImpersonatorToken string `mapstructure:"impersonator"`
+	APIURL            string `mapstructure:"apiURL"`
```
gmgigi96, line 6: instead of apiURL in the tag, call it "api_url"

pkg/storage/fs/cback/options.go
```diff
@@ -0,0 +1,7 @@
+package cback
+
+// Options for the CBACK module
+type Options struct {
+	ImpersonatorToken string `mapstructure:"impersonator"`
```
gmgigi96, line 5: instead of impersonator in the tag, call it "token"

gmgigi96 (changes requested):

pkg/storage/fs/cback/cback.go
KyleFearne: Will be checking if User is inContext, else returning a UserRequired error.

pkg/storage/fs/cback/cback.go
KyleFearne: Checked if user is inContext, if not will return UserRequired error.

pkg/storage/fs/cback/cback.go
KyleFearne: Now using user.Id

pkg/storage/fs/cback/utilities.go
KyleFearne: Changed accordingly, and changed return type of function fileSystem to ([]*fsReturn, error)

pkg/storage/fs/cback/utilities.go
```diff
@@ -0,0 +1,360 @@
...
+		return nil, err
+	}
+
+	if body != nil {
+		req.Header.Add("Content-Type", "application/json")
+	}
+
+	req.Header.Add("accept", `application/json`)
+
+	client := &http.Client{}
```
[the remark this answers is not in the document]
KyleFearne: Created a 'client' of type *http.Client in the cback struct, then used rhttp.GetHTTPClient() with default options to initialize it in the 'New' function.

pkg/storage/fs/cback/utilities.go
KyleFearne: Checking if the code is NOT an Okay code (resp.StatusCode > 200 && resp.StatusCode >= 300), if this is so the switch statement is invoked which returns a different error according to the code. If the return status is OKAY, it will simply return the body, and nil error.

pkg/storage/fs/cback/utilities.go
KyleFearne: `	/*If there is no error, but also no match found in the backup path the response is nil.
	This means that the LSFolder function will know that no match has been found using the exact path,
	and will therefore start checking if there is a substring of the backup job included in the inputted path*/`

pkg/storage/fs/cback/utilities.go
KyleFearne: Discussed with Gianmaria

Issue #3113, opened 2022-08-01
Mentix OCM Auth Provider Normalization


Daniel-WWU-IT:
The Mentix OCM Auth Provider currently lacks provider normalization; this can lead to incorrect domain mismatches.

labkode:
@mirekys can you take this over?

Issue #2288, opened 2021-11-17
UserId idp may have prefix


michielbdejong:
The [idp field of a UserId](https://github.com/cs3org/cs3apis/blob/9648574fa9963be02dc3bd2700c0639ee8708492/cs3/identity/user/v1beta1/resources.proto#L35-L37) should identify the identity provider of the user, but it's unclear from the current documentation of cs3apis whether this should be done with a domain (e.g. 'cernbox.cern.ch') or with an origin (e.g. 'https://cernbox.cern.ch').

Recently, the users.demo.json file in the ocmd example switched from the first to the second format.

In https://github.com/cs3org/cs3apis/pull/159 I proposed to make the second format the prescribed one, but we don't want to make this a breaking change, so reva should start supporting origin-format for userid idp _in addition to_ supporting domain-format as it does currently.

We should to make the provider authorizer interceptor accept both formats, I'll create a PR.

redblom:
@michielbdejong We want to finalize this (in the context of https://github.com/cs3org/reva/pull/3121).
Can you confirm that the proposed format of the idp field, as origin (eg. https://cernbox.cern.ch) is definite ?

michielbdejong:
In https://github.com/cs3org/cs3apis/pull/159#discussion_r766662560 @wkloucek proposed a 3rd option, namely a full URL. So the options are:

1) domain, e.g. "cernbox.cern.ch"
2) origin, e.g. "https://cernbox.cern.ch"
3) full URL, e.g. "https://cernbox.cern.ch/some/path"

I'm OK with whatever you prefer!

PR #3121, opened 2022-08-03
Normalize domains in mentix provider authorizer driver

mirekys:
Fixes https://github.com/cs3org/reva/issues/3113


redblom:
@mirekys This PR is to solve https://github.com/cs3org/reva/issues/3113 right? That issue was created because invites could not be accepted because the providers could not be authorized. Please take note of the following:

When the user idp of the sender of the forward invite is a domein rather than a url (eg. `cesnet.cz` vs. `http://cesnet.cz`) the provider authorization also fails at the receiving end (where the token has been generated). This is because from the bare domain the host name will not be resolved in the current implementation: https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/internal/http/services/ocmd/invites.go#L241-L248 It will result in an empty string (but no error!). So `ProviderInfo.Domain` will be an empty string. This definitely gives issues with the current mesh sites' users conf as we found out last week.
Previously the `recipientProvider` parameter was actually normalized but this has been removed (https://github.com/cs3org/reva/pull/2751). See also https://github.com/cs3org/reva/issues/2288.
So I guess this `recipientProvider` needs to be normalized once more. And solve https://github.com/cs3org/reva/issues/2288 in the process.

mirekys:
You're right, it also needs normalization, thanks for pointing that out. If I understood the situation correctly, there is yet no exact specification on how the provider domains should look, and Reva should accept both the bare domain name & full URL, while the full URL is the preferable variant in the future? I'll update the PR accordingly

mirekys:
I just simply removed the code and pass the provider as is, as it should be a responsibility of [IsProviderAllowed](https://github.com/cs3org/reva/pull/3121/commits/782742f49ea4c670438886a1e5d82023bfec5e18#diff-2341cc591cded70abda25bfcf8df337380486f7d5e3de784e3dc41171b6221bbR100) endpoint to normalize value before checking

redblom:
> I just simply removed the code and pass the provider as is, as it should be a responsibility of [IsProviderAllowed](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R171) endpoint to [normalize](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R99-R114) value before checking

Makes sense. 
Could `IsProviderAllowed` also log some meaningful err data like the actual input so it's easier to spot the root of an issue? I guess since we currently have no specification on eg. user idp domain vs. url we can not log suggestions in that regard.

mirekys:
It already returns errors if domain is not found amongst mesh providers (together with input domain value) [here](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R190), which then gets logged [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/internal/http/services/ocmd/invites.go#L260), is this insufficient or something else needs to also be logged?

redblom:
> It already returns errors if domain is not found amongst mesh providers (together with input domain value) [here](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R190), which then gets logged [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/internal/http/services/ocmd/invites.go#L260), is this insufficient or something else needs to also be logged?
> 
> EDIT: udated other current mentix errors to provide more context

Yeah I like that ! (there's obviously some personal preference from me here but I like this better ;)

redblom:
> I just simply removed the code and pass the provider as is, as it should be a responsibility of [IsProviderAllowed](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R171) endpoint to [normalize](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R99-R114) value before checking

@mirekys @labkode in the `ForwardInvite` request the recipientProvider that comes from (is) the context user idp still must be normalized also. This could be done [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/json/json.go#L203) and [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/memory/memory.go#L106). 
But we can also let the user manager normalize the user Idp when setting it in the first place. What would be the preferred way?

mirekys:
> > I just simply removed the code and pass the provider as is, as it should be a responsibility of [IsProviderAllowed](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R171) endpoint to [normalize](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R99-R114) value before checking
> 
> @mirekys @labkode in the `ForwardInvite` request the recipientProvider that comes from (is) the context user idp still must be normalized also. This could be done [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/json/json.go#L203) and [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/memory/memory.go#L106). But we can also let the user manager normalize the user Idp when setting it in the first place. What would be the preferred way?

I would vote for handling it in a single place (user manager), rather than doing it in all places idp is being used, but not sure whether it won't break something somewhere

redblom:
I'm also for that. Best to do this in a separate PR though. I can take that one, in 2 weeks as I'm currently on holidays. But would be good to get an opinion from @labkode also, he's back from holidays in 2 weeks as well.

redblom:
@mirekys 

> > I just simply removed the code and pass the provider as is, as it should be a responsibility of [IsProviderAllowed](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R171) endpoint to [normalize](https://github.com/cs3org/reva/pull/3121/files#diff-52e4ab07056f109253b95874955420517a73b65281f985e534b7873dbb914139R99-R114) value before checking
> 
> @mirekys @labkode in the `ForwardInvite` request the recipientProvider that comes from (is) the context user idp still must be normalized also. This could be done [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/json/json.go#L203) and [here](https://github.com/cs3org/reva/blob/5eb6e708af1f7cdd4ae1f950610b0fceff96c28e/pkg/ocm/invite/manager/memory/memory.go#L106). But we can also let the user manager normalize the user Idp when setting it in the first place. What would be the preferred way?

I realized, and tested, that this is not needed after all. So simply removing as you've already done in `internal/http/services/ocmd/invites.go` will suffice indeed.

Next the [`getOCMHost`](https://github.com/cs3org/reva/blob/46294b45ae525b59c47c9a1a147fdadd51aa5131/pkg/ocm/provider/authorizer/mentix/mentix.go#L231) method however currently also fails. Here [`ocmHost.Host`](https://github.com/cs3org/reva/blob/46294b45ae525b59c47c9a1a147fdadd51aa5131/pkg/ocm/provider/authorizer/mentix/mentix.go#L238) will return an empty string after the host field is parsed into a url.
If we return the normalized domain (from the host field) here as well then everything seems to be in order.

mirekys:
@antoonp 
I must admit that I'm getting kinda confused about acceptable formats of values
of each different provider properties. As I understood from this PR and
https://github.com/cs3org/cs3apis/pull/159#discussion_r766662560,
we have to deal with the following provider-related properties:

- `Domain`
- `Host`
- `User IDP`

that can have 3 possible values:

- FQDN (without protocol)
- URI without path
- full URI

I start thinking it would be cleaner to determine the required formats on each property,
and then validate/convert to a given format, when setting the property (rather then forcing
full URLs on all properties used throughout the code). So it would be safer to expect from
everywhere, that if a given property is set on a provider, it is in an expected format.

I think `Domain` should be just FQDNs, `User IDP` should be a full URI, and `Host` is defined as https://github.com/cs3org/cs3apis/blob/main/cs3/ocm/provider/v1beta1/resources.proto#L64, so it should probably also be required to always have a full valid URL there

michielbdejong:
Oops, I used the GitHub online merge tool to merge this branch into my debug branch, and apparently it somehow allowed me to add commits to https://github.com/mirekys/reva/commits/fix-3113 - that's scary!

michielbdejong:
Force push to someone else's GitHub account is also scary but looks like it did put the branch back in its correct state :/

micbar:
@michielbdejong The Ci has changed on the master branch. You need to rebase this PR.

michielbdejong:
I'm still not sure why I'm allowed to push to @mirekys's branch, but I merged `origin/master` into it as @micbar requested.

michielbdejong:
While testing this code, I think there is one more change missing:
In [this block](https://github.com/cs3org/reva/blob/fdced3305f289fd9cb59684a24ce0a97b31d7f0b/pkg/ocm/provider/authorizer/mentix/mentix.go#L266-L270) in the mentix driver for ocm provider authorizer,  `service.Host` is converted from a URL to a FQDN, but looking at for instance this gocdb edit screen:
Screenshot 2022-12-09 at 16 41 32

I think the "service host" is already a FQDN, and parsing it in this way changes it to `""`. So I think it's better to just `return s.Host, nil` here (and same for the `json` and `open` drivers).

michielbdejong:
@labkode I think we should merge this PR asap

labkode (approved):

PR #3123, opened 2022-08-03
Allow stating links with no permissions

kobergj:
This is needed for alias links. 

Before it wasn't allowed to stat a link that has empty permission assigned. This needs to be possible as alias links need to be resolved by their receivers

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/token.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/token.go
@@ -20,7 +20,6 @@ package sharees
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/token.go
```diff
@@ -158,12 +154,11 @@ func handleBasicAuth(ctx context.Context, c gateway.GatewayAPIClient, token, pw
...
 
-func getTokenStatInfo(ctx context.Context, client gateway.GatewayAPIClient, token string) (*provider.StatResponse, error) {
-	return client.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{
-		ResourceId: &provider.ResourceId{
-			StorageId: utils.PublicStorageProviderID,
-			SpaceId:   utils.PublicStorageSpaceID,
-			OpaqueId:  token,
-		},
-	}})
+func getTokenStatInfo(ctx context.Context, client gateway.GatewayAPIClient, token string) (*link.GetPublicShareResponse, error) {
```
C0rby, line 157: I think it makes sense to update the name of this function since it doesn't return a `StatResponse` anymore.

C0rby (approved):
One minor thing. Looks good otherwise.

internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/token.go
kobergj: Good point :+1:  renamed to `getPublicShare` as this is literally what it does.

PR #3132, opened 2022-08-08
[tests-only] Allow setting quota when creating personal spaces

kobergj:
Previously it was not possible setting a quota using the `CreateHome` handler


fschade (approved):
LGTM

PR #3134, opened 2022-08-08
[tests-only] Emit Tag lifecycle events

fschade:
emit events for tag delete and create.

### DIFF
--- a/internal/http/services/owncloud/ocdav/tags.go
+++ b/internal/http/services/owncloud/ocdav/tags.go
@@ -28,6 +28,7 @@ import (
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/errors"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/net"
 	"github.com/cs3org/reva/v2/pkg/appctx"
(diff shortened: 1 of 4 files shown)


internal/http/services/owncloud/ocdav/tags.go
```diff
@@ -70,33 +71,57 @@ func (h *TagHandler) Handler(s *svc) http.Handler {
...
+			Tags: newtags,
+			Ref: &provider.Reference{
+				ResourceId: rid,
+				Path:       ".",
+			},
+			Executant: info.Owner,
+		}); err != nil {
+			log.Error().Err(err).Msg("Failed to publish TagsAdded event")
+		}
+
```
kobergj, line 94: It could lead to problems to send the event within the modify func. If there is an error happening when actually setting the tags the event would still be send, even though there where no tags added.

Better: 
- Change `h.modifyTag` to take an `ev interface{}` as parameter. 
- in `h.modifyTag` if tags where successfully set and `ev != nil` => Publish(ev)`

kobergj (changes requested):
Small change needed I fear..

kobergj (approved):

PR #3135, opened 2022-08-08
Add locking to xattrs Get() and Remove()

dragonchaser:
In case of other processes having a Write lock that needs to be respected.

refs https://github.com/owncloud/ocis/issues/4251


dragonchaser:
This PR does not resolve the issue reported in https://github.com/owncloud/ocis/issues/4251 reverting to draft for now

PR #3136, opened 2022-08-09
fix the cs3 share manager for space shares

micbar:
# Description

Space shares need to be looked up by resourceID.

### DIFF
--- a/pkg/share/manager/cs3/cs3.go
+++ b/pkg/share/manager/cs3/cs3.go
@@ -315,13 +315,42 @@ func (m *Manager) ListShares(ctx context.Context, filters []*collaboration.Filte
 	if !ok {
 		return nil, errtypes.UserRequired("error getting user from context")
 	}
-
-	createdShareIds, err := m.indexer.FindBy(&collaboration.Share{},
-		indexer.NewField("OwnerId", userIDToIndex(user.Id)),
-		indexer.NewField("CreatorId", userIDToIndex(user.Id)),
+	var rIDs []*provider.ResourceId
+	if len(filters) != 0 {
+		grouped := share.GroupFiltersByType(filters)
+		for _, g := range grouped {
+			for _, f := range g {
+				if f.GetResourceId() != nil {
+					rIDs = append(rIDs, f.GetResourceId())
+				}
+			}
+		}
+	}
+	var (
+		createdShareIds []string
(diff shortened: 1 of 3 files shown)


pkg/share/manager/cs3/cs3.go
```diff
@@ -315,11 +315,31 @@ func (m *Manager) ListShares(ctx context.Context, filters []*collaboration.Filte
 	if !ok {
 		return nil, errtypes.UserRequired("error getting user from context")
 	}
+	var rID *provider.ResourceId
+	if len(filters) != 0 {
+		grouped := share.GroupFiltersByType(filters)
+		for _, f := range grouped {
+			for _, f := range f {
```
kobergj, line 322: please rename one of those `f`s

pkg/share/manager/cs3/cs3.go
```diff
@@ -315,11 +315,31 @@ func (m *Manager) ListShares(ctx context.Context, filters []*collaboration.Filte
...
+				if f.GetResourceId() != nil {
+					rID = f.GetResourceId()
```
kobergj, line 324: Is it guaranteed that `grouped` contains maximal one filter where `f.ResourceID != nil`? Otherwise this will return random results..

kobergj (changes requested):
Small question and remark...

pkg/share/manager/cs3/cs3.go
micbar: fixed by using a slice of rIDs

pkg/share/manager/cs3/cs3.go
```diff
@@ -315,11 +315,38 @@ func (m *Manager) ListShares(ctx context.Context, filters []*collaboration.Filte
...
+	if len(rIDs) != 0 {
+		var shareIDs []string
+		for _, rID := range rIDs {
+			shareIDs, err = m.indexer.FindBy(&collaboration.Share{},
+				indexer.NewField("ResourceId", resourceIDToIndex(rID)),
+			)
+			if err != nil {
+				continue
+			}
+			createdShareIds = append(createdShareIds, shareIDs...)
```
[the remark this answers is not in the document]
micbar: we should return when the first error occurs.

kobergj (approved):

PR #3137, opened 2022-08-09
[tests-only] Improve Tags pkg

kobergj:
Allow returning a slice of tags with new `AsSlice` method. Adjust unit tests to test also slices.


fschade (approved):
LGTM

PR #3141, opened 2022-08-10
check listGrants permission when listing shares

micbar:
# Description

We now check the ListGrants permission when listing outgoing shares. If this permission is set, users can list all shares in all spaces.


kobergj (approved):
Nice fix :+1:

PR #3152, opened 2022-08-15
Disable cache

butonic:
We no longer cache not found responses to prevent concurrent requests interfering with put requests.

related: https://github.com/owncloud/ocis/issues/4251


kobergj (approved):
LGTM

PR #3154, opened 2022-08-16
dataproviders now return file metadata

butonic:
Dataprovider drivers can now return file metadata. When the resource info contains a file id, the mtime or an etag, these will be included in the response as the corresponding http headers.


dragonchaser (approved):
lgtm

PR #3167, opened 2022-08-22
disable cache

dragonchaser:
This disables the cache because of race conditions happening in the tests in combination with the ocis search extension.

https://github.com/owncloud/ocis/issues/4251


butonic (approved):
ok for now to get rid of flakyness. a proper distributed cache is needed anyway

PR #3170, opened 2022-08-23
[tests-only] make test macOs support

fschade:
macos arm does not support bou.ke/monkey natively, now those machines can use `make test-macos` instead but without race detection for packages using monkey.


kulmann (approved):
:heart:

PR #3171, opened 2022-08-23
load, dump and storageprovider fixes for migration

butonic:
This PR
- adds `cs3` share manager `Dump()`
- adds `jsoncs3` share manager `Load()`
- some sharemanager fixes to allow uploading a non existing file when an IfUnmodifiedSince header is set


pkg/share/manager/jsoncs3/jsoncs3.go
```diff
@@ -697,3 +699,74 @@ func (m *Manager) UpdateReceivedShare(ctx context.Context, receivedShare *collab
...
+}
+
+// Load imports shares and received shares from channels (e.g. during migration)
+func (m *Manager) Load(ctx context.Context, shareChan <-chan *collaboration.Share, receivedShareChan <-chan share.ReceivedShareWithUser) error {
+	log := appctx.GetLogger(ctx)
+	if err := m.initialize(); err != nil {
+		return err
+	}
+
+	var mu sync.Mutex
```
kobergj, line 717: As both go-routines operate on different data structures, they would not need to share the same lock, do they? The locks could be created within each go routine, then they would not block each other.

pkg/share/manager/jsoncs3/jsoncs3_test.go
```diff
@@ -147,6 +150,58 @@ var _ = Describe("Jsoncs3", func() {
...
 
+	Describe("Load", func() {
+		It("loads shares including state and mountpoint information", func() {
+
+			sharesChan := make(chan *collaboration.Share)
+			receivedChan := make(chan sharespkg.ReceivedShareWithUser)
+
+			share := &collaboration.Share{
+				Id: &collaboration.ShareId{OpaqueId: "1iaeiae$vlcvlcvlc!pzbpzbpzb"},
+				// FIXME we may have to deal with importing existing share ids ... without a storage or provider prefix
```
kobergj, line 161: If this `FIXME` is important I would suggest moving it to the actual code, instead of a test file noone will be looking at until it breaks.

pkg/share/manager/jsoncs3/shareid/shareid.go
```diff
@@ -20,24 +20,26 @@ package shareid
...
 func Decode(id string) (string, string, string) {
-	parts := strings.SplitN(id, "^", 2)
-	if len(parts) == 1 {
+	parts := strings.SplitN(id, IDDelimiter, 3)
+	switch len(parts) {
+	case 1:
 		return "", "", parts[0]
+	case 2:
+		return parts[0], parts[1], ""
+	default:
```
kobergj, line 42: Please add a `case 0:` Otherwise this will panic when feeded with broken data

kobergj (changes requested):
Nipicks as always...

pkg/share/manager/cs3/cs3.go
```diff
@@ -206,6 +206,57 @@ func (m *Manager) Load(ctx context.Context, shareChan <-chan *collaboration.Shar
...
+// Dump exports shares and received shares to channels (e.g. during migration)
+func (m *Manager) Dump(ctx context.Context, shareChan chan<- *collaboration.Share, receivedShareChan chan<- share.ReceivedShareWithUser) error {
+	log := appctx.GetLogger(ctx)
+	if err := m.initialize(); err != nil {
+		return err
+	}
+
+	shareids, err := m.storage.ReadDir(ctx, "shares")
+	if err != nil {
+		log.Error().Err(err).Msg("error fetching shares")
```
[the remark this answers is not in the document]
butonic:
```suggestion
		return err
```

pkg/share/manager/cs3/cs3.go
```diff
@@ -756,6 +807,43 @@ func granteeToIndex(grantee *provider.Grantee) (string, error) {
 	}
 }
 
+// indexToGrantee trues to unparse a grantee in a metadata dir
```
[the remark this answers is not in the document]
butonic:
```suggestion
// indexToGrantee tries to unparse a grantee in a metadata dir
```

pkg/share/manager/jsoncs3/jsoncs3.go
```diff
@@ -697,3 +699,74 @@ func (m *Manager) UpdateReceivedShare(ctx context.Context, receivedShare *collab
...
+			mu.Lock()
+			if err := m.Cache.Add(context.Background(), s.GetResourceId().GetStorageId(), s.GetResourceId().GetSpaceId(), s.Id.OpaqueId, s); err != nil {
+				log.Error().Err(err).Interface("share", s).Msg("error persisting share")
+			} else {
+				log.Info().Str("storageid", s.GetResourceId().GetStorageId()).Str("spaceid", s.GetResourceId().GetSpaceId()).Str("shareid", s.Id.OpaqueId).Msg("imported share")
+			}
+			if err := m.CreatedCache.Add(ctx, s.GetCreator().GetOpaqueId(), s.Id.OpaqueId); err != nil {
+				log.Error().Err(err).Interface("share", s).Msg("error persisting created cache")
+			} else {
+				log.Info().Str("creatorid", s.GetCreator().GetOpaqueId()).Str("shareid", s.Id.OpaqueId).Msg("updated created cache")
```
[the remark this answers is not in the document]
butonic: Load and Dump are used when migrating on the CLI. In my experience, admins really want to see migration progress, what worked and what did not work. That is why I made this verbose.

pkg/share/manager/jsoncs3/shareid/shareid.go
butonic: SplitN will always return at least one element.

pkg/share/manager/jsoncs3/jsoncs3_test.go
butonic: can be dropped, updateShareID() implements this

pkg/share/manager/jsoncs3/jsoncs3.go
butonic: hm, do I even need a mutex? the go routine will pull an element from the channel one by one. The caches Add() functions do *not* have their own lock ... so a lock in Load won't protect against concurrent updates via Share(), Unshare() or UpdateShare(), ... ah, no! it does not need to.

AFAICT the locks in Load are unnecessary. The CLI will instantiate its own instance of the manager and sync with the running share manager service using the cs3 backend.

anyway, I can drop the mutex completely from LOAD, can't I?

kobergj: Yes I think so. `Add` methods are only called by one go routine. If further syncing is needed I would build that directly into the `Add` method. But not needed for now.

pkg/share/manager/jsoncs3/jsoncs3.go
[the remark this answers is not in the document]
kobergj: That is exactly the problem. How should I spot an error as admin if my terminal is full of success messages?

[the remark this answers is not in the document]
butonic: k, decreasing to debug

kobergj (changes requested):

kobergj (approved):
Now I like it :+1:

PR #3172, opened 2022-08-24
[tests-only]Bump core commit id for tests

SwikritiT:
Part of: https://github.com/owncloud/QA/issues/761


SwikritiT:
@phil-davis These tests are currently failing in master passes in edge and ocis 
```
apiSharePublicLink1/createPublicLinkShare.feature:141
apiSharePublicLink1/createPublicLinkShare.feature:142
```
These tests were touched by PR https://github.com/owncloud/core/commit/75da839eb41d710557886339ff3934e566fe328e#
Any idea on why they might be failing here?

phil-davis:
> Any idea on why they might be failing here?

core PR https://github.com/owncloud/core/pull/40264 added a feature to oC10 "allow editing of public link shared single files". That was reverted out of core master yesterday by PR https://github.com/owncloud/core/pull/40312 - and the matching test changes were also reverted.

So the test results in reva might also change. Have a look in the history of the expected-failures file. There should be an old entry for these test scenarios failing. You can put that back into expected-failures.

individual-it (approved):

phil-davis (approved):

PR #3175, opened 2022-08-24
[tests-only] keep etag consistent while uploading

fschade:
propagate etag only when post processing is finished.

### DIFF
--- a/pkg/storage/utils/decomposedfs/decomposedfs.go
+++ b/pkg/storage/utils/decomposedfs/decomposedfs.go
@@ -199,7 +199,7 @@ func (fs *Decomposedfs) Postprocessing(ch <-chan interface{}) {
 				failed = true
 				keepUpload = true
 			case events.PPOutcomeContinue:
-				if err := upload.Finalize(up); err != nil {
+				if err := up.Finalize(); err != nil {
 					log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("could not finalize upload")
 					keepUpload = true // should we keep the upload when assembling failed?
 					failed = true
@@ -244,7 +244,8 @@ func (fs *Decomposedfs) Postprocessing(ch <-chan interface{}) {
 				log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("Failed to set scan results")
 				continue
 			}
-
+		default:
+			log.Error().Interface("event", ev).Msg("Unknown event")
 		}
 	}
 
--- a/pkg/storage/utils/decomposedfs/upload/processing.go
+++ b/pkg/storage/utils/decomposedfs/upload/processing.go
@@ -25,7 +25,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strconv"
+	"time"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
@@ -38,6 +38,8 @@ import (
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/lookup"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/options"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/xattrs"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/filelocks"
 	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
@@ -214,16 +216,7 @@ func Get(ctx context.Context, id string, lu *lookup.Lookup, tp Tree, fsRoot stri
 	ctx = appctx.WithLogger(ctx, &sub)
 
 	up := buildUpload(ctx, info, info.Storage["BinPath"], infoPath, lu, tp, pub, async, tknopts)
-
-	oldsize, err := strconv.ParseUint(info.MetaData["oldsize"], 10, 64)
-	if err == nil {
-		up.oldsize = &oldsize
-	}
-
-	// add node if possible - only in async case
-	if n, err := node.ReadNode(ctx, lu, info.Storage["SpaceRoot"], info.Storage["NodeId"], false); err == nil && async {
-		up.Node = n
-	}
+	up.versionsPath = info.MetaData["versionsPath"]
 	return up, nil
 }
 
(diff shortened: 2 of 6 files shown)


pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -1173,12 +1173,24 @@ func (n *Node) FindStorageSpaceRoot() error {
...
 
 // UnmarkProcessing removes the processing flag from the node
 func (n *Node) UnmarkProcessing() error {
-	return n.RemoveMetadata(xattrs.StatusPrefix)
+	return n.MetaChange(func() error { return n.RemoveMetadata(xattrs.StatusPrefix) })
+}
+
+// MetaChange executes an operation on the node without changing the underlying files mtime
+func (n *Node) MetaChange(op func() error) error {
+	fi, err := os.Stat(n.InternalPath())
```
dragotin, line 1186: I think this stat call is not needed. The provided func should (and will) be prepared to that error. Im most cases it is just another (time consuming) stat.

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -228,23 +228,32 @@ func (fs *Decomposedfs) Postprocessing(ch <-chan interface{}) {
...
-				log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("Failed to set scan results")
-				continue
-			}
-
+		default:
+			log.Error().Interface("event", ev).Msg("Unknown event")
+
+			// NOTE: Listening to the VirusscanFinished event enables the server to store the "lastScanned" date
+			// For a nice UX this should be displayed by the user.
+			// However enabling it requires the decision if the etag should change after a virusscan (and proper testing to verify this)
```
dragotin, line 236: Currently an ETag change would cause a re-download of the file for most clients. We need a solution for metadata-only changes, but currently we are not there. Compare this fairly old issue: https://github.com/owncloud/client/issues/4755

I'd say we stay with the current approach to NOT change the ETag after this kind of metadata has changed.

dragotin (approved):

pkg/storage/utils/decomposedfs/upload/processing.go
```diff
@@ -292,21 +284,61 @@ func CreateNodeForUpload(upload *Upload) (*node.Node, error) {
...
+
+		vfi, err := os.Stat(old.InternalPath())
+		if err != nil {
+			return nil, err
+		}
+
+		// When the if-match header was set we need to check if the
+		// etag still matches before finishing the upload.
+		if ifMatch, ok := upload.Info.MetaData["if-match"]; ok {
+			targetEtag, err := node.CalculateEtag(n.ID, vfi.ModTime())
```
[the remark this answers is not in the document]
kobergj: Mmh. What if another process has already edited the file but didn't update the etag yet? However this timeframe is very short, so we could theoretically go with this risk.

pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -208,10 +175,70 @@ func (upload *Upload) FinishUpload(_ context.Context) error {
...
+		}
+		defer f.Close()
+
+		r1 := io.TeeReader(f, sha1h)
+		r2 := io.TeeReader(r1, md5h)
+
+		_, err = io.Copy(adler32h, r2)
+		if err != nil {
+			log.Info().Err(err).Msg("error copying checksums")
+		}
```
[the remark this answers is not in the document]
kobergj: Valid points but out of the scope for this PR. Changing checksum logic will need alignment with all clients.

pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -208,10 +175,70 @@ func (upload *Upload) FinishUpload(_ context.Context) error {
...
+		case "adler32":
+			err = upload.checkHash(parts[1], adler32h)
+		default:
+			err = errtypes.BadRequest("unsupported checksum algorithm: " + parts[0])
+		}
+		if err != nil {
+			Cleanup(upload, true, false)
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
kobergj: Mmh. I don't quite get you. How should it know the checksum if it wasn't send with the request?

pkg/storage/utils/decomposedfs/node/node.go
kobergj: The goal of `MetaChange` method is to change the xattrs of a file without changing its mtime.  That means I need to
- get current mtime and store in a variable
- do the change
- overwrite new mtime with stored one

I don't get how to get current mtime without a stat? Or maybe I got you wrong?

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj: I'm afraid that is NOT the current approach. The etag is calculated from the mtime of the file. If any extended attribute is added the mtime is changed and therefore the etag changes.

pkg/storage/utils/decomposedfs/upload/processing.go
```diff
@@ -292,21 +284,61 @@ func CreateNodeForUpload(upload *Upload) (*node.Node, error) {
...
+				[>
+				return strings.HasPrefix(attributeName, xattrs.GrantPrefix) || // for grants
+				strings.HasPrefix(attributeName, xattrs.MetadataPrefix) || // for arbitrary metadata
+				strings.HasPrefix(attributeName, xattrs.FavPrefix) || // for favorites
+				strings.HasPrefix(attributeName, xattrs.SpaceNameAttr) || // for a shared file
+			*/
+		}); err != nil {
 			return nil, err
 		}
 	}
```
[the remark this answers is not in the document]
kobergj: Added a lock. Please recheck...

pkg/storage/utils/decomposedfs/decomposedfs.go
dragotin: Not true. When *only* an xattr is added or changed, the mtime is NOT changed, but only the ctime! As long as we only look on the mtime, there would not be a change.

pkg/storage/utils/decomposedfs/node/node.go
dragotin: I am sorry, I overlooked that. Forget it.

pkg/storage/utils/decomposedfs/upload/upload.go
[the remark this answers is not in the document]
dragotin: I created 

https://github.com/owncloud/ocis/issues/4529

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj: Reenabled functionality after yesterdays discovery.

dragotin (approved):

PR #3184, opened 2022-09-01
Exclude vendor folder from docker build context

michielbdejong:
I got an 'inconsistent vendoring' error when running 'docker build -f Dockerfile.revad .'
This change made the error go away.
Is there a reason not to exclude the the vendor folder from the docker build context?

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,5 +4,6 @@ docs
 grpc-tests
 tests
 tools
+vendor
 vendor-bin
 *.md


butonic:
@wkloucek I think this makes sense, pinging you because you are more familiar with docker containers than I am.

butonic:
@michielbdejong can you add a changelog to make CI happy?

michielbdejong:
Closing as stale.

PR #3189, opened 2022-09-02
[tests-only] Default group quotas

kobergj:
Allows setting quotas per group

### DIFF
--- a/pkg/bytesize/bytesize.go
+++ b/pkg/bytesize/bytesize.go
@@ -1,0 +1,109 @@
+// Copyright 2018-2022 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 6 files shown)


pkg/bytesize/bytesize.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Bytes converts the ByteSize to an uint64
+func (b ByteSize) Bytes() uint64 {
+	return uint64(b)
+}
+
+// String converts the ByteSize to a string
+func (b ByteSize) String() string {
+	return strconv.Itoa(int(b))
```
C0rby, line 108: I think `strconv.FormatUint(uint64(b), 10)` might be safer here in case of big values in b but `strconv.Itoa(int(b))` is probably good enough 99.99% of the time.

C0rby (approved):

pkg/bytesize/bytesize.go
kobergj: :+1: Let's go with 100% solution then

rhafer (approved):

PR #3192, opened 2022-09-05
enable space member to update share

C0rby:
Enabled space members to update shares which they have not created themselves.

@butonic, did I understand this new approach correctly?

### DIFF
--- a/pkg/share/manager/jsoncs3/jsoncs3.go
+++ b/pkg/share/manager/jsoncs3/jsoncs3.go
@@ -28,12 +28,15 @@ import (
 	"github.com/pkg/errors"
 	"google.golang.org/genproto/protobuf/field_mask"
 
+	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	collaboration "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
+	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/v2/pkg/share"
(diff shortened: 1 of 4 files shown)


pkg/share/manager/jsoncs3/jsoncs3.go
```diff
@@ -355,6 +355,13 @@ func (m *Manager) GetShare(ctx context.Context, ref *collaboration.ShareReferenc
 	if share.IsCreatedByUser(s, user) || share.IsGrantedToUser(s, user) {
 		return s, nil
 	}
+
+	shares := m.Cache.ListSpace(s.ResourceId.StorageId, s.ResourceId.SpaceId)
+	if _, ok := shares.Shares[s.GetId().GetOpaqueId()]; ok {
+		// Members of a space are also allowed to list the share.
```
micbar, line 361: Members of a space are not always allowed to list the shares. We need to check the `ListGrants` permission.

pkg/share/manager/jsoncs3/jsoncs3.go
```diff
@@ -425,7 +432,11 @@ func (m *Manager) UpdateShare(ctx context.Context, ref *collaboration.ShareRefer
 
 	user := ctxpkg.ContextMustGetUser(ctx)
 	if !share.IsCreatedByUser(s, user) {
-		return nil, errtypes.NotFound(ref.String())
+		shares := m.Cache.ListSpace(s.ResourceId.StorageId, s.ResourceId.SpaceId)
+		// Members of a space are also allowed to list the share.
```
micbar, line 436: same

pkg/share/manager/jsoncs3/jsoncs3.go
micbar: but the ocs part has a validation step. Maybe that is sufficient.

micbar: We could also decide that the share manager always returns the shares and the ocs filters it. This is the case for the `in-memory` and `cs3` share managers.

C0rby: > We could also decide that the share manager always returns the shares and the ocs filters it.

I don't like this. The caller of the share service shouldn't receive shares which the current user shouldn't be able to see.

> This is the case for the in-memory, json and cs3 share managers.

The managers are also stating the files to check the permissions. At least that was the case when I implemented that a couple months ago.

C0rby: > Members of a space are not always allowed to list the shares. We need to check the `ListGrants` permission.

@micbar @butonic, I think we should maybe have a short call about this, because this is not how it is implemented right now.

micbar: call has happened, we decided to add permission checks to the share manager if the calling user is neither `creator` nor `grantee`.

kobergj (approved):

micbar (approved):

PR #3196, opened 2022-09-06
[tests-only] On demand and on access scanning

kobergj:
Adds possibility to start postprocessing steps on demand. At first virus scan should be supported

Also enables on access scanning (scan before download) when configured


fschade (approved):
LGTM ❤️

PR #3204, opened 2022-09-07
Make FlockFile function private to not expose it for misuse.

dragotin:
This function should rather be private to not be "misused" from outside the module.


C0rby (approved):

butonic (approved):

changelog/unreleased/makeflockfileprivate.md
```diff
@@ -0,0 +1,8 @@
+Enhancement: Make the function flockFile private.
```
butonic, line 1:
```suggestion
Enhancement: Make the function flockFile private
```

PR #3214, opened 2022-09-09
Revamp cback storage driver + http service

gmgigi96:


labkode:
@gmgigi96 is this still needed?

gmgigi96:
> @gmgigi96 is this still needed?

Yes, it is, but let's wait before merging it

PR #3232, opened 2022-09-13
parse reference without path as id only

butonic:
alternative to https://github.com/owncloud/ocis/pull/4573


butonic:
closing because it would leak the path in this scenario:
* a is a manager of space s
* a shares `/secret/folder` with b
* b shares `/Shares/folder/file.txt` via link
* the response should not leak `/secret/folder/file.txt`

PR #3240, opened 2022-09-14
We now only encode &, < and > in PROPFIND PCDATA

butonic:
Signed-off-by: Jörn Friedrich Dreyer <jfd@butonic.de>


internal/http/services/owncloud/ocdav/prop/prop.go
```diff
@@ -58,13 +59,78 @@ func EscapedNS(namespace string, local string, val string) PropertyXML {
...
+// > elsewhere, they must be escaped using either numeric character references
+// > or the strings " &amp; " and " &lt; " respectively. The right angle
+// > bracket (>) may be represented using the string " &gt; ", and must, for
+// > compatibility, be escaped using either " &gt; " or a character reference
+// > when it appears in the string " ]]> " in content, when that string is not
+// > marking the end of a CDATA section.
+//
+// The code ignores errors as the legacy Escaped() does
 // TODO properly use the space
 func Escaped(key, val string) PropertyXML {
```
kobergj, line 99: I would suggest to move this to a dedicated pkg. We want the search REPORT endpoint to be consistent with the PROPFIND endpoint. But we can't import it from `internal` path

butonic: moving some of the ocdav specific stuff to `pkg` makes sense, right. extracting a dedicated oc flavoured webdav package as `github.com/owncloud/go-webdav` that others can reuse is the right way to go.

But not as part of this PR.

internal/http/services/owncloud/ocdav/prop/prop.go
```diff
@@ -58,13 +59,78 @@ func EscapedNS(namespace string, local string, val string) PropertyXML {
...
+		i += width
+		switch r {
+		case '&':
+			esc = escAmp
+		case '<':
+			esc = escLT
+		case '>':
+			esc = escGT
+		default:
+			if !isInCharacterRange(r) || (r == 0xFFFD && width == 1) {
```
rhafer, line 115: Nit: Would you mind changing the 0xFFFD to the utf.RuneError  constant? Same for the escFFFD variable defined above.

internal/http/services/owncloud/ocdav/prop/prop.go
```diff
@@ -58,13 +59,78 @@ func EscapedNS(namespace string, local string, val string) PropertyXML {
...
+// > bracket (>) may be represented using the string " &gt; ", and must, for
+// > compatibility, be escaped using either " &gt; " or a character reference
+// > when it appears in the string " ]]> " in content, when that string is not
+// > marking the end of a CDATA section.
+//
+// The code ignores errors as the legacy Escaped() does
 // TODO properly use the space
 func Escaped(key, val string) PropertyXML {
+	w := new(bytes.Buffer)
+	s := []byte(val)
```
rhafer, line 101: Nit: If we talk about (micro-)optimization here. This might be a bit better:
```
s := []byte(val)
w := bytes.NewBuffer(make([]byte,0,len(s)))
```

rhafer (changes requested):

rhafer (approved):

PR #3242, opened 2022-09-15
[tests-only] Read group quotas from file

kobergj:
We decided to read group quotas from file rather than from EnvVar. Otherwise it is not possible to use group names like `plutonium-lovers`


fschade (approved):
LGTM

PR #3250, opened 2022-09-19
gateway: use go-micro store to cache

butonic:
This PR replaces the in memory caches of the gateway with the [go-micro store interface](https://github.com/go-micro/go-micro/blob/master/store/store.go). It allows sharing the cache using redis or etcd for redundant deployments and the correct timing of stat cache invalidation because the dataprovider can now invalidate the cache when the upload has finished.

first step for https://github.com/owncloud/ocis/issues/4427

- [x] allow datagateway to clean cache when file is available


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1089,8 +1089,7 @@ func (s *svc) getStorageProviderClient(_ context.Context, p *registry.ProviderIn
 		return nil, err
 	}
 
-	// TODO: reinstate cache as soon there is a implementation for a non-in-memory cache
-	return c, nil
+	return s.cache.StorageProviderClient(c), nil
```
butonic, line 1092: we are using a noop cache by default and ttls default to 0, also disabling the cache. We need a shared cache to make the dataprovider invalidate the stat cache when a file is written.

pkg/rhttp/datatx/manager/tus/tus.go
```diff
@@ -117,6 +124,7 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 					},
 					Path: utils.MakeRelativePath(filepath.Join(info.MetaData["dir"], info.MetaData["filename"])),
 				}
+				datatx.InvalidateCache(owner, ref, m.statCache)
```
aduffeck, line 127: Putting the cache invalidation here means that it won't happen if event publishing is disabled. I don't think those should be tied to each other.

butonic: correct, digging into it

aduffeck (approved):

PR #3252, opened 2022-09-20
add spaceshared event

dragonchaser:
We added an event that is emmitted when somebody shares a space.

### DIFF
--- a/changelog/unreleased/add-spaceshared-event.md
+++ b/changelog/unreleased/add-spaceshared-event.md
@@ -1,0 +1,7 @@
+Enhancement: Add SpaceShared event
+
+We added an event that is emmitted when somebody shares a space.
+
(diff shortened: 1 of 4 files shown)


butonic (approved):

changelog/unreleased/add-spaceshared-event.md
```diff
@@ -0,0 +1,7 @@
+Enhancement: Add SpaceShared event
+
+We added an event that is emmitted when somebody shares a space.
+
+https://github.com/owncloud/ocis/pull/4564
```
micbar, line 5: @butonic @dragonchaser Please check during reviews that the first link is always the reva pull request. This changelog entry is broken, cannot be rendered because the first id doesn't resolve inside of reva

PR #3255, opened 2022-09-21
ocdav: Properly escape oc:name in propfind response

rhafer:
The oc:name property in the ocdav propfind response might contain XML special characters. We now apply the proper escaping on that property

https://github.com/owncloud/ocis/issues/4474


C0rby (approved):
Nice!

PR #3261, opened 2022-09-22
Metadatastore

butonic:
Decomposedfs can now store metadata in JSON
- [ ] refactor all xattr access into node specific methods
- [ ] introduce MetadataStore interface and pass it down to Node instances
- [ ] move xattr access to MetadataStore implementation
- [ ] make sure tests still work
- [ ] implement new JSON backed Metadata store


butonic:
@micbar @dragotin hm ... writing metadata to the file stubs works, but directories currently are used to represend the child hierarchy. Where do we put their metadata? we could store the children as a list in the json file ... 🤔

butonic:
superseded by https://github.com/cs3org/reva/pull/3657

PR #3272, opened 2022-09-26
[tests-only] Add blobstore metrics

dragonchaser:
Enhancement: Add s3ng metrics

We have added rx/tx metrics to the s3ng blobstore driver.

### DIFF
--- a/pkg/storage/fs/s3ng/blobstore/blobstore.go
+++ b/pkg/storage/fs/s3ng/blobstore/blobstore.go
@@ -30,6 +30,7 @@ import (
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 // Blobstore provides an interface to an s3 compatible blobstore
@@ -38,6 +39,20 @@ type Blobstore struct {
 
 	bucket string
 }
+
+// PrometheusAwareReader provides an interace to an prometheus aweare Reader
+type PrometheusAwareReader struct {
+	r io.Reader
+	m *prometheus.CounterVec
+}
+
+// PrometheusAwareReadCloser provides an interface to a prometheus aware ReadCloser
+type PrometheusAwareReadCloser struct {
+	r io.ReadCloser
+	m *prometheus.CounterVec
+}
+
+var metrics = NewMetrics()
 
 // New returns a new Blobstore
 func New(endpoint, region, bucket, accessKey, secretKey string) (*Blobstore, error) {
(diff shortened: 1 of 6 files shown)


pkg/storage/fs/s3ng/blobstore/blobstore.go
```diff
@@ -62,8 +77,31 @@ func New(endpoint, region, bucket, accessKey, secretKey string) (*Blobstore, err
...
+func (p *PrometheusAwareReadCloser) Close() error {
+	return p.r.Close()
+}
+
 // Upload stores some data in the blobstore under the given key
 func (bs *Blobstore) Upload(node *node.Node, reader io.Reader) error {
+	reader = &PrometheusAwareReader{
+		r: reader,
+		m: metrics.Tx,
+	}
```
kobergj, line 104: why can't we use the `PrometheusAwareReadCloser`? A `ReadCloser` is also a `Reader`?

pkg/storage/fs/s3ng/blobstore/metrics.go
```diff
@@ -0,0 +1,57 @@
...
+		}, []string{}),
+		Tx: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: Namespace,
+			Subsystem: Subsystem,
+			Name:      "Tx",
+			Help:      "Storage access tx",
+		}, []string{}),
+	}
+	_ = prometheus.Register(m.Rx)
+	_ = prometheus.Register(m.Tx)
```
kobergj, line 54: This will ignore the errors. Should we use `MustRegister` (to crashloop the pod) or at least log the error so we are aware of it?

kobergj (changes requested):
Some open questions

pkg/storage/fs/s3ng/blobstore/blobstore.go
dragonchaser: Because it will require lots of changes, which will add cause lots of refactoring when we merge `edge` into our branch again. I tried to avoid this.

pkg/storage/fs/s3ng/blobstore/metrics.go
dragonchaser: Makes sense, will adapt the code.

pkg/storage/fs/s3ng/blobstore/blobstore.go
kobergj: No, I mean just here in this line. Why don't we do 
```suggestion
	reader = &PrometheusAwareReadCloser{
		r: reader,
		m: metrics.Tx,
	}
```
instead? Then `PrometheusAwareReader` would be unused and could be removed...

kobergj (approved):

fschade:
@labkode, ci seems to be pretty flaky today. Known on your side?

PR #3282, opened 2022-09-27
better user names in wopi apps

micbar:
# Description

Use better usernames in WOPI apps.

- Display Name for authenticated users
- Don't set the owner's userid when editing a publicshare, that could lead to wrong assumtions


glpatcern:
Strikingly similar to https://github.com/cs3org/reva/pull/3280, which I was working on a couple of days ago.

The `userid` parameter though is needed by the wopiserver IIRC...

micbar:
> Changed the PR to keep it

butonic (approved):

PR #3286, opened 2022-09-29
add optional existingLockID to RefreshLockRequest in the decomposedFS

micbar:
Enhancement: Make Refresh Lock operation WOPI compliant

We now support the WOPI compliant `UnlockAndRelock` operation. This has been implemented in the DecomposedFS. To make use of it, we need a compatible WOPI server.

https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/unlockandrelock


glpatcern (approved):
Looks good to me, yet considering the pure `refreshLock` logic also "correct" if the existing lock is simply overwritten (the eosfs way).

Will patch that one as well to take the new argument into account of course, but could you please push this to `master`?

micbar:
@glpatcern please check out https://github.com/cs3org/reva/pull/3289 for the master branch.

Unfortunately we have no `PreconditionFailed` Return code on master branch.

butonic (approved):

PR #3288, opened 2022-09-29
oidc driver: get user from user provider

gmgigi96:
As some claim used in the oidc driver are not standard, if the uid or the gid are not available from the response of the provider, resolve the user from the user provider.

### DIFF
--- a/pkg/auth/manager/oidc/oidc.go
+++ b/pkg/auth/manager/oidc/oidc.go
@@ -359,7 +359,7 @@ func (am *mgr) resolveUser(ctx context.Context, claims map[string]interface{}) e
 	}
 
 	// take the properties of the mapped target user to override the claims
-	claims["preferred_username"] = value
+	claims["preferred_username"] = getUserByClaimResp.GetUser().Username
 	claims[am.c.IDClaim] = getUserByClaimResp.GetUser().GetId().OpaqueId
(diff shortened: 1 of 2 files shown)


pkg/auth/manager/oidc/oidc.go
```diff
@@ -320,32 +330,41 @@ func (am *mgr) resolveUser(ctx context.Context, claims map[string]interface{}) e
 			return errtypes.PermissionDenied("no user mapping found for the given group claim(s)")
 		}
 		for _, m := range intersection {
-			username = am.oidcUsersMapping[m.(string)].Username
+			value = am.oidcUsersMapping[m.(string)].Username
 		}
+		resolve = true
+	} else if uid == 0 || gid == 0 {
+		claim = "mail"
```
glpatcern, line 337: Are you sure this is `mail` and not `email` ?

gmgigi96: Yes https://github.com/cs3org/reva/blob/master/pkg/user/manager/ldap/ldap.go#L218, for example

glpatcern: I see, thanks - what is quite misleading is that just the line after we have `email` as claim from the OIDC token just parsed. Maybe just add a comment to say this is a claim used against the user provider, otherwise it's fine.

glpatcern (approved):

PR #3291, opened 2022-09-29
Enhancement: improve the wopi app driver

micbar:
# Description

The wopi app driver supports more options

We now generate a folderurl that is used in the wopi protocol to provide an endpoint to go back from the app to the containing folder in the file list. In addition to that we now inlclude the UI_LLCC parameter in the app open URL.

https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#ui_llcc

### DIFF
--- a/pkg/storage/utils/templates/templates.go
+++ b/pkg/storage/utils/templates/templates.go
@@ -33,31 +33,42 @@ import (
 
 	"github.com/Masterminds/sprig"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	providerv1beta1 "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/v2/pkg/storagespace"
 	"github.com/pkg/errors"
 )
 
 // UserData contains the template placeholders for a user.
 // For example {{.Username}} or {{.Id.Idp}}
-type UserData struct {
-	*userpb.User
-	Email EmailData
-}
+type (
+	UserData struct {
+		*userpb.User
+		Email EmailData
+	}
 
-// SpaceData contains the templace placeholders for a space.
-// For example {{.SpaceName}} {{.SpaceType}} or {{.User.Id.OpaqueId}}
-type SpaceData struct {
-	*UserData
-	SpaceType string
-	SpaceName string
-}
+	// SpaceData contains the templace placeholders for a space.
+	// For example {{.SpaceName}} {{.SpaceType}} or {{.User.Id.OpaqueId}}
+	SpaceData struct {
+		*UserData
+		SpaceType string
+		SpaceName string
+	}
 
-// EmailData contains mail data
-// split into local and domain part.
-// It is extracted from splitting the username by @.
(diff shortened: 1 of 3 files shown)


glpatcern:
This is also a welcome improvement in `master`. How are you going to generate the `PrivateLinkBaseURL`?

butonic (approved):

pkg/storage/utils/templates/templates.go
```diff
@@ -33,31 +33,42 @@ import (
...
+	EmailData struct {
+		Local  string
+		Domain string
+	}
+
+	// ResourceData contains the ResourceInfo
+	// ResourceData.ResourceId is a stringified form of ResourceInfo.Id
+	ResourceData struct {
+		ResourceInfo *providerv1beta1.ResourceInfo
+		ResourceId   string
```
micbar, line 69: This is a template variable, the capital ID would be confusing to people.

micbar:
> How are you going to generate the PrivateLinkBaseURL?
I plan to use the private link which not be user facing in the future in ocis but ownCloud Web will redirect.

pkg/storage/utils/templates/templates.go
```diff
@@ -95,6 +106,24 @@ func WithSpacePropertiesAndUser(u *userpb.User, spaceType string, spaceName stri
...
 
+// WithResourceInfo renders template stings with ResourceInfo variables
+func WithResourceInfo(i *providerv1beta1.ResourceInfo, tpl string) string {
+	tpl = clean(tpl)
+	data := newResourceData(i)
+	// compile given template tpl
+	t, err := template.New("tpl").Funcs(sprig.TxtFuncMap()).Parse(tpl)
+	if err != nil {
+		err := errors.Wrap(err, fmt.Sprintf("error parsing template: fileinfoandresourceid_template:%+v tpl:%s", data, tpl))
+		panic(err)
```
kobergj, line 117: Please don't panic. Return the error or ignore it.

pkg/storage/utils/templates/templates.go
```diff
@@ -95,6 +106,24 @@ func WithSpacePropertiesAndUser(u *userpb.User, spaceType string, spaceName stri
...
+	}
+	b := bytes.Buffer{}
+	if err := t.Execute(&b, data); err != nil {
+		err := errors.Wrap(err, fmt.Sprintf("error executing template: fileinfoandresourceid_template:%+v tpl:%s", data, tpl))
+		panic(err)
```
kobergj, line 122: See above.

kobergj (changes requested):

pkg/storage/utils/templates/templates.go
micbar: ok, that means we should also refactor the other templates?
IMO it was designed that way that you cannot start the serivice with a bad template.

kobergj: It would still not start if we return the error, right?

Apparently I didn't review the first part. If there are already panics we can keep these for now and refactor later.

kobergj (approved):

PR #3303, opened 2022-10-03
Backport from edge: added support for language locales in apps

glpatcern:
The goal here is to backport the minimal logic from `edge` to support different locales in WOPI apps, and drive this by configuration rather than by query parameters.

In addition, the `DC_LLCC` parameter should be used for Office 365 to compensate issues with incorrect dates computations depending on the locale:

https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/online/discovery#dc_llcc

### DIFF
--- a/changelog/unreleased/lang-supp-in-app.md
+++ b/changelog/unreleased/lang-supp-in-app.md
@@ -1,0 +1,10 @@
+Enhancement: added support for configuring language locales in apps
+
+This is a partial backport from edge: we introduce a language option
+in the appprovider, which if set is passed as appropriate parameter
+to the external apps in order to force a given localization. In particular,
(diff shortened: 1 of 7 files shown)


glpatcern:
@micbar as you authored https://github.com/cs3org/reva/pull/3291 you may want to see a different flavor for it:
* here we have based this on server-side configuration rather than client-side (opaque) parameters: we plan to force a locale in our deployment, otherwise the default behavior is autodetection by the application engine.
* for Office365, the parameters are `ui=<UI_LLCC>` and `rs=<DC_LLCC>`. `<UI_LLCC>` is a "placeholder", the query parameter names are to be taken from the `/hosting/discovery` XML endpoint.

PR #3304, opened 2022-10-03
Add http service to send email for shares

gmgigi96:
This PR adds an http service, exposed by default under `/mailer`, that for a given list of share ids, will send the email to the recipient of these shares.

Example:
`curl -X POST localhost:9143/mailer -u <user> -d "id=<share_id>"`
`{"recipients":["<email>"]}`


micbar:
JFYI: there is already an event based solution available https://github.com/owncloud/ocis/tree/master/services/notifications

labkode:
@micbar thanks, that is the next step. We have already a NATS cluster and we plan to create the hooks with the same payload as you have in edge so the transition will be easier.

PR #3305, opened 2022-10-03
Add description to public links + internal public links

gmgigi96:
This PR adds the possibility to set a description to a public link, allowing an user to describe the usefulness of the public link.
It also adds an option to set the public links for internal usage (for example for being using by apps). When a public link is internal, this will not be exposed in the list of public links.

### DIFF
--- a/changelog/unreleased/publiclink-description.md
+++ b/changelog/unreleased/publiclink-description.md
@@ -1,0 +1,3 @@
+Enhancement: Add description to public link
+
+https://github.com/cs3org/reva/pull/3305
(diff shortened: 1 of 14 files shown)


labkode (changes requested):
One more field in DB `hidden(boolean)` so the normal shares queries from the web will always be on type: `select * from oc_share where type = public link and hidden=false`, today we have codimd, but tomorrow we may have other apps and then we would need to touch again the normal query (not good).

glpatcern:
As just discussed with Gianmaria: let's call the extra field `internal` (bool), defaults = `false`, also in the CS3APIs. Of course will have it "exposed" for users to use it (if they discover/reverse engineer outside the web UI), that's fine after all.

gmgigi96:
The hidden in the db requires to modify the cs3apis, as now is not anymore cernbox specific. As we discussed with @glpatcern we could add a bool field named Internal, to specify that that public link is used for internal purpose (in the context of an application, for example), and so because was not created by the user, is hidden in the list. What do you think?

labkode:
Go ahead!

glpatcern (approved):

PR #3313, opened 2022-10-05
Fix content type so OCM verification works

labkode:
OCM verification is string comparison, and `param=value` is not recognised by the media parser as valid content-type.


gmgigi96:
LGTM!

PR #3332, opened 2022-10-10
Add settings to allow enabling TLS for grpc services

rhafer:
For grpc services TLS can be enabled per revad instance in the new
`[grpc.tls_settings]` section:

```
[grpc.tls_settings]`
enabled = true
certificate = "<path/to/cert.pem>"
key = "<path/to/key.pem>"
```

If `certificate` or `key` are empty a temporary self-signed certificate
is generated upon startup. The default for `enabled` is currently
`"false"`, i.e. transport security is disabled.

The configuration of grpc clients is currently only possible in the
`[shared.grpc_client_options]` section (not on a per client level):

```
[shared.grpc_client_options]
tls_mode = "on"
tls_cacert = "</path/to/cafile.pem>"

```

`tls_mode` has 3 possible values. `off` (the default) disables transport
security for the clients. `insecure` allows to use transport security,
but disables certificate verification (to be used with the autogenerated
self-signed certificates). `on` enables transport security.

`tls_cacert` is only relevant when `tls_mode` is set to `on` and allows
specifying a ca file/bundle to be used for verifying server
certificates.

For config examples see the toml files in tests/oc-integration-tests/local


rhafer:
The cs3api-validator tests are still failing with TLS enabled, because the validator doesn't know the new TLS related flags yet. It's a bit of a hen and egg issue. I guess it's easiest to keep TLS disabled for the tests and enable it after this change got merged (and the validator is updated to use the options).

changelog/unreleased/grpc-tls-settings.md
```diff
@@ -0,0 +1,42 @@
...
+to `true`, but no certificate and key files are supplied reva will generate
+temporary self-signed certificates at startup (this requires to also configure
+the clients to disable certificate verification, see below).
+
+The client side can be configured via the shared section. Set this to configure the CA for
+verifying server certificates:
+
+```toml
+[shared.grpc_client_options]
+tls_mode = "on"
```
rhafer, line 23: @micbar @C0rby Different to my initial approach, where I just had a `tls_insecure` boolean this  is introducing a third state for the client's tls setting which allows to completely turn off TLS (e.g. for debugging, or for isolated deployments where TLS is not really needed).

This is somewhat different to the way we configure TLS clients in other places, but I think we need this kind of switch. Are you ok with this, or I go back to the `tls_insecure=true/false` setting without allowing to turn TLS off.

C0rby: Yeah, I think it could make sense to turn TLS off in certain situations.
This means we need to adjust those other places but we can do that in separate PRs.

changelog/unreleased/grpc-tls-settings.md
```diff
@@ -0,0 +1,42 @@
...
+
+To disable server certificate verification (e.g. when using the autogenerated self-signed certificates)
+set:
+
+```toml
+[shared.grpc_client_options]
+tls_mode = "insecure"
+```
+
+To switch of TLS for the clients (which is also the default):
```
C0rby, line 35:
```suggestion
To switch off TLS for the clients (which is also the default):
```

C0rby (approved):

PR #3336, opened 2022-10-12
[tests-only] Return when running in errors on ListContainers 

kobergj:
Fixes part II of https://github.com/owncloud/KoKo22/issues/143


micbar (approved):
LGTM

PR #3339, opened 2022-10-12
[tests-only] Touch on 0 byte uploads

kobergj:
Touch 0 byte files instead uploading them (according to tus logic)

### DIFF
--- a/internal/http/services/owncloud/ocdav/put.go
+++ b/internal/http/services/owncloud/ocdav/put.go
@@ -150,6 +150,24 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	if length == 0 {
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocdav/put.go
```diff
@@ -150,6 +150,24 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	if length == 0 {
+		tfRes, err := client.TouchFile(ctx, &provider.TouchFileRequest{
+			Ref: ref,
+		})
+		if err != nil {
+			log.Error().Err(err).Msg("error sending grpc stat request")
```
micbar, line 158: shouldn't that be grpc TouchFile Request?

micbar (approved):
LGTM

PR #3373, opened 2022-10-19
[tests-only] fix high s3ng memory usage

wkloucek:
happened after #3272, fix is the same as #1940


kobergj (approved):

micbar:
@wkloucek Do we need to backport to edge?

PR #3376, opened 2022-10-19
add request id to important http handlers

micbar:
# Add request-id to logs in

- [x] ocdav
- [x] ocs
- [x] dataprovider
- [x] archiver


rhafer (commented):
I am completely against this, but wouldn't it be a lot cleaner to add the middleware in the same way how all the other middlewares are handled in pkg/rhttp/rhttp.go? Instead of touching every single service individually?

C0rby (approved):

PR #3397, opened 2022-10-26
Further improve lock contention

aduffeck:
This PR reduces lock contention by adding a xattr write-through cache for nodes.
    
The cache functions as a request-level xattr cache and reduces the number of *xattr syscalls being made.

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -386,12 +386,11 @@ func (n *Node) Owner() *userpb.UserId {
 func (n *Node) readOwner() (*userpb.UserId, error) {
 	owner := &userpb.UserId{}
 
-	rootNodePath := n.SpaceRoot.InternalPath()
 	// lookup parent id in extended attributes
 	var attr string
 	var err error
 	// lookup ID in extended attributes
-	attr, err = xattrs.Get(rootNodePath, xattrs.OwnerIDAttr)
+	attr, err = n.SpaceRoot.Xattr(xattrs.OwnerIDAttr)
 	switch {
 	case err == nil:
 		owner.OpaqueId = attr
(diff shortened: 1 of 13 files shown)


pkg/storage/utils/decomposedfs/node/xattrs.go
```diff
@@ -0,0 +1,84 @@
...
+// Xattr returns an extended attribute of the node. If the attributes have already
+// been cached it is not read from disk again.
+func (n *Node) Xattr(key string) (string, error) {
+	if n.xattrsCache == nil {
+		b, err := xattr.Get(n.InternalPath(), key)
+		if err != nil {
+			return "", err
+		}
+		return string(b), nil
+	}
```
kobergj, line 78: If I understood correctly, that means the cache is only used once I called `n.Xattrs()` once. If I call `n.Xattr(key)` multiple times in a row, the attribute will be read from disk every time. Is this intended behaviour?

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -435,26 +402,17 @@ func (n *Node) readOwner() (*userpb.UserId, error) {
 	}
 
 	// lookup IDP in extended attributes
-	attr, err = xattrs.Get(rootNodePath, xattrs.OwnerIDPAttr)
-	switch {
-	case err == nil:
-		owner.Idp = attr
-	case xattrs.IsAttrUnset(err):
-		// ignore
```
kobergj, line 443: Why don't we need this check any more?

pkg/storage/utils/filelocks/filelocks.go
```diff
@@ -85,6 +85,10 @@ func acquireLock(file string, write bool) (*flock.Flock, error) {
 		return nil, ErrPathEmpty
 	}
 
+	if _, err = os.Stat(file); err != nil {
+		return nil, err
+	}
```
kobergj, line 90: Why do we need to stat the file when we are not using its result?

pkg/storage/utils/decomposedfs/node/xattrs.go
aduffeck: This is actually the intended behavior, yes. There are use cases where only single attributes are fetched and no `listxattr` is ever called. Warming up the cache on single reads would mean making superfluous list calls in such cases.
We can't optimize for all cases at the same time, but the list calls are what is expensive, esp. regarding the lock contention so I think this is the better approach.

pkg/storage/utils/filelocks/filelocks.go
aduffeck: This is only about making sure that the file that is to be locked actually exists. I don't think we want to lock non-existent files.

pkg/storage/utils/decomposedfs/node/node.go
aduffeck: Hm, CI seems to be happy without it, but I removed this check intermittently during a bigger refactoring which I rolled back again, but I now agree that it should probably still be there. Let me try to add it back.

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -963,29 +907,25 @@ func (n *Node) IsDisabled() bool {
...
-	return n.SetMetadata(xattrs.ChecksumPrefix+csType, string(h.Sum(nil)))
+	return n.SetXattr(xattrs.ChecksumPrefix+csType, string(h.Sum(nil)))
 }
 
 // UnsetTempEtag removes the temporary etag attribute
 func (n *Node) UnsetTempEtag() (err error) {
-	err = xattrs.Remove(n.InternalPath(), xattrs.TmpEtagAttr)
-	if xattrs.IsAttrUnset(err) {
-		return nil
-	}
```
kobergj, line 987: This attrunset check is also removed. Not sure if this is critical

kobergj (approved):

pkg/storage/utils/decomposedfs/node/node.go
aduffeck: This error is only logged further up the stack, and it shouldn't occur in the first place.

PR #3401, opened 2022-10-26
AppProvider: drop hardcoded config for CodiMD

glpatcern:
This required to have some extra logic to handle the mime types <-> file extensions mapping, with a PR due in https://github.com/cubewise-code


pkg/app/provider/wopi/wopi.go
```diff
@@ -265,7 +259,7 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
 		}
 	}
 
-	log.Info().Msg(fmt.Sprintf("wopi: returning app URL %s", appFullURL))
+	log.Info().Msgf("wopi: returning app URL %s", appFullURL)
```
micbar, line 262: Structured logging should be always prefered over `Msgf` or `Sprintf` usage. In this case you should add the url with `.Str("url", appFullURL)`

glpatcern: Makes sense, thanks. Will have a look at a few others, the logging in this part of the software was/is pretty incomplete...

labkode (approved):

PR #3402, opened 2022-10-27
Block users

gmgigi96:
This PR allows an operator to set a list of users (username) that are banned for every operation in reva.
This requires that every service is restarted to get the new list if a user is added!

Example:
Given the storage-references example.
Add in gateway.toml:
```
[shared]
blocked_users = ["einstein"]
```

```
$ cmd/reva/reva -host localhost:19000 -insecure
reva-cli v1.19.0-124-ga4ffa8aa (rev-a4ffa8aa)
Please use `exit` or `Ctrl-D` to exit this program.
>> login basic
username: einstein
password: error: code=CODE_PERMISSION_DENIED msg="user is blocked" support_trace="ac860997b02743ef52b9d5bdab4311d0"
```

Even with a valid token:
```
>> whoami
rpc error: code = PermissionDenied desc = user einstein blocked
```

### DIFF
--- a/internal/grpc/interceptors/auth/auth.go
+++ b/internal/grpc/interceptors/auth/auth.go
@@ -33,6 +33,7 @@ import (
 	"github.com/cs3org/reva/pkg/sharedconf"
 	"github.com/cs3org/reva/pkg/token"
 	tokenmgr "github.com/cs3org/reva/pkg/token/manager/registry"
+	"github.com/cs3org/reva/pkg/user"
 	"github.com/cs3org/reva/pkg/utils"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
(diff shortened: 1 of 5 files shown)


internal/grpc/interceptors/auth/auth.go
```diff
@@ -50,6 +51,7 @@ type config struct {
 	TokenManager  string                            `mapstructure:"token_manager"`
 	TokenManagers map[string]map[string]interface{} `mapstructure:"token_managers"`
 	GatewayAddr   string                            `mapstructure:"gateway_addr"`
+	blockedUsers  []string
```
labkode, line 54: Uppercase

labkode:
@gmgigi96 looks good, can you just add an example on the config change?

labkode (approved):

PR #3432, opened 2022-11-03
remove share jail check

dragonchaser:
We have removed the share jail check.

fixes https://github.com/owncloud/ocis/issues/4945


butonic (approved):
References are currently not supported for now, they can be brought back when we change the Storage interface to use a Reference instead of a path.

PR #3436, opened 2022-11-07
Allow updating to internal link

kobergj:
Allow updating a link to a private link. See https://github.com/owncloud/ocis/issues/4979

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
@@ -36,7 +36,6 @@ import (
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/publicshare"
 	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
-	"github.com/cs3org/reva/v2/pkg/storage/utils/grants"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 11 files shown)


kobergj:
@micbar could you check my solution? It seems to a problem of `private links` vs. `denials`. Not sure my approach is the perfect one

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -355,9 +356,17 @@ func (h *Handler) updatePublicShare(w http.ResponseWriter, r *http.Request, shar
 		return
 	}
 
-	if !sufficientPermissions(statRes.Info.PermissionSet, newPermissions) {
-		response.WriteOCSError(w, r, http.StatusNotFound, "no share permission", nil)
-		return
+	// empty permissions mean private link here - NOT denial. Hence we need an extra check
```
micbar, line 359:
```suggestion
	// empty permissions mean internal link here - NOT denial. Hence we need an extra check
```

micbar (dismissed):
LGTM

kobergj:
@micbar needed some more changes to make this work. Please double check

micbar:
I cannot get the PR to work locally. The alias link cannot resolve. I am missing the alias=true prop in the response.

kobergj:
@micbar I needed to pass an `islink` parameter to the `RoleFromResourcePermissions` func. That's not nice but it is the only way we can differentiate between "denial" and "internal link". 

Please recheck and retest

micbar:
> Please recheck and retest

Works fine now.

micbar (approved):
:+1:

PR #3439, opened 2022-11-08
Refactor move error handling

butonic:
part of https://github.com/owncloud/ocis/issues/5009

### DIFF
--- a/internal/http/services/owncloud/ocdav/move.go
+++ b/internal/http/services/owncloud/ocdav/move.go
@@ -35,18 +35,14 @@ import (
 	"github.com/cs3org/reva/v2/pkg/rhttp/router"
 	"github.com/cs3org/reva/v2/pkg/storagespace"
 	"github.com/cs3org/reva/v2/pkg/utils"
-	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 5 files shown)


internal/http/services/owncloud/ocdav/move.go
```diff
@@ -26,203 +26,167 @@ import (
...
-		w.WriteHeader(http.StatusBadRequest)
-		return
+		return http.StatusBadRequest, err
 	}
 
 	for _, r := range nameRules {
 		if !r.Test(dstPath) {
-			w.WriteHeader(http.StatusBadRequest)
-			return
+			return http.StatusBadRequest, fmt.Errorf("path violates naming rules")
```
dragonchaser, line 56: can we have a reference in the error message where the rules are listed?

dragonchaser (changes requested):

internal/http/services/owncloud/ocdav/move.go
butonic: Hm, no: if a user reports this error he has a broken client. It is still possible to search for this error string and then follow `nameRules` to its definition. Pointing to source code will only get outdated.

We could turn `nameRules` into a map with rules and error messages that tell the client exactly what he did wrong, but that is outside the scope of this PR

PR #3441, opened 2022-11-09
Dav unit tests

butonic:
We added unit tests to cover more ocdav handlers:
- delete, increases coverage from 3.1% to 9.8%

part of https://github.com/owncloud/ocis/issues/5022

### DIFF
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
-	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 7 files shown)


internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -226,6 +227,13 @@ func NewWith(conf *Config, fm favorite.Manager, ls LockSystem, _ *zerolog.Logger
 	if err := s.davHandler.init(conf); err != nil {
 		return nil, err
 	}
+	if gwc == nil {
+		var err error
+		s.gwClient, err = pool.GetGatewayServiceClient(s.c.GatewaySvc)
```
aduffeck, line 232: How about moving this default behavior into `New` and just erroring out in here if the given client is `nil`?

aduffeck (changes requested):

internal/http/services/owncloud/ocdav/ocdav.go
butonic: I would have to duplicate it to `pkg/micro/ocdav/service.go` as well, otherwise we would change the behavior of the micro ocdav service, requiring a major version bump. So I put the fallback here, to make passing in a gwclient optional.

aduffeck (approved):
lgtm

PR #3443, opened 2022-11-10
simplify gateway client usage

butonic:
Since ocdav only fetches a gateway client once on startup we no longer need to check the error case whenever we call getClient() and can actually use the client of the service. Deleting all these unused error checks brings ocdav coverage from 9.8% to 10.1%

part of https://github.com/owncloud/ocis/issues/5022

### DIFF
--- a/internal/http/services/owncloud/ocdav/ocdav.go
+++ b/internal/http/services/owncloud/ocdav/ocdav.go
@@ -322,10 +322,6 @@ func (s *svc) Handler() http.Handler {
 	})
 }
 
-func (s *svc) getClient() (gateway.GatewayAPIClient, error) {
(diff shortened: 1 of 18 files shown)


internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -322,10 +322,6 @@ func (s *svc) Handler() http.Handler {
 	})
 }
 
-func (s *svc) getClient() (gateway.GatewayAPIClient, error) {
-	return s.gwClient, nil
-}
-
```
butonic, line 328: we never returned an error ... so getting rid of this

dragonchaser (approved):

PR #3445, opened 2022-11-10
add ocdav delete error case unit tests

butonic:
we now cover error cases in the ocdav delete handler in unit tests. increases coverage from 10.1% to 11.0%

part of https://github.com/owncloud/ocis/issues/5009

### DIFF
--- a/internal/http/services/owncloud/ocdav/delete.go
+++ b/internal/http/services/owncloud/ocdav/delete.go
@@ -76,7 +76,8 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
 	case res.Status.Code == rpc.Code_CODE_OK:
 		return http.StatusNoContent, nil
 	case res.Status.Code == rpc.Code_CODE_NOT_FOUND:
-		return http.StatusNotFound, errtypes.NotFound("Resource not found") // mimic the oc10 error message
+		//lint:ignore ST1005 mimic the exact oc10 error message
+		return http.StatusNotFound, fmt.Errorf("Resource not found")
 	case res.Status.Code == rpc.Code_CODE_PERMISSION_DENIED:
 		status = http.StatusForbidden
 		if lockID := utils.ReadPlainFromOpaque(res.Opaque, "lockid"); lockID != "" {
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -76,7 +76,8 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
 	case res.Status.Code == rpc.Code_CODE_OK:
 		return http.StatusNoContent, nil
 	case res.Status.Code == rpc.Code_CODE_NOT_FOUND:
-		return http.StatusNotFound, errtypes.NotFound("Resource not found") // mimic the oc10 error message
+		//lint:ignore ST1005 mimic the exact oc10 error message
+		return http.StatusNotFound, fmt.Errorf("Resource not found")
```
C0rby, line 80: If you aren't using string formats, you should use `errors.New("Resource not found")`.

internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -95,10 +96,10 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
 			// return not found error so we do not leak existence of a file
 			// TODO hide permission failed for users without access in every kind of request
 			// TODO should this be done in the driver?
-			return http.StatusNotFound, errtypes.NotFound("Resource not found") // mimic the oc10 error message
+			//lint:ignore ST1005 mimic the exact oc10 error message
+			return http.StatusNotFound, fmt.Errorf("Resource not found")
```
C0rby, line 100: Same here.

internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -95,10 +96,10 @@ func (s *svc) handleDelete(ctx context.Context, w http.ResponseWriter, r *http.R
...
 		}
-		// TODO path might be empty or relative...
-		return status, fmt.Errorf("permission denied to delete %v", ref.Path)
+		return status, fmt.Errorf("") // mimic the oc10 error messages which have an empty message in this case
```
C0rby, line 102: Same here, even though this  seems fishy? Why did oc10 have an empty message here?

internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -121,7 +122,7 @@ func (s *svc) handleSpacesDelete(w http.ResponseWriter, r *http.Request, spaceID
 	// we get a relative reference coming from the space root
 	// so if the path is "empty" we a referencing the space
 	if ref.GetPath() == "." {
-		return http.StatusForbidden, errtypes.PermissionDenied("deleting spaces via dav is not allowed")
+		return http.StatusMethodNotAllowed, fmt.Errorf("deleting spaces via dav is not allowed")
```
C0rby, line 125: And here

internal/http/services/owncloud/ocdav/delete.go
butonic: oc10 throws a lot of NotFoundException exceptions without a specific error message: https://github.com/owncloud/core/search?q=throw+new+NotFoundException

aduffeck (approved):

PR #3461, opened 2022-11-14
More dav unit tests

butonic:
brings ocdav coverage to 16.8%
- adds xml body to some bad requests
- starts reorganizing the ocdav unit tests:
  - per endpoint
    - per method
      - successful case
      - error cases
- fixes https://github.com/owncloud/ocis/issues/4332

### DIFF
--- a/internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
+++ b/internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
@@ -28,8 +28,9 @@ import (
 
 	cs3gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	cs3user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	cs3rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	cs3storageprovider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
-	typesv1beta1 "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
+	cs3types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/net"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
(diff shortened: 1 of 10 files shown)


internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
```diff
@@ -120,13 +134,383 @@ var _ = Describe("ocdav", func() {
...
+				return p == "/" || strings.HasPrefix(p, "/users")
+			})).Return(&cs3storageprovider.ListStorageSpacesResponse{
+				Status:        status.NewOK(ctx),
+				StorageSpaces: []*cs3storageprovider.StorageSpace{userspace},
+			}, nil)
+		})
+
+		Describe("PROPFIND", func() {
+
+			var rr *httptest.ResponseRecorder
```
aduffeck, line 189: How about defining those in a `var` block at the `Context` or the `Describe` level so they can be reused?

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
```diff
@@ -120,13 +134,383 @@ var _ = Describe("ocdav", func() {
...
+
+					client.On("Stat", mock.Anything, mock.MatchedBy(func(req *cs3storageprovider.StatRequest) bool {
+						return req.Ref.Path == "/users/username"
+					})).Return(&cs3storageprovider.StatResponse{
+						Status: status.NewNotFound(ctx, "resource not found"),
+					}, nil)
+
+					handler.Handler().ServeHTTP(rr, req)
+					Expect(rr).To(HaveHTTPStatus(http.StatusMultiStatus))
+					Expect(rr).To(HaveHTTPBody("<d:multistatus xmlns:s=\"http://sabredav.org/ns\" xmlns:d=\"DAV:\" xmlns:oc=\"http://owncloud.org/ns\"><d:response><d:href>/webdav</d:href><d:propstat><d:prop><oc:name>username</oc:name><d:resourcetype></d:resourcetype><d:getcontentlength>0</d:getcontentlength><oc:favorite>0</oc:favorite></d:prop><d:status>HTTP/1.1 200 OK</d:status></d:propstat><d:propstat><d:prop><oc:file-parent></oc:file-parent></d:prop><d:status>HTTP/1.1 404 Not Found</d:status></d:propstat></d:response></d:multistatus>"), "Body must not be empty")
```
aduffeck, line 230: I wonder how static the response xml is. Maybe we could just do something like `ContainSubstring("<d:status>HTTP/1.1 404 Not Found</d:status>")` instead to prevent false failures?

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
```diff
@@ -120,13 +134,383 @@ var _ = Describe("ocdav", func() {
...
+						return utils.ResourceEqual(req.Ref, &cs3storageprovider.Reference{
+							ResourceId: userspace.Root,
+							Path:       "./existingfolder",
+						})
+					})).Return(&cs3storageprovider.DeleteResponse{
+						Status: status.NewOK(ctx),
+					}, nil)
+
+					handler.Handler().ServeHTTP(rr, req)
+					Expect(rr).To(HaveHTTPStatus(http.StatusNoContent))
```
aduffeck, line 344: The expectation doesn't seem to match the test description (created vs. no content)

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
```diff
@@ -120,13 +134,383 @@ var _ = Describe("ocdav", func() {
...
+				rr = httptest.NewRecorder()
+				req, err = http.NewRequest("PROPFIND", basePath, strings.NewReader(""))
+				Expect(err).ToNot(HaveOccurred())
+				req = req.WithContext(ctx)
+
+			})
+			When("the gateway returns a file list", func() {
+				It("returns a multistatus with the file info", func() {
+
+					client.On("Stat", mock.Anything, mock.MatchedBy(func(req *cs3storageprovider.StatRequest) bool {
```
aduffeck, line 204: This mock is done quite often. Might be worth to add a helper function for that.

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
```diff
@@ -120,13 +134,383 @@ var _ = Describe("ocdav", func() {
...
+
+					// MKCOL needs to check if the resource already exists to return the correct status
+					client.On("Stat", mock.Anything, mock.MatchedBy(func(req *cs3storageprovider.StatRequest) bool {
+						return req.Ref.Path == "/users/username/newfolder"
+					})).Return(&cs3storageprovider.StatResponse{
+						Status: status.NewAborted(ctx, errors.New("parent does not exist"), ""),
+					}, nil)
+
+					handler.Handler().ServeHTTP(rr, req)
+					Expect(rr).To(HaveHTTPStatus(http.StatusPreconditionFailed))
```
aduffeck, line 290: The description says "conflict", which one is correct?

aduffeck (changes requested):
LGTM besides the nitpicks

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: moved

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: introduced, but could  not be used as often

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: I'll use sth like this:
```

					Expect(rr).To(HaveHTTPBody(
						And(
							ContainSubstring("<s:exception>Sabre\\DAV\\Exception\\MethodNotAllowed</s:exception>"),
							ContainSubstring("<s:message>The resource you tried to create already exists</s:message>"))),
						"Body must contain sabredav exception and message")
```

butonic: and the d:status is part of a propfind multistatus response xml, which is different then the http status of the response...

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: sorted out: 
- cs3 Aborted -> http precondition failed (eg mismatching lock id or etag)
- cs3 failed precondition -> http conftlict (missing intermediate dir) or method not allowed (dir already exists)

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: good catch

internal/http/services/owncloud/ocdav/ocdav_blackbox_test.go
butonic: added another mockPathStat for the other cases

labkode:
@butonic could you add to master as well?

aduffeck (approved):

butonic:
@labkode we started refactoring the error handling, the added tests make use of the error returned by the handlers ... we also moved to ginkgo/v2 ... urgh ... so I cannot even just add the test because they make use of DescribeTable which is only in ginkgo v2 ...

most of these cases are covered by the acceptance tests, we just want to develop unit tests to improve the coverage and then refactor the handlers further.

PR #3467, opened 2022-11-16
Enable gocritic linter in golangci-lint and solve issues (STACKED ON #3465)

vascoguita:


vascoguita:
The current Github Runner doesn't have enough RAM to run this linter.
Should work after I deploy a new runner for production.

PR #3468, opened 2022-11-16
Decomposedfs readdir including stat

butonic:
We now fetch the stat info of all children using `Readdir` instead of `Readdirnames` and subsequent `Stat` calls. While it does not remove the tree size aggregation penalty during uploads it does help in large directories.

see https://github.com/owncloud/ocis/issues/5061


vascoguita:
Hi, for some reason the CI is failing in Drone for everyone. :/
I'm also trying to understand what's going on.

aduffeck (approved):

butonic:
Readdir only returns the stat info of the symlink, not the file or folder

PR #3469, opened 2022-11-16
Forbidden overwrite in public links with drop zone

gmgigi96:
Closes https://github.com/cs3org/reva/issues/2480


labkode (dismissed):

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -214,6 +217,10 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	if userInCtxHasUploaderRole(ctx) {
+		ref.Path = randomizePath(ref.Path)
+	}
```
phil-davis, line 222: @gmgigi96 @labkode this looks like every file uploaded to an upload-only public link will get the `randomizePath` `_uuid` added to the name.

Even for the first person who uploads "file.txt", the actual file will be saved as "file_`uuid`.txt".

Am I reading the code correctly?

Is that the intended implementation?

cc: @micbar

micbar: @labkode after clarification with @tbsbdr we want to support the normal behavior like the users already know from their OS. That means the first file should just be `file.txt` and the second file `file(2).txt` and so on.

labkode: @micbar be aware that behaviour is difficult (if not impossible) to be enforced atomically (one has to block the whole folder to avoid race conditions). Is that a risk you're willing to take?

glpatcern: Let's not forget that in terms of UX it is meaningful NOT to give an "ordering". The typical use case we have (for years) is administrative personnel handling personal/sensitive documentation, and by no means the filename should suggest some sort of ordering or (worse) versioning, as a numbered suffix might do.

micbar: @labkode @glpatcern  I spoke with @tbsbdr and @dragotin again. We need to achieve a compromise.
A uuidv4 is just too long and looks "nerdy". That will never be accepted by non-physicists. For normal users, it just looks like a bug. They see a random number and are scared.

We could come together with a short 5 character string. Would that be possible?

phil-davis: See discussion in https://github.com/owncloud/ocis/issues/5074
IMO it will be easier to keep the discussion in that issue.

labkode: @micbar the users that use the DropZone are not physicists but administrative people to: obtain passports, tendering contracts, medical records, etc  ... we speak from the experience of our user-base. A short 5 character string is a good compromise.

gmgigi96:
@phil-davis can you take a look at the failing tests?

PR #3482, opened 2022-11-22
Decomposedfs propagate diff

butonic:
We now propagate the size diff instead of calculating the treesize. This fixes the slower upload speeds in large folders.

fixes https://github.com/owncloud/ocis/issues/5061

requires https://github.com/cs3org/reva/pull/3473

### DIFF
--- a/pkg/storage/utils/decomposedfs/revisions.go
+++ b/pkg/storage/utils/decomposedfs/revisions.go
@@ -126,7 +126,7 @@ func (fs *Decomposedfs) DownloadRevision(ctx context.Context, ref *provider.Refe
 	switch {
 	case err != nil:
 		return nil, errtypes.InternalError(err.Error())
-	case !rp.ListFileVersions || !rp.RestoreFileVersion || !rp.InitiateFileDownload: // TODO add explicit permission in the CS3 api?
+	case !rp.ListFileVersions || !rp.InitiateFileDownload: // TODO add explicit permission in the CS3 api?
 		f, _ := storagespace.FormatReference(ref)
 		if rp.Stat {
 			return nil, errtypes.PermissionDenied(f)
--- a/pkg/storage/utils/decomposedfs/revisions.go
+++ b/pkg/storage/utils/decomposedfs/revisions.go
@@ -21,7 +21,6 @@ package decomposedfs
 import (
 	"context"
 	"io"
-	iofs "io/fs"
 	"os"
 	"path/filepath"
 	"strings"
@@ -99,6 +98,7 @@ func (fs *Decomposedfs) ListRevisions(ctx context.Context, ref *provider.Referen
 }
 
 // DownloadRevision returns a reader for the specified revision
+// FIXME the CS3 api should explicitly allow initiating revision and trash download, a related issue is https://github.com/cs3org/reva/issues/1813
 func (fs *Decomposedfs) DownloadRevision(ctx context.Context, ref *provider.Reference, revisionKey string) (io.ReadCloser, error) {
 	log := appctx.GetLogger(ctx)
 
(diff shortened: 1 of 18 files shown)


pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -757,50 +770,62 @@ func (t *Tree) Propagate(ctx context.Context, n *node.Node) (err error) {
...
-					Uint64("treesize", treeSize).
-					Uint64("calculatedTreeSize", calculatedTreeSize).
-					Msg("parent size matches calculated size, not updating")
-			}
-
-			if updateTreeSize {
-				// update the tree time of the parent node
-				if err = n.SetTreeSize(calculatedTreeSize); err != nil {
-					sublog.Error().Err(err).Uint64("calculatedTreeSize", calculatedTreeSize).Msg("could not update treesize of parent node")
+				if sizeDiff > 0 {
```
aduffeck, line 808: Alternative approach to get rid of the `if` (assuming that the `int64` cast is safe):

```
newSize = uint64(int64(newSize) + sizeDiff)
```

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -684,83 +701,187 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) (err error) {
...
+		file, err := os.Create(versionsPath)
 		if err != nil {
-			sublog.Info().Err(err).Msg("Decomposedfs: failed to copy xattrs")
+			discardBlob()
+			sublog.Err(err).Str("version", versionsPath).Msg("could not create version node")
+			return errtypes.InternalError("could not create version node")
+		}
+		fi, err := file.Stat()
+		if err != nil {
+			file.Close()
```
aduffeck, line 798: We could `defer file.Close()` instead of closing manually here and in line 804.

pkg/storage/utils/decomposedfs/tree/tree.go
butonic: but this allows handling bigger treesizes

aduffeck (approved):
lgtm

pkg/storage/utils/decomposedfs/revisions.go
```diff
@@ -191,23 +207,65 @@ func (fs *Decomposedfs) RestoreRevision(ctx context.Context, ref *provider.Refer
...
-		versionsPath := fs.lu.InternalPath(spaceID, kp[0]+node.RevisionIDDelimiter+fi.ModTime().UTC().Format(time.RFC3339Nano))
+		// revisions are stored alongside the actual file, so a rename can be efficient and does not cross storage / partition boundaries
+		newRevisionPath := fs.lu.InternalPath(spaceID, kp[0]+node.RevisionIDDelimiter+fi.ModTime().UTC().Format(time.RFC3339Nano))
+
+		// touch new revision
+		if file, err := os.Create(newRevisionPath); err != nil {
+			return err
+		} else if err := file.Close(); err != nil {
+			return err
+		}
```
dragotin, line 218: At any of the error returns following here, a new revision file in inconsistent stage is left behind. Shouldn't we have a `defer` that removes the file in case of error and closes the descriptor?

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -684,83 +701,189 @@ func (upload *fileUpload) FinishUpload(ctx context.Context) (err error) {
...
-	// who will become the owner?  the owner of the parent actually ... not the currently logged in user
-	err = n.WriteAllNodeMetadata()
+		// basic node metadata
+		attrs[xattrs.ParentidAttr] = n.ParentID
+		attrs[xattrs.NameAttr] = n.Name
+		oldSize = 0
+	}
+
+	// update node metadata with new blobid etc
+	err = n.SetXattrsWithLock(attrs, lock)
```
dragotin, line 849: this `n` was defined fife or six screens up (for me, not on 4k) - I have an idea that it might be a node. Maybe the node of the new file? Maybe the node of the parent dir? (why not?) - or maybe not a Node, but a file counter called n? 
I remember times where C programmers thought that it is efficient to use one char vars which turned out to be a stupid idea that was expensive and annoying... I am really puzzled that we repeat that today. :sheep:

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -207,6 +225,21 @@ func Set(filePath string, key string, val string) (err error) {
 	return xattr.Set(filePath, key, []byte(val))
 }
 
+// SetWithLock an extended attribute key to the given value with an existing lock
+func SetWithLock(filePath string, key string, val string, fileLock *flock.Flock) (err error) {
+	// check the file is write locked
+	switch {
+	case fileLock == nil:
+		return errors.New("no lock provided")
+	case fileLock.Path() != filePath+".flock":
```
dragotin, line 234: There is a func to get the lock file name for a file - should be used here. This cancatenation is not nice.

pkg/storage/utils/decomposedfs/xattrs/xattrs.go
```diff
@@ -246,6 +279,20 @@ func SetMultiple(filePath string, attribs map[string]string) (err error) {
...
 
+	return SetMultipleWithLock(filePath, attribs, fileLock)
+}
+
+// SetMultipleWithLock allows setting multiple key value pairs at once with an existing lock
+func SetMultipleWithLock(filePath string, attribs map[string]string, fileLock *flock.Flock) (err error) {
+	switch {
+	case fileLock == nil:
+		return errors.New("no lock provided")
+	case fileLock.Path() != filePath+".flock":
```
dragotin, line 290: repeated here.

pkg/storage/utils/decomposedfs/revisions.go
butonic: agreed

PR #3483, opened 2022-11-22
Reduce number of parallel runs in Drone CI

vascoguita:


micbar:
@vascoguita why did you do this? This will increase the CI time dramatically.

vascoguita:
It was a mistake, I already removed the merge.

PR #3029, opened 2022-07-05
[tests-only] Experimental

kobergj:
To keep track of changes between edge and experimental branch. Not to run into the same issues as with master and edge

DO NOT MERGE FOR NOW - NEEDS GO FROM MANAGEMENT

PR #3531, opened 2022-12-08
Async Postprocessing

kobergj:
brings postprocessing to edge

Mainly backports from experimental branch https://github.com/cs3org/reva/pull/3029

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,4 @@
 issues:
-  exclude:
-    - "Error return value of .* is not checked"
   exclude-rules:
     - path: internal/http/interceptors/log/log.go
       text: "SA1019:"
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,4 +1,6 @@
 issues:
+  exclude:
+    - "Error return value of .* is not checked"
   exclude-rules:
     - path: internal/http/interceptors/log/log.go
       text: "SA1019:"
--- a/internal/http/services/owncloud/ocdav/propfind/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind/propfind.go
@@ -1476,6 +1476,14 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 		}
 	}
 
+	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
+		response.Propstat = append(response.Propstat, PropstatXML{
+			Status: "HTTP/1.1 425 TOO EARLY",
+			Prop:   propstatOK.Prop,
+		})
+		return &response, nil
+	}
+
 	if len(propstatOK.Prop) > 0 {
 		response.Propstat = append(response.Propstat, propstatOK)
 	}
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -68,6 +68,9 @@ const (
 
 	// RootID defines the root node's ID
 	RootID = "root"
+
+	// ProcessingStatus is the name of the status when processing a file
+	ProcessingStatus = "processing"
 )
 
 // Node represents a node in the tree and provides methods to get a Parent or Child instance
@@ -90,7 +93,7 @@ type Node struct {
 type PathLookup interface {
 	InternalRoot() string
 	InternalPath(spaceID, nodeID string) string
-	Path(ctx context.Context, n *Node) (path string, err error)
+	Path(ctx context.Context, n *Node, hasPermission func(*Node) bool) (path string, err error)
 }
 
 // New returns a new instance of Node
@@ -123,6 +126,33 @@ func (n *Node) ChangeOwner(new *userpb.UserId) (err error) {
 	}
 
 	return
+}
+
+// SetMetadata populates a given key with its value.
+// Note that consumers should be aware of the metadata options on xattrs.go.
+func (n *Node) SetMetadata(key string, val string) (err error) {
+	nodePath := n.InternalPath()
+	if err := xattrs.Set(nodePath, key, val); err != nil {
+		return errors.Wrap(err, "Decomposedfs: could not set extended attribute")
+	}
+	return nil
+}
+
+// RemoveMetadata removes a given key
+func (n *Node) RemoveMetadata(key string) (err error) {
+	if err = xattrs.Remove(n.InternalPath(), key); err == nil || xattrs.IsAttrUnset(err) {
+		return nil
+	}
+	return err
+}
+
+// GetMetadata reads the metadata for the given key
+func (n *Node) GetMetadata(key string) (val string, err error) {
+	nodePath := n.InternalPath()
+	if val, err = xattrs.Get(nodePath, key); err != nil {
+		return "", errors.Wrap(err, "Decomposedfs: could not get extended attribute")
+	}
+	return val, nil
 }
 
 // WriteAllNodeMetadata writes the Node metadata to disk
@@ -632,7 +662,7 @@ func (n *Node) AsResourceInfo(ctx context.Context, rp *provider.ResourcePermissi
 	case returnBasename:
 		fn = n.Name
 	default:
-		fn, err = n.lu.Path(ctx, n)
+		fn, err = n.lu.Path(ctx, n, func(*Node) bool { return true })
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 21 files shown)


internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -1476,6 +1476,14 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 		}
 	}
 
+	if status := utils.ReadPlainFromOpaque(md.Opaque, "status"); status == "processing" {
```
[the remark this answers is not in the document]
kobergj: Good idea :+1: https://github.com/cs3org/cs3apis/pull/190

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -478,535 +262,50 @@ func (fs *Decomposedfs) PurgeExpiredUploads(purgedChan chan<- tusd.FileInfo) err
...
-	// attributes that will change
-	attrs := map[string]string{
-		xattrs.BlobIDAttr:   newNode.BlobID,
-		xattrs.BlobsizeAttr: strconv.FormatInt(newNode.Blobsize, 10),
-
-		// update checksums
-		xattrs.ChecksumPrefix + "sha1":    string(sha1h.Sum(nil)),
-		xattrs.ChecksumPrefix + "md5":     string(md5h.Sum(nil)),
-		xattrs.ChecksumPrefix + "adler32": string(adler32h.Sum(nil)),
-	}
```
butonic, line 782: here we create a map of attributes that we later set using a the write lock we aquired above.

we later add the checksums to this map, as well as parentid and name.

All this got lost when moving the code to processing.go CreateForUpload().

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -478,535 +262,50 @@ func (fs *Decomposedfs) PurgeExpiredUploads(purgedChan chan<- tusd.FileInfo) err
...
-			return errtypes.InternalError("could not stat version node")
-		}
-		newMtime = fi.ModTime()
-
-		// copy blob metadata to version node
-		err = xattrs.CopyMetadataWithSourceLock(targetPath, versionsPath, func(attributeName string) bool {
-			return strings.HasPrefix(attributeName, xattrs.ChecksumPrefix) ||
-				attributeName == xattrs.BlobIDAttr ||
-				attributeName == xattrs.BlobsizeAttr
-		}, lock)
```
butonic, line 811: we copy only the necessary blob metadata to the version. shares always stay in the ectual node. this got lost as well

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -478,535 +262,50 @@ func (fs *Decomposedfs) PurgeExpiredUploads(purgedChan chan<- tusd.FileInfo) err
...
-		xattrs.ChecksumPrefix + "adler32": string(adler32h.Sum(nil)),
-	}
-
-	// create version node with current metadata
-
-	var newMtime time.Time
-	// if file already exists
-	if versionsPath != "" {
-		// touch version node
-		file, err := os.Create(versionsPath)
```
butonic, line 790: Instead of moving the existing node we create a new node. Thit fixes some nasty bugs, search for `// This move drops all metadata!!! We copy it below with CopyMetadata` in the old code which has moved to `updateExistingNode` in this PR.

You should use a `initNewRevisionNode`

butonic (changes requested):
the rewrite of Upload in https://github.com/cs3org/reva/pull/3482/commits/fced7bdf510e96459f6b20523681ab2a566ce2b9 got lost.

pkg/storage/utils/decomposedfs/upload.go
kobergj: Mmmh. Maybe I misunderstand you.
- patentid and name is written in `WriteAllNodeMetadata` within the `CreateNodeForUpload` func https://github.com/cs3org/reva/blob/a4825d8b9ecc89a1635a9c536b7a578195cb3739/pkg/storage/utils/decomposedfs/upload/processing.go#L276-L279
- checksums are written when the upload is finished. Before the bytes are not yet there https://github.com/cs3org/reva/blob/a4825d8b9ecc89a1635a9c536b7a578195cb3739/pkg/storage/utils/decomposedfs/upload/upload.go#L238-L241

Not sure what is missing?

pkg/storage/utils/decomposedfs/upload.go
kobergj: Ah. Thats new. I still use the old code https://github.com/cs3org/reva/blob/a4825d8b9ecc89a1635a9c536b7a578195cb3739/pkg/storage/utils/decomposedfs/upload/processing.go#L374-L386

So only Name, BlobID and BlobSize need to be copied?

.golangci.yaml
```diff
@@ -1,4 +1,6 @@
 issues:
+  exclude:
+    - "Error return value of .* is not checked"
```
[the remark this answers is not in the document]
kobergj: Yes. This linting message is nothing but annoying. If a developer decides that an  error doesn't need to be checked, she should be able to do so without annoying linting messages

pkg/storage/utils/decomposedfs/upload.go
kobergj: I know pass a `map[string]string` to the `CreateNodeForUpload` function so it only needs to write once

pkg/storage/utils/decomposedfs/upload.go
kobergj: I'm creating only one lock now and pass it around. This will be much better performance wise

pkg/storage/utils/decomposedfs/upload.go
kobergj: Got it. I'm only copying BlobID, BlobSize and Mtime from old node

pkg/storage/utils/decomposedfs/upload.go
```diff
@@ -478,535 +262,50 @@ func (fs *Decomposedfs) PurgeExpiredUploads(purgedChan chan<- tusd.FileInfo) err
...
-		return err
+		return nil, err
 	}
-	defer file.Close()
-
-	for _, partialUpload := range uploads {
-		fileUpload := partialUpload.(*fileUpload)
 
-		src, err := os.Open(fileUpload.binPath)
+	idRegexp := regexp.MustCompile(".*/([^/]+).info")
```
C0rby, line 293: This can/should be moved to a global variable so that we don't recompile the same pattern over and over again.

.golangci.yaml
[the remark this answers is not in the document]
kobergj: In my opinion there is no need for linting this at all. Any error can be ignored if the developer sees fit. There is no additional explanation needed to do so. Furthermore I do not think that you can "accidentally" miss an error in go. The common pattern of `if err != nil` is everywhere, so you need to actively decide to ignore an error.

Additionally this error can be removed by actively ignoring the error like `_ := MyFunc()`. This however just makes the code ugly and is explicitly annoying in `defer` statements.

So I would argue we can savely ignore this linting error message, but I will follow teams decision if I am overruled.

pkg/storage/utils/decomposedfs/upload.go
kobergj: Excellently spotted :+1: Moved to global var

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -125,6 +128,33 @@ func (n *Node) ChangeOwner(new *userpb.UserId) (err error) {
 	return
 }
 
+// SetMetadata populates a given key with its value.
+// Note that consumers should be aware of the metadata options on xattrs.go.
+func (n *Node) SetMetadata(key string, val string) (err error) {
```
aduffeck, line 133: Is there any reason to bring those `[Set|Remove|Get]Metadata` methods back? They are bypassing the node's xattrs cache, potentially leading to inconsistent data.
I think we should use the `node.*Xattr` methods instead.

aduffeck (changes requested):

pkg/storage/utils/decomposedfs/node/node.go
kobergj: Leftover from merge. Removed the `*Metadata` funcs and now use `*Xattr` instead

.golangci.yaml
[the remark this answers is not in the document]
C0rby: Yeah, I don't need a comment on obvious lines but sometimes it isn't obvious.

> The common pattern of if err != nil is everywhere, so you need to actively decide to ignore an error.

We also have lot's of places with `w.Write` where write can fail and returns an error but we never check those.
I'd bet these were not intentional. It's just very rare that these calls fail.

I'm also in favor of `_ := DoSomething()` because if there is just a line `DoSomething()` where the function returns an error I don't see if the error was ignored on purpose.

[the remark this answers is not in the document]
kobergj: Okok. You overruled me. Removed it again.

aduffeck (approved):
lgtm

PR #3534, opened 2022-12-08
Fix nil node on err

butonic:
When returning `nil, err`, `n` is set to nil causing a panic in `n.ID += revisionSuffix`.


kobergj:
Is this urgent? It will make me need to rebase https://github.com/cs3org/reva/pull/3531 again which is quite a pain. It should be ready for review shortly.

butonic:
I don't think so. can wait for https://github.com/cs3org/reva/pull/3531

kobergj:
Added the fix to my async upload branch: https://github.com/cs3org/reva/pull/3556

PR #3536, opened 2022-12-09
[tests-only] Bump core commit id for tests (master)

saw-jan:
Bump core commit id for tests

Part of
https://github.com/owncloud/QA/issues/785


kulmann:
Wondering if this recurring task still makes sense. I was under the impression that the reva `master` doesn't run `drone` at all anymore. Who can confirm? @micbar ?

saw-jan:
> Wondering if this recurring task still makes sense. I was under the impression that the reva `master` doesn't run `drone` at all anymore. Who can confirm? @micbar ?

Yeah, heard the same. Looks like the drone is still running for PRs to master. 
Should we continue bumping commit-id till the drone is dropped or cease to do so?

CC @phil-davis

micbar:
please rebase. the CI has changed on master.

phil-davis:
> please rebase. the CI has changed on master.

rebased - I will look at what needs to be deleted from expected-failures files etc.

phil-davis (approved):

PR #3526, opened 2022-12-07
Efss backend fixes

michielbdejong:
Continuation of https://github.com/cs3org/reva/pull/3524

PR #3551, opened 2022-12-12
Fix #3550: OCM share fails for json driver

### LINKED ISSUE #3550: OCM sharing fails when reva is configured to use users json file
https://github.com/cs3org/reva/pull/3526 introduces a new `owner` parameter format (`opaqueId@webDAVHost`) in the share request for the ocm share nextcloud driver. The ocmd shares service at the receiving end will try to parse this owner value into the user opaqueID and the webDAV host again. 
But this is not implemented for the ocm share json driver. So when reva is configured to use the users json file sharing fails 'in panic' due to an `index out of range` error.

redblom:
Fixes implementation omission in ocm share json driver in https://github.com/cs3org/reva/pull/3526

### DIFF
--- a/changelog/unreleased/fix-3550.md
+++ b/changelog/unreleased/fix-3550.md
@@ -1,4 +1,4 @@
-Fix: Fixes implementation omission of #3526
+Bugfix: Fixes implementation omission of #3526
 
 In #3526 a new value format of the owner parameter of the ocm share request was introduced.
 This change was not implemented in the json driver. This change fixes that.
--- a/changelog/unreleased/fix-3550.md
+++ b/changelog/unreleased/fix-3550.md
@@ -1,0 +1,5 @@
+Fix: Fixes implementation omission of #3526
+
+In #3526 a new value format of the owner parameter of the ocm share request was introduced.
+This change was not implemented in the json driver. This change fixes that.
+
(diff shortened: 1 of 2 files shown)


redblom:
Ps. Tested locally. Both sharing and transfers successful.

michielbdejong (approved):
Thanks! Sorry if this broke for you! We were in a bit of a hurry to merge last week. :)

One remark: I notice you used `fmt.Sprintf("%s@%s", userID.OpaqueId, userID.Idp)` instead of `fmt.Sprintf("%s@%s", userID.OpaqueId, sm.webDAVHost)` which may be fine depending on which service is used as the WebDAV server, right?

I guess if you run Reva's `ocdav` module then `userID.Idp` works? 
Does this work with the ocdav module?

redblom:
> One remark: I notice you used `fmt.Sprintf("%s@%s", userID.OpaqueId, userID.Idp)` instead of `fmt.Sprintf("%s@%s", userID.OpaqueId, sm.webDAVHost)` which may be fine depending on which service is used as the WebDAV server, right?
> 
> I guess if you run Reva's `ocdav` module then `userID.Idp` works? Does this work with the ocdav module?

Well, originally this was the 'meshProvider' (user.idp), this fix still works for this driver with ocdav. The webDAVHost is not available in the json driver. 
Is the ratio behind (sending) this (new) webDAVHost config by the nextcloud driver discussed somewhere? Can't seem to find that.

glpatcern (approved):

PR #3552, opened 2022-12-12
Suppress tusd logs

kobergj:
fixes https://github.com/owncloud/ocis/issues/4803 by passing a Logger to the tusd package

### DIFF
--- a/changelog/unreleased/suppress-tusd-logs.md
+++ b/changelog/unreleased/suppress-tusd-logs.md
@@ -1,0 +1,5 @@
+Enhancement: Suppress tusd logs
+
+`tusd` package would log specific messages regardless of loglevel. This is now changed
+
+https://github.com/cs3org/reva/pull/3552


micbar (approved):
Good catch.

butonic:
well, yeah, this disables the log ... but there is no way to reenable it. I guess we need to invest in upstream patches that fetch the logger from the context. or maybe fetching the logger from the context is not a good pattern? Then again, I don't see a good way of changing the log level for different users ... maybe using logging for user specific logs is the wrong approach. Maybe that is better done with tracing and span attributes.

That should in theory allow us to pass a logger to the handlers, like the tus handler, wrap it with a trace middleware and have the best of both worlds?

Not part of this PR, though ...

kobergj:
Why is there no way to reenable it? We pass a custom logger. I expect it to log through the logger we provided. Otherwise this is a upstream bug (either in basic log or zerolog package.

pkg/rhttp/datatx/manager/tus/tus.go
```diff
@@ -99,6 +100,7 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 	config := tusd.Config{
 		StoreComposer:         composer,
 		NotifyCompleteUploads: true,
+		Logger:                log.New(appctx.GetLogger(context.Background()), "", 0),
```
micbar, line 103: hmm, now that the discussion is on, let me throw in my 2 cents. Can we pass in the active context instead of creating it on the fly because the active context holds the requestID.

kobergj: There is no active context. This happens on service init

PR #3553, opened 2022-12-12
Implement CS3APIs datatx name change PullTransfer -> CreateTransfer

redblom:
This breaking change implements the CS3APIs name change PullTransfer -> CreateTransfer (requires: https://github.com/cs3org/cs3apis/pull/195).

### DIFF
--- a/internal/grpc/services/gateway/ocmshareprovider.go
+++ b/internal/grpc/services/gateway/ocmshareprovider.go
@@ -30,7 +30,6 @@ import (
 	ocm "github.com/cs3org/go-cs3apis/cs3/sharing/ocm/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	datatx "github.com/cs3org/go-cs3apis/cs3/tx/v1beta1"
-	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/pkg/ctx"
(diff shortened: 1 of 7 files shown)


internal/grpc/services/gateway/ocmshareprovider.go
```diff
@@ -388,22 +388,22 @@ func (s *svc) UpdateReceivedOCMShare(ctx context.Context, req *ocm.UpdateReceive
 							},
 						},
 					}
-					req := &datatx.PullTransferRequest{
+					req := &datatx.CreateTransferRequest{
 						SrcTargetUri:  srcTargetURI,
 						DestTargetUri: destTargetURI,
 						Opaque:        opaqueObj,
```
glpatcern, line 394: As we're renaming those methods, what about cleaning the API as well? that is promoting this opaque info into a proper property of the `CreateTransferRequest` object?

glpatcern (changes requested):
Hi there, I see you sorted out Reva with the rebase, at this point let's make things clearer - we're also cleaning up the other OCM properties including having `datatx` as a protocol type, see https://github.com/cs3org/OCM-API/pull/57

Therefore I propose you to also add the opaque parameters as real ones.

internal/grpc/services/gateway/ocmshareprovider.go
redblom: Share id is promoted to real property.

glpatcern (approved):

glpatcern (changes requested):

glpatcern (approved):

PR #3564, opened 2022-12-19
Fix Tags pkg

kobergj:
adds an option to create the tags struct from a `[]string`


fschade:
just a idea:

```go
type Tags struct {
	t       []string
}

func NewTags(in ...string) *Tags {
	t := &Tags{}
	
	for _, tgs := range in {
		t.t = append(t.t, strings.Split(tgs, ",")...)
	}
	
	return t
}
```

then we do not have to differentiate between `string` & `[]string`. Same for `AddList `.
What do you think

kobergj:
Excellent idea :+1: Just named it `New` as it will anyway be called as `tags.New`

fschade (approved):
LGTM 🥳

PR #3566, opened 2022-12-20
Migrate the litmusOcisSpacesDav test from Drone to GitHub Actions

vascoguita:


Makefile
```diff
@@ -123,11 +119,11 @@ build-reva-docker: off
 
 .PHONY: test
 test: off
-	go test -coverprofile coverage.out -race $$(go list ./... | grep -v /tests/integration)
```
micbar, line 126: No coverage anymore?

vascoguita: Hugo asked me to remove it since Codacy no longer uses that file.
Today I noticed that Codacy doesn't have coverage reports since I migrated it to the Codacy/GitHub Integration.
I'll have to fix that.

vascoguita: Aparently, the correct way to do it is to use the Codacy GitHub Action for reporting coverage:
https://github.com/marketplace/actions/codacy-coverage-reporter
It's not very elegant since we are forced to simultaneously use the GitHub integration and the GitHub Action instead of having everything done in one place.

gmgigi96 (approved):

Makefile
vascoguita: I've just setup the test coverage reports with Codacy: https://github.com/cs3org/reva/pull/3608
The GitHub Actions generate a new report on each push and send it to Codacy.
The coverage of the master branch can be observed here: https://app.codacy.com/gh/cs3org/reva/dashboard

PR #3569, opened 2022-12-22
list shares jail even when no shares received

C0rby:
Changes spaces listing to always include the shares jail, even when no shares where received.
If there are no received shares the shares jail will have the etag value `DECAFC00FEE`.


butonic (approved):

rhafer:
@C0rby  this seems to break the litmus tests in ocis. See: https://drone.owncloud.com/owncloud/ocis/18127/38/6

PR #3571, opened 2022-12-28
More Async Upload Fixes

kobergj:
Fixes https://github.com/owncloud/ocis/issues/5278 and https://github.com/owncloud/ocis/issues/5279

### DIFF
--- a/pkg/storage/utils/decomposedfs/upload/upload.go
+++ b/pkg/storage/utils/decomposedfs/upload/upload.go
@@ -263,15 +263,13 @@ func (upload *Upload) FinishUpload(_ context.Context) error {
 		}
 	}
 
-	if upload.async {
-		// handle postprocessing asynchronously but inform there is something in progress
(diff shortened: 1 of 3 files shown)


pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -263,15 +263,13 @@ func (upload *Upload) FinishUpload(_ context.Context) error {
 		}
 	}
 
-	if upload.async {
-		// handle postprocessing asynchronously but inform there is something in progress
-		return upload.tp.Propagate(upload.Ctx, n, upload.sizeDiff)
```
micbar, line 268: not needed anymore? where do we propagate now?

kobergj: We still propagate. I just switched the code part so we need havae the call to `upload.tp.Propagate` only once

micbar (approved):
CI in ocis confirms the fixes.

PR #3580, opened 2023-01-04
Share Expired Event

kobergj:
Adds an event when shares expire. Also adds a helper function to impersonate an user.


pkg/utils/grpc.go
```diff
@@ -0,0 +1,41 @@
...
+
+	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
+	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	revactx "github.com/cs3org/reva/v2/pkg/ctx"
+	"google.golang.org/grpc/metadata"
+)
+
+// Impersonate returns an authenticated reva context and the user it represents
+func Impersonate(userID *user.UserId, gwc gateway.GatewayAPIClient, machineAuthAPIKey string) (context.Context, *user.User, error) {
```
micbar, line 15: Nice util. Is this func currently used somewhere?
And because we have a sunny day, i would love to see a unit test for this func 😄

kobergj: I'm using it in the ocis pendant of this PR.

Not sure what I am supposed to unit test here. It does requests to `GetUser` and `Authenticate`. If I mock them away there is nothing left to be tested...

micbar: hmm, i see.  lets merge it.

micbar (approved):

PR #3609, opened 2023-01-18
Space Management Permissions

kobergj:
implements "manage properties" and "disable space" permissions (naming wip)


pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// AssemblePermissions is used to assemble file permissions
+func (p Permissions) AssemblePermissions(ctx context.Context, n *node.Node) (provider.ResourcePermissions, error) {
+	return p.item.AssemblePermissions(ctx, n)
+}
+
+// CreateSpace returns true when the user is allowed to create the space
+func (p Permissions) CreateSpace(ctx context.Context, spaceid string) bool {
+	return p.checkPermission(ctx, "create-space", spaceRef(spaceid))
```
butonic, line 42: I would prefer to use a naming scheme for permissions that is similar to [ms graph permissions](https://learn.microsoft.com/en-us/graph/permissions-overview?tabs=http#permissions-naming-pattern): `{resource}.{operation}.{constraint}`. Here we want to check if the user has the permission to create project spaces
```suggestion
	return p.checkPermission(ctx, "drive.create.all", spaceRef(spaceid))
```

When using spicedb, this code would change to sth like 
```go
resp, err := client.CheckPermission(ctx, &pb.CheckPermissionRequest{
		Resource: organization,
		Permission: "drive.create.all",
		Subject: user,
	})
```
In SpiceDB we would model our permissions in relation to an organization: see the `can_admin` example at https://authzed.com/blog/check-it-out/

Anyway, naming is hard ... and ms may not have gotten it right as well. While they have eg. `Files.ReadWrite.All` the resources this affects are actually `drive` and `driveItem`, so the actual permissions should be `Drive.ReadWrite.All` and `DriveItem.ReadWrite.All`. But they also don't have explicit permissions to manage drives.

@C0rby just educated me on scopes vs permissions:

The permission check here is one of two checks that should ultimately be done:
1. a scope check to make sure the user gave his consent, if not redirect him to the idp and ask for the scope. way out of scope of this PR though ;-) the scope check can be done as early as in the ocis graph service.
2. an actual permissions check if the user is allowed to access a resource or has a certain role. This is done when the concrete subject, permission and resource ids have been determined.

These are two different things which we need to clarify in the docs.

Trying to wrap my head around how this relates to customizing roles and permissions relationships with spacedb: 
- we would define permissions in a spacedb schema as a spicedb permission in an organization/tenant/instance? class
- the composition of permissions that belong to a role, aka user defined roles, see https://authzed.com/blog/user-defined-roles/ for how that is modeled with spicedb

It seems to me that we can reuse the permissions naming scheme from the graph api for organizational permissions.

pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// SetSpaceQuota returns true when the user is allowed to change the spaces quota
+func (p Permissions) SetSpaceQuota(ctx context.Context, spaceid string) bool {
+	return p.checkPermission(ctx, "set-space-quota", spaceRef(spaceid))
```
butonic, line 47: why do we need to differentiate between set quota and manage space properties?

pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// ManageSpaceProperties returns true when the user is allowed to change space properties (name/subtitle)
+func (p Permissions) ManageSpaceProperties(ctx context.Context, spaceid string) bool {
+	return p.checkPermission(ctx, "manage-space-properties", spaceRef(spaceid))
```
butonic, line 52:
```suggestion
	return p.checkPermission(ctx, "drive.manage", spaceRef(spaceid))
```

pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// SpaceAbility returns true when the user is allowed to enable/disable the space
+func (p Permissions) SpaceAbility(ctx context.Context, spaceid string) bool {
+	return p.checkPermission(ctx, "space-ability", spaceRef(spaceid))
```
butonic, line 57:
```suggestion
	return p.checkPermission(ctx, "space.toggle", spaceRef(spaceid))
```
or `toggleEnabled`

pkg/storage/utils/decomposedfs/spacepermissions.go
butonic, line 47: why do we need to differentiate between set quota and manage space properties?
```suggestion
	return p.checkPermission(ctx, "space.writequota", spaceRef(spaceid))
```

pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// ListAllSpaces returns true when the user is allowed to list all spaces
+func (p Permissions) ListAllSpaces(ctx context.Context) bool {
+	return p.checkPermission(ctx, "list-all-spaces", nil)
```
butonic, line 62:
```suggestion
	return p.checkPermission(ctx, "drive.read.all", nil)
```

pkg/storage/utils/decomposedfs/spacepermissions.go
```diff
@@ -0,0 +1,129 @@
...
+	case ctxpkg.ContextMustGetUser(ctx).GetId().GetOpaqueId():
+		return true
+	default:
+		return p.ListAllSpaces(ctx)
+	}
+}
+
+// DeleteAllSpaces returns true when the user is allowed to delete all spaces
+func (p Permissions) DeleteAllSpaces(ctx context.Context) bool {
+	return p.checkPermission(ctx, "delete-all-spaces", nil)
```
butonic, line 80:
```suggestion
	return p.checkPermission(ctx, "drive.delete.all", nil)
```

butonic (approved):
We should merge this without renaming existing permissions. Maybe you can prefix the new permissions with `Drive.` and then use `(Read)(Write){Property}`, eg. `Drive.ReadWriteEnabled` and  `Drive.ReadWriteQuota` for permissions on specific properties.? Renaming existing permissions should be deferred until moving to spicedb.

pkg/storage/utils/decomposedfs/spacepermissions.go
kobergj: As discussed following new naming conventions for new permissions while old ones stay the same

pkg/storage/utils/decomposedfs/spacepermissions.go
kobergj: That is a question I cannot answer. It has been a requirement for the ticket.

PR #3611, opened 2023-01-19
Revamp OCM invitation workflow

gmgigi96:


glpatcern (dismissed):

glpatcern (approved):

glpatcern:
@michielbdejong from this point on we're on the new OCM endpoints, therefore you can start adapting the ownCloud/Nextcloud sciencemesh apps code according to https://github.com/cs3org/OCM-API/pull/54 and our discussion. I think we could tag a Reva version after you give a green light that the apps are OK.

PR #3612, opened 2023-01-20
Add event code for group changes.

ainmosni:
This MR adds the boilerplate to support emiting group change events.

### DIFF
--- a/pkg/events/groups.go
+++ b/pkg/events/groups.go
@@ -87,7 +87,7 @@ type GroupFeature struct {
 // GroupFeatureChanged is emitted when a group feature was changed
 type GroupFeatureChanged struct {
 	Executant *user.UserId
-	UserID    string
+	GroupID   string
 	Features  []GroupFeature
 }
 
--- a/pkg/events/groups.go
+++ b/pkg/events/groups.go
@@ -77,3 +77,23 @@ func (GroupMemberRemoved) Unmarshal(v []byte) (interface{}, error) {
 	err := json.Unmarshal(v, &e)
 	return e, err
 }
+
+// GroupFeature represents a group feature
+type GroupFeature struct {
+	Name  string
+	Value string
+}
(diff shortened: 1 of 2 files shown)


pkg/events/groups.go
```diff
@@ -77,3 +77,23 @@ func (GroupMemberRemoved) Unmarshal(v []byte) (interface{}, error) {
...
+type GroupFeature struct {
+	Name  string
+	Value string
+}
+
+// GroupFeatureChanged is emitted when a group feature was changed
+type GroupFeatureChanged struct {
+	Executant *user.UserId
+	UserID    string
+	Features  []GroupFeature
```
kobergj, line 91: We don't need a timestamp?

kobergj (approved):

pkg/events/groups.go
ainmosni: Good question, I used the `UserFeatureChanged` as a template, and it also does not have a timestamp field.

pkg/events/groups.go
```diff
@@ -77,3 +77,23 @@ func (GroupMemberRemoved) Unmarshal(v []byte) (interface{}, error) {
...
+// GroupFeature represents a group feature
+type GroupFeature struct {
+	Name  string
+	Value string
+}
+
+// GroupFeatureChanged is emitted when a group feature was changed
+type GroupFeatureChanged struct {
+	Executant *user.UserId
+	UserID    string
```
kobergj, line 90: No `GroupID`?

ainmosni: Oops, guess I wasn't awake enough, fixed now.

pkg/events/groups.go
kobergj: Oki. Then it is consistent at least :smile:

kobergj (approved):

PR #3623, opened 2023-01-25
Copy p.Services to pi.Services in mentix

michielbdejong:
When testing Reva with Mentix in the ScienceMesh, I found that without this, the services struct is missing and an error is thrown.

### DIFF
--- a/pkg/ocm/provider/authorizer/mentix/mentix.go
+++ b/pkg/ocm/provider/authorizer/mentix/mentix.go
@@ -177,6 +177,7 @@ func (a *authorizer) IsProviderAllowed(ctx context.Context, pi *ocmprovider.Prov
 		for _, p := range providers {
 			if p.Domain == normalizedDomain {
 				providerAuthorized = true
+				pi.Services = p.Services
 				break
 			}
 		}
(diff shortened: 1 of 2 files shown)


gmgigi96:
The `IsProviderAllowed` is supposed to just check that a given provider is allowed, not have some side effects.
Where did you get the error? The problem might be somewhere else...

mirekys:
Here are all current usages of the `IsProviderAllowed` API:

1. From HTTP interceptor middleware:
https://github.com/cs3org/reva/blob/605cec43f46ebef9846f50041dcf82af5027b138/internal/http/interceptors/providerauthorizer/providerauthorizer.go#L86-L99

2. From HTTP OCM AcceptInvite handler
https://github.com/cs3org/reva/blob/14c9c5f8cfd4c78bb108d6f6cb5469c440f097ec/internal/http/services/ocmd/invites.go#L84-L101

3. From HTTP OCM CreateShare handler
https://github.com/cs3org/reva/blob/14c9c5f8cfd4c78bb108d6f6cb5469c440f097ec/internal/http/services/ocmd/shares.go#L111-L127

In case of 1., you can see no services are passed into the check as part of `ProviderInfo` struct - probably the better way would be to provide it in input where `IsProviderAllowed` is called?

michielbdejong:
> provide it in input where `IsProviderAllowed` is called

> [no] need to re-check `IsProviderAllowed` in 2.,3.

Good points, I think both make sense!

gmgigi96:
> > provide it in input where `IsProviderAllowed` is called
> 
> > [no] need to re-check `IsProviderAllowed` in 2.,3.
> 
> Good points, I think both make sense!

Since the last changes to the `ocmd` service, all the exposed endpoints are unprotected. This means that the `providerauthorizer` interceptor is always skipped, making it useless.
So the checks in 2 and 3 are still need.

michielbdejong:
Agreed on avoiding side effects. Also copying data from `p` to `pi` is dangerous because it might influence the check and give a false match. Let's see how we can do this better.

The problem is that it's failing to look up the ocm host of `p` which is from `a.providers`.
This is unrelated to the `pi` that is passed into the function.

I'll debug the `getOCMProviders` function, apparently it's not successfully reading in the json?

PR #3639, opened 2023-02-03
[tests-only][full-ci]Bump commit id for tests

SwikritiT:
Part of: https://github.com/owncloud/QA/issues/797


phil-davis (approved):
LGTM. Note: the test scenarios are being cleaned-up, irrelevant test scenarios removed from the oCIS test code. So that is why quite a lot of items are removed from expected-failures. They are test scenarios that are not relevant to ocis/reva anyway.

PR #3641, opened 2023-02-04
Dont share versions

micbar:
# Description

Hide file versions for share receivers

We needed to change the visibility of file versions and hide them to share receivers. Space Editors can still see and restore file versions.

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,4 +1,4 @@
 # The test runner source for API tests
-APITESTS_COMMITID=f4ae082265c54013f28e7facb892d6fdc943e827
+APITESTS_COMMITID=2cd625777122e292c92b30bc0ec0c6f4fc29a178
 APITESTS_BRANCH=dont-share-versions
 APITESTS_REPO_GIT_URL=https://github.com/owncloud/ocis.git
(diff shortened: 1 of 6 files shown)


changelog/unreleased/hide-file-versions-shares.md
```diff
@@ -0,0 +1,5 @@
+Change: Hide file versions for share receivers
+
+We needed to change the visibility of file versions and hide them to share receivers. Space Editors can still see and restore file versions.
```
butonic, line 3: Could you elaborate how this is intended to evolve in the future? The change is substaintial as it breaks the CI but solves a very specific problem. Maybe this should be configurable?

butonic (approved):

changelog/unreleased/hide-file-versions-shares.md
micbar: This is decided to be the new default product behavior.

micbar:
needs fixes to tests

.drone.env
```diff
@@ -1,4 +1,4 @@
 # The test runner source for API tests
-APITESTS_COMMITID=76cc388e4546d4e588515c6c7d624829a674163a
-APITESTS_BRANCH=master
+APITESTS_COMMITID=2cd625777122e292c92b30bc0ec0c6f4fc29a178
+APITESTS_BRANCH=dont-share-versions
```
kobergj, line 3: This is temporarily because of some hen/egg problem?

kobergj (approved):

PR #3646, opened 2023-02-08
Update OCM shares to last version of CS3APIs

gmgigi96:
This PR updates the OCM share implementation after the changes in https://github.com/cs3org/cs3apis/pull/199, https://github.com/cs3org/OCM-API/pull/54 and https://github.com/cs3org/OCM-API/pull/57.

In particular:
* Introduced the concept of allowing multiple protocols in the OCM shares
* Refactor the `ocmcore` and `ocmshareprovider`, simplifying the single drivers, now just repositories, and implemented common logic at provider level.
* Implemented `WebDAV` access method for an OCM share (`Webapp` and `Transfer` later in an other PR)
* Add command `ocm-share-get <id>` and `ocm-share-get-received <id>` to get the information of a single OCM share
* Written integration tests

The `sharedSecret` returned with the WebDAV protocol is still with owner permissions on the home folder that created the share. Will follow up an other PR with a proper OCM scoped token.


gmgigi96:
@michielbdejong can you review? This is touching even the `nextcloud` driver for the ocm shares

internal/grpc/services/gateway/ocmshareprovider.go
```diff
@@ -267,8 +205,8 @@ func (s *svc) UpdateReceivedOCMShare(ctx context.Context, req *ocm.UpdateReceive
 					panic("gateway: error updating a received share: the share is nil")
 				}
 
-				if share.GetShare().ShareType == ocm.Share_SHARE_TYPE_TRANSFER {
```
glpatcern, line 270: @redblom for your info this is now how an OCM share is identified to be a `datatx` one

glpatcern (approved):
OK, let's go ahead with this large change that already includes several unit tests, but we need to test in QA now.

The transfer (or `datatx`) OCM shares where also refactored and will need some testing as well in the context of ScienceMesh (cc @redblom).

All OCM-related APIs are much clearer now, and a CLI is also available, which eases further developments.

PR #3649, opened 2023-02-09
Store extended attributes in files in stead of in extended attributes

aduffeck:
We added a new metadata backend for the decomposed storage driver that uses an additional `.ini` file to store file metadata. This allows scaling beyond some filesystem specific xattr limitations.


butonic (approved):
Yuck, the missing abstraction of a node causes a lot of noise and making the linter happy causes even more noise. The prefix usage is sprinkled over all decomposedfs files ... we really need to move that to the node as in https://github.com/cs3org/reva/pull/3261/commits/87a0d23e02c971378ce3275b3cb700bdae67ff27 to truly decouple the implementation. The current xattr backend interface only ties the code further together by introducing methods like `IsMetaFile(path string) bool`, `UsesExternalMetadataFile() bool` and especially `MetadataPath(path string) string`.

I'm ok with merging it as is, as it introduces a new experimental backend for metadata which frees us from xattr limitations in some corner cases. But before touching this again we should refactor this mess ASAP.

pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -984,54 +979,50 @@ func (t *Tree) readRecycleItem(ctx context.Context, spaceID, key, path string) (
...
+	trashItem = filepath.Join(t.lookup.InternalRoot(), "spaces", lookup.Pathify(spaceID, 1, 2), "trash", lookup.Pathify(key, 4, 2))
+	resolvedTrashItem, err := filepath.EvalSymlinks(trashItem)
+	if err != nil {
+		return
+	}
+	deletedNodePath, err = filepath.EvalSymlinks(filepath.Join(resolvedTrashItem, path))
+	if err != nil {
+		return
 	}
+	nodeIDRegep := regexp.MustCompile(`.*/nodes/([^.]*).*`)
```
C0rby, line 993: This should be a global variable to not compile the regex everytime

C0rby (approved):
LGTM. Just one small nitpick

kobergj (approved):
LGTM :+1: But can we run a [full-ci] pipeline with this changes on ocis before merging? If you did already please ignore the comment.

butonic:
> LGTM 👍 But can we run a [full-ci] pipeline with this changes on ocis before merging? If you did already please ignore the comment.

running in https://drone.owncloud.com/owncloud/ocis/19330

butonic:
ocis full-ci green, too https://drone.owncloud.com/owncloud/ocis/19330

labkode:
Hi, I just want to give my 2 cents on this.

Keeping a metadata file in sync with the file is referring to is **very** hard task. Atomicity is paramount to avoid race conditions. 
- How are you preventing two concurrent updates to the metadata  of the file?
- The use of `flock` is process-dependant. The storage layer dealing with files is usually one single process . How do you `flock`.
- The use of `flock` it's has always a pain in the ass on top of NFS (what enterprises usually want). `fnctl` should be used in these cases.

PR #3650, opened 2023-02-12
FreeBSD xattr support

joramkruijer:
Since #3559, reva compiles perfectly fine on FreeBSD, but runs into problems with the `xattr` on FreeBSD:

`Failed to set all xattrs: xattr.Set   user.ocis.blobsize: invalid argument`

This MR adds support for FreeBSD's xattr namespaces. On Linux, every namespace is accessible, but in FreeBSD (and in FreeBSD jails) this is not the case: the `user` part of the namespace is the only namespace writable, and is the implied namespace. 

Setting a namespace manually, i.e. `user.ocis` will have the system refuse to use said namespace. Omitting the `user.` part of the namespace on FreeBSD solves this bug. The complete namespace will still be `user.ocis`, as FreeBSD prepends this part automatically.

Many thanks to @BlackNovaTech for help with the debugging.

### DIFF
--- a/changelog/unreleased/freebsd-xattr-support.md
+++ b/changelog/unreleased/freebsd-xattr-support.md
@@ -1,0 +1,5 @@
+Bugfix: FreeBSD xattr support
+
+We now properly handle FreeBSD xattr namespaces by leaving out the `user.` prefix. FreeBSD adds that automatically.
+
+https://github.com/cs3org/reva/pull/3650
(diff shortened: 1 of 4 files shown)


micbar:
@joramkruijer Thanks for the PR!

@butonic FYI

butonic (approved):

joramkruijer:
@butonic Thanks, it sure was an interesting thing to solve! I've added the changelog entry you proposed, I think it covers the fix perfectly.

butonic (approved):

PR #3655, opened 2023-02-15
remove expired spaces grants on access

C0rby:
When I implement the spaces membership expiration I forgot to actually check the expiration and remove expired grants.
This adds these missing bits. 

/cc @ScharfViktor when this was merged and reva was updated in oCIS then your tests should work.


C0rby:
@micbar, can you tell me why the changelog step is failing? Do I need to change the order of the PR URLs or something else?

micbar:
> can you tell me why the changelog step is failing? Do I need to change the order of the PR URLs or something else?

I think the PR always needs to be in the first line of the list because it tries to match the id

micbar (approved):

PR #3657, opened 2023-02-16
[WIP] Metadata store

butonic:
I'm fed up with the node metdata mess ...


butonic:
ok ... afaict we need te intreduce interfaces for trash, trash nodes, revisions and revision nodes before we can introduce a real node metadata abstraction.

butonic:
I am no longer working on this. Feel free to pick it up. I'll happily review.

PR #3671, opened 2023-02-21
Quota

C0rby:
- Added a global max quota option to limit how much quota can be assigned to spaces.
- Added a quota parameter to CreateHome. This is a prerequisite for setting a default quota per usertypes.


kobergj (approved):

micbar:
@aduffeck @C0rby For these kind of changes, we need to also think about the EOS driver in the future.

PR #3686, opened 2023-02-24
Return no users when user disabled.

ainmosni:
This alters the `GetUserByClaim` method to return not found if the user is disabled, via either the property, or the local group way.

### DIFF
--- a/pkg/utils/ldap/identity.go
+++ b/pkg/utils/ldap/identity.go
@@ -41,6 +41,7 @@ type userConfig struct {
 	scopeVal            int
 	Filter              string     `mapstructure:"user_filter"`
 	Objectclass         string     `mapstructure:"user_objectclass"`
+	EnabledProperty     string     `mapstructure:"user_enabled_property"`
 	Schema              userSchema `mapstructure:"user_schema"`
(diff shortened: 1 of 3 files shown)


pkg/utils/ldap/identity.go
```diff
@@ -151,6 +155,22 @@ func (i *Identity) Setup() error {
 		return fmt.Errorf("error configuring group substring filter type: %w", err)
 	}
 
+	dm := i.User.DisableMechanism
+	if dm == "" || dm == "none" || dm == "attribute" || dm == "group" {
+		if dm == "attribute" || dm == "group" {
+			if i.User.EnabledProperty == "" {
```
rhafer, line 161: I think we need to support the case we `dm == "group"` and `i.User.EnabledProperty==""`. I.e. only evaluating the enabled state based on the group membership.

rhafer: BTW, I think this would more readable if it used a `switch` statement instead of having those nested `if`s with lots of conditions.

rhafer (changes requested):

rhafer (approved):

Issue #2739, opened 2022-04-12
OCM WebDAV service


ishank011:
# What?

The Open Cloud Mesh (OCM) is a protocol that enables sharing of data across various cloud providers, given that they agree on the APIs that they'll be exposing to users from these other providers. For different reva deployments, each of which might connect to different storage backends and authentication providers, this is supposed to work out of the box.

# The workflow
## Invitation workflow
1. Suppose user `Alice` at `CERN` wants to share data with `Bob` at `CESNET`. Both the reva deployments at these two organizations will come to know of the other sites by either reading from a json config file or connecting to a central database, we support both the options. Let's say for now that both the sites have the json files with the following content:
```
[
	{
		"name": "cernbox",
		"services": [
			{
				"endpoint": {
					"type": {
						"name": "OCM"
					},
					"path": "https://cernbox.cern.ch/ocm/"
				},
			},
			{
				"endpoint": {
					"type": {
						"name": "Webdav"
					},
					"path": "https://cernbox.cern.ch/remote.php/webdav/",
				},
			},
		]
	},
	{
		"name": "oc-cesnet",
		"services": [
			{
				"endpoint": {
					"type": {
						"name": "OCM"
					},
					"path": "https://cesnet.cz/ocm/",
				},
			},
			{
				"endpoint": {
					"type": {
						"name": "Webdav"
					},
					"path": "https://cesnet.cz/remote.php/webdav/"
				},
			},
		]
	}
]
```

2. Now both the sites know of each other, but not about the individual users they have. So `Alice@CERN` will send a `/invites/forward` request to `Bob@CESNET`, asking Bob if he wants to begin sharing of data.
3. Now if `Bob` accepts the invite, the reva deployment at CESNET will send some metadata about Bob to CERN and then Alice can search for Bob and begin sharing of data with him.

*Note that the invitation workflow is already implemented and not under the scope of this project. I've described it only for developers to get a better understanding of how OCM works.*

## OCM sharing
1. Now the reva deployment at CERN, and hence Alice, knows about Bob@CESNET. So she'll send a `POST /shares` request to Reva@CERN with the following parameters:
```
Type: OCM-Share
Path: /users/alice/myFolder
Permissions: Editor
Sharee: Bob@CESNET
```

[...]

PR #3691, opened 2023-02-27
Add OCM scope and webdav endpoint

gmgigi96:
This PR adds the OCM scope and the ocmshares authentication, to authenticate the federated user to use the OCM shared
resources.
It also adds the (unprotected) webdav endpoint used to interact with the shared resources as described in https://github.com/cs3org/reva/issues/2739.

Closes https://github.com/cs3org/reva/issues/2739


glpatcern (dismissed):
Tested for CS3

glpatcern (approved):

PR #3692, opened 2023-02-28
Create and list OCM shares in OCS layer

gmgigi96:
This PR adds the support of OCM shares in the OCS layer, allowing a client to create, listing the created shares and the received shares transparently.


glpatcern (approved):
Already demonstrated in a live demo

PR #3693, opened 2023-02-28
enforce the publiclink.create permission

C0rby:
Added checks for the "publiclink.create" permission when creating or updating public links.

### DIFF
--- a/pkg/permission/permission.go
+++ b/pkg/permission/permission.go
@@ -27,6 +27,8 @@ const (
 	ListAllSpaces string = "list-all-spaces"
 	// CreateSpace is the hardcoded name for the create space permission
 	CreateSpace string = "create-space"
+	// WritePublicLink is the hardcoded name for the PublicLink.Write permission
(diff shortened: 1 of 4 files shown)


pkg/permission/permission.go
```diff
@@ -27,6 +27,8 @@ const (
 	ListAllSpaces string = "list-all-spaces"
 	// CreateSpace is the hardcoded name for the create space permission
 	CreateSpace string = "create-space"
+	// CreatePublicLink is the hardcoded name for the publiclink.create permission
+	CreatePublicLink string = "publiclink.create"
```
kobergj, line 31: We should decide on on how we want to name permission. Last new implemented permissions go for `Drive.ReadWriteEnabled` (uppercase) therefore we should make this uppercase too, or change existing permissions to go lowercase.

kobergj (changes requested):

kobergj (approved):

PR #3695, opened 2023-03-01
Create OCM share from sciencemesh service

gmgigi96:
As in the recent refactor of the OCM service the endpoint /ocm/share was removed, this PR re-add it back.

Closes https://github.com/pondersource/sciencemesh-php/issues/166


glpatcern (dismissed):
@michielbdejong now `/ocm/send` is `/sciencemesh/create-share` !

internal/http/services/sciencemesh/sciencemesh.go
```diff
@@ -108,7 +113,7 @@ func (s *svc) Prefix() string {
 }
 
 func (s *svc) Unprotected() []string {
-	return nil
+	return []string{"/create-share"}
```
michielbdejong, line 116: Why Unprotected? All endpoints including this one should be protected by HTTP Basic Auth with username/password which gets checked by the Auth Manager, right?

internal/http/services/sciencemesh/share.go
```diff
@@ -0,0 +1,200 @@
...
+func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
+	log := appctx.GetLogger(r.Context())
+
+	req, err := getCreateShareRequest(r)
+	if err != nil {
+		reqres.WriteError(w, r, reqres.APIErrorInvalidParameter, "invalid parameters", err)
+		return
+	}
+
+	// FIXME: the client should be already authenticated
```
michielbdejong, line 80: Hm, maybe this is where authentication is checked?

michielbdejong (dismissed):
LGTM!

Maybe you can double-check that this endpoint returns a 401 when you hit it without any authentication.

internal/http/services/sciencemesh/sciencemesh.go
gmgigi96: Because the original "/ocm/send" has some auth parameters in the request. To keep it compatible i left the authentication in the /create-share request. If it's protected this requires double auth.
But I'm fine we can remove the authentication phase in this endpoint, and leave it protected.

glpatcern: Reviving this: I fully support removing the "custom auth" implemented via auth parameters in the request payload. It was there for some legacy reasons but it makes more sense to use the auth interceptors and make the endpoint protected.

michielbdejong:
FYI I merged this branch into https://github.com/cs3org/reva/tree/sciencemesh-dev

internal/http/services/sciencemesh/sciencemesh.go
```diff
@@ -116,7 +120,7 @@ func (s *svc) Prefix() string {
...
```
glpatcern, line 123:
```suggestion
	return nil
```

internal/http/services/sciencemesh/share.go
```diff
@@ -0,0 +1,200 @@
...
+	SourcePath        string `json:"sourcePath" validate:"required"`
+	TargetPath        string `json:"targetPath" validate:"required"`
+	Type              string `json:"type"`
+	Role              string `json:"role" validate:"oneof=viewer editor"`
+	RecipientUsername string `json:"recipientUsername" validate:"required"`
+	RecipientHost     string `json:"recipientHost" validate:"required"`
+	// FIXME: the client should not authenticate here
+	LoginType     string `json:"loginType" validate:"required"`
+	LoginUsername string `json:"loginUsername" validate:"required"`
+	LoginPassword string `json:"loginPassword" validate:"required"`
```
glpatcern, line 67:
```suggestion
```

internal/http/services/sciencemesh/share.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+	if err != nil {
+		reqres.WriteError(w, r, reqres.APIErrorInvalidParameter, "unexpected error", err)
+		return
+	}
+	if res.Status.Code != rpc.Code_CODE_OK {
+		reqres.WriteError(w, r, reqres.APIErrorUnauthenticated, res.Status.Message, errors.New(res.Status.Message))
+		return
+	}
+
```
glpatcern, line 94:
```suggestion
```

glpatcern (changes requested):
As there were no objections, let's go ahead and fix the authentication as discussed, and then we merge this.

gmgigi96:
> As there were no objections, let's go ahead and fix the authentication as discussed, and then we merge this.

The PR is now ready

glpatcern (approved):

PR #3697, opened 2023-03-02
Optimize file metadata

aduffeck:
This PR improves the ini backend for file metadata:
  - Improve performance
  - Optionally use a reva cache for storing the metadata, which helps
    tremendously when using distributed file systems, for example
  - Allow for using different metadata backends for different storages

We also switched the s3ng integration tests to the ini backend so we cover both the xattrs and the ini backend at the same time.

Finally it adds support for redis sentinel caches (https://github.com/owncloud/ocis/issues/5645)


butonic (approved):
renaming the xattrs package to metadata and adding the Metadata interface to PathLookup makes sense. I see how you are sneaking in a Metadata backend into the Node. ;-)

Still needs some hoops for the revisions and trash, but this is a step forward. ❤️

PR #3710, opened 2023-03-08
Fix error handling in DeleteStorage space

rhafer:
Return a permission denied error when the user is not able to list the to be delete space.


kobergj (approved):
Thanky for fixing :bow:

PR #3714, opened 2023-03-09
respect the max quota when updating the space quota

C0rby:
When the space quota should be updated to '0' i.e. unlimited and a global max quota was configured we only give the space the max quota amount instead of unlimited storage.


micbar (approved):
Looks good.

PR #3716, opened 2023-03-09
properly enforce the password on writeable public links

C0rby:
@ScharfViktor found some bugs in the first implementation of this feature.
This PR fixes these issues and properly handles share creation and updates.

### DIFF
--- a/pkg/publicshare/publicshare.go
+++ b/pkg/publicshare/publicshare.go
@@ -24,6 +24,7 @@ import (
 	"crypto/sha256"
 	"crypto/sha512"
 	"encoding/hex"
+	"errors"
 	"time"
 
 	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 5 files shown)


pkg/publicshare/publicshare.go
```diff
@@ -213,8 +219,8 @@ func IsCreatedByUser(share link.PublicShare, user *user.User) bool {
 }
 
 // IsWriteable checks if the grant for a publicshare allows writes or uploads.
-func IsWriteable(g *link.Grant) bool {
-	p := g.Permissions.Permissions
-	return p.CreateContainer || p.Delete || p.InitiateFileUpload ||
+func IsWriteable(perm *link.PublicSharePermissions) bool {
+	p := perm.GetPermissions()
+	return p != nil && p.CreateContainer || p.Delete || p.InitiateFileUpload ||
 		p.Move || p.AddGrant || p.PurgeRecycle || p.RestoreFileVersion || p.RestoreRecycleItem
```
kobergj, line 225: Wouldn't this still panic when `p == nil`? It will evaluate `p != nil && p.Createcontainer` (which is false), then continue with evaluating `p.Delete` (which panics). Or do I miss some magic here?

kobergj (changes requested):

pkg/publicshare/publicshare.go
C0rby: I think you're right. I'm missing parentheses there but let me check this quickly to confirm.

C0rby: Fixed it.

kobergj (approved):

PR #3728, opened 2023-03-14
Migrate to msgpack

aduffeck:


butonic (approved):

pkg/storage/utils/decomposedfs/migrator/migrator.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, migration := range allMigrations {
+		s := m.states[migration]
+		switch s.State {
+		case "succeeded":
+			m.log.Info().Msg("Skipping finished migration " + migration + "...")
+			continue
+		case "":
+			m.log.Info().Msg("Running migration " + migration + "...")
+		default:
+			m.log.Info().Msg("Re-running migration " + migration + "...")
```
butonic, line 87: we should probably drop any logs that show up during startup unless a migration actually needs to happen. We can think of passing down a logger or better just context the whole chain of services to clean this up in a subsequent PR.

But for now lets silence these logs, ok?

aduffeck: Sure, I'll push an according commit soon.

butonic (approved):

PR #3739, opened 2023-03-20
Try to rename the file after upload instead of copying.

dragotin:
If renaming fails for whatever reason, it falls back to copying as before. This comes for the cost of an interface change towards file name instead of Reader.
Plus some fixed tests.


butonic (approved):

butonic:
@dragotin you'll have to do a make go-generate to update the mocks and check them in.

Issue #3677, opened 2023-02-22
d.SharedSecret undefined


michielbdejong:
This line throws a nil pointer exception when Marie tries to accept a transfer-create from Einstein: https://github.com/cs3org/reva/blame/94ced01208864b66b21d2c54977c6df548d37974/internal/grpc/services/gateway/ocmshareprovider.go#L429

michielbdejong:
CC @gmgigi96 @redblom

redblom:
See https://github.com/pondersource/sciencemesh-php/issues/159#issuecomment-1440419702

redblom:
Results of the local regular! share tests so far:
1. In `gateway.ocmshareprovider.createOCMReference` the secret is extracted in `getTransferProtocol` through `TransferOptions`, this should be `WebdavOptions`. Consider creating a `getWebdavProtocol` method.
2. The targetURL created in `gateway.ocmshareprovider.createOCMReference` has an incorrect format. The host part also contains the idp schema. Subsequently in `gateway.webdavstorageprovider.ExtractEndpointInfo` the endpoint extraction fails. The host part in the target url should lack the schema.

Result of local transfer share tests:
1. The transfer protocol is not send with the request, and thus the share will not be recognized as a transfer.

redblom:
Can be closed. Dealt with in PR https://github.com/cs3org/reva/pull/3847.

PR #3743, opened 2023-03-23
Use webdav protocol when creating ocm reference.

redblom:
This fixes the 'd.SharedSecret undefined' issue of https://github.com/cs3org/reva/issues/3677

### DIFF
--- a/changelog/unreleased/fix-createocmreference.md
+++ b/changelog/unreleased/fix-createocmreference.md
@@ -1,0 +1,3 @@
+Bugfix: Fix createOCMReference - correct the protocol
+
+https://github.com/cs3org/reva/pull/3743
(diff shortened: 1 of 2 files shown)


glpatcern (changes requested):
I have the impression that there's some misunderstanding on which OCM `protocol` to use here, @redblom can we have a chat also with @gmgigi96?

redblom:
After testing again with the latest ocm version implementation: a regular ocm share will not need to be accepted anymore. It is accessible 'by default'. The `gateway.ocmshareprovider.createOCMReference` method has become obsolete.
For transfers however we still need acceptance of the (datatx) share.

redblom:
This is dealt with in https://github.com/cs3org/reva/pull/3847

PR #3744, opened 2023-03-23
Add UserTypeProperty to get user type from ldap user entry

Excds:
Adding an LDAP attribute UserTypeProperty to be able to distinguish between member and guest users.

### DIFF
--- a/pkg/utils/ldap/identity.go
+++ b/pkg/utils/ldap/identity.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"strings"
 
+	identityUser "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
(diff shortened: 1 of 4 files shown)


rhafer (approved):

rhafer (changes requested):
Please use the full URL of the PR in the changelog

butonic (approved):

pkg/utils/ldap/identity.go
```diff
@@ -685,3 +692,16 @@ func (i *Identity) getGroupAttributeFilter(attribute, value string) (string, err
 		value,
 	), nil
 }
+
+// GetUserType is used to set the proper UserType from ldap entry string
```
kobergj, line 696:
```suggestion
// GetUserType is used to get the proper UserType from ldap entry string
```

kobergj (approved):

micbar (approved):

PR #3749, opened 2023-03-24
Fix persistence of received share update #3747

### LINKED ISSUE #3747: UpdateReceivedShare of pkg ocm json driver does not actually persist the updated field
Using the ocm share json driver, when accepting an ocm share the state field of the model is not actually updated to `accepted` (ie. not saved). Although the `State` field of the retrieved share is updated, https://github.com/cs3org/reva/blob/31fb861daa204f326d6844876146ace03bbf7f83/pkg/ocm/share/repository/json/json.go#L510
the mrg model is not. And thus the update is not persisted. And listing the received share still shows it as `pending`.

This https://github.com/cs3org/reva/blob/31fb861daa204f326d6844876146ace03bbf7f83/pkg/ocm/share/repository/json/json.go#L495
does not return a pointer to the model and thus the model is not updated.

redblom:
Update the model before persisting it.
Fixes https://github.com/cs3org/reva/issues/3747

### DIFF
--- a/changelog/unreleased/fix-3747.md
+++ b/changelog/unreleased/fix-3747.md
@@ -1,0 +1,3 @@
+Bugfix: Fix persisting updates of received shares in json driver
+
+https://github.com/cs3org/reva/pull/3749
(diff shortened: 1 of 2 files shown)


michielbdejong:
Great! I'll try it out
CC @mahdibaghbani

glpatcern (approved):

PR #3755, opened 2023-03-27
change language code validation

micbar:
Bugfix: Fix app provider language validation

This changes the validation to only look at the first part (tag) of the language code and ignore the second part (sub-tag).


labkode:
@glpatcern something worth cherry-picking to master?

glpatcern:
Yep, looked at it already 👀 and no, the way we pass the language in master is actually different as we read it from the configuration, and we're assuming the config is correct (i.e. not having the "subtag"). We use `en-GB` in prod.

kobergj (approved):

PR #3757, opened 2023-03-28
Support OCM v1.0 schema

glpatcern:
Following https://github.com/cs3org/cs3apis/pull/206, we add the fields to ensure backwards compatibility with OCM v1.0.

However, if we receive a request where the `protocol.options` undocumented object is not empty, we bail out for now. Also, we always send out an empty `protocol.options`, and use the newly defined properties that (will) come with OCM v1.1. This is reflected in the unit tests as well.

Supporting interoperability with existing OCM v1.0 implementations (notably Nextcloud 25) may come in the future once the undocumented options are fully reverse engineered.

Also added `viewMode` to the webapp protocol and adapted the SQL layer.

### DIFF
--- a/internal/http/services/ocmd/protocols.go
+++ b/internal/http/services/ocmd/protocols.go
@@ -80,7 +80,7 @@ type Webapp struct {
 
 // ToOCMProtocol convert the protocol to a ocm Protocol struct.
 func (w *Webapp) ToOCMProtocol() *ocm.Protocol {
-	return ocmshare.NewWebappProtocol(w.URITemplate, utils.GetAPViewMode(w.ViewMode))
+	return ocmshare.NewWebappProtocol(w.URITemplate, utils.GetAppViewMode(w.ViewMode))
 }
 
 // Datatx contains the parameters for the Datatx protocol.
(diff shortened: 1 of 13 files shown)


internal/http/services/ocmd/protocols.go
```diff
@@ -73,11 +75,12 @@ func (w *WebDAV) ToOCMProtocol() *ocm.Protocol {
 // Webapp contains the parameters for the Webapp protocol.
 type Webapp struct {
 	URITemplate string `json:"uriTemplate" validate:"required"`
+	ViewMode    string `json:"viewMode" validate:"required,dive,required,oneof=view read write"`
 }
 
 // ToOCMProtocol convert the protocol to a ocm Protocol struct.
 func (w *Webapp) ToOCMProtocol() *ocm.Protocol {
-	return ocmshare.NewWebappProtocol(w.URITemplate)
+	return ocmshare.NewWebappProtocol(w.URITemplate, utils.GetAPViewMode(w.ViewMode))
```
gmgigi96, line 83: There is a typo here `GetAPViewMode` -> `GetAppViewMode`

gmgigi96 (changes requested):
Can you also update the init script for creating the webapp table? https://github.com/cs3org/reva/blob/master/pkg/ocm/share/repository/sql/init.sql#L72

gmgigi96 (approved):

PR #3758, opened 2023-03-28
check for password during update public link

micbar:
Bugfix: Fix public links with enforced password

Fix the public link update operation in the case that a password is enforced.

### DIFF
--- a/pkg/publicshare/manager/json/json.go
+++ b/pkg/publicshare/manager/json/json.go
@@ -343,8 +343,20 @@ func (m *manager) UpdatePublicShare(ctx context.Context, u *user.User, req *link
 		old, _ := json.Marshal(share.Permissions)
 		new, _ := json.Marshal(req.Update.GetGrant().Permissions)
 
-		if m.writeableShareMustHavePassword && publicshare.IsWriteable(req.GetUpdate().GetGrant().GetPermissions()) && !share.PasswordProtected {
+		if m.writeableShareMustHavePassword &&
+			publicshare.IsWriteable(req.GetUpdate().GetGrant().GetPermissions()) &&
+			(!share.PasswordProtected && req.GetUpdate().GetGrant().GetPassword() == "") {
 			return nil, publicshare.ErrShareNeedsPassword
+		}
+
+		if req.GetUpdate().GetGrant().GetPassword() != "" {
+			passwordChanged = true
+			h, err := bcrypt.GenerateFromPassword([]byte(req.Update.GetGrant().Password), m.passwordHashCost)
+			if err != nil {
+				return nil, errors.Wrap(err, "could not hash share password")
+			}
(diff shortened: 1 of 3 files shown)


pkg/publicshare/manager/json/json.go
```diff
@@ -343,10 +343,22 @@ func (m *manager) UpdatePublicShare(ctx context.Context, u *user.User, req *link
 		old, _ := json.Marshal(share.Permissions)
 		new, _ := json.Marshal(req.Update.GetGrant().Permissions)
 
-		if m.writeableShareMustHavePassword && publicshare.IsWriteable(req.GetUpdate().GetGrant().GetPermissions()) && !share.PasswordProtected {
+		if m.writeableShareMustHavePassword &&
+			publicshare.IsWriteable(req.GetUpdate().GetGrant().GetPermissions()) &&
+			(!share.PasswordProtected && req.GetUpdate().GetGrant().GetPassword() == "") {
```
kobergj, line 348: I don't understand why we check for `!share.PasswordProtected`. Why is it relevant? 

```suggestion
		if m.writeableShareMustHavePassword &&
			publicshare.IsWriteable(req.GetUpdate().GetGrant().GetPermissions()) &&
			req.GetUpdate().GetGrant().GetPassword() == "" {
```

pkg/publicshare/manager/json/json.go
```diff
@@ -343,10 +343,22 @@ func (m *manager) UpdatePublicShare(ctx context.Context, u *user.User, req *link
...
 			return nil, publicshare.ErrShareNeedsPassword
 		}
 
+		if req.GetUpdate().GetGrant().GetPassword() != "" {
```
kobergj, line 352: Why do we change permissions and the password together? Shouldn't a password change be done within the `link.UpdatePublicShareRequest_Update_TYPE_PASSWORD` case?

kobergj (changes requested):
confused

micbar:
> confused

I was too at the beginning. Let me explain.

From the backend side, things always look different. 

In the web frontend you have the following situation:

1) You create a link. During that, you cannot specify permissions, it just uses the default and no password
2) You want to change the permissions to "Editor". Now the Frontend side validation kicks in ans says: "hey, that doesn't work without a password!" and prompts a password. You  provide one and violà, we have a request that updates the permissions and the password together. IMO this makes sense.

pkg/publicshare/manager/json/json.go
micbar: this is relevant when the password has already been set before. Then we are able to change the permissions without a new password.

micbar: You only need to provide a new password when we have no existing password.

kobergj (approved):

PR #3759, opened 2023-03-30
Set correct share type when listing shares

Excds:
This fixes so that we can differentiate between guest/member users for shares.

# Description

## List Shares

When listing shares on a file/folder with a GET on the following endpoint: /ocs/v1.php/apps/files_sharing/api/v1/shares the entities in the response have a shareType of 0 for all users, no matter if they are guests or not at the time of the request (already implemented like that). However, we need an additional attribute share_with_user_type for the shares where the user represented by the share_with* attributes is a guest (value 1) or not a guest (value 0) at the time of the request. The reason for this additional attribute instead of a shareType 0 vs 4 is that we don't want the need to migrate shares when a guest user becomes a non-guest or vice versa.

## List sharees (share candidates)

when searching for potential sharees on a file/folder with a GET on the following endpoint:
/ocs/v2.php/apps/files_sharing/api/v1/sharees?search=ein the entities in the response have a shareType and a userType. shareType needs to be 0 and userType needs to be 1 for users that are guests at the time of the request. We pulled that out of [demo.owncloud.com](http://demo.owncloud.com/)

@Excds @kobergj FYI

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
@@ -100,15 +100,17 @@ func (h *Handler) FindSharees(w http.ResponseWriter, r *http.Request) {
 }
 
 func (h *Handler) userAsMatch(u *userpb.User) *conversions.MatchData {
-	shareType := conversions.ShareTypeUser
-	if u.Id.Type == userpb.UserType_USER_TYPE_GUEST {
-		shareType = conversions.ShareTypeGuestUser
+	var ocsUserType int
+	if u.Id.Type == userpb.UserType_USER_TYPE_GUEST || u.Id.Type == userpb.UserType_USER_TYPE_LIGHTWEIGHT {
+		ocsUserType = 1
 	}
 
 	return &conversions.MatchData{
 		Label: u.DisplayName,
 		Value: &conversions.MatchValueData{
-			ShareType: int(shareType),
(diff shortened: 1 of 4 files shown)


micbar (approved):

internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
```diff
@@ -100,10 +100,15 @@ func (h *Handler) FindSharees(w http.ResponseWriter, r *http.Request) {
...
+	var userType int
+	if u.Id.Type == userpb.UserType_USER_TYPE_GUEST || u.Id.Type == userpb.UserType_USER_TYPE_LIGHTWEIGHT {
+		userType = 1
+	}
+
 	return &conversions.MatchData{
 		Label: u.DisplayName,
 		Value: &conversions.MatchValueData{
-			ShareType: int(conversions.ShareTypeUser),
+			ShareType: userType,
```
dragonchaser, line 111: I find this a bit confusing here, there is a fork from @c0rby here that defines a struct for usertypes https://github.com/C0rby/go-cs3apis/blob/main/cs3/identity/user/v1beta1/resources.pb.go#L27-L45 that seems to be still consumend (though not consistent with https://github.com/cs3org/go-cs3apis)... can we resolve that or use a different variable name here?

micbar: different API, this is a custom ocs property.

Changed the naming and added comment.

PR #3762, opened 2023-03-31
Denial and Resharing Default capability

diocas:
2 missing capabilities from upstream


glpatcern (commented):
What is the difference between `Resharing` and `ResharingDefault`?

diocas:
Resharing allows resharing, Resharing Default sets wether the other roles (editor, reader, etc) enable resharing or not (if not, you can use the custom permissions to set it.

glpatcern (dismissed):

gmgigi96 (approved):

PR #3767, opened 2023-04-03
ldap: Fix binary UUID handling in GetUserGroups

rhafer:
The LDAP backend for the users service didn't correctly decode binary UUIDs when looking up a user's group memberships.

### DIFF
--- a/pkg/utils/ldap/identity.go
+++ b/pkg/utils/ldap/identity.go
@@ -358,7 +358,19 @@ func (i *Identity) GetLDAPUserGroups(log *zerolog.Logger, lc ldap.Client, userEn
 		// FIXME this makes the users groups use the cn, not an immutable id
 		// FIXME 1. use the memberof or members attribute of a user to get the groups
 		// FIXME 2. ook up the id for each group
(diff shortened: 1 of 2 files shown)


pkg/utils/ldap/identity.go
```diff
@@ -358,7 +358,19 @@ func (i *Identity) GetLDAPUserGroups(log *zerolog.Logger, lc ldap.Client, userEn
...
 		// FIXME 2. ook up the id for each group
-		groups = append(groups, entry.GetEqualFoldAttributeValue(i.Group.Schema.ID))
+		var groupID string
+		if i.Group.Schema.IDIsOctetString {
+			raw := entry.GetEqualFoldRawAttributeValue(i.Group.Schema.ID)
+			if value, err := uuid.FromBytes(raw); err == nil {
+				groupID = value.String()
+			} else {
+				return nil, err
+			}
```
kobergj, line 368:
```suggestion
			value, err := uuid.FromBytes(raw)
			if err != nil {
				return nil, err
			}
			groupID = value.String()
```

kobergj (approved):
Small suggestion because I hate `else`s :smile: But definitely not critical.

PR #3768, opened 2023-04-04
wopi driver: force preview mode for external or non-owner users

glpatcern:


micbar:
@glpatcern Interesting issue. Was that fixed otherwise?

glpatcern:
@micbar we're going to fix this from the frontend only - we need the frontend to drive this so it can listen to postMessages. My first attempt to have that simple switch in the backend is not complete... But the issue per se is valid, we had several users reporting that they'd prefer not to `touch` a file that is not theirs but just open it in preview mode.

micbar:
Couldn't we do that with another file action? "open in <AppName> (Preview)"?

glpatcern:
We can, yes, and the backend already supports that by passing `view_mode=preview` on the `/app/open` call.

This is in `master` of course... I think at some point we need to at least merge the appprovider!

PR #3775, opened 2023-04-05
Fix #3774

### LINKED ISSUE #3774: Provide debug information about OCM requests/responses
In https://github.com/pondersource/sciencemesh-php/issues/180#issuecomment-1497024666 I'm running into a situation where Reva CLI says "invalid parameters" and Revad logs say nothing at all.
Spelonking the code, it looks like the error is actually outside our system (a 400 response from the OCM receiver), so this definitely feels like something that should be logged not only in development but even in production.

michielbdejong:
Add debug logs for the OCM /shares call.
Note that this does not resolve the problem for the other calls that this client makes, such as invitations and notifications.

### DIFF
--- a/pkg/ocm/client/client.go
+++ b/pkg/ocm/client/client.go
@@ -214,7 +214,7 @@ func (c *OCMClient) NewShare(ctx context.Context, endpoint string, r *NewShareRe
 	// defer resp.Body.Close()
 	// body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		log.Debug().Msgf("OCM response from: %d; could not read body")
+		log.Debug().Msgf("OCM response from: %d; could not read body", resp.StatusCode)
 	}
 	log.Debug().Msgf("OCM response from: %d %s", resp.StatusCode, respBody)
 
--- a/pkg/ocm/client/client.go
+++ b/pkg/ocm/client/client.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/cs3org/reva/internal/http/services/ocmd"
+	"github.com/cs3org/reva/pkg/appctx"
 	"github.com/cs3org/reva/pkg/errtypes"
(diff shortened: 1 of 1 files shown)


glpatcern:
Nice, if you add the changelog we can merge this right away (modulo the errors that we have to understand)

pkg/ocm/client/client.go
```diff
@@ -194,6 +196,7 @@ func (c *OCMClient) NewShare(ctx context.Context, endpoint string, r *NewShareRe
 	if err != nil {
 		return nil, err
 	}
+	log.Debug().Msgf("OCM send to %s: %s", url, body)
```
gmgigi96, line 199: Can you add the url and body as key-values in the logs, instead of including directly in the message?
Something like `log.Debug().Str("url", url).Msg("<something>")`

pkg/ocm/client/client.go
```diff
@@ -206,6 +209,14 @@ func (c *OCMClient) NewShare(ctx context.Context, endpoint string, r *NewShareRe
...
+	respBody, err := io.ReadAll(resp.Body)
+
+	// defer resp.Body.Close()
+	// body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Debug().Msgf("OCM response from %s: %d; could not read body", url, resp.StatusCode)
+	}
+	log.Debug().Msgf("OCM response from %s: %d %s", url, resp.StatusCode, respBody)
 
 	return c.parseNewShareResponse(resp)
```
gmgigi96, line 221: Not sure this is going to work then, after consuming all the body.
You can think of adding the log in the `parseNewShareReponse`function.

gmgigi96 (changes requested):

michielbdejong:
FYI I included this branch into https://github.com/cs3org/reva/tree/sciencemesh-dev
I'll fix the requested changes!

PR #3780, opened 2023-04-06
Resolve Makefile error on Ubuntu

### LINKED ISSUE #3773: Makefile depends on bash
Since https://github.com/cs3org/reva/commit/0374913e39e043a1af7bf10179911531c3c191b6 it is no longer possible to build revad on Ubuntu, because it uses dash instead of bash as the shell, which doesn't support the `[[ ... ]]]` syntax.

MahdiBaghbani:
Makefile will be using bash as it's shell instead of sh which is defaulted to dash in Ubuntu and doesn't support `[[` syntax.

Fixes #3773

### DIFF
--- a/changelog/unreleased/fix-makefile.md
+++ b/changelog/unreleased/fix-makefile.md
@@ -3,5 +3,5 @@ Bugfix: Fix Makefile error on Ubuntu
 I've fixed Makefile using sh which is defaulted to dash in ubuntu, dash doesn't support
 `[[ ... ]]` syntax and Makefile would throw `/bin/sh: 1: [[: not found` errors.
 
-https://github.com/cs3org/reva/issues/3773
 https://github.com/cs3org/reva/pull/3780
+https://github.com/cs3org/reva/issues/3773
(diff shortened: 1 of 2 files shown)


changelog/unreleased/fix-makefile.md
```diff
@@ -0,0 +1,7 @@
+Bugfix: Fix Makefile error on Ubuntu
+
+I've fixed Makefile using sh which is defaulted to dash in ubuntu, dash doesn't support
+`[[ ... ]]` syntax and Makefile would throw `/bin/sh: 1: [[: not found` errors.
+
+https://github.com/cs3org/reva/issues/3773
+https://github.com/cs3org/reva/pull/3780
```
gmgigi96, line 7: The link to the pull request should be the first one

gmgigi96 (changes requested):

changelog/unreleased/fix-makefile.md
MahdiBaghbani: ah, I followed the TEMPLATE and put issue first, fixing it now.

gmgigi96 (approved):

PR #3788, opened 2023-04-13
Make Resharing Configurable

kobergj:
Makes resharing configurable via the capablity

Only `ocs` service has access to the capabilty. It can pass this easily to `conversion` package. However even though placed inside the `internal` folder, other services also import `conversion` but do not have access to the capabilty. 

These rogue imports of `conversions` package were hardcoded to `resharing=false` and need to be checked one by one before merging:
- [x] internal/http/services/ocmd/shares.go
    - returns error now
- [x] pkg/cbox/utils/conversions.go
    - fixed by proper conversion
- [x] ~pkg/publicshare/manager/owncloudsql/conversions.go~
- [x] ~pkg/share/manager/owncloudsql/conversions.go~
- [x] ~pkg/storage/fs/owncloudsql/filecache/filecache.go~
- [x] ~pkg/storage/fs/owncloudsql/owncloudsql.go~
- [x] pkg/storage/utils/eosfs/eosfs.go
    - defaulted to false - package needs a refactoring
- [x] cmd/reva/ocm-share-create.go
    - defaulted to false. Only applies when creating shares via cli
- [x] cmd/reva/share-create.go
    - defaulted to false. Only applies when creating shares via cli
- [ ] ...

### DIFF
--- a/cmd/reva/ocm-share-create.go
+++ b/cmd/reva/ocm-share-create.go
@@ -37,6 +37,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// default for resharing
+var _resharing = false
+
 func ocmShareCreateCommand() *command {
 	cmd := newCommand("ocm-share-create")
 	cmd.Description = func() string { return "create OCM share to a user or group" }
@@ -166,11 +169,11 @@ func ocmShareCreateCommand() *command {
 func getOCMSharePerm(p string) (*ocm.SharePermissions, int, error) {
 	if p == viewerPermission {
 		return &ocm.SharePermissions{
-			Permissions: conversions.NewViewerRole(false).CS3ResourcePermissions(),
+			Permissions: conversions.NewViewerRole(_resharing).CS3ResourcePermissions(),
 		}, 1, nil
 	} else if p == editorPermission {
 		return &ocm.SharePermissions{
-			Permissions: conversions.NewEditorRole(false).CS3ResourcePermissions(),
+			Permissions: conversions.NewEditorRole(_resharing).CS3ResourcePermissions(),
 		}, 15, nil
 	}
 	return nil, 0, errors.New("invalid rol: " + p)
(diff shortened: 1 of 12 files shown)


pkg/cbox/utils/conversions.go
```diff
@@ -128,20 +128,8 @@ func SharePermToInt(p *provider.ResourcePermissions) int {
...
-			return conversions.NewEditorRole().CS3ResourcePermissions()
-		}
-		return conversions.NewFileEditorRole().CS3ResourcePermissions()
-	case 4:
-		return conversions.NewUploaderRole().CS3ResourcePermissions()
-	default:
-		// TODO we may have other options, for now this is a denial
-		return &provider.ResourcePermissions{}
-	}
+	perms, _ := conversions.NewPermissions(p)
```
micbar, line 131: I do not understand the refactoring here. Can you explain?

kobergj: It was converting an `int` to a `*provider.ResourcePermissions`. But it was using a custom implementation. This is bad because it will break when somethin changes.
Other service do what we do now. Use `NewPermission` to create a new permissions struct, then use `CS3ResourcePermissions()` to convert it to the `*provider.ResourcePermissions`.
I can implement a custom logic, but I think going with the default is less error prone.

cmd/reva/ocm-share-create.go
```diff
@@ -166,11 +166,11 @@ func ocmShareCreateCommand() *command {
 func getOCMSharePerm(p string) (*ocm.SharePermissions, int, error) {
 	if p == viewerPermission {
 		return &ocm.SharePermissions{
-			Permissions: conversions.NewViewerRole().CS3ResourcePermissions(),
+			Permissions: conversions.NewViewerRole(false).CS3ResourcePermissions(),
```
dragonchaser, line 169: can we use constants here for readability?

dragonchaser (approved):
lgtm, just nitpicking about the constant :)

cmd/reva/ocm-share-create.go
kobergj: Good point :+1: Fixed it

pkg/cbox/utils/conversions.go
micbar: ok, let us see what happens.

micbar (approved):

PR #3792, opened 2023-04-14
Add prometheus metrics for active transfers.

ainmosni:
This PR adds simple gauges for active uploads and downloads. These
can then be picked up by the prometheus handler.

### DIFF
--- a/pkg/rhttp/datatx/manager/simple/simple.go
+++ b/pkg/rhttp/datatx/manager/simple/simple.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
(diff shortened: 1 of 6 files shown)


pkg/rhttp/datatx/manager/simple/simple.go
```diff
@@ -87,8 +101,16 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 
 		switch r.Method {
 		case "GET", "HEAD":
+			downloadsActive.Add(1)
```
butonic, line 104: only count downloads for GET, not HEAD?

pkg/rhttp/datatx/manager/simple/simple.go
```diff
@@ -40,6 +42,18 @@ import (
 	"github.com/cs3org/reva/v2/pkg/utils"
 )
 
+// Keep track of active uploads/downloads.
```
butonic, line 45: this is the 'simple' upload, pleas also check the spaces and tus datatx managers

pkg/rhttp/datatx/manager/simple/simple.go
ainmosni: I was considering that, but thought that HEAD requests would be short enough to not matter. I'll do it for GET so that the graphs are less jittery.

pkg/rhttp/datatx/manager/tus/tus.go
```diff
@@ -155,6 +160,10 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 		case "DELETE":
 			handler.DelFile(w, r)
 		case "GET":
+			metrics.DownloadsActive.Add(1)
+			defer func() {
+				metrics.DownloadsActive.Sub(1)
+			}()
```
kobergj, line 166: Have you tested this? I think this the wrong place for the download counter. Afair the `GetFile` method will only return files that are currently being uploaded.

kobergj (changes requested):

pkg/rhttp/datatx/manager/tus/tus.go
ainmosni: I tested this, and the download worked fine, it was the POST/PATCH that didn't work as expected, as the upload is chunked. Added the counter to the PATCH as well.

dragonchaser (approved):
lgtm

PR #3809, opened 2023-04-20
pass down context for tracing

butonic:
rewrote this as a pr to pass the context deep down into decomposedfs metadata backends and trace syscall durations

### DIFF
--- a/pkg/storage/cache/filemetadata.go
+++ b/pkg/storage/cache/filemetadata.go
@@ -39,5 +39,5 @@ func NewFileMetadataCache(store string, nodes []string, database, table string, 
 
 // RemoveMetadata removes a reference from the metadata cache
 func (c *fileMetadataCache) RemoveMetadata(path string) error {
-	return c.Delete(path)
+	return c.s.Delete(path)
 }
--- a/pkg/storage/cache/filemetadata.go
+++ b/pkg/storage/cache/filemetadata.go
@@ -39,5 +39,5 @@ func NewFileMetadataCache(store string, nodes []string, database, table string, 
 
 // RemoveMetadata removes a reference from the metadata cache
 func (c *fileMetadataCache) RemoveMetadata(path string) error {
-	return c.s.Delete(path)
+	return c.Delete(path)
 }
--- a/changelog/unreleased/trace-syscalls.md
+++ b/changelog/unreleased/trace-syscalls.md
@@ -1,0 +1,6 @@
+Enhancement: Trace decomposedfs syscalls
+
+To investigate performance characteristics of the underlying storage when the system is under load we wrapped the related syscalls in decomposedfs with trace spans.
+
+https://github.com/cs3org/reva/pull/3809
+
--- a/internal/grpc/services/gateway/gateway.go
+++ b/internal/grpc/services/gateway/gateway.go
@@ -191,10 +191,10 @@ func New(m map[string]interface{}, ss *grpc.Server) (rgrpc.Service, error) {
 		c:                        c,
 		dataGatewayURL:           *u,
 		tokenmgr:                 tokenManager,
-		statCache:                cache.GetStatCache(c.StatCacheStore, c.StatCacheNodes, c.StatCacheDatabase, "stat", time.Duration(c.StatCacheTTL)*time.Second, c.StatCacheSize),
-		providerCache:            cache.GetProviderCache(c.ProviderCacheStore, c.ProviderCacheNodes, c.ProviderCacheDatabase, "provider", time.Duration(c.ProviderCacheTTL)*time.Second, c.ProviderCacheSize),
-		createHomeCache:          cache.GetCreateHomeCache(c.CreateHomeCacheStore, c.CreateHomeCacheNodes, c.CreateHomeCacheDatabase, "createHome", time.Duration(c.CreateHomeCacheTTL)*time.Second, c.CreateHomeCacheSize),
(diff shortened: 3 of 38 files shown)


Excds (approved):
LGTM

micbar:
This needs a rebase.

butonic:
when updating ocis this needs:
```
diff --git a/ocis/pkg/command/decomposedfs.go b/ocis/pkg/command/decomposedfs.go
index 16a431e54d..9bec16e359 100644
--- a/ocis/pkg/command/decomposedfs.go
+++ b/ocis/pkg/command/decomposedfs.go
@@ -113,7 +113,7 @@ func check(c *cli.Context) error {
                })
 
        treeSize, err := walkTree(ctx, tree, lu, n, repairFlag)
-       treesizeFromMetadata, err := n.GetTreeSize()
+       treesizeFromMetadata, err := n.GetTreeSize(c.Context)
        if err != nil {
                fmt.Printf("failed to read treesize of node: %s: %s\n", n.ID, err)
        }
@@ -123,14 +123,14 @@ func check(c *cli.Context) error {
                if repairFlag {
                        fmt.Printf("Fixing tree size for node: %s. Calculated treesize: %d\n",
                                n.ID, treeSize)
-                       n.SetTreeSize(treeSize)
+                       n.SetTreeSize(c.Context, treeSize)
                }
        }
        return nil
 }
 
 func walkTree(ctx context.Context, tree *tree.Tree, lu *lookup.Lookup, root *node.Node, repair bool) (uint64, error) {
-       if root.Type() != provider.ResourceType_RESOURCE_TYPE_CONTAINER {
+       if root.Type(ctx) != provider.ResourceType_RESOURCE_TYPE_CONTAINER {
                return 0, errors.New("can't travers non-container nodes")
        }
        children, err := tree.ListFolder(ctx, root)
@@ -141,14 +141,14 @@ func walkTree(ctx context.Context, tree *tree.Tree, lu *lookup.Lookup, root *nod
 
        var treesize uint64
        for _, child := range children {
-               switch child.Type() {
+               switch child.Type(ctx) {
                case provider.ResourceType_RESOURCE_TYPE_CONTAINER:
                        subtreesize, err := walkTree(ctx, tree, lu, child, repair)
                        if err != nil {
                                fmt.Printf("error calculating tree size of node: %s: %s\n", child.ID, err)
                                return 0, err
                        }
-                       treesizeFromMetadata, err := child.GetTreeSize()
+                       treesizeFromMetadata, err := child.GetTreeSize(ctx)
                        if err != nil {
                                fmt.Printf("failed to read tree size of node: %s: %s\n", child.ID, err)
                                return 0, err
@@ -163,19 +163,19 @@ func walkTree(ctx context.Context, tree *tree.Tree, lu *lookup.Lookup, root *nod
                                if repair {
                                        fmt.Printf("Fixing tree size for node: %s. Calculated treesize: %d\n",
                                                child.ID, subtreesize)
-                                       child.SetTreeSize(subtreesize)
+                                       child.SetTreeSize(ctx, subtreesize)
                                }
                        }
                        treesize += subtreesize
                case provider.ResourceType_RESOURCE_TYPE_FILE:
-                       blobsize, err := child.GetBlobSize()
+                       blobsize, err := child.GetBlobSize(ctx)
                        if err != nil {
                                fmt.Printf("error reading blobsize of node: %s: %s\n", child.ID, err)
                                return 0, err
                        }
                        treesize += blobsize
                default:
-                       fmt.Printf("Ignoring type: %v, node: %s %s\n", child.Type(), child.Name, child.ID)
+                       fmt.Printf("Ignoring type: %v, node: %s %s\n", child.Type(ctx), child.Name, child.ID)
                }
        }
 
@@ -215,7 +215,7 @@ func dumpCmd(cfg *config.Config) *cli.Command {
                                return err
                        }
 
-                       attribs, err := backend.All(path)
+                       attribs, err := backend.All(c.Context, path)
                        if err != nil {
                                fmt.Println("Error reading attributes")
                                return err
@@ -244,7 +244,7 @@ func getCmd(cfg *config.Config) *cli.Command {
                                return err
                        }
 
-                       attribs, err := backend.All(path)
+                       attribs, err := backend.All(c.Context, path)
                        if err != nil {
                                fmt.Println("Error reading attributes")
                                return err
@@ -297,7 +297,7 @@ func setCmd(cfg *config.Config) *cli.Command {
                                }
                        }
 
-                       err = backend.Set(path, c.String("attribute"), []byte(v))
+                       err = backend.Set(c.Context, path, c.String("attribute"), []byte(v))
                        if err != nil {
                                fmt.Println("Error setting attribute")
                                return err
```

pkg/storage/cache/filemetadata.go
```diff
@@ -39,5 +39,5 @@ func NewFileMetadataCache(store string, nodes []string, database, table string,
 
 // RemoveMetadata removes a reference from the metadata cache
 func (c *fileMetadataCache) RemoveMetadata(path string) error {
-	return c.Delete(path)
+	return c.s.Delete(path)
```
aduffeck, line 42: Why do we need this change? `c.Delete` sets `microstore.DeleteFrom` options like the database and table which `c.s.Delete` doesn't. Maybe it's not necessary but then I wonder why `c.Delete` does it in the first place..

aduffeck (approved):
lgtm except for that cache change.

pkg/storage/cache/filemetadata.go
butonic: hm the implementations might forget to use the table and database that was configured when initializing the store. I'll drop this change as it is unrelated.

ainmosni:
LGTM

PR #3816, opened 2023-04-25
decomposedfs: write metadata once

butonic:
decomposedfs now aggregates metadata when creating directories and spaces into a single write.


pkg/storage/utils/decomposedfs/tree/tree_test.go
```diff
@@ -279,10 +281,6 @@ var _ = Describe("Tree", func() {
 					_, err := os.Stat(trashPath)
 					Expect(err).To(HaveOccurred())
 				})
-
-				It("does not try to delete a blob from the blobstore", func() {
-					env.Blobstore.AssertNotCalled(GinkgoT(), "Delete", mock.AnythingOfType("*node.Node"))
-				})
```
butonic, line 285: hm ... were we testing a bug? Or was there a reason for asserting NO delete?

PR #3823, opened 2023-04-27
Deny Access Correctly for Groups

kobergj:
Checks also for groups when denying access to resources

Fixes https://github.com/owncloud/ocis/issues/5483


micbar (approved):
LGTM

PR #3824, opened 2023-04-27
Serverless services

javfg:
This PR adds Serverless services to Reva. It is the first step before the big PR for the Notifications Service.

This adds a new type of service which doesn't have a listener (neither http or grp). As the Notifications service is communicating via NATS only with other backend services, it doesn't need to be exposed.

It could later on be used to implement other services which don't need http/grp endpoints, like metrics.

### DIFF
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -106,6 +106,9 @@ func run(mainConf map[string]interface{}, coreConf *coreConf, logger *zerolog.Lo
 		log.Panic(err)
 	}
 	listeners := initListeners(watcher, servers, logger)
+	if serverless != nil {
+		watcher.SL = serverless
+	}
 
 	start(mainConf, servers, serverless, listeners, logger, watcher)
 }
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -157,6 +157,7 @@ func initServerless(mainConf map[string]interface{}, log *zerolog.Logger) *rserv
 		serverless, err := getServerless(mainConf["serverless"], log)
 		if err != nil {
 			log.Error().Err(err).Msg("error")
+			os.Exit(1)
 		}
 		return serverless
 	}
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -33,6 +33,7 @@ import (
 	"github.com/cs3org/reva/pkg/registry/memory"
 	"github.com/cs3org/reva/pkg/rgrpc"
 	"github.com/cs3org/reva/pkg/rhttp"
+	"github.com/cs3org/reva/pkg/rserverless"
 	"github.com/cs3org/reva/pkg/sharedconf"
 	rtrace "github.com/cs3org/reva/pkg/trace"
 	"github.com/cs3org/reva/pkg/utils"
@@ -93,13 +94,20 @@ func run(mainConf map[string]interface{}, coreConf *coreConf, logger *zerolog.Lo
 	initCPUCount(coreConf, logger)
 
 	servers := initServers(mainConf, logger)
+	serverless := initServerless(mainConf, logger)
+
+	if len(servers) == 0 && serverless == nil {
+		logger.Info().Msg("nothing to do, no grpc/http/serverless enabled_services declared in config")
+		os.Exit(1)
+	}
+
 	watcher, err := initWatcher(logger, filename)
 	if err != nil {
 		log.Panic(err)
 	}
 	listeners := initListeners(watcher, servers, logger)
 
-	start(mainConf, servers, listeners, logger, watcher)
+	start(mainConf, servers, serverless, listeners, logger, watcher)
 }
 
 func initListeners(watcher *grace.Watcher, servers map[string]grace.Server, log *zerolog.Logger) map[string]net.Listener {
@@ -141,11 +149,19 @@ func initServers(mainConf map[string]interface{}, log *zerolog.Logger) map[strin
 		servers["grpc"] = s
 	}
 
-	if len(servers) == 0 {
-		log.Info().Msg("nothing to do, no grpc/http enabled_services declared in config")
-		os.Exit(1)
-	}
 	return servers
+}
+
+func initServerless(mainConf map[string]interface{}, log *zerolog.Logger) *rserverless.Serverless {
+	if isEnabledServerless(mainConf) {
+		serverless, err := getServerless(mainConf["serverless"], log)
+		if err != nil {
+			log.Error().Err(err).Msg("error")
+		}
+		return serverless
+	}
+
+	return nil
 }
 
 func initTracing(conf *coreConf) {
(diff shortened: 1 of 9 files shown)


labkode:
@javfg can you add an example of how the config would look like for an example service?

gmgigi96:
This is awesome, we could use the "serverless" services for running some internal reva procedures!

Regarding the code, I have a comment regarding the management of the services. In the HTTP and GRPC services, revad is handling properly the graceful shutdown when a signal is received (check https://github.com/cs3org/reva/blob/master/cmd/revad/internal/grace/grace.go#L258), so every service does not have to implement its own signal handler.

Also, how do you deal with errors in a single service? We don't have any way of propagating this to the Serverless struct. If we have a way then, do we need to restart the single service? Do we need to panic the entire reva?

javfg:
> This is awesome, we could use the "serverless" services for running some internal reva procedures!
> 
> Regarding the code, I have a comment regarding the management of the services. In the HTTP and GRPC services, revad is handling properly the graceful shutdown when a signal is received (check https://github.com/cs3org/reva/blob/master/cmd/revad/internal/grace/grace.go#L258), so every service does not have to implement its own signal handler.

I looked into this initially and thought about adding some way to deal with the shutdown for serverless in `grace.go` too, but in the end I did not because I'm not sure if there is a way to deal with them in a unified way.

Maybe we could discuss this during next week? I thought perhaps using context (which serverless services don't have right now), and maybe using `context.WithCancel` to start the services?

> Also, how do you deal with errors in a single service? We don't have any way of propagating this to the Serverless struct. If we have a way then, do we need to restart the single service? Do we need to panic the entire reva?

Not dealing with them at all (a broken service crashes the whole Reva instance), and indeed this is something that is missing too. I'm not sure about how we handle them in rhttp/rgrpc. Another thing to discuss, I could use some input.

javfg:
> @javfg can you add an example of how the config would look like for an example service?

Sure! Here is the config for the notifications service:

```toml
[serverless.services.notifications]
nats_address = ""
nats_token = "TOKEN"
nats_template_subject = "reva-notifications-template"
nats_notification_subject = "reva-notifications-notification"
nats_trigger_subject = "reva-notifications-trigger"
storage_driver = "sql"
grouping_interval = 5
grouping_maxsize = 5
```

labkode:
@javfg thanks!

@gmgigi96 has a good point. Even if they are server less, I expect that these services will create network connections to upstream services and maybe open local files. These cases should be handled gracefully if possible. 

Adding the following methods shouldn't be difficult. We can split he interface into two: Gracer and Listener
```
Stop() error
GracefulStop() error
```

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,103 @@
...
+// Start starts the serverless service collection.
+func (s *Serverless) Start() error {
+	return s.registerServices()
+}
+
+func (s *Serverless) registerServices() error {
+	for svcName := range s.conf.Services {
+		if s.isServiceEnabled(svcName) {
+			newFunc := Services[svcName]
+			svc, err := newFunc(s.conf.Services[svcName], &s.log)
```
gmgigi96, line 85: Here I would pass a logger containing the service name

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				return errors.Wrapf(err, "serverless service %s could not be initialized", svcName)
+			}
+
+			go func() {
+				svc.Start()
```
gmgigi96, line 91: this can be simplified with just `go svc.Start()`

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,103 @@
...
+	_, ok := Services[svcName]
+	return ok
+}
+
+// Start starts the serverless service collection.
+func (s *Serverless) Start() error {
+	return s.registerServices()
+}
+
+func (s *Serverless) registerServices() error {
```
gmgigi96, line 81: call this `registerAndStartServices`

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,103 @@
...
+func New(m interface{}, l zerolog.Logger) (*Serverless, error) {
+	conf := &config{}
+	if err := mapstructure.Decode(m, conf); err != nil {
+		return nil, err
+	}
+
+	n := &Serverless{
+		conf:     conf,
+		log:      l,
+		services: map[string]Service{},
```
gmgigi96, line 66: Why do you need this map?

cmd/revad/runtime/runtime.go
```diff
@@ -141,13 +149,21 @@ func initServers(mainConf map[string]interface{}, log *zerolog.Logger) map[strin
...
-		os.Exit(1)
-	}
 	return servers
 }
 
+func initServerless(mainConf map[string]interface{}, log *zerolog.Logger) *rserverless.Serverless {
+	if isEnabledServerless(mainConf) {
+		serverless, err := getServerless(mainConf["serverless"], log)
+		if err != nil {
+			log.Error().Err(err).Msg("error")
```
gmgigi96, line 159: you should abort here

gmgigi96 (changes requested):

pkg/rserverless/rserverless.go
javfg: Good idea! I'll also add this to `rhttp`. gRPC services each handle their own logger, it seems.

pkg/rserverless/rserverless.go
javfg: Indeed, this requires clarification.

This is there to follow the other types of services ([`rhttp`](https://github.com/cs3org/reva/blob/master/pkg/rhttp/rhttp.go#L69) and [`rgrpc`](https://github.com/cs3org/reva/blob/master/pkg/rgrpc/rgrpc.go#L122C5)). It is unused right now, but I'll make use of it for the graceful stop.

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,174 @@
...
+				}
+			} else {
+				s.log.Info().Msgf("service %s stopped", svcName)
+				stoppedServices[svcName] = true
+			}
+		}(svcName, svc)
+	}
+
+	count := 9 // one second less than the grace watcher deadlne
+	ticker := time.NewTicker(time.Second)
```
gmgigi96, line 120: I don't understand this for-loop, you can do the same using a `context.WithDeadline`, passed also to the `Close` method (see the previous comment)

pkg/rserverless/rserverless.go
```diff
@@ -0,0 +1,174 @@
...
+			}
+
+			go svc.Start()
+
+			s.services[svcName] = svc
+
+			s.log.Info().Msgf("serverless service enabled: %s", svcName)
+		} else {
+			message := fmt.Sprintf("serverless service %s does not exist", svcName)
+			return errors.New(message)
```
gmgigi96, line 169: Use `fmt.Errorf`

gmgigi96 (changes requested):

gmgigi96 (approved):

PR #3833, opened 2023-04-29
perf(mime): Prevent overriding mime type mappings in RegisterMime

iamrajiv:
## Description

This PR updates the **`RegisterMime`** function in the **`mime`** package to prevent overriding mime type mappings. The updated function checks if the given extension already exists in the **`mimes`** map before adding it. If the extension is already registered, the function returns an error. This ensures that existing mime type mappings are not accidentally overwritten.

To implement this change, I added an **`error`** return value to the **`RegisterMime`** function signature and modified the function body to use the **`LoadOrStore`** method of the **`sync.Map`** to atomically check for the existence of the extension in the map and add it if it does not already exist.

### DIFF
--- a/pkg/mime/mime.go
+++ b/pkg/mime/mime.go
@@ -19,6 +19,7 @@
 package mime
 
 import (
+	"fmt"
 	"path"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


pkg/mime/mime.go
```diff
@@ -36,9 +37,13 @@ func init() {
 
 // RegisterMime is a package level function that registers
 // a mime type with the given extension.
-// TODO(labkode): check that we do not override mime type mappings?
```
glpatcern, line 39: The original `TODO` suggests that it may or may not be a problem to override a type mapping. As such, a change here would be justified only by some evidence that there is a real problem to be fixed. This is not the case as far as I know.

Furthermore, any caller of this function would need to be amended to handle the error now returned.

glpatcern:
I agree with closing the PR. Letting alone that the code
```
// handle the error here
log.Printf("Failed to register mime type for extension '%s': %v\n", e, err)
// return the error, or continue processing
return err // or continue
```
looks to have been written by ChatGPT, the point I raised remains that this "fix" attempts to fix a non-problem.

Issue #3846, opened 2023-05-04
Update data transfers for current OCM implementation


redblom:
Some adjustments for handling data transfers need to be implemented in accordance with the current OCM implementation. 
This will also allow for some refactoring.

Includes https://github.com/cs3org/reva/issues/3808, https://github.com/cs3org/reva/issues/3677

PR #3847, opened 2023-05-04
Data transfers new ocm impl

redblom:
Concerns https://github.com/cs3org/reva/issues/3846
. remove unnecessary creation of ocm reference in ocmshareprovider when accepting an ocm share
. set transfer protocol when creating transfer type ocm share
. refactor transfer endpoints for new ocm impl
. refactor/cleanup gateway.ocmshareprovider.UpdateReceivedOCMShare() code
. refactor data transfers folder config
. new transfers config setting 'remove_on_cancel'
. implement transfer destination path
. update datatx example toml
. update cli ocm-share-update-received with path flag


glpatcern (approved):
Nice work! It would be even nicer to not have opaque fields, but I think we can go ahead with this.

redblom:
> Nice work! It would be even nicer to not have opaque fields, but I think we can go ahead with this.

Yes, let's save that for the next CS3APIs upgrade.

PR #3848, opened 2023-05-04
The DeleteStorageSpace method response code has been changed

2403905:
The DeleteStorageSpace method response status has been changed to 'not found' if no one spaces in a list or errtypes.IsNotFound. That allows us to handle the status properly.  Related to https://github.com/owncloud/ocis/issues/5031

### DIFF
--- a/changelog/unreleased/fix-propagation.md
+++ b/changelog/unreleased/fix-propagation.md
@@ -1,0 +1,5 @@
+Bugfix: Fix propagation
+
+Fix propagation in concurrency scenarios
+
+https://github.com/cs3org/reva/pull/3845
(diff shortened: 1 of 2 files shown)


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -599,13 +599,18 @@ func (s *service) DeleteStorageSpace(ctx context.Context, req *provider.DeleteSt
...
+		case errtypes.IsNotFound:
+			st = status.NewNotFound(ctx, "not found when deleting space")
+		case errtypes.PermissionDenied:
 			st = status.NewPermissionDenied(ctx, err, "permission denied")
 		case errtypes.BadRequest:
 			st = status.NewInvalid(ctx, err.Error())
 		default:
 			st = status.NewInternal(ctx, "error deleting space: "+req.Id.String())
 		}
+		if len(spaces) == 0 {
```
dragonchaser, line 611: can't we write `len(spaces) > 0` in line 599 and put this in an else branch, would be easier to read IMHO. (just nitpicking sorry :))

dragonchaser (changes requested):

dragonchaser (approved):

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -609,6 +611,10 @@ func (s *service) DeleteStorageSpace(ctx context.Context, req *provider.DeleteSt
 		return &provider.DeleteStorageSpaceResponse{
 			Status: st,
 		}, nil
+	} else if len(spaces) != 1 {
```
kobergj, line 614:
```suggestion
	} else if len(spaces) == 1 {
```
If we need to handle `len(spaces) > 1` we should a separate case for it.

kobergj (changes requested):

PR #3857, opened 2023-05-08
Fix nil-pointer exception on trashbin purge

dragonchaser:
We have fixed a nil-pointer-exception, when purging files from the trashbin that do not have a parent (any more)

fixes https://github.com/owncloud/ocis/issues/6245


micbar (approved):
LGTM

kobergj (approved):

PR #3866, opened 2023-05-09
Fix public link lookup performance

butonic:
Fix inefficient path based space lookup for public links.

public links use `/public/{token}` path based references. This PR checks the configured mountpoint of a space to skip storage providers that don't have a matching mountpoint, when looking up spaces for a path.


butonic:
cc @fschade this fixes public link access performance

aduffeck (approved):

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -241,6 +241,11 @@ func (s *svc) ListStorageSpaces(ctx context.Context, req *provider.ListStorageSp
 		// TODO check for allowed filters
 		filters["mask"] = mask
 	}
+	path := utils.ReadPlainFromOpaque(req.Opaque, "path")
+	if mask != "" {
```
butonic, line 245:
```suggestion
	if path != "" {
```

Issue #3855, opened 2023-05-08
Adapt to historical way of constructing WebDAV URL


michielbdejong:
We need to adapt the behaviour of Reva to https://github.com/cs3org/OCM-API/issues/70

gmgigi96:
In particular: we need to get the specification of how OC10 and NC currently build the URL AND use the `sharedSecret`, so to 1) implement it in Reva as a route and 2) document it in the OCM API specs in the description of the (deprecated) webdav sharing payload.

glpatcern:
A bit more context for why we need this - as I got lost myself :) and I'll need some confirmations here:

**Assumption**: the ScienceMesh app in OC10/NC would create OCM shares, which would be entirely channeled through Reva, such that the traffic goes like:

Site_A [ OC10 <--> Reva_A ] <-----> [ Reva_B <--> NC ] Site_B

Or in other words, the SM app in OC10/NC shall ensure that traffic is proxied through Reva, and then we can always assume that remote traffic is Reva-to-Reva - which provides the extra flexibility to drive the evolution of the ScienceMesh.

**Implementation**: for OC10/NC, the existing logic to browse a "remote" share is used as that's their native way. Therefore, the three arrows above seem to work as follows, from right to left, assuming that Marie@Site_B is accessing a share served by Einstein@Site_A:
1. NC@Site_B makes a "remote" access to the share (using the "historical way" we discuss here) to Reva_B (i.e. to the local Reva).
2. Reva_B forwards the request to Reva_A using the new way specified in OCM v1.1, i.e. using the whole webDAV path.
3. Reva_A serves the request by accessing OC10@A via the WebDAV storage driver.

Crucially, NC@Site_B shall not make a remote access to OC10@Site_A, bypassing Reva. Is all that right?

My doubt here is that an OCM share created via Reva through the SM app is actually persisted as an external share exactly like any other OCM share created by OC10/NC, and as such it seems it is "natively" or "externally" accessible by OC10/NC. How is it actually different? And how do we persist the OCM v1.1 properties such as the whole WebDAV path (and in the future the webapp path) that Reva_B needs to request to Reva_A in step 2?

michielbdejong:
No, an easier way to understand it is that only the OCM create call goes through the reva's. After that, all interaction (WebDAV but actually now that I think of it, als OCM notifications) are directly between the two EFSSs.

That part is working (almost).

[...]

PR #3868, opened 2023-05-09
Implement historical way of constructing OCM WebDAV URL 

gmgigi96:
Expose the expected WebDAV endpoint for OCM by OC10 and Nextcloud as described in https://github.com/cs3org/OCM-API/issues/70#issuecomment-1538551138 to allow Reva providers to participate to ScienceMesh.

In addition, expose the OCM discovery endpoint as `/ocm-provider` at the root level.

Closes https://github.com/cs3org/reva/issues/3855

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -503,7 +503,7 @@ func (s *svc) newPropRaw(key, val string) *propertyXML {
 }
 
 func supportLegacyOCMAccess(ctx context.Context, md *provider.ResourceInfo) {
-	ocm10 := ctx.Value(ctxOCM10).(bool)
+	ocm10, _ := ctx.Value(ctxOCM10).(bool)
 	if ocm10 {
 		// the path is something like /<token>/...
 		// we need to strip the token part as this


internal/http/services/owncloud/ocdav/ocdav.go
```diff
@@ -232,7 +234,45 @@ func (s *svc) Handler() http.Handler {
...
+		case "ocm-provider":
+			// Reverse proxy of /ocm/ocm-provider, assuming it is exposed in localhost. When not configured, returns 404.
+			// TODO(lopresti) this as well as other endpoints here should go to some common place for
+			// routes served from the root, including the reverseproxy service (which if configured
+			// together with ocdav would just override it!).
+			proto := "https"
+			if s.c.Insecure {
+				proto = "http"
+			}
+			ocmEP := fmt.Sprintf("%s://localhost/ocm/ocm-provider", proto)
```
gmgigi96, line 246: I think this will use the default ports, but this is not always true. For example at CERN this is exposed in the port 9143...

glpatcern: yes, too many assumptions with this approach - we go for a separate microservice

gmgigi96:
@glpatcern can you update the ocmd example adding the new service please?

examples/nextcloud-integration/revad.toml
```diff
@@ -126,9 +126,11 @@ driver = "nextcloud"
 user_layout = "{{.Username}}"
 
 [http.services.ocmd]
-prefix = "ocm"
+
+[http.services.ocmprovider]
 provider = "Reva-Nextcloud"
 endpoint = "http://localhost"
+webdav_root = "/public.php/webdav"
```
gmgigi96, line 133: This should be `/remote.php/dav/ocm` to enable reva to serve ocm resources using the old method

glpatcern: Right! and `/ocm` is actually appended as it's the prefix anyway

internal/http/services/ocmprovider/ocmprovider.go
```diff
@@ -0,0 +1,177 @@
...
+	return nil
+}
+
+func (s *svc) Prefix() string {
+	// this is hardcoded as per OCM specifications
+	return "/ocm-provider"
+}
+
+func (s *svc) Unprotected() []string {
+	return []string{"."}
```
gmgigi96, line 153: This should be `/`

glpatcern (changes requested):
OK, I tested my part and it's all fine, and I understand that Michiel successfully used the webdav access part, so we are good to go.

The cosmetic changes I propose are just to acknowledge that these are not temporary "fixes" and are there to stay potentially for a long time - at least until there's a OC10 site in ScienceMesh, and NC implements OCM v1.1.

glpatcern (dismissed):

glpatcern (approved):

PR #3873, opened 2023-05-10
Fix parsing of grappa response

gmgigi96:


glpatcern (approved):
A subtle but important fix!

PR #3875, opened 2023-05-10
use relative references when accessing public resources

butonic:


kobergj (approved):
LGTM but wants a changelog

PR #3877, opened 2023-05-10
fix the empty exact list while searching for a sharee #4265

2403905:
Fixed a bug in the sharing api, it always returns an empty exact list while searching for a sharee

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
@@ -20,6 +20,7 @@ package sharees
 
 import (
 	"net/http"
+	"strings"
 
 	grouppb "github.com/cs3org/go-cs3apis/cs3/identity/group/v1beta1"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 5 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
```diff
@@ -132,3 +143,11 @@ func (h *Handler) groupAsMatch(g *grouppb.Group) *conversions.MatchData {
 func (h *Handler) getAdditionalInfoAttribute(u *userpb.User) string {
 	return templates.WithUser(u, h.additionalInfoAttribute)
 }
+
+func (h *Handler) isExactUser(match *conversions.MatchData, term string) bool {
```
kobergj, line 147:
```suggestion
func (h *Handler) isExactMatch(match *conversions.MatchData, term string) bool {
```
Not user specific, so naming is a bit off. But not critical

kobergj (approved):

internal/http/services/owncloud/ocs/handlers/apps/sharing/sharees/sharees.go
2403905: Already changed.

PR #3880, opened 2023-05-11
decomposedfs no longer deadlocks when cache is disabled

butonic:
We fixed a bug in the decomposedfs that causes the propagation to deadlock if caching is disabled.


aduffeck (approved):
lgtm. Needs a rebase when the ci has been fixed.

PR #3885, opened 2023-05-12
use context to pass reader for locked files

butonic:
We now pass a context to lower level functions which allows passing a reader for already locked files.

alternative approach to https://github.com/cs3org/reva/pull/3880


butonic:
fixed with https://github.com/cs3org/reva/pull/3886

PR #3887, opened 2023-05-12
trace uploads through datagateway

butonic:
used to produce the traces for https://github.com/owncloud/ocis/issues/6296


dragonchaser (approved):
changelog missing otherwise, lgtm

Issue #2104, opened 2021-09-24
Add resource size info to transfer type OCM share


redblom:
At WP4.4 it is discussed that it would be useful for a receiver of a transfer type OCM share to know the size of the transfer before accepting it. Although the size cannot be guaranteed, a sender could still add more files to a transfer folder which would also be transferred after acceptance in case of a transfer retry, the receiving end could limit the transfer to the expected size (data cap). Or, from the end user viewpoint, have an idea of what to expect.
For the receiver it is not possible to stat the resource to be transferred before accepting it because the webdav reference to it is only created after accepting the share. Thus, the size info must be send by the creator of the resource as a specific size parameter.

The initiator of the datatx type OCM share adds the current size to the opaque field of the request:
https://github.com/cs3org/reva/blob/44c9fa189af323a877168a63d77a078caaabda20/cmd/reva/transfer-create.go#L112
```
		createShareReq := &ocm.CreateOCMShareRequest{
			Opaque: &types.Opaque{
				Map: map[string]*types.OpaqueEntry{
					"permissions": {
						Decoder: "plain",
						Value:   []byte(strconv.Itoa(pint)),
					},
					"name": {
						Decoder: "plain",
						Value:   []byte(statRes.Info.Path),
					},
					"protocol": {
						Decoder: "plain",
						Value:   []byte("datatx"),
					},
					"size": {
						Decoder: "plain",
						Value:   []byte("60000000"),
					},
				},
			},
```
Next we add the size as an option of the protocol section of the OCM share request:
https://github.com/cs3org/reva/blob/44c9fa189af323a877168a63d77a078caaabda20/pkg/ocm/share/manager/json/json.go#L257
```
		protocol := map[string]interface{}{
			"name": "webdav",
			"options": map[string]string{
				"permissions": pm,
				"token":       ctxpkg.ContextMustGetToken(ctx),
				"size": size,
			},
		}
```

And the datatx driver will be called specifying the size in a new options parameter:
https://github.com/cs3org/reva/blob/44c9fa189af323a877168a63d77a078caaabda20/pkg/datatx/datatx.go#L30
```
type Manager interface {
	// StartTransfer initiates a transfer job and returns a TxInfo object including a unique transfer id, and error if any.
	StartTransfer(
		ctx context.Context, 
		srcRemote string, 
		srcPath string, 
		srcToken string, 
		destRemote string, 
		destPath string, 
		destToken string, 
		options map[string]interface{}) (*datatx.TxInfo, error)
[...]

PR #3891, opened 2023-05-16
Provide total transfer size with the datatx protocol

redblom:
Sets the `size` property of the OCM share's datatx protocol (https://github.com/cs3org/reva/issues/2104).

### DIFF
--- a/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
+++ b/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
@@ -44,6 +44,7 @@ import (
 	"github.com/cs3org/reva/pkg/rgrpc/status"
 	"github.com/cs3org/reva/pkg/rgrpc/todo/pool"
 	"github.com/cs3org/reva/pkg/sharedconf"
+	"github.com/cs3org/reva/pkg/storage/utils/walker"
 	"github.com/cs3org/reva/pkg/utils"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
@@ -71,6 +72,7 @@ type service struct {
 	client     *client.OCMClient
 	gateway    gateway.GatewayAPIClient
 	webappTmpl *template.Template
+	walker     walker.Walker
 }
 
 func (c *config) init() {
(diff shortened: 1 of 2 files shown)


internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -210,13 +214,38 @@ func (s *service) getWebappProtocol(share *ocm.Share) *ocmd.Webapp {
...
-	// TODO discover the size
+	var size uint64
+	// get the path of the share
+	statRes, err := s.gateway.Stat(ctx, &providerpb.StatRequest{
+		Ref: &providerpb.Reference{
+			ResourceId: share.ResourceId,
+		},
+	})
+	if err != nil {
+		panic(err)
```
glpatcern, line 225: Shouldn't we just fail the request here rather than killing the server?

glpatcern (changes requested):
Looks good! Only one comment

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
redblom: `Panic` doesn't seem to kill the server but the current execute flow indeed.
In this case I was kind of following `getWebappProtocol`, and it sure is bad if this particular request would fail. And it also meant I didn't need to add error handling further up.
But (maybe like you ... ?) I'm actually not too fond of this `panic` reaction which is used at more places. Is there a protocol for when we should use this?

glpatcern: I appreciate we agree on the principle ;-) let me check what we do elsewhere, I don't think there's a strict "protocol" for using `panic()` vs logging the error - I'd go for the latter indeed (and we know that logging could be vastly improved in several corners of the code base).

gmgigi96: If at that point of the code you expect to find the transfer protocol, for example because you have already checked before, is fine

gmgigi96 (approved):

gmgigi96 (approved):

glpatcern (approved):
OK, it appears that this use of `panic` to just assert something that was already validated is relatively common in Go-lang (though not so widespread in the Reva code base).

PR #3895, opened 2023-05-16
Add missing expiry in shares

dragonchaser:


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -1411,6 +1411,22 @@ func (h *Handler) createCs3Share(ctx context.Context, w http.ResponseWriter, r *
...
+			return nil, &ocsError{
+				Code:    response.MetaBadRequest.StatusCode,
+				Message: "could not parse expiry timestamp on this item",
+				Error:   err,
+			}
+		}
+		req.Grant.Expiration = &types.Timestamp{
+			Seconds: uint64(ts.Unix()),
+			Nanos:   0,
+		}
```
kobergj, line 1427: There is `utils.TimeToTS` which could do this for you. But not critical.

kobergj (approved):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
dragonchaser: Thx, did not know that, incorporated it.

PR #3896, opened 2023-05-16
Dont lose revisions when replacing

aduffeck:
We no longer delete-and-upload targets of copy operations but ratheradd a new version with the source content.

This makes "overwrite when copying" behave the same as "overwrite when uploading".

Overwriting when moving a file still deletes the old file (moves it into the trash) and replaces the whole file including the revisions of the source file.

Fixes https://github.com/owncloud/ocis/issues/4797

### DIFF
--- a/changelog/unreleased/fix-preflight-requests.md
+++ b/changelog/unreleased/fix-preflight-requests.md
@@ -2,4 +2,4 @@ Bugfix: fix preflight requests
 
 The datagateway now correctly overwrites the preconfigured CORS related headers with the headers returned by a dataprovider.
 
-https://github.com/cs3org/reva/pull/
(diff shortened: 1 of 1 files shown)


micbar (approved):

micbar (changes requested):
The tests indicate that we change some expected behavior

aduffeck:
@micbar The failing tests expect a file to be turned into a folder when a fold is copied onto (into? over?) them. Is that really the expected behavior here or should we rather adapt the tests?
Doing this operation isn't possible using the web UI and tools like `cp` also bail out in cases like that.

micbar:
@aduffeck @individual-it Ok we need to change the test expectations.

micbar (approved):

PR #3898, opened 2023-05-17
Add Graph User Capabilities

kobergj:
Adds capabilities if users are writeable in LDAP


rhafer (approved):

internal/http/services/owncloud/ocs/data/capabilities.go
```diff
@@ -80,8 +80,16 @@ type CapabilitiesCore struct {
 
 // CapabilitiesGraph holds the graph capabilities
 type CapabilitiesGraph struct {
-	PersonalDataExport     ocsBool  `json:"personal-data-export" xml:"personal-data-export" mapstructure:"personal_data_export"`
-	ReadOnlyUserAttributes []string `json:"read_only_user_attributes" xml:"read_only_user_attributes" mapstructure:"read_only_user_attributes"`
+	PersonalDataExport ocsBool               `json:"personal-data-export" xml:"personal-data-export" mapstructure:"personal_data_export"`
+	User               CapabilitiesGraphUser `json:"user" xml:"user" mapstructure:"user"`
```
kulmann, line 84: What do you think about matching the key with the graph endpoint? I.e. `Users` instead of `User`?

kulmann: I'm indifferent, both make sense. Just came to my mind. Maybe you have an opinion.

kobergj: `users` is correct. It is also what was written in the ticket. My mistake. Thanks for finding!

PR #3899, opened 2023-05-17
harden uploads

butonic:
Uploads now check response headers for a file id and omit a subsequent stat request which might land on a storage provider that does not yet see the new file due to latency, eg. when NFS caches direntries.


micbar (approved):
Awesome fix!

PR #3903, opened 2023-05-19
Clean IDCache properly

kobergj:
~Needs more work. Cache can't work like this on distributed systems~

Decided to use a (configurable) go-micro store instead of the hardcoded inmemory implementation

### DIFF
--- a/pkg/storage/utils/decomposedfs/tree/tree.go
+++ b/pkg/storage/utils/decomposedfs/tree/tree.go
@@ -234,6 +234,9 @@ func (t *Tree) Move(ctx context.Context, oldNode *node.Node, newNode *node.Node)
 		}
 	}
 
+	// remove cache entry in any case to avoid inconsistencies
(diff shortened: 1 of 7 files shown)


pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -416,6 +416,10 @@ func (t *Tree) ListFolder(ctx context.Context, n *node.Node) ([]*node.Node, erro
 
 // Delete deletes a node in the tree by moving it to the trash
 func (t *Tree) Delete(ctx context.Context, n *node.Node) (err error) {
+	path := filepath.Join(n.ParentPath(), n.Name)
+	// remove entry from cache immediately to avoid inconsistencies
+	_ = t.idCache.Delete(path)
```
aduffeck, line 421: Little nitpick: Wouldn't it be safer to invalidate the cache after actually having deleted the file? Otherwise the cache could be filled again by another goroutine or instance between here and line 499, no?

butonic (approved):

PR #3905, opened 2023-05-19
Remove config for invite_link

glpatcern:
This mini-PR drops invite_link from the OCM-related config.

~~The link template is still defined in a constant but not exposed in the config, as it depends on Mentix and should not be changed.~~

Edit: the link is not any longer a template and it is instead constructed out of the needed parameters.
Also, the `provider_domain` and `mesh_directory_url` configuration options are now mandatory, and the `sciencemesh` service won't start without it.

### DIFF
--- a/internal/http/services/sciencemesh/token.go
+++ b/internal/http/services/sciencemesh/token.go
@@ -24,7 +24,6 @@ import (
 	"html/template"
 	"mime"
 	"net/http"
-	"strings"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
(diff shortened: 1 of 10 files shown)


internal/http/services/sciencemesh/token.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/cs3org/reva/pkg/smtpclient"
 )
 
-const defaultInviteLink = "{{.MeshDirectoryURL}}?token={{.Token}}&providerDomain={{.User.Id.Idp}}"
+const defaultInviteLinkPrefix = "{{.MeshDirectoryURL}}?token={{.Token}}&providerDomain="
```
gmgigi96, line 41: Actually we can remove this template and simplify how we generate this link, to just a string concatenations.

gmgigi96 (changes requested):

internal/http/services/sciencemesh/token.go
glpatcern: you mean end up with something like `meshURL + "?token=" + token + "&providerDomain=" + provider` ?

gmgigi96: Exactly

gmgigi96: this can be done here https://github.com/cs3org/reva/blob/580a83f0ae0b19e12e25bed057d9b977733a9c25/internal/http/services/sciencemesh/token.go#L137

gmgigi96 (approved):

PR #3911, opened 2023-05-22
Use defer to invalidate ID Cache

kobergj:
Follows suggestion from here: https://github.com/cs3org/reva/pull/3903


aduffeck (approved):
:+1:

PR #3917, opened 2023-05-24
Prevent last Space manager from leaving

micbar:


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/spaces.go
```diff
@@ -125,6 +125,31 @@ func (h *Handler) addSpaceMember(w http.ResponseWriter, r *http.Request, info *p
...
+		return
+	}
+
+	lgRes, err := providerClient.ListGrants(ctx, &provider.ListGrantsRequest{Ref: &ref})
+	if err != nil || lgRes.Status.Code != rpc.Code_CODE_OK {
+		response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "error listing space grants", err)
+		return
+	}
+
+	if len(lgRes.Grants) == 1 || !isSpaceManagerRemaining(lgRes.Grants, grantee) {
```
kobergj, line 148: This means I cannot add a member to my space when there is only one member. 
```suggestion
	if !isSpaceManagerRemaining(lgRes.Grants, grantee) {
```
`isSpaceManagerRemaining` assumes that the grantee does not have manager permissions. This should be fine in most cases but we will do unnecessary requests. Maybe we skip all checks when the grantee gets Manager permissions?

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/spaces.go
```diff
@@ -125,6 +125,31 @@ func (h *Handler) addSpaceMember(w http.ResponseWriter, r *http.Request, info *p
...
+		response.WriteOCSError(w, r, http.StatusForbidden, "can't remove the last manager", nil)
```
kobergj, line 149: Log message is not correct. The manager is not removed

kobergj (changes requested):
Needs tuning

kobergj (approved):

PR #3926, opened 2023-06-01
Service Accounts

kobergj:
Introduces service-account auth manager

ATTENTION: BREAKS OCIS TEMPORARILY 

check https://github.com/owncloud/ocis/pull/6427 before merging

### DIFF
--- a/pkg/auth/manager/serviceaccounts/serviceaccounts.go
+++ b/pkg/auth/manager/serviceaccounts/serviceaccounts.go
@@ -39,7 +39,6 @@ func (m *manager) Configure(config map[string]interface{}) error {
 	// only inmem authenticator for now
 	a := &inmemAuthenticator{make(map[string]string)}
 	for _, s := range c.ServiceUsers {
-		// TODO: hash secrets
 		a.m[s.ID] = s.Secret
 	}
 	m.authenticate = a.Authenticate
@@ -71,6 +70,7 @@ func (m *manager) Authenticate(ctx context.Context, userID string, secret string
 		Id: &userpb.UserId{
 			OpaqueId: userID,
 			Type:     userpb.UserType_USER_TYPE_SERVICE,
+			Idp:      "none",
 		},
 	}, scope, nil
 }
@@ -80,7 +80,9 @@ type inmemAuthenticator struct {
 }
 
 func (a *inmemAuthenticator) Authenticate(userID string, secret string) error {
-	// TODO: hash secrets
+	if secret == "" || a.m[userID] == "" {
+		return errors.New("unknown user")
+	}
 	if a.m[userID] == secret {
 		return nil
 	}
--- a/pkg/auth/manager/serviceaccounts/serviceaccounts.go
+++ b/pkg/auth/manager/serviceaccounts/serviceaccounts.go
@@ -68,7 +68,10 @@ func (m *manager) Authenticate(ctx context.Context, userID string, secret string
 	}
 	return &userpb.User{
 		// TODO: more details for service users?
-		Id: &userpb.UserId{OpaqueId: userID},
+		Id: &userpb.UserId{
+			OpaqueId: userID,
+			Type:     userpb.UserType_USER_TYPE_SERVICE,
+		},
 	}, scope, nil
 }
 
--- a/pkg/auth/manager/serviceaccounts/serviceaccounts.go
+++ b/pkg/auth/manager/serviceaccounts/serviceaccounts.go
@@ -1,0 +1,85 @@
+package serviceaccounts
+
+import (
+	"context"
+
+	authpb "github.com/cs3org/go-cs3apis/cs3/auth/provider/v1beta1"
+	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
+
+	"github.com/cs3org/reva/v2/pkg/auth"
+	"github.com/cs3org/reva/v2/pkg/auth/manager/registry"
+	"github.com/cs3org/reva/v2/pkg/auth/scope"
+	"github.com/mitchellh/mapstructure"
+	"github.com/pkg/errors"
+)
+
+type conf struct {
+	ServiceUsers []serviceuser `mapstructure:"service_accounts"`
+}
+
+type serviceuser struct {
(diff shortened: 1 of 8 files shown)


gmgigi96:
Interesting! Which is the context of this?

kobergj:
@gmgigi96 sorry, I completely missed your comment. Shame on me :disappointed: 

We want to have so called `service-accounts`. We want to use these for inter service communication, instead of impersonating users like we do currently.

Reason for this is that we do not have an active user in some parts of the code, but need to get spaces/filemetadata/users/groups/....

pkg/utils/grpc.go
```diff
@@ -35,13 +24,13 @@ func GetUser(userID *user.UserId, gwc gateway.GatewayAPIClient, machineAuthAPIKe
 	return getUserResponse.GetUser(), nil
 }
 
-// ImpersonateUser impersonates the given user
-func ImpersonateUser(usr *user.User, gwc gateway.GatewayAPIClient, machineAuthAPIKey string) (context.Context, error) {
-	ctx := revactx.ContextSetUser(context.Background(), usr)
+// ImpersonateServiceUser impersonates the given user
```
rhafer, line 27: Hm, this doc comment is rather misleading, isn't it?

pkg/utils/grpc.go
```diff
@@ -35,13 +24,13 @@ func GetUser(userID *user.UserId, gwc gateway.GatewayAPIClient, machineAuthAPIKe
...
+func ImpersonateServiceUser(serviceUserID string, gwc gateway.GatewayAPIClient, serviceUserSecret string) (context.Context, error) {
```
rhafer, line 28: I wonder if `ImpersonateServiceUser` is really a good name. Isn't it more a `GetServiceUserContext()` function now?

pkg/auth/manager/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,88 @@
...
+// Configure parses the map conf
+func (m *manager) Configure(config map[string]interface{}) error {
+	c := &conf{}
+	if err := mapstructure.Decode(config, c); err != nil {
+		return errors.Wrap(err, "error decoding conf")
+	}
+	// only inmem authenticator for now
+	a := &inmemAuthenticator{make(map[string]string)}
+	for _, s := range c.ServiceUsers {
+		// TODO: hash secrets
```
rhafer, line 42: If we want to hash the password. The password hashes should already be supplied via config.  Hashing it here wouldn't have much benefit.

pkg/auth/manager/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	scope, err := scope.AddOwnerScope(nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	return &userpb.User{
+		// TODO: more details for service users?
+		Id: &userpb.UserId{
+			OpaqueId: userID,
+			Type:     userpb.UserType_USER_TYPE_SERVICE,
```
rhafer, line 73: I am not entirely sure, but we might want to add some value for `Idp`. The cs3 API has this as a required attribute.

rhafer (changes requested):
Overall this looks fine to me. Just a new nits and questions.

pkg/utils/grpc.go
kobergj: Thats what it is, yes :+1:

pkg/auth/manager/serviceaccounts/serviceaccounts.go
kobergj: I was thinking about memory leaks or something like this. Prolly not needed here. Removed the comment

pkg/auth/manager/serviceaccounts/serviceaccounts.go
kobergj: Added a `"none"` value for the idp because it has none

pkg/auth/manager/serviceaccounts/serviceaccounts.go
rhafer: Did you? :thinking:

kobergj: Yes? https://github.com/cs3org/reva/pull/3926/commits/662ac1dd0061068ce8228211b9ecea7676b55d6f#diff-a94eec5ee3afc75706a275742376a4b8d296b2c9624afeaa72ef5d81f4205111L83

kobergj: There was another one. Sorry. Now It's gone :)

rhafer: Ah, didn't notice the other one ;-)

rhafer (approved):
IMO it's good now for a start.

kobergj:
Temporarilly converted to draft so it doesn't get merged before https://github.com/cs3org/reva/pull/4133 (that one is urgent for the web team)

pkg/auth/manager/serviceaccounts/serviceaccounts.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	}, scope, nil
+}
+
+type inmemAuthenticator struct {
+	m map[string]string
+}
+
+func (a *inmemAuthenticator) Authenticate(userID string, secret string) error {
+	if a.m[userID] == secret {
```
rhafer, line 83: Ugh, just noticed that this will happily validate an empty user and password. So it's possible to get a token using

```
authRes, err := gatewayClient.Authenticate(ctx, &gateway.AuthenticateRequest{
        Type:         "serviceaccounts",
        ClientId:    "",
        ClientSecret: "",
})

```

rhafer (changes requested):

pkg/auth/manager/serviceaccounts/serviceaccounts.go
kobergj: Thanks. Missed that. Fixed now :+1:

rhafer (approved):

PR #3934, opened 2023-06-05
fix wrong space editor permission

micbar:
# Description

A space editor should not have the permission to purge the trash bin.

### DIFF
--- a/changelog/unreleased/fix-space-editor-permissions.md
+++ b/changelog/unreleased/fix-space-editor-permissions.md
@@ -1,0 +1,5 @@
+Bugfix: Fix the space editor permissions
+
+We fixed the permissions of a space editor which accidentally granted the permission to purge the trash bin.
+
+https://github.com/cs3org/reva/pull/3934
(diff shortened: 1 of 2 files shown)


phil-davis (approved):

phil-davis:
Various tests now fail, for example:
https://drone.owncloud.com/cs3org/reva/1697/12/6
```
Failed step: And as user "Brian" folder "/Shares/tmp" should contain a property "ocs:share-permissions" with value "15"
        Property "ocs:share-permissions" found with value "7", expected "#^15$#" or "#^15$#"
```

If this is the new requirement, then tests need adjusting. The "source" test scenarios are in ocis, so maybe the known-failing tests get added to expected-failures here with and issue raised in ocis. Then we adjust the test scenarios in ocis as we pull this code into ocis. Then we pull the updated ocis tests back into reva (and the expected-failures can be deleted)

micbar:
@phil-davis There must be something wrong. That should not touch sharing at all.

kobergj (approved):
:+1: Good enough for me. Note that It won't affect existing space editors

micbar:
Let's see, there were some code changes missing. Now it should work and pass the testsuite.

kobergj (approved):

phil-davis (approved):

PR #3936, opened 2023-06-05
[tests-only] Updated reference OCM config

glpatcern:


gmgigi96:
Although it is not required for OCM to work properly and this is for now only used from the CERNBox web UI to enable the apps, LGTM!

glpatcern:
Superseded by #3962

PR #3948, opened 2023-06-07
handle the bad request status

2403905:
Handle the bad request status for the CreateStorageSpace function


phil-davis:
This should fix https://github.com/owncloud/ocis/issues/6414 
Needs merging, and then pulling latest reva into ocis...

PR #3956, opened 2023-06-08
Adding missing spanKind to tracing.

Excds:
This adds missing spanKind to tracing. I _think_ I found all of them, but please point out if I didn't.


internal/grpc/interceptors/auth/auth.go
```diff
@@ -115,7 +115,10 @@ func NewUnary(m map[string]interface{}, unprotected []string, tp trace.TracerPro
 		span := trace.SpanFromContext(ctx)
 		defer span.End()
 		if !span.SpanContext().HasTraceID() {
-			ctx, span = tp.Tracer(tracerName).Start(ctx, "grpc auth unary")
+			spanOpts := []trace.SpanStartOption{
```
butonic, line 118: I think we should not add a new tracer for every middleware... 🤔

glpatcern:
Obsoleted

PR #3963, opened 2023-06-12
Bugfix: treesize interger overflows

rhafer:
Reading the treesize was parsing the string value as a signed integer while setting the treesize used unsigned integers this could cause failures (out of range errors) when reading very large treesizes.

This PR is mainly a workaround for scenarios where the treesize propagation would result in negative treesize, which then are stored as unsigned integers in the xattrs. 

https://github.com/owncloud/enterprise/issues/5783

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -948,11 +948,11 @@ func (n *Node) IsDisabled() bool {
 
 // GetTreeSize reads the treesize from the extended attributes
 func (n *Node) GetTreeSize() (treesize uint64, err error) {
-	s, err := n.XattrInt64(prefixes.TreesizeAttr)
+	s, err := n.XattrUint64(prefixes.TreesizeAttr)
 	if err != nil {
 		return 0, err
 	}
-	return uint64(s), nil
+	return s, nil
 }
 
 // SetTreeSize writes the treesize to the extended attributes
--- a/pkg/storage/utils/decomposedfs/tree/tree.go
+++ b/pkg/storage/utils/decomposedfs/tree/tree.go
@@ -794,12 +794,17 @@ func (t *Tree) Propagate(ctx context.Context, n *node.Node, sizeDiff int64) (err
 				}
 			case err != nil:
 				return err
+			case sizeDiff > 0:
+				newSize = treeSize + uint64(sizeDiff)
+			case uint64(-sizeDiff) > treeSize:
+				// The sizeDiff is larger than the current treesize. Which would result in
+				// a negative new treesize. Something must have gone wrong with the accounting.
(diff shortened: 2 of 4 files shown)


pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -794,12 +794,17 @@ func (t *Tree) Propagate(ctx context.Context, n *node.Node, sizeDiff int64) (err
...
+					Msg("Error when updating treesize of parent node. Updated treesize < 0. Reestting to 0")
+				newSize = 0
 			default:
-				if sizeDiff > 0 {
-					newSize = treeSize + uint64(sizeDiff)
-				} else {
-					newSize = treeSize - uint64(-sizeDiff)
-				}
+				newSize = treeSize - uint64(-sizeDiff)
 			}
```
dragotin, line 808: I think this case statement makes the code harder to understand. I'd rather prefer to have it in the old if - else with an additional condition in the `sizeDiff < 0` branch.

kobergj (approved):
Looks good to me :+1:

pkg/storage/utils/decomposedfs/tree/tree.go
kobergj: I dislike nested if/elses. I think new solution is cleaner in that regard. (just my opinion - not critical)

pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -948,7 +948,7 @@ func (n *Node) IsDisabled() bool {
 
 // GetTreeSize reads the treesize from the extended attributes
 func (n *Node) GetTreeSize() (treesize uint64, err error) {
-	s, err := n.XattrInt64(prefixes.TreesizeAttr)
+	s, err := n.XattrUint64(prefixes.TreesizeAttr)
```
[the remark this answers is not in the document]
rhafer: Yes. Fixed.

pkg/storage/utils/decomposedfs/tree/tree.go
rhafer: Yeah, I also found the nested if inside the switch harder to read. Especially as I would need to add an additional if branch (or turn the nested if into a nested switch ...)

It might be more readable if I turn this into two separate switch commands though. Let me try ...

rhafer: > I'd rather prefer to have it in the old if - else with an additional condition in the `sizeDiff < 0` branch.

@dragotin Played around with this a bit. And adding another branch to the if statement will actually make the linter complain about the ifElseChain that should be turned into a switch statement ;-)

I guess I'll just leave it as is for now. (Splitting up the switch also doesn't help much with readability I think, as it will require an outer if condition. :man_shrugging:

PR #3966, opened 2023-06-12
add space metadata to ocs shares

micbar:
# Description

Enhancement: Add space metadata to ocs shares list

We needed to add the space ID and the space alias of the orgininal space to the share metadata when we are listing shares. This is needed to directly navigate to the orgiginating space location.


kobergj (changes requested):
found typos and have a question

PR #3970, opened 2023-06-14
fix enforce-password issue for public links

2403905:
The related issue [https://github.com/owncloud/ocis/issues/6476](https://github.com/owncloud/ocis/issues/6476) [https://github.com/owncloud/ocis/issues/6401](https://github.com/owncloud/ocis/issues/6401)

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
@@ -23,12 +23,14 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	permissionsv1beta1 "github.com/cs3org/go-cs3apis/cs3/permissions/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
+	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
 	"github.com/rs/zerolog/log"
 
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocs/conversions"
@@ -36,7 +38,6 @@ import (
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/publicshare"
-	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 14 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -548,6 +558,40 @@ func (h *Handler) removePublicShare(w http.ResponseWriter, r *http.Request, shar
...
+	if pk == nil {
+		return enforce
+	}
+	p, err := conversions.NewPermissions(decreasePermissionsIfNecessary(*pk))
+	if err != nil {
+		return enforce
+	}
+	if h.publicPasswordEnforced.EnforcedForReadOnly &&
+		p == conversions.PermissionRead {
+		enforce = true
```
kobergj, line 577: Looks correct now. But I have one little nitpick:
We could `return true` here because the `enforce` value will never be changed back to `false` again. If we do this for all `if` cases we can completely get rid of the `enforce` var

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/shares.go
```diff
@@ -103,6 +104,14 @@ type ocsError struct {
 	Message string
 }
 
+type passwordEnforced struct {
+	Enforced                   bool
```
kobergj, line 108: Is `Enforced` actually used? If not I'm fine with just deleting it.

kobergj (changes requested):
Just some small nitpicks as always. Rest looks :+1:

.drone.env
```diff
@@ -1,4 +1,4 @@
 # The test runner source for API tests
-APITESTS_COMMITID=fc686d8cd827fe7b2bea55894c781fb72ad24865
+APITESTS_COMMITID=3cdc3bc48599af7f598f58e8dd87e910a2fca3b6
```
phil-davis, line 2: @2403905 this change can be removed - it is reporting a git conflict. 
The commit id was updated late last week in PR #3984

tests/oc-integration-tests/drone/frontend-global.toml
```diff
@@ -72,12 +72,13 @@ multiple = true
 supports_upload_only = true
 
 [http.services.ocs.capabilities.capabilities.files_sharing.public.password]
-enforced = true
+enforced = false
```
phil-davis, line 75: IMO this is fine. In reva we can run the CI with public.password enforcement off.
In ocis CI, in the future, we will be able to run with various "system parameter" settings, and can have test scenarios that check what happens when various public.password items are enforced.

kobergj (approved):

tests/oc-integration-tests/drone/frontend-global.toml
2403905: As I understand we can't use the env adjustments in the reva test. We have to cover the OCIS_SHARING_PUBLIC_WRITEABLE_SHARE_MUST_HAVE_PASSWORD configuration the ocis tests.

phil-davis: Yes, we now have ways to reset env variables and restart ocis during tests on ocis. So we can do that sort of thing in ocis CI. We don't have any framework to do that is reva with `toml` files etc.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -198,7 +201,7 @@ func (h *Handler) createPublicLinkShare(w http.ResponseWriter, r *http.Request,
 		Metadata: map[string]string{
 			"name":      r.FormValue("name"),
 			"quicklink": r.FormValue("quicklink"),
-			// "password": r.FormValue("password"),
+			// "password": password,
```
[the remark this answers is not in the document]
2403905: sure

PR #3980, opened 2023-06-15
drop O_CREATE for messagepack backend

butonic:
The decomposedfs messagepack backend now dies when trying to propagate metadata to a non existing parent


rhafer (approved):

butonic:
closing in favor of https://github.com/cs3org/reva/pull/4033

PR #3982, opened 2023-06-15
fixed couple of smaller space lookup issue

rhafer:
Fix ListFolder to have the correct SpaceRoot in the results
    
ListFolder passed the wrong node for the SpaceRoot parameter to ReadNode()

also:
Fix ReadNode call in NodeFromSpaceID()
    
Pass the correct ID (the spaced id) to ReadNode().


pkg/storage/utils/decomposedfs/lookup/lookup.go
```diff
@@ -164,7 +164,7 @@ func Pathify(id string, depth, width int) string {
 
 // NodeFromSpaceID converts a resource id without an opaque id into a Node
 func (lu *Lookup) NodeFromSpaceID(ctx context.Context, id *provider.ResourceId) (n *node.Node, err error) {
-	node, err := node.ReadNode(ctx, lu, id.SpaceId, id.OpaqueId, false, nil, false)
+	node, err := node.ReadNode(ctx, lu, id.SpaceId, id.SpaceId, false, nil, false)
```
aduffeck, line 167: The change itself looks good to me, but how about changing the signature of `NodeFromSpaceID` to
```
func (lu *Lookup) NodeFromSpaceID(ctx context.Context, spaceID string) (n *node.Node, err error)
```
while we're at it to better reflect what's going on?
The current signature poses the risk that it's gonna be used wrongly in the future when people shove in a resource ID and expect the opaque id to be considered.

aduffeck (approved):

PR #3985, opened 2023-06-16
Fix concurrent access to the lock map

aduffeck:
Fix concurrent access to the lock map by using a sync.Map to store the granular locks.


kobergj (approved):
Nice :+1:

PR #3995, opened 2023-06-19
convert by-user-id index to messagepack

butonic:


pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -300,28 +303,75 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
 		allMatches := map[string]string{}
-		indexPath := filepath.Join(fs.o.Root, "indexes", "by-user-id", requestedUserID.GetOpaqueId())
+		indexPath := filepath.Join(fs.o.Root, "indexes", "by-user-id", requestedUserID.GetOpaqueId()+".mpk")
 		fi, err := os.Stat(indexPath)
 		if err == nil {
 			allMatches, err = fs.spaceIDCache.LoadOrStore("by-user-id:"+requestedUserID.GetOpaqueId(), fi.ModTime(), func() (map[string]string, error) {
-				path := filepath.Join(fs.o.Root, "indexes", "by-user-id", requestedUserID.GetOpaqueId(), "*")
-				m, err := filepath.Glob(path)
+				// Acquire a read log on the index file
+				f, err := lockedfile.Open(indexPath)
```
aduffeck, line 311: Could we get away without the read lock? Reading a slightly outdated state would be fine I think, I just wonder if we could read an inconsistent state but I actually don't think so....

pkg/storage/utils/decomposedfs/spaces.go
```diff
@@ -300,28 +303,75 @@ func (fs *Decomposedfs) ListStorageSpaces(ctx context.Context, filter []*provide
...
+				return links, nil
 			})
-		}
-		if err != nil {
-			return nil, err
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			indexPath := filepath.Join(fs.o.Root, "indexes", "by-user-id", requestedUserID.GetOpaqueId())
```
aduffeck, line 342: This migration code could also go into a decomposedfs migration which would make this code leaner and cleaner.
That would also make sure that `UserIDToUserAndGroups` can always read from messagepack.

aduffeck (commented):
How about we force-refresh the index during startup? I'm just a little worried that spaces end up missing from the index when something goes wrong while updating the index. It's kinda self-healing when the according index is touched again, but esp. for the user cache this could take a while or not happen at all.

micbar:
@butonic @aduffeck Please describe any impact to existing installations.

Does the ocis instance need to be in "maintenance" mode?

aduffeck:
@micbar yes, with the current implementation the first instance running the new code will start the migration which removes the old indexes. Older instances would thus not be able to list spaces anymore.

Is that a problem? We could make it populate both indexes at first for a smoother upgrade, but I'm not sure the added complexity is justified.

micbar (approved):
LGTM

PR #3996, opened 2023-06-19
update CODEOWNERS for master branch

labkode:


phil-davis (commented):
Do you want us at JankariTech to keep maintaining the acceptance tests for master branch?

It is much easier to do that if we can be code-owners of things like `.drone.env` and `tests` ...

PR #4011, opened 2023-06-23
Improved logging of HTTP requests

glpatcern:


gmgigi96:
I would include these info as a trace level, they will be used only during development for debugging. I don't see the point of exposing them on production, even because those headers will contain the reva token, authentication secrets, etc...

glpatcern:
> I would include these info as a trace level, they will be used only during development for debugging.

Good point, we have now two log messages, one is `info` with minimal information (client host, URI, method, response status) and one is `trace` with all details.

internal/http/interceptors/log/log.go
```diff
@@ -55,7 +55,7 @@ func (h loggingHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	logger := makeLogger(w)
 	url := *req.URL
 	h.handler.ServeHTTP(logger, req)
-	writeLog(log, req, url, t, logger.Status(), logger.Size())
+	writeLog(log, req, url, t, logger.Status(), logger.Size(), w.Header())
```
labkode, line 58: it should be `logger.Header()` and not `w.Header()`.

The logger records the response before is flushed out, so we need to take the response headers from there.

labkode (changes requested):

labkode (approved):

PR #4012, opened 2023-06-23
fix mtime if 0 size file uploaded

2403905:


internal/http/services/owncloud/ocdav/put.go
```diff
@@ -168,6 +168,20 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
+		if mtime := r.Header.Get(net.HeaderOCMtime); mtime != "" {
+			smRes, err := client.SetArbitraryMetadata(ctx, &provider.SetArbitraryMetadataRequest{
```
kobergj, line 172: This means we need to do another call to the storage provider which is not nice. Can we instead send the `mtime` with the `TouchFileRequest`? I'm thinking about putting it in the opaque map, then extract it in the storage provider and pass it to decomposedfs. Similar like we do with the `markprocessing` variable. 

See https://github.com/cs3org/reva/blob/edge/pkg/storage/utils/decomposedfs/decomposedfs.go#L619 and https://github.com/cs3org/reva/blob/edge/internal/grpc/services/storageprovider/storageprovider.go#L671

We would need to change the interface but I think that is better than the extra call

kobergj (changes requested):

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -655,7 +656,10 @@ func (fs *Decomposedfs) TouchFile(ctx context.Context, ref *provider.Reference,
 	if err := n.CheckLock(ctx); err != nil {
 		return err
 	}
-	return fs.tp.TouchFile(ctx, n, markprocessing)
+	if err = n.SetXattrString(prefixes.MetadataPrefix+mtime, mtime); err != nil {
+		return err
+	}
```
kobergj, line 661: This should be done by the `tree` just like setting `markprocessing`. 

See here: https://github.com/cs3org/reva/pull/4012/files#diff-f46b4dbde0970bc0eeea2d36b8966b3fb084404cf52a006a12da0050fb8db936R156

kobergj (changes requested):

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -154,9 +154,20 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	opaqueMap := map[string]*typespb.OpaqueEntry{}
+	if mtime := r.Header.Get(net.HeaderOCMtime); mtime != "" {
+		opaqueMap[net.HeaderOCMtime] = &typespb.OpaqueEntry{
```
kobergj, line 159: use `utils.AppendPlainToOpaque`

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -191,21 +202,9 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
-
-	if mtime := r.Header.Get(net.HeaderOCMtime); mtime != "" {
-		opaqueMap[net.HeaderOCMtime] = &typespb.OpaqueEntry{
-			Decoder: "plain",
-			Value:   []byte(mtime),
-		}
-
-		// TODO: find a way to check if the storage really accepted the value
-		w.Header().Set(net.HeaderOCMtime, "accepted")
+	opaqueMap[net.HeaderUploadLength] = &typespb.OpaqueEntry{
```
kobergj, line 205: also here. use `utils.AppendPlainToOpaque`

pkg/storage/fs/owncloudsql/owncloudsql.go
```diff
@@ -830,15 +830,25 @@ func (fs *owncloudsqlfs) TouchFile(ctx context.Context, ref *provider.Reference,
 	if err != nil {
 		return err
 	}
-	mtime := time.Now().Unix()
+	storageMtime := time.Now().Unix()
+	mt := storageMtime
+	if mtime != "" {
+		t, err := strconv.Atoi(mtime)
```
kobergj, line 836: Can we use `time.Parse` instead?

kobergj (changes requested):

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -154,9 +154,20 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	opaqueMap := map[string]*typespb.OpaqueEntry{}
```
kobergj, line 157: I would suggest to change this into
```suggestion
	var opaque *typespb.Opaque
```

Then use `utils.AppendPlainToOpaque` to add `HeaderOCMTime` and `HeaderUploadLength`. This way the opaque magic is hidden in `utils` package

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj, line 661: This can be removed now?

kobergj (changes requested):

kobergj (approved):

PR #4014, opened 2023-06-26
En-/Disable DEPTH:inifinity in PROPFIND

dragonchaser:
Enhancement: En-/Disable DEPTH:inifinity in PROPFIND

We have added the ability to en-/disable DEPTH:infinitiy in PROPFIND requests for spaces

https://github.com/owncloud/ocis/issues/4188

### DIFF
--- a/internal/http/services/owncloud/ocdav/config/config.go
+++ b/internal/http/services/owncloud/ocdav/config/config.go
@@ -1,0 +1,88 @@
+package config
+
+import "github.com/cs3org/reva/v2/pkg/sharedconf"
+
+// Config holds the config options that need to be passed down to all ocdav handlers
+type Config struct {
+	Prefix string `mapstructure:"prefix"`
+	// FilesNamespace prefixes the namespace, optionally with user information.
+	// Example: if FilesNamespace is /users/{{substr 0 1 .Username}}/{{.Username}}
+	// and received path is /docs the internal path will be:
+	// /users/<first char of username>/<username>/docs
+	FilesNamespace string `mapstructure:"files_namespace"`
+	// WebdavNamespace prefixes the namespace, optionally with user information.
+	// Example: if WebdavNamespace is /users/{{substr 0 1 .Username}}/{{.Username}}
+	// and received path is /docs the internal path will be:
+	// /users/<first char of username>/<username>/docs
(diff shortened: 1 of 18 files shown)


kobergj (changes requested):

micbar:
I was under the impression that this is a configurable option. Why do we disable it completely?

butonic (approved):

internal/http/services/owncloud/ocdav/config/config.go
```diff
@@ -0,0 +1,86 @@
...
+	PublicURL                   string                            `mapstructure:"public_url"`
+	FavoriteStorageDriver       string                            `mapstructure:"favorite_storage_driver"`
+	FavoriteStorageDrivers      map[string]map[string]interface{} `mapstructure:"favorite_storage_drivers"`
+	Version                     string                            `mapstructure:"version"`
+	VersionString               string                            `mapstructure:"version_string"`
+	Edition                     string                            `mapstructure:"edition"`
+	Product                     string                            `mapstructure:"product"`
+	ProductName                 string                            `mapstructure:"product_name"`
+	ProductVersion              string                            `mapstructure:"product_version"`
+	AllowPropfindDepthInfinitiy bool                              `mapstructure:"allow_depth_infinity"`
```
kobergj, line 33: Are we sure we want this to be a breaking change? If we reverse the logic and call it `DisablePropfindDepthInfinity` we would have achieved the same, but would not have a breaking a change...

kobergj (changes requested):

internal/http/services/owncloud/ocdav/config/config.go
dragonchaser: According to the ticket... https://github.com/owncloud/ocis/issues/4188#issuecomment-1198112265

micbar: IMO the change is only breaking for the CI. We should enable it for the CI and disable it for the default deployment.

kobergj (approved):

aduffeck (approved):

dragonchaser:
reverted to draft, this seems to be causeing a nil-pointer issue when testing with ocis details: https://github.com/owncloud/ocis/pull/6842

PR #4013, opened 2023-06-26
Add WebDAV permission for PurgeRecyle

micbar:
# Description

We added the permission "PurgeRecycle" to the WebDAV permissions list. I is represented by the capital letter ``P``.

PR #4018, opened 2023-06-29
[tests-only][full-ci] Bump ocis commit id for tests 

S-Panta:
This PR bumps ocis commit id for tests
Part of: https://github.com/owncloud/QA/issues/815

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,4 +1,4 @@
 # The test runner source for API tests
-APITESTS_COMMITID=c8f8bc55e22597bc3d53b63b2f38f345b4814b07
-APITESTS_BRANCH=master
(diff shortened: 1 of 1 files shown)


saw-jan:
Another PR to bump ocis: https://github.com/cs3org/reva/pull/4013. Same failures

micbar:
@saw-jan Could it be, that we should not use the spaceContext in reva at all?

saw-jan:
> @saw-jan Could it be, that we should not use the spaceContext in reva at all?

Yeah, I think the hooks got loaded from one of the PRs in oCIS ( (I think in https://github.com/owncloud/ocis/pull/6627))
Will push the fix to the oCIS repo

S-Panta:
The bump is done in PR https://github.com/cs3org/reva/pull/4013, so closing this.

PR #4020, opened 2023-06-29
Add local setup example

javfg:
Adds a new example to the collection which sets up reva with a local storage.


glpatcern:
@javfg I guess we want to make use of the new mono-process config that was just merged, right? so you can have a single docker container

javfg:
Yes, I'll check with @gmgigi96 to translate this into the new config!

PR #4031, opened 2023-07-03
Dump config on SIGUSR1

gmgigi96:
Now that the configuration has some randomness in the allocated addresses for each services, default configs, templates that needs to be resolved, the actual (resolved) config could not be clearly visible to the operator.
This PR add an option to the reva runtime to dump the configuration on a file (default to `/tmp/reva-dump.toml` and configurable)  when the process receives a SIGUSR1 signal. Eventual errors are logged in the log.

### DIFF
--- a/cmd/revad/runtime/runtime.go
+++ b/cmd/revad/runtime/runtime.go
@@ -22,10 +22,14 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
+	"os/signal"
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 
+	"github.com/BurntSushi/toml"
 	"github.com/pkg/errors"
 
 	"github.com/cs3org/reva/cmd/revad/pkg/config"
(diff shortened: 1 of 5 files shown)


labkode (dismissed):

cmd/revad/runtime/runtime.go
```diff
@@ -115,15 +119,49 @@ func New(config *config.Config, opt ...Option) (*Reva, error) {
...
+		<-sigs
+		r.dumpConfig()
+	}()
+}
+
+func (r *Reva) dumpConfig() {
+	cfg := r.config.Dump()
+	// TODO: merge this config with the one received from
+	// the running services
+	out := r.config.Core.ConfigDumpFile
```
labkode, line 150: Is this system default or hardcoded to `/tmp`?

gmgigi96: this is configurable in the `core` config:

```
[core]
config_dump_file = "..."
```

the default is `/tmp/reva-dump.toml`

labkode: @gmgigi96 can you append filename to https://pkg.go.dev/os#TempDir?

Default is `/tmp` but if someones overwrites the env variable is a good practice in UNIX to obey

gmgigi96: @labkode done!

labkode (approved):

PR #4036, opened 2023-07-04
[WIP] Do not track files as empty files

butonic:
Currently, the messagepack metadata uses an empty file to track the mtime of a node. The rest of the data is kept in a `.mpk` file. This requires creating three files when a new file is uploaded:
1. the empty file
2. the `.mlock` file to prevent other processes to mlock the file
3. creating a temp file and then renaming it to the `.mpk` file

Furthermore, AFAICT we are leaking file handdles because we never close the empty node files after os.Creating them:
```
_, err := os.Create(nodePath) // the file handtle is just ignored?
```

We should be able to get rid of the empty node file, but we may have to take into account the metadata backend.

maybe we can just always use the Metadata path instead of the InternalPath ... but a bigger challenge is handling trashed files because we have to not only move them to the trash, but when using the messagepack backend the directory with the children ... hm ... or we leave that in the tree? the mpk file is gone ...

urgh its too late already ....

### DIFF
--- a/pkg/storage/utils/decomposedfs/node/node.go
+++ b/pkg/storage/utils/decomposedfs/node/node.go
@@ -542,9 +542,8 @@ func (n *Node) SetMtimeString(mtime string) error {
 
 // SetMtime sets the mtime and atime of a node
 func (n *Node) SetMtime(mtime time.Time) error {
-	nodePath := n.InternalPath()
 	// updating mtime also updates atime
-	return os.Chtimes(nodePath, mtime, mtime)
+	return os.Chtimes(n.lu.MetadataBackend().MetadataPath(n.InternalPath()), mtime, mtime)
 }
 
 // SetEtag sets the temporary etag of a node if it differs from the current etag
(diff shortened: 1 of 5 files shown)


pkg/storage/utils/decomposedfs/node/node.go
```diff
@@ -888,7 +887,7 @@ func (n *Node) GetTMTime(ctx context.Context) (time.Time, error) {
 
 // GetMTime reads the mtime from disk
 func (n *Node) GetMTime() (time.Time, error) {
-	fi, err := os.Lstat(n.InternalPath())
+	fi, err := os.Stat(n.lu.MetadataBackend().MetadataPath(n.InternalPath()))
```
aduffeck, line 890: could we store the mtime in the metadata itself? we propably got the metadata for the node already anyway so sparing the roundtrip to the storage would be desirable. Would also simplify propagating the tree size and tree modification time into one operation.

butonic: we certainly can and should. I just did not go in that direction first, because I stumbled over the complexity of moving all files and folders belonging to a node into trash: the dir (or unneded file) with children, the metadata file and maybe lock files?

butonic:
I am no longer working on this. Feel free to pick it up. I'll happily review.

PR #4041, opened 2023-07-05
Improve error checking code and fix sql script

javfg:


pkg/notification/db_changes.sql
```diff
@@ -41,11 +41,12 @@ COMMIT;
 CREATE INDEX `cbox_notifications_ix0` ON `cbox_notifications` (`ref`);
 
 CREATE INDEX `cbox_notification_recipients_ix0` ON `cbox_notification_recipients` (`notification_id`);
-CREATE INDEX `cbox_notification_recipients_ix1` ON `cbox_notification_recipients` (`user_name`);
+CREATE INDEX `cbox_notification_recipients_ix1` ON `cbox_notification_recipients` (`recipient`);
 
 -- changes for added notifications on oc shares
 
 ALTER TABLE cernboxngcopy.oc_share ADD notify_uploads BOOL DEFAULT false;
```
glpatcern, line 48: Can you remove the reference to the db name (`cernboxngcopy`) so we make the script generic?

glpatcern (changes requested):

pkg/notification/db_changes.sql
glpatcern: (that was my comment as well when we reviewed the main PR. Maybe time to fix this before we go to prod, as then it's going to be impossible? :) )

Not that we properly respected a naming convention, but the tables can just be created without any prefix.

javfg: Done!

pkg/notification/db_changes.sql
```diff
@@ -19,33 +19,34 @@
...
 
-CREATE INDEX `cbox_notification_recipients_ix0` ON `cbox_notification_recipients` (`notification_id`);
-CREATE INDEX `cbox_notification_recipients_ix1` ON `cbox_notification_recipients` (`user_name`);
+CREATE INDEX `notification_recipients_ix0` ON `notification_recipients` (`notification_id`);
+CREATE INDEX `notification_recipients_ix1` ON `notification_recipients` (`recipient`);
 
 -- changes for added notifications on oc shares
 
 ALTER TABLE cernboxngcopy.oc_share ADD notify_uploads BOOL DEFAULT false;
+ALTER TABLE cernboxngcopy.oc_share ADD notify_uploads_extra_recipients VARCHAR(2048);
```
glpatcern, line 49: Great! Still this one refers to the DB name...

javfg: Fixed. As long as the script is running with the proper database selected beforehand, that shouldn't be needed.

glpatcern: That's correct, and the script must be generic - the choice of the db is for the person deploying the script!
But.. did you push? :)

javfg: I did, but to the test branch :laughing:  Pushed here now.

gmgigi96 (approved):

glpatcern (approved):
Modulo the actual schema change on the `oc_share` we quickly discussed, this is so far OK

PR #4044, opened 2023-07-06
Added /app/notify endpoint for logging/tracking apps

glpatcern:
The new endpoint serves to probe the health state of apps such as Microsoft Office Online, and it is expected to be called by the frontend upon successful loading of the document by the underlying app.


glpatcern:
> Don't we have this info in the wopi driver? Like a first request that an app is doing that we can intercept?

We do, but it turns out it's not enough. This goes with https://its.cern.ch/jira/browse/CERNBOX-3406 and we already have evidence that only the frontend can tell whether the app loaded correctly.

gmgigi96 (approved):

PR #4051, opened 2023-07-10
decomposedfs: Set treesize to zero when creating a space

rhafer:
Set the initial treesize of a new storage space to zero instead of leaving it unset until the first item is uploaded.


butonic (approved):

butonic:
The code still needs to handle cases when the treesize is unset, right?

rhafer:
> The code still needs to handle cases when the treesize is unset, right?

@butonic Not, sure. Currently, when treesize is unset, Propagate() will trigger a recalculation of the treesize (using the blobsize/treesize values of its direct children). Do you want to change that? (We could set it to `0` there unconditionally)

aduffeck (approved):

PR #4056, opened 2023-07-11
fix move

2403905:


internal/http/services/owncloud/ocdav/move.go
```diff
@@ -217,6 +217,20 @@ func (s *svc) handleMove(ctx context.Context, w http.ResponseWriter, r *http.Req
...
+		// do not allow overwriting spaces
+		// check if it is a root?
+		if dstStatRes.GetInfo().GetPath() == "." {
+			log.Warn().Bool("overwrite", overwrite).Msg("overwriting spaces via dav is not allowed ")
+			w.WriteHeader(http.StatusPreconditionFailed) // 412, see https://tools.ietf.org/html/rfc4918#section-9.9.4
+			return
+		}
+		// validate if the destination can be overviewed
+		if dstStatRes.GetInfo().GetPath() != "" && dstStatRes.GetInfo().GetMimeType() != "" &&
+			(srcStatRes.GetInfo().GetPath() != dstStatRes.GetInfo().GetPath() || srcStatRes.GetInfo().GetType() != dstStatRes.GetInfo().GetType()) {
```
kobergj, line 229: I don't quite get this logic. What are we trying to check here?

Also I think `srcStatRes.GetInfo().GetType() != dstStatRes.GetInfo().GetType()` is wrong. As discussed in standup we want to allow moving a file to a folder

kobergj (changes requested):

internal/http/services/owncloud/ocdav/move.go
2403905: It is the wrong destination I think. Because the destination for a file has to be a file regarding the specs
http://www.webdav.org/specs/rfc4918.html#n-example---move-of-a-non-collection

internal/http/services/owncloud/ocdav/move.go
```diff
@@ -218,6 +218,13 @@ func (s *svc) handleMove(ctx context.Context, w http.ResponseWriter, r *http.Req
 			return
 		}
 
+		// do not allow overwriting spaces
+		if dstStatRes.GetInfo().GetPath() == "." {
```
kobergj, line 222: Is that enough? Don't we need to check for
```suggestion
		if dstStatRes.GetId().GetSpaceId() == dstStatRes.GetId().GetOpaqueId() && dstStatRes.GetInfo().GetPath() == "." {
```

kobergj (changes requested):
More a question :smile:

pkg/app/provider/wopi/wopi.go
```diff
@@ -256,6 +256,7 @@ func (p *wopiProvider) GetAppURL(ctx context.Context, resource *provider.Resourc
 	if language != "" {
 		urlQuery.Set("ui", language)      // OnlyOffice
 		urlQuery.Set("lang", language)    // Collabora
+		urlQuery.Set("rs", language)      //
```
kobergj, line 259: Did you smuggle another Bugfix into this? :rofl:

kobergj (approved):

PR #4058, opened 2023-07-11
Fix removing grants

aduffeck:
Fix removing grants from the messagepack space index.


rhafer:
Testing with ocis in: https://github.com/owncloud/ocis/pull/6748

rhafer (approved):

PR #4059, opened 2023-07-12
Fix invalid space index format from migration

aduffeck:
The space index migration wrongly migrated to a `map[string][]byte` data structure while the proper data structure is `map[string]string`. This PR adds another migration which fixes that issue.

Fixes https://github.com/owncloud/ocis/issues/6781


kobergj (approved):
:+1: LGTM

PR #4061, opened 2023-07-12
Fixes on notifications

glpatcern:
This follows #4041

@javfg can you have a look? I didn't test it yet


pkg/notification/db_changes.sql
```diff
@@ -45,9 +45,5 @@ CREATE INDEX `notification_recipients_ix1` ON `notification_recipients` (`recipi
 
 -- changes for added notifications on oc shares
 
-ALTER TABLE oc_share ADD notify_uploads BOOL DEFAULT false;
+ALTER TABLE oc_share ADD notify_uploads INTEGER DEFAULT 0 NOT NULL;
```
gmgigi96, line 48: Why an Integer?

glpatcern: Because I suggested - a bit late indeed - to have a more future-proof column for this, imaging that in the future we want to enable notifications for different types of events (and even the name is misleading then). But I realize after talking to @javfg that we have to review the whole stack, CS3APIs included. Sorry for not having picked up this when we reviewed the CS3API change...

At this stage, for consistency this is to be reverted to `BOOL`, but I'd still like to check what it entails to change it properly across the stack. Remember that once we apply this change in prod, we create a column for 490K entries - which itself takes quite a bit of time and takes an exclusive lock on the table! - and it would be a pity to have to change that or to need to add yet another column afterwards.

gmgigi96: This is why we should normalize this table :)

glpatcern: I agree!

labkode (approved):

javfg (approved):

PR #4062, opened 2023-07-12
Centralize HTTP routing

gmgigi96:
Different HTTP services handle the HTTP routing with different methods, using a big switch statement and checking the next segment in the URL, or using external libraries like chi.
Also each service was required to defined a prefix used as root path for all the endpoint the service was exposing. This represents a limit for services exposing on different root path, like `ocm`, which should expose on both `/ocm` and `/ocm-provider`, or `ocdav` on which the default prefix is the empty string.

In order to streamline the process of creation of new services, a library has been added handling all the HTTP routing in a common way, so that each service is only responsible to implement its business logic and define in a declarative way which are the endpoints and method exposed.
The new routing implementation is based on a radix tree, and supports wildcards:
 - `:key`, which is matching a single path segment
 - `*key`, which is matching everything

The actual values for the matches can be retrieved using the method `ParamsFromRequest`.

All the registered endpoints are logged at init time, reporting which ones are unprotected:
```
2023-07-18 17:04:54.549 DBG pkg/rhttp/rhttp.go:159 > GET        /app/list (unprotected) pid=24243 pkg=http
2023-07-18 17:04:54.549 DBG pkg/rhttp/rhttp.go:159 > POST       /app/new pid=24243 pkg=http
```

All the existing HTTP services have been migrated to new structure, but `ocdav` that requires some more work, and is left as future work.


butonic:
Why not stick to a common go http router / mux like chi or even a framework like fiber? I don't think reinventing the wheel is helpful.

PR #4070, opened 2023-07-17
Selectable propagators

aduffeck:
This PR adds an experimental asynchronous propagator which propagates changes asynchronously outside of the request scope. It also combines changes that happen shortly after each other into one change.

The propagator can be configured, the original `sync` propagator is still the default.

### DIFF
--- a/changelog/unreleased/async-propagator.md
+++ b/changelog/unreleased/async-propagator.md
@@ -1,5 +1,5 @@
 Enhancement: async propagation (experimental)
 
-decomposedfs can now be configured to propagat treetime/treesize changes asynchronously.
+decomposedfs can now be configured to propagate treetime/treesize changes asynchronously.
 
 https://github.com/cs3org/reva/pull/4070
--- a/changelog/unreleased/async-propagator.md
+++ b/changelog/unreleased/async-propagator.md
@@ -1,0 +1,5 @@
+Enhancement: async propagation (experimental)
+
+decomposedfs can now be configured to propagat treetime/treesize changes asynchronously.
+
+https://github.com/cs3org/reva/pull/4070
--- a/pkg/storage/utils/decomposedfs/lookup/lookup.go
+++ b/pkg/storage/utils/decomposedfs/lookup/lookup.go
@@ -44,6 +44,20 @@ func init() {
 	tracer = otel.Tracer("github.com/cs3org/reva/pkg/storage/utils/decomposedfs/lookup")
 }
 
+// PathLookup defines the interface for the lookup component
+type PathLookup interface {
+	NodeFromResource(ctx context.Context, ref *provider.Reference) (*node.Node, error)
+	NodeFromID(ctx context.Context, id *provider.ResourceId) (n *node.Node, err error)
+
+	InternalRoot() string
+	InternalPath(spaceID, nodeID string) string
+	Path(ctx context.Context, n *node.Node, hasPermission node.PermissionFunc) (path string, err error)
+	MetadataBackend() metadata.Backend
+	ReadBlobSizeAttr(ctx context.Context, path string) (int64, error)
+	ReadBlobIDAttr(ctx context.Context, path string) (string, error)
+	TypeFromPath(ctx context.Context, path string) provider.ResourceType
+}
+
 // Lookup implements transformations from filepath to node and back
 type Lookup struct {
 	Options *options.Options
--- a/pkg/storage/utils/decomposedfs/tree/propagator/async.go
+++ b/pkg/storage/utils/decomposedfs/tree/propagator/async.go
@@ -296,7 +296,21 @@ func (p AsyncPropagator) propagate(ctx context.Context, spaceID, nodeID string, 
 	}
 
 	if !n.IsSpaceRoot(ctx) { // This does not seem robust as it checks the space name property
-		p.propagateChange(ctx, n.SpaceID, n.ParentID, pc, log)
+		p.queuePropagation(ctx, n.SpaceID, n.ParentID, pc, log)
+	}
+
+	// Check for a changes dir that might have been added meanwhile and pick it up
+	if _, err = os.Open(changeDirPath); err == nil {
+		err = os.Rename(changeDirPath, processingPath)
+		if err != nil {
+			// This can fail in 2 ways
+			// 1. source does not exist anymore as it has already been propagated by another goroutine
+			//    -> ignore, as the change is already being processed
+			// 2. target already exists because a previous propagation is still running
+			//    -> ignore, the previous propagation will pick the new changes up
+			return
+		}
+		p.propagate(ctx, spaceID, nodeID, false, log)
 	}
 }
 
--- a/pkg/storage/utils/decomposedfs/tree/propagator/async.go
+++ b/pkg/storage/utils/decomposedfs/tree/propagator/async.go
@@ -40,6 +40,8 @@ import (
 	"github.com/shamaton/msgpack/v2"
 )
 
+const propagationDelay = 5 * time.Second
+
 type AsyncPropagator struct {
 	treeSizeAccounting bool
 	treeTimeAccounting bool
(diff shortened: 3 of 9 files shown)


changelog/unreleased/async-propagator.md
```diff
@@ -0,0 +1,5 @@
+Enhancement: async propagation (experimental)
+
+decomposedfs can now be configured to propagat treetime/treesize changes asynchronously.
```
kobergj, line 3:
```suggestion
decomposedfs can now be configured to propagate treetime/treesize changes asynchronously.
```

pkg/storage/utils/decomposedfs/lookup/lookup.go
```diff
@@ -44,6 +44,20 @@ func init() {
...
+
+	InternalRoot() string
+	InternalPath(spaceID, nodeID string) string
+	Path(ctx context.Context, n *node.Node, hasPermission node.PermissionFunc) (path string, err error)
+	MetadataBackend() metadata.Backend
+	ReadBlobSizeAttr(ctx context.Context, path string) (int64, error)
+	ReadBlobIDAttr(ctx context.Context, path string) (string, error)
+	TypeFromPath(ctx context.Context, path string) provider.ResourceType
+}
+
```
kobergj, line 60: Can we get rid of the interface entirely? I don't see a reason for keeping it - Reimplementing such a big interface will not be done anyways...

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
+	"github.com/google/renameio"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/rogpeppe/go-internal/lockedfile"
+	"github.com/rs/zerolog"
+	"github.com/shamaton/msgpack/v2"
+)
+
+const propagationDelay = 5 * time.Second
```
kobergj, line 43: Shouldn't this be configurable? I imagine we want to play with it to find the best value

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+
+type AsyncPropagator struct {
```
kobergj, line 45: exported type `AsyncPropagator` should have comment or be unexported.

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+	treeSizeAccounting bool
+	treeTimeAccounting bool
+	lookup             lookup.PathLookup
+}
+
+type Change struct {
```
kobergj, line 51: exported type `Change` should have comment or be unexported.

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+	SyncTime time.Time
+	SizeDiff int64
+}
+
+func NewAsyncPropagator(treeSizeAccounting, treeTimeAccounting bool, lookup lookup.PathLookup) AsyncPropagator {
```
kobergj, line 56: exported func `NewAsyncPropagator` should have comment or be unexported.

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+		for {
+			if doSleep {
+				time.Sleep(5 * time.Minute)
+			}
+			doSleep = true
+			log.Debug().Msg("scanning for stale .processing dirs")
+
+			entries, err := filepath.Glob(changesDirPath + "/**/*")
+			if err != nil {
+				continue
```
kobergj, line 86: Don't we want to log errors that happen during propagation?

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+			}
+
+			for _, e := range entries {
+				changesDirPath := e
+				entry, err := os.Stat(changesDirPath)
+				if err != nil {
+					continue
+				}
+				// recover all dirs that haven't been propagated within 3 minutes
+				if !entry.IsDir() || time.Now().Before(entry.ModTime().Add(3*time.Minute)) {
```
kobergj, line 96: Can you extract `3*time.Minute` in a global var? Then it is easier to spot

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+							return
+						}
+						changesDirPath += ".processing"
+					}
+
+					log.Debug().Str("dir", changesDirPath).Msg("propagating stale .processing dir")
+					parts := strings.SplitN(entry.Name(), ":", 2)
+					now := time.Now()
+					_ = os.Chtimes(changesDirPath, now, now)
+					p.propagate(context.Background(), parts[0], strings.TrimSuffix(parts[1], ".processing"), true, *log)
```
kobergj, line 114: We need to check for `len(parts)` first. Otherwise this can panic under strange circumstances

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+				}()
+			}
+		}
+	}()
+
+	return p
+}
+
+func (p AsyncPropagator) Propagate(ctx context.Context, n *node.Node, sizeDiff int64) error {
```
kobergj, line 123: exported func `Propagate` should have comment or be unexported

pkg/storage/utils/decomposedfs/tree/propagator/async.go
```diff
@@ -0,0 +1,427 @@
...
+		return
+	}
+
+	_, subspan := tracer.Start(ctx, "write changes file")
+	ready := false
+	triggerPropagation := false
+	_ = os.MkdirAll(filepath.Dir(filepath.Dir(changePath)), 0700)
+	err = os.Mkdir(filepath.Dir(changePath), 0700)
+	triggerPropagation = err == nil || os.IsExist(err) // only the first goroutine, which succeeds to create the directory, is supposed to actually trigger the propagation
+	for retries := 0; retries <= 500; retries++ {
```
kobergj, line 167: `500` retries seems absolutely random. Is there a reason for this?

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/cs3org/reva/v2/pkg/appctx"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/lookup"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata/prefixes"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
+	"github.com/rogpeppe/go-internal/lockedfile"
+)
+
+type SyncPropagator struct {
```
kobergj, line 36: exported type `SyncPropagator` should have comment or be unexported

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	treeSizeAccounting bool
+	treeTimeAccounting bool
+	lookup             lookup.PathLookup
+}
+
+func NewSyncPropagator(treeSizeAccounting, treeTimeAccounting bool, lookup lookup.PathLookup) SyncPropagator {
```
kobergj, line 42: exported func `NewSyncPropagator` should have comment or be unexported

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
```diff
@@ -0,0 +1,206 @@
...
+	return SyncPropagator{
+		treeSizeAccounting: treeSizeAccounting,
+		treeTimeAccounting: treeTimeAccounting,
+		lookup:             lookup,
+	}
+}
+
+func (p SyncPropagator) Propagate(ctx context.Context, n *node.Node, sizeDiff int64) error {
```
kobergj, line 50: exported func `Propagate` should have comment or be unexported

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+		}()
+
+		if n, err = n.Parent(ctx); err != nil {
+			sublog.Error().Err(err).
+				Msg("Propagation failed. Could not read parent node.")
+			return err
+		}
+
+		// TODO none, sync and async?
```
kobergj, line 107: comment can be removed - we are in sync case

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
kobergj, line 50: Can we reuse the `Propagate` method from the sync propagator in the async propagator? Or at least extract some parts of it to a common function. I can see lots of duplication in these two methods....

kobergj (changes requested):
some questions and nitpicks

pkg/storage/utils/decomposedfs/lookup/lookup.go
aduffeck: Maybe. But it's not only about enabling reimplementations but also about decoupling dependencies (just removing it currently leads to import cycles) and increasing testability (albeit I'm not sure whether the latter is the case here.
We can take a look at that, the lookup interface has been a nuisance a few times already, but I'd rather we do that as a separate change as it's gonna be a bigger refactoring.

pkg/storage/utils/decomposedfs/tree/propagator/async.go
aduffeck: Well, the thing here is that we *really* want to write the change to disk and not lose this update, but failing to write is an expected case when changes directory has been moved by a propagation goroutine meanwhile and this could happen multiple times as well.

At the same time we can't keep trying indefinitely as we might pile up goroutines and run out of resources at some point.

pkg/storage/utils/decomposedfs/tree/propagator/sync.go
aduffeck: Phew... I just tried but there are lots of subtle changes and slightly different dependencies. I would suggest we defer this refactoring to a later point in time so we can get the feature in for testing now.

kobergj (approved):

PR #4072, opened 2023-07-18
revad: Allow to specify a shutdown timeout

rhafer:
Wnen setting 'graceful_shutdown_timeout' revad with try to shutdown in a graceful manner when receiving an INT or TERM signal (similar to how it already behaves on SIGQUIT). This allows ongoing operations to complete before exiting.

If the shutdown didn't finish before 'graceful_shutdown_timeout' seconds the process will exit with an error code (1).

### DIFF
--- a/cmd/revad/internal/grace/grace.go
+++ b/cmd/revad/internal/grace/grace.go
@@ -36,13 +36,14 @@ import (
 // Watcher watches a process for a graceful restart
 // preserving open network sockets to avoid packet loss.
 type Watcher struct {
-	log       zerolog.Logger
-	graceful  bool
-	ppid      int
-	lns       map[string]net.Listener
-	ss        map[string]Server
-	pidFile   string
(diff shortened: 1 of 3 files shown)


cmd/revad/internal/grace/grace.go
```diff
@@ -36,13 +36,14 @@ import (
...
-	pidFile   string
-	childPIDs []int
+	log                     zerolog.Logger
+	graceful                bool
+	ppid                    int
+	lns                     map[string]net.Listener
+	ss                      map[string]Server
+	pidFile                 string
+	childPIDs               []int
+	gracefulShutdownTimeout int
```
kobergj, line 46: Should we make this a `time.Duration`? That is what it is after all and we could configure it like `12s`

kobergj (approved):
Looks good but found a small typo and have a question. Nothing critical though...

PR #4077, opened 2023-07-24
Handle target in OpenInApp response

glpatcern:
This PR adds the `OpenInApp.target` and `AppProviderInfo.action` properties to the respective responses (from `/app/open` and `/app/list`), to support different app integrations. Builds on top of https://github.com/cs3org/cs3apis/pull/212.

In addition, the archiver was extended to use the name of the file/folder as opposed to `download`, and to include a query parameter to override the archive type, as it will be used in an upcoming app, see #4084.

### DIFF
--- a/internal/http/services/appprovider/appprovider.go
+++ b/internal/http/services/appprovider/appprovider.go
@@ -24,10 +24,11 @@ import (
 	"net/http"
 	"path"
 
+	apppb "github.com/cs3org/go-cs3apis/cs3/app/provider/v1beta1"
 	appregistry "github.com/cs3org/go-cs3apis/cs3/app/registry/v1beta1"
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
-	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	storagepb "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	typespb "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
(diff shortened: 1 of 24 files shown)


labkode:
Rebase needed

internal/http/services/archiver/handler.go
```diff
@@ -229,13 +230,27 @@ func (s *svc) Handler() http.Handler {
 			return
 		}
 
-		userAgent := ua.Parse(r.Header.Get("User-Agent"))
+		archTypeParams, ok := v["arch_type"] // optional, either "tar" or "zip"
```
gmgigi96, line 233: Use https://pkg.go.dev/net/url#Values.Get to simplify the code

glpatcern: OK!

internal/http/services/appprovider/appprovider.go
```diff
@@ -512,3 +522,14 @@ func resolveViewMode(res *provider.ResourceInfo, vm string) gateway.OpenInAppReq
 	}
 	return viewMode
 }
+
+func appTargetToString(t apppb.Target) string {
```
labkode, line 526: In case the the target is `invalid` you'll be opening it with `iframe`?

glpatcern: Yeah, I thought about that, after all any future app provider would have to specify the target so let's not assume a default: I return empty string now (i.e. likely failing on the frontend)

glpatcern: Not sure it makes much sense given the context (I'd have preferred a default value in protobuf, but it's an enum so it must have a value). But I don't have a strong opinion, so ok I changed it to `invalid`.

labkode (approved):

labkode:
Breaking change:
Default changes from "download_archive" http prefix to "archiver".

PR #4078, opened 2023-07-25
fix the default document language for OnlyOffice

2403905:
https://github.com/owncloud/enterprise/issues/5807


pkg/app/provider/wopi/wopi.go
```diff
@@ -470,3 +470,22 @@ func getEtherpadExtensions(appURL string) map[string]map[string]string {
 	}
 	return appURLs
 }
+
+// TODO Find better solution
+// This conversion was made because no other way to set the default document language to OnlyOffice was found.
+func covertLangTag(lang string) string {
```
phil-davis, line 476: I suppose that the function name was supposed to be `convertLangTag`
But it doesn't matter - it is just a function name.

PR #4084, opened 2023-07-27
Implementation of an app provider for Overleaf

abaldacchino:
This PR adds an app provider for Overleaf as a standalone http service. 
The app provider currently consists of support for the export to Overleaf feature, which when called returns a URL to Overleaf that prompts Overleaf to download the appropriate resource making use of the Archiver service, and upload the files to a user's Overleaf account.


internal/http/services/overleaf/overleaf.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// Setting up archiver request
+	archHTTPReq, err := rhttp.NewRequest(ctx, http.MethodGet, s.conf.ArchiverURL, nil)
+	if err != nil {
+		reqres.WriteError(w, r, reqres.APIErrorServerError, "overleaf: error setting up http request", nil)
+		return
+	}
+
+	archQuery := archHTTPReq.URL.Query()
+	archQuery.Add("id", resource.Id.StorageId+"!"+resource.Id.OpaqueId)
```
gmgigi96, line 191: Don't build the resource id manually. Use [this](https://github.com/cs3org/reva/blob/5bd1923565619e21f51728019456da010e1d5689/pkg/utils/resourceid/owncloud.go#L61C6-L61C28) function instead

pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -1176,18 +1176,15 @@ func (c *Client) mapToFileInfo(ctx context.Context, kv, attrs map[string]string,
 		if mtimesec, err = strconv.ParseUint(stimeSplit[0], 10, 64); err == nil {
 			mtimeSet = true
 		}
+
```
gmgigi96, line 1179: remove this file and rebase on master

internal/http/services/overleaf/overleaf.go
```diff
@@ -0,0 +1,293 @@
...
+				reqres.WriteError(w, r, reqres.APIErrorServerError, "error marshalling JSON response", err)
+				return
+			}
+			w.Header().Set("Content-Type", "application/json")
+			return
+		}
+	}
+
+	// TODO: generate and use a more restricted token
+	restrictedToken := token
```
gmgigi96, line 181: For generating a restricted token, use a similar code to [this](https://github.com/cs3org/reva/blob/5bd1923565619e21f51728019456da010e1d5689/pkg/share/cache/warmup/cbox/cbox.go#L97-L119).

For the scope, you can use [this](https://github.com/cs3org/reva/blob/5bd1923565619e21f51728019456da010e1d5689/pkg/auth/scope/resourceinfo.go#L125C21-L125C21) function to get a restricted scope for the resource you want to archive. As role set it to viewer, and the initial scope as nil.

gmgigi96 (changes requested):

gmgigi96:
This can be in the future an app provider, once we will define an `Export` and `Import` in the CS3APIs to handle cases like overleaf that do not allow any storage integration. Until that it's fine to keep it as a standalone http service.

gmgigi96 (approved):

gmgigi96 (approved):
Congrats @abaldacchino for your first merged PR in reva!

PR #4098, opened 2023-08-03
Allow naming nats connections

kobergj:
Allows naming nats connections

ATTENTION: Breaking change for ocis. Merge shortly before ocis PR


micbar:
this is still a draft. Should we review?

kobergj:
Yes please. I kept it on draft because it will temporarily break edge compatibility. I'd prefer to merge it after ocis part (https://github.com/cs3org/reva/pull/4098)  is reviewed too, so downtime is as short as possible.

2403905 (approved):

PR #4099, opened 2023-08-04
Log upload errors

aduffeck:
We fixed a problem where problems with uploading blobs to the blobstore weren't logged.

Fixes https://github.com/owncloud/ocis/issues/6962


kobergj:
Don't we also need a log for the async upload case? Or is this covered already?

aduffeck:
@kobergj in the async-uploads case the errors are logged at https://github.com/cs3org/reva/blob/edge/pkg/storage/utils/decomposedfs/decomposedfs.go#L271 already.

aduffeck:
Superseded by https://github.com/cs3org/reva/pull/4100

PR #4100, opened 2023-08-04
S3ng include md5 checksum on put

wkloucek:
Bugfix: S3ng include md5 checksum on put

We've fixed the S3 put operation of the S3ng storage to include a md5 checksum.

This md5 checksum is needed when a bucket has a retention period configured (see https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html).

### DIFF
--- a/pkg/storage/fs/s3ng/blobstore/blobstore.go
+++ b/pkg/storage/fs/s3ng/blobstore/blobstore.go
@@ -71,9 +71,9 @@ func (bs *Blobstore) Upload(node *node.Node, source string) error {
 	}
 	defer reader.Close()
 
-	_, err1 := bs.client.PutObject(context.Background(), bs.bucket, bs.path(node), reader, node.Blobsize, minio.PutObjectOptions{ContentType: "application/octet-stream"})
+	_, err = bs.client.PutObject(context.Background(), bs.bucket, bs.path(node), reader, node.Blobsize, minio.PutObjectOptions{ContentType: "application/octet-stream"})
 
-	if err1 != nil {
+	if err != nil {
 		return errors.Wrapf(err, "could not store object '%s' into bucket '%s'", bs.path(node), bs.bucket)
 	}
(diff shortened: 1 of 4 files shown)


pkg/storage/fs/s3ng/blobstore/blobstore.go
```diff
@@ -71,9 +71,9 @@ func (bs *Blobstore) Upload(node *node.Node, source string) error {
 	}
 	defer reader.Close()
 
-	_, err1 := bs.client.PutObject(context.Background(), bs.bucket, bs.path(node), reader, node.Blobsize, minio.PutObjectOptions{ContentType: "application/octet-stream"})
+	_, err = bs.client.PutObject(context.Background(), bs.bucket, bs.path(node), reader, node.Blobsize, minio.PutObjectOptions{ContentType: "application/octet-stream", SendContentMd5: true})
```
butonic, line 74: This might be a performance problem. `SendContentMd5: true` causes minio-go to reread the whole file before making the actual putobject request as it needs to send a `Content-MD56` header. Maybe a more recent version of minio-go supports sending that as a trailer.

butonic: nah ... we can't send our MD5 with this lib ... the `SendContentMd5` flag will always calculate the hash itself and there is no way to sneak in a `Content-MD5` header.

butonic: let's try to get a fix into upstream: https://github.com/minio/minio-go/pull/1867

butonic (approved):

butonic:
we cannot use the precalculated md5 hash when large files are sent using multipart upload. For now stick to the lib calculating the hash. We can work on https://github.com/minio/minio-go/pull/1867 to make it use a user provided hash when it does not use multipart uploads.

PR #4104, opened 2023-08-09
mime: support mixed-case file extensions

glpatcern:
Port of https://github.com/cs3org/reva/pull/4103

### DIFF
--- a/pkg/mime/mime.go
+++ b/pkg/mime/mime.go
@@ -47,8 +47,7 @@ func Detect(isDir bool, fn string) string {
 		return defaultMimeDir
 	}
 
-	ext := path.Ext(fn)
-	ext = strings.TrimPrefix(ext, ".")
+	ext := strings.ToLower(strings.TrimPrefix(path.Ext(fn), "."))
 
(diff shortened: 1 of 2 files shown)


pkg/mime/mime.go
```diff
@@ -47,8 +47,7 @@ func Detect(isDir bool, fn string) string {
 		return defaultMimeDir
 	}
 
-	ext := path.Ext(fn)
-	ext = strings.TrimPrefix(ext, ".")
+	ext := strings.ToLower(strings.TrimPrefix(path.Ext(fn), "."))
```
micbar, line 50: This is not needed because line 55 is doing a comparison with lowercase

micbar (approved):

micbar (changes requested):
Not needed.

pkg/mime/mime.go
glpatcern: Ah! I "jumped" on it to try and keep those parts in sync, thanks for noting. I'll close the PR.

PR #4107, opened 2023-08-09
Return locks on PROPFIND when requested

kobergj:
Locks would not be properly returned when another prop was requested. This fixes it.


lookacat (approved):
LGTM (as far as my backend/go knowledge goes)

butonic (approved):

PR #4110, opened 2023-08-10
Add tracing propagation to events

ainmosni:
This adds a TraceParent field to the event envelope, adds context as
a parameter to the Publish function, adds functionality to extract
the traceparent from the context, and added a method to the envelope
to recover the tracing info and add it to the context.

This also passes context (empty if no context was present in the calling
function) to Publish wherever it is called.

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -19,11 +19,13 @@
 package events
 
 import (
+	"context"
 	"log"
 	"reflect"
 
 	"github.com/google/uuid"
 	"go-micro.dev/v4/events"
+	"go.opentelemetry.io/otel/propagation"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


pkg/events/events.go
```diff
@@ -134,10 +142,30 @@ func ConsumeAll(s Consumer, group string) (<-chan Event, error) {
...
+}
+
+// getTraceParentFromCtx will return a traceparent from the context if it exists.
+// it will be a string as specificied here: https://www.w3.org/TR/trace-context/
+// If no trace info in the context, the return will be an empty string
+func getTraceParentFromCtx(ctx context.Context) string {
+	mc := propagation.MapCarrier{}
+	tc := propagation.TraceContext{}
+	tc.Inject(ctx, &mc)
+	return mc["traceparent"]
```
Excds, line 170: This should use the constant?

kobergj (approved):
LGTM :+1:

pkg/events/events.go
ainmosni: The constant there is not the same as the other one, this is the key that is described in the wc3 spec and what is used by the propagation functions. I think using the same constant will give the wrong intent.

Excds: Ah, ok.

PR #4112, opened 2023-08-14
fix: downloading remote files gives HTTP 500

MahdiBaghbani:
related to https://github.com/sciencemesh/nc-sciencemesh/issues/34


glpatcern (approved):
Good finding! :+1:

PR #4117, opened 2023-08-17
Grant locking

butonic:


butonic:
why are we testing that restoring a revision with the same mtime as the current file deletes the current file? shouldn't they just swap places? I think this test is weird.

```gherkin
  Scenario: upload the same file twice with the same mtime and no version after restoring                                                            # /drone/src/tmp/testrunner/tests/acceptance/features/coreApiVersions/fileVersions.feature:267
    Given user "Alice" has uploaded file "filesForUpload/textfile.txt" to "file.txt" with mtime "Thu, 08 Aug 2019 04:18:13 GMT" using the WebDAV API # FeatureContext::userHasUploadedFileToWithMtimeUsingTheWebdavApi()
    And user "Alice" has uploaded file "filesForUpload/textfile.txt" to "file.txt" with mtime "Thu, 08 Aug 2019 04:18:13 GMT" using the WebDAV API   # FeatureContext::userHasUploadedFileToWithMtimeUsingTheWebdavApi()
    When user "Alice" restores version index "1" of file "/file.txt" using the WebDAV API                                                            # FilesVersionsContext::userRestoresVersionIndexOfFile()
    Then the HTTP status code should be "204"                                                                                                        # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And the version folder of file "/file.txt" for user "Alice" should contain "0" element                                                           # FilesVersionsContext::theVersionFolderOfFileShouldContainElements()
      could not find 0 version element(s) in 
      <?xml version="1.0"?>
      <d:multistatus xmlns:s="http://sabredav.org/ns" xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns"><d:response><d:href>/remote.php/dav/meta/1284d238-aa92-42ce-bdc4-0b0000009157$6d58d130-d156-103d-90fa-1357bb3e8dca%21bc2fba86-ae17-40e3-8d7c-4df74ea37af8/</d:href><d:propstat><d:prop><d:getetag/></d:prop><d:status>HTTP/1.1 404 Not Found</d:status></d:propstat></d:response><d:response><d:href>/remote.php/dav/meta/1284d238-aa92-42ce-bdc4-0b0000009157$6d58d130-d156-103d-90fa-1357bb3e8dca%21bc2fba86-ae17-40e3-8d7c-4df74ea37af8/v/bc2fba86-ae17-40e3-8d7c-4df74ea37af8.REV.2023-08-17T14:30:58.705779052Z</d:href><d:propstat><d:prop><d:getetag>"3e829cdcdaed9b04535a474921e2492a"</d:getetag></d:prop><d:status>HTTP/1.1 200 OK</d:status></d:propstat></d:response></d:multistatus>
      
      Failed asserting that 1 matches expected 0.
```

changelog/unreleased/grant-locking.md
```diff
@@ -0,0 +1,7 @@
+Bugfix: fix jsoncs3 atomic persistence
+
+The jsoncs3 shame manager now uses etags instead of mtimes to determine when metadata needs to be updated.
```
micbar, line 3:
```suggestion
The jsoncs3 share manager now uses etags instead of mtimes to determine when metadata needs to be updated.
```

pkg/share/manager/jsoncs3/providercache/providercache.go
```diff
@@ -146,221 +158,336 @@ func (c *Cache) Add(ctx context.Context, storageID, spaceID, shareID string, sha
...
-		span.SetStatus(codes.Error, fmt.Sprintf("Failed to stat the provider cache: %s", err.Error()))
-		log.Error().Err(err).Msg("Failed to stat the provider cache")
-		return err
-	}
-	// check mtime of /users/{userid}/created.json
-	if utils.TSToTime(info.Mtime).After(mtime) {
-		span.AddEvent("updating cache")
-		//  - update cached list of created shares for the user in memory if changed
-		createdBlob, err := c.storage.SimpleDownload(ctx, jsonPath)
+		err = downloadFunc()
```
kobergj, line 457: Shouldn't we have some retry logic here? We execute the same func without waiting twice. Why would we expect different results? Why do we retry only once?

There a multiple occurences of this, so I assume this is a deliberate decision?

pkg/storage/utils/metadata/cs3.go
```diff
@@ -161,6 +164,14 @@ func (cs3 *CS3) Upload(ctx context.Context, req UploadRequest) error {
 			IfMatch: req.IfMatchEtag,
 		}
 	}
+	if len(req.IfNoneMatch) > 0 {
+		if req.IfNoneMatch[0] == "*" {
+			ifuReq.Options = &provider.InitiateFileUploadRequest_IfNotExist{}
+		}
+		// else {
+		//   the http upload will carry all if-not-match etags
+		// }
```
kobergj, line 173: remove?

pkg/storage/utils/metadata/cs3.go
```diff
@@ -188,21 +199,38 @@ func (cs3 *CS3) Upload(ctx context.Context, req UploadRequest) error {
...
 
 	md, _ := metadata.FromOutgoingContext(ctx)
 	httpReq.Header.Add(ctxpkg.TokenHeader, md.Get(ctxpkg.TokenHeader)[0])
 	resp, err := cs3.dataGatewayClient.Do(httpReq)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	if err := errtypes.NewErrtypeFromHTTPStatusCode(resp.StatusCode, httpReq.URL.Path); err != nil {
+		// resp.Body.Close()
```
kobergj, line 220: remove?

pkg/storage/utils/metadata/cs3.go
```diff
@@ -258,49 +286,156 @@ func (cs3 *CS3) SimpleDownload(ctx context.Context, downloadpath string) (conten
...
+			return err
+		}
+
+		if err = resp.Body.Close(); err != nil {
+			return err
+		}
+		return nil
+	}
+	err = downloadFunc()
+	// TODO add retries
```
kobergj, line 433: not needed here?

kobergj (changes requested):
Looks good. Just some questions

pkg/share/manager/jsoncs3/providercache/providercache.go
butonic: NotFound or NotModified errors are expected as part of normal operations. Only for the other errors are we going to retry the download. In theory that should not be necessary at all. We are just being cautious.

pkg/storage/utils/metadata/cs3.go
butonic: I left it in to explain the missing else case ...

pkg/storage/utils/metadata/cs3.go
butonic: No, not really ... I put it there because we added `Download()` with more options. And one option in the future could be built in retries ... we can drop the comment.

pkg/share/manager/jsoncs3/providercache/providercache.go
kobergj: Not talking about the error switch, thats fine and clear imo. I'm about:
```
	err = downloadFunc()
	if err != nil {
		err = downloadFunc()
       }
```
We are executing the same func without any checks twice. There is also no retry counter. So we just retry twice within milliseconds.

butonic: cleaned up and tested locally and with the helm chart

kobergj (approved):

PR #4133, opened 2023-08-23
Add Metadata to lock

kobergj:
Adds `ownername` and `locktime` to the lock information

### DIFF
--- a/internal/http/services/owncloud/ocdav/locks.go
+++ b/internal/http/services/owncloud/ocdav/locks.go
@@ -42,6 +42,7 @@ import (
 	ctxpkg "github.com/cs3org/reva/v2/pkg/ctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
 	"github.com/cs3org/reva/v2/pkg/rgrpc/todo/pool"
+	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/google/uuid"
 	"go.opentelemetry.io/otel/attribute"
 )
@@ -190,11 +191,18 @@ func (cls *cs3LS) Create(ctx context.Context, now time.Time, details LockDetails
 	// see: http://www.webdav.org/specs/rfc2518.html#n-lock-tokens
 	token := uuid.New()
 
+	u := ctxpkg.ContextMustGetUser(ctx)
+
+	// add metadata via opaque
+	o := utils.AppendPlainToOpaque(nil, "lockownername", u.GetDisplayName())
+	o = utils.AppendPlainToOpaque(o, "locktime", now.Format(time.RFC3339))
+
 	r := &provider.SetLockRequest{
 		Ref: details.Root,
 		Lock: &provider.Lock{
-			Type: provider.LockType_LOCK_TYPE_EXCL,
-			User: details.UserID, // no way to set an app lock? TODO maybe via the ownerxml
+			Opaque: o,
+			Type:   provider.LockType_LOCK_TYPE_EXCL,
+			User:   details.UserID, // no way to set an app lock? TODO maybe via the ownerxml
 			//AppName: , // TODO use a urn scheme?
 			LockId: lockTokenPrefix + token.String(), // can be a token or a Coded-URL
 		},
@@ -278,6 +286,10 @@ type LockDetails struct {
 	// ZeroDepth is whether the lock has zero depth. If it does not have zero
 	// depth, it has infinite depth.
 	ZeroDepth bool
+	// OwnerName is the name of the lock owner
+	OwnerName string
+	// Locktime is the time the lock was created
+	Locktime time.Time
 }
 
 func readLockInfo(r io.Reader) (li lockInfo, status int, err error) {
@@ -440,7 +452,8 @@ func (s *svc) lockReference(ctx context.Context, w http.ResponseWriter, r *http.
 	}
 
 	u := ctxpkg.ContextMustGetUser(ctx)
-	token, ld, now, created := "", LockDetails{UserID: u.Id, Root: ref, Duration: duration}, time.Now(), false
+	token, now, created := "", time.Now(), false
+	ld := LockDetails{UserID: u.Id, Root: ref, Duration: duration, OwnerName: u.GetDisplayName(), Locktime: now}
 	if li == (lockInfo{}) {
 		// An empty lockInfo means to refresh the lock.
 		ih, ok := parseIfHeader(r.Header.Get(net.HeaderIf))
@@ -566,6 +579,13 @@ func writeLockInfo(w io.Writer, token string, ld LockDetails) (int, error) {
 	if href != "" {
 		lockdiscovery.WriteString(fmt.Sprintf("  <d:lockroot><d:href>%s</d:href></d:lockroot>\n", prop.Escape(href)))
 	}
+	if ld.OwnerName != "" {
+		lockdiscovery.WriteString(fmt.Sprintf("  <d:ownername>%s</d:ownername>\n", prop.Escape(ld.OwnerName)))
+	}
+	if !ld.Locktime.IsZero() {
+		lockdiscovery.WriteString(fmt.Sprintf("  <d:locktime>%s</d:locktime>\n", prop.Escape(ld.Locktime.Format(time.RFC3339))))
+	}
+
 	lockdiscovery.WriteString("</d:activelock></d:lockdiscovery></d:prop>")
 
 	return fmt.Fprint(w, lockdiscovery.String())
--- a/internal/http/services/owncloud/ocdav/locks.go
+++ b/internal/http/services/owncloud/ocdav/locks.go
@@ -194,6 +194,7 @@ func (cls *cs3LS) Create(ctx context.Context, now time.Time, details LockDetails
 	u := ctxpkg.ContextMustGetUser(ctx)
 
 	// add metadata via opaque
+	// TODO: upate cs3api: https://github.com/cs3org/cs3apis/issues/213
 	o := utils.AppendPlainToOpaque(nil, "lockownername", u.GetDisplayName())
 	o = utils.AppendPlainToOpaque(o, "locktime", now.Format(time.RFC3339))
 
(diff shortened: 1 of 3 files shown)


glpatcern:
@kobergj FWIW the [Lock object](https://cs3org.github.io/cs3apis/#cs3.storage.provider.v1beta1.Lock) already features the owner as a `user` object, and the `expiration` as opposed to the lock time because at the time we agreed it's better to directly encode the expiration time. Are you sure you need to encode those extra properties as opaque parameters?

kobergj:
@glpatcern it only contains the `ownerID`. Web would need to make another call to get the user name. Also we need the lock time to show since when the file is locked.

Both values are only needed for the client. That's why I encoded them in the opaque. The server doesn't really need them. 

This was the simplest solution how to get this values to the client that I could come up with. I'm happy to implement another solution if you have an idea.

internal/http/services/owncloud/ocdav/locks.go
```diff
@@ -566,6 +579,13 @@ func writeLockInfo(w io.Writer, token string, ld LockDetails) (int, error) {
 	if href != "" {
 		lockdiscovery.WriteString(fmt.Sprintf("  <d:lockroot><d:href>%s</d:href></d:lockroot>\n", prop.Escape(href)))
 	}
+	if ld.OwnerName != "" {
+		lockdiscovery.WriteString(fmt.Sprintf("  <d:ownername>%s</d:ownername>\n", prop.Escape(ld.OwnerName)))
+	}
+	if !ld.Locktime.IsZero() {
+		lockdiscovery.WriteString(fmt.Sprintf("  <d:locktime>%s</d:locktime>\n", prop.Escape(ld.Locktime.Format(time.RFC3339))))
```
butonic, line 586: You cannot use the RFC DAV namespace when extending the xml. For that we need to use the owncloud namespace. Since we are writing the xml ourself, I think we can hardcode `oc`.
```suggestion
		lockdiscovery.WriteString(fmt.Sprintf("  <oc:ownername>%s</oc:ownername>\n", prop.Escape(ld.OwnerName)))
	}
	if !ld.Locktime.IsZero() {
		lockdiscovery.WriteString(fmt.Sprintf("  <oc:locktime>%s</oc:locktime>\n", prop.Escape(ld.Locktime.Format(time.RFC3339))))
```

internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -1635,6 +1635,17 @@ func activeLocks(log *zerolog.Logger, lock *provider.Lock) string {
...
+
+	if un := utils.ReadPlainFromOpaque(lock.Opaque, "lockownername"); un != "" {
+		activelocks.WriteString("<d:ownername>")
+		activelocks.WriteString(un)
+		activelocks.WriteString("</d:ownername>")
+	}
+	if lt := utils.ReadPlainFromOpaque(lock.Opaque, "locktime"); lt != "" {
+		activelocks.WriteString("<d:locktime>")
+		activelocks.WriteString(lt)
+		activelocks.WriteString("</d:locktime>")
```
butonic, line 1647:
```suggestion
		activelocks.WriteString("<oc:ownername>")
		activelocks.WriteString(un)
		activelocks.WriteString("</oc:ownername>")
	}
	if lt := utils.ReadPlainFromOpaque(lock.Opaque, "locktime"); lt != "" {
		activelocks.WriteString("<oc:locktime>")
		activelocks.WriteString(lt)
		activelocks.WriteString("</oc:locktime>")
```

butonic (changes requested):
change the xml namespace, pls

butonic:
In general I would also prefer if web would cache more information like user displaynames ... they rarely change. With the graph api we can fine tune these requests ... but not part of this PR...

internal/http/services/owncloud/ocdav/locks.go
kobergj: Good point. Changed :+1:

internal/http/services/owncloud/ocdav/propfind/propfind.go
kobergj: Good point. Changed :+1:

glpatcern:
I see, thanks. For the username I guess it's fine as it's essentially an optimization. For the lock time, I'd challenge the usefulness of that information, also considering that wopi apps often recreate locks (`UnlockAndRelock` in the WOPI API), therefore the lock time keeps being pushed forward. Same reasoning applies to the expiration of course, but IMHO it's better to know that "the lock may expire in x minutes with no activity" rather than "the lock was created x minutes ago".

internal/http/services/owncloud/ocdav/locks.go
```diff
@@ -190,11 +191,18 @@ func (cls *cs3LS) Create(ctx context.Context, now time.Time, details LockDetails
 	// see: http://www.webdav.org/specs/rfc2518.html#n-lock-tokens
 	token := uuid.New()
 
+	u := ctxpkg.ContextMustGetUser(ctx)
+
+	// add metadata via opaque
+	o := utils.AppendPlainToOpaque(nil, "lockownername", u.GetDisplayName())
```
butonic, line 197: Here you are persisting the displayname ... but that might change over time. The lock owner is already persisted by id.

Either the web ui has to look up the displayname whan showing the lock or the server has to look up the user displayname when trying to render the lock. 

```suggestion
```

internal/http/services/owncloud/ocdav/locks.go
```diff
@@ -190,11 +191,18 @@ func (cls *cs3LS) Create(ctx context.Context, now time.Time, details LockDetails
...
+	o = utils.AppendPlainToOpaque(o, "locktime", now.Format(time.RFC3339))
+
 	r := &provider.SetLockRequest{
 		Ref: details.Root,
 		Lock: &provider.Lock{
-			Type: provider.LockType_LOCK_TYPE_EXCL,
-			User: details.UserID, // no way to set an app lock? TODO maybe via the ownerxml
+			Opaque: o,
+			Type:   provider.LockType_LOCK_TYPE_EXCL,
+			User:   details.UserID, // no way to set an app lock? TODO maybe via the ownerxml
```
butonic, line 205: I think wo should write a creation time for the lock in the storage driver. A creation date might however not be supported by every storage provider, so this is optional and the web UI will need to be able to deal with missing info here . Same goes for the `ownername` property. It might be missing and the client then needs to fetch (and cache that user metadata).

butonic (changes requested):

kobergj:
Yeah, that is actually a good point. But not for me to decide unfortunately. @tbsbdr is that something we want to consider? Showing lock expiration time instead of lock creation time?

internal/http/services/owncloud/ocdav/locks.go
kobergj: That would mean 
A) The server needs to make another call on propfinds. Possibly multiple when there are different files locked.
B) The server needs to allow `GetUserByID` for non admin users. I would try to avoid this as it is leaking data.

Locks are usually short lived, so the chance the username changes while the lock is active is very low. And even if, it wouldn't bring much confusion. Hence I was going for the simplest solution.

Would you prefer option A?

internal/http/services/owncloud/ocdav/locks.go
kobergj: The client can't fetch data for a userID as a normal user is not allowed to get users by ID. Only admins can. 

Of course it is always wise for the webui to have a workaround if fields are missing. 

We create the expiration time also in the ocdav handler. So I put the creation time also here to have it consistent. As of now the storage provider just stores the lock data and doesn't add any additional information. Breaking this would mean there a two different sources of truth for the lock details: The ocdav handler AND the storage provider which is confusing.

internal/http/services/owncloud/ocdav/locks.go
butonic: There should already be a cache for A) as propfinds already have to fetch user data to fill the `oc:owner-display-name` property 👀 oh well it does not do that:
```
sublog.Debug().Msg("TODO fetch user displayname")
```

maybe we don't even need that in web any longer as the owner/manager is part of the space ... It is only in the code for web ...

B) How is a `GetUserByID` call leaking when you can already search for sharing recipients? Listing shares also has the problvem that we have to fetch user metadata data for every share. Another endpoint where I would like to omit the metadata. If you already have the userid I think it is ok to allow users to fetch the displayname, username and avatar. the same data you can fetch via the share recipients lookup.

We could even maintain a list of users the currently logged in user has recently interacted with. On the server as well as in the browser... at some point in the future...

For now, I'd go with A) as it will only cause a user lookup when a resource is locked.

kobergj: You search for recipients by name. That means you know the name already. If you just know the ID you can't get any more information on this user as of now. This is good because it is secure. But we can discuss this in the future if you want.

Are you really SURE we want to have the extra load on the server just to avoid a username that is outdated in an edge case for not more than 1 Minute? It will make the code more complicated, the logic more complicated and the put extra pressure on (already slow) propfind calls. 

I can't see any benefit in that.

internal/http/services/owncloud/ocdav/locks.go
butonic: Yes, but in contrast to the expiration time the creation time should be determined automatically when writing the lock. The Lock() call can be made by other services than ocdav, eg. the wopiserver. 

WebDAV actually sets a timeout using a timespan, eg `Second-604800` (see https://greenbytes.de/tech/webdav/draft-reschke-webdav-locking-latest.html#rfc.section.2.5.7) and that needs to be relative to the creation date (and requires two propertios to be stored). So we decided back then that calculating the expiry time and storing only one datetime made more sense.

But we do loose the creation time. So, storing the creation date and the timeout in seconds seperately allow use to capture more details. @glpatcern I think we should reconsider and add this to the next iteration of the CS3 API.

glpatcern: Well, after all the storage provider could already store all properties on `SetLock` (`creationdate = now()`, and `timeout = expirationdate - now()`). But I see you may want to expose both values when getting the Lock. Not sure how much it adds though...

tbsbdr:
Showing lock expiration time is totally fine for me 👍

tbsbdr:
Showing lock expiration time is totally fine for me 👍

internal/http/services/owncloud/ocdav/locks.go
kobergj: I added a ticket to the cs3api: https://github.com/cs3org/cs3apis/issues/213

We can do further discussion there.

butonic (approved):

PR #4136, opened 2023-08-25
Fix some tests

michielbdejong:


michielbdejong:
Actually now that i think about it, in [b71cffc](https://github.com/cs3org/reva/pull/4136/commits/b71cffc0f637be637414a132300552203edbe481) we changed the code instead of changing the mock, maybe we should have changed the mock to match what the current code does?

michielbdejong:
@MahdiBaghbani you can open a GitPod on this branch and run
```
go test -v github.com/cs3org/reva/pkg/ocm/share/repository/nextcloud
```
One thing you could try is to invert the change, so instead of making the owner & creator a UserId in the code, make it a user object in the mock!

pkg/ocm/share/repository/nextcloud/nextcloud.go
```diff
@@ -182,23 +186,27 @@ func (sm *Manager) GetShare(ctx context.Context, user *userpb.User, ref *ocm.Sha
...
 		Id: altResult.ID,
-		ResourceId: &provider.ResourceId{
-			OpaqueId:  altResult.ResourceID.OpaqueID,
-			StorageId: sm.mountID,
-		},
+		// ResourceId: &provider.ResourceId{
+		// 	OpaqueId:  altResult.ResourceID.OpaqueID,
+		// 	StorageId: sm.mountID,
+		// },
+		ResourceId: nil,
```
glpatcern, line 193: This was added to fix the access to remote resources, if you remove it the access is broken. Please revert ;)

pkg/ocm/share/repository/nextcloud/nextcloud.go
```diff
@@ -415,16 +423,19 @@ func (sm *Manager) UpdateReceivedShare(ctx context.Context, user *userpb.User, s
 			State: altResult.State,
 		}, nil
 	}
+	fmt.Printf("received json: %s\n", respBody)
+	fmt.Printf("received altResultShare: %+v\n", altResultShare)
+
```
glpatcern, line 428: That is very temporary and should be reverted too...

glpatcern (changes requested):
I guess it's better to go as @michielbdejong did initially, i.e. adapt the code to manipulate userIds as likely you don't have (nor do you need) the other metadata coming with `User`.

Nevertheless, there are two changes to be reverted.

glpatcern:
Superseded

glpatcern:
@michielbdejong @MahdiBaghbani just to keep track of today's work: eventually we reverted this because it broke the code, so you need to change the mock and keep the code. Can you prepare another PR?

PR #4147, opened 2023-09-01
Add the password policies

2403905:


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -457,23 +464,30 @@ func (h *Handler) updatePublicShare(w http.ResponseWriter, r *http.Request, shar
 	}
 
 	// Password
-	newPassword, ok := r.Form["password"]
+	_, ok = r.Form["password"]
+	newPassword := strings.TrimSpace(r.FormValue("password"))
```
kobergj, line 468: Why is it necessary to trim spaces? I wouldn't do any magic on the password the user provided

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -457,23 +464,30 @@ func (h *Handler) updatePublicShare(w http.ResponseWriter, r *http.Request, shar
...
 	if h.enforcePassword(permKey) {
-		if (!ok && !share.PasswordProtected) || (ok && len(strings.TrimSpace(newPassword[0])) == 0) {
+		if !ok && !share.PasswordProtected || ok && len(newPassword) == 0 {
 			response.WriteOCSError(w, r, response.MetaBadRequest.StatusCode, "missing required password", err)
 			return
 		}
 	}
 
+	// skip validation if the clear password scenario
+	if err := h.passwordValidator.Validate(newPassword); ok && len(newPassword) > 0 && err != nil {
```
kobergj, line 478: `len(newPassword) > 0` already implies that `ok == true` so the check for `ok` can be omited. Also same comment as above: Check `len(password) > 0` first because there is no need to validate the password if there is none.

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"unicode/utf8"
+)
+
+var defaultSpecialCharacters = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
```
kobergj, line 11: Please prefix unexported global vars with `_`. They are much easier to spot then.

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+	minUpperCaseCharacters   int
+	minDigits                int
+	minSpecialCharacters     int
+	allowedSpecialCharacters string
+	digits                   *regexp.Regexp
+	specialCharacters        *regexp.Regexp
+}
+
+func NewPasswordPolicies(minCharacters, minLowerCaseCharacters, minUpperCaseCharacters, minDigits, minSpecialCharacters int,
+	allowedSpecialCharacters string) (Validator, error) {
```
kobergj, line 29: Remove `allowedSpecialCharacters` and always use the default. I don't see a usecase for custom special characters.

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+		minSpecialCharacters:     minSpecialCharacters,
+		allowedSpecialCharacters: allowedSpecialCharacters,
+		digits:                   digits,
+		specialCharacters:        specialCharacters,
+	}, nil
+}
+
+func NewDefaultPasswordPolicies() (Validator, error) {
+	return NewPasswordPolicies(0, 0, 0, 0, 0, "")
+}
```
kobergj, line 53: Do we really need this? If so it should hold sane defaults and not `0` everywhere. But we could also remove it and rely on the caller to pass correct values.

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"unicode/utf8"
+)
+
+var defaultSpecialCharacters = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
+
+type Validator interface {
```
kobergj, line 13: exported type `Validator` should have comment or be unexported

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+	Validate(str string) error
+}
+
+type Policies struct {
```
kobergj, line 17: exported type `Policies` should have comment or be unexported

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+	minLowerCaseCharacters   int
+	minUpperCaseCharacters   int
+	minDigits                int
+	minSpecialCharacters     int
+	allowedSpecialCharacters string
+	digits                   *regexp.Regexp
+	specialCharacters        *regexp.Regexp
+}
+
+func NewPasswordPolicies(minCharacters, minLowerCaseCharacters, minUpperCaseCharacters, minDigits, minSpecialCharacters int,
```
kobergj, line 28: exported func `NewPasswordPolicies` should have comment or be unexported

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+		digits:                   digits,
+		specialCharacters:        specialCharacters,
+	}, nil
+}
+
+func NewDefaultPasswordPolicies() (Validator, error) {
+	return NewPasswordPolicies(0, 0, 0, 0, 0, "")
+}
+
+func (s Policies) Validate(str string) error {
```
kobergj, line 55: exported func `Validate` should have comment or be unexported

pkg/password/password_policies.go
```diff
@@ -0,0 +1,167 @@
...
+		return allErr
+	}
+	return nil
+}
+
+func (s Policies) validateCharacters(str string) error {
+	if s.count(str) < s.minCharacters {
+		if s.minCharacters == 1 {
+			return fmt.Errorf("at least one character is required")
+		}
```
kobergj, line 90: No need for this if case imo. The error message below is sufficient (also for other validate funcs)

pkg/password/password_policies.go
2403905: I discussed with Tobias and Michael. So we have to keep it and the escaping rules have to be documented.

kobergj: Really? Can't we try to convince them that this is a bad idea? Nobody will use it and the ones that use it will run into issues...

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
2403905: fixed.

pkg/password/password_policies.go
```diff
@@ -0,0 +1,152 @@
...
+	allowedSpecialCharacters string) (Validator, error) {
+
+	digits := regexp.MustCompile("[0-9]")
+	if len(allowedSpecialCharacters) > 0 {
+		_defaultSpecialCharacters = allowedSpecialCharacters
+	}
+	specialCharacters, err := regexp.Compile("[" + regexp.QuoteMeta(_defaultSpecialCharacters) + "]")
+	if err != nil {
+		return nil, err
+	}
```
kobergj, line 41: Please don't overwrite the default. This can lead to race conditions or misconfigurations. Do it other way around instead:
```suggestion
	if len(allowedSpecialCharacters) == 0 {
		allowedSpecialCharacters = _defaultSpecialCharacters
	}
	specialCharacters, err := regexp.Compile("[" + regexp.QuoteMeta(allowedSpecialCharacters) + "]")
	if err != nil {
		return nil, err
	}
```

kobergj (changes requested):
One more small change would be cool. Rest looks perfect :+1:

kobergj (approved):

PR #4185, opened 2023-09-12
Refurbish the grpc and http plugins for eos

ffurano:
Refurbish the grpc and http plugins for eos


cmd/reva/test.go
```diff
@@ -0,0 +1,99 @@
...
+package main
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+	"time"
+)
+
+var testCommand = func() *command {
```
gmgigi96, line 28: The reva CLI is meant to be used by end users, so I would keep this test cmd out. Maybe in a `benchmark` folder, where later we can add more benchmarks

cmd/reva/test.go
```diff
@@ -0,0 +1,99 @@
...
+			return nil
+		}
+
+		elapsedmkdir := time.Since(start)
+
+		start = time.Now()
+
+		for i := 0; i < 1000; i++ {
+
+			b, err := executeCommand(uploadCommand(), "-protocol", "simple", "/tmp/1MFile", "/home/testperf/file-"+strconv.FormatInt(int64(i), 10))
```
gmgigi96, line 50: go already provides a way for running benchmarks https://pkg.go.dev/testing#hdr-Benchmarks

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -16,6 +16,11 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
+// NOTE: compile the grpc proto with these commands
+// and do not ask any questions, I don't have the answer
+// protoc ./Rpc.proto --go-grpc_out=.
```
gmgigi96, line 21: we can add these commands in the makefile

gmgigi96 (changes requested):

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -650,7 +666,7 @@ func (c *Client) GetFileInfoByPath(ctx context.Context, auth eosclient.Authoriza
 	log.Info().Str("func", "GetFileInfoByPath").Str("uid,gid", auth.Role.UID+","+auth.Role.GID).Str("path", path).Msg("")
 
 	// Initialize the common fields of the MDReq
-	mdrq, err := c.initMDRequest(ctx, auth)
+	mdrq, err := c.initMDRequest(context.Background(), auth)
```
gmgigi96, line 669: I think would be better if here we still create an empty context but with the (eventual) cancel signal from the `ctx` passed as parameters.
In this way, if the caller cancel the context (for a timeout, CTRL-C, etc...), the grpc call to EOS will be cancelled.

I'm not aware of something similar in the standard library, so we should define it in reva (in `ctxpkg`).

gmgigi96 (changes requested):

pkg/eosclient/eosgrpc/eosgrpc.go
glpatcern: Actually this is not to be executed at every build, rather it stays "frozen" until we sync with eos

labkode:
What is the issue of re-using existing context?

glpatcern:
For the test command, true that we want to expose the CLI at some point but at that point we can comment this `test` out as it's useful only for some QA tests now. If we really think it can be useful in the future, then we move it somewhere so that it is not accessible to users.

@gmgigi96 can you approve?

glpatcern:
> What is the issue of re-using existing context?

I'm with @ffurano to understand this - apparently the GRPC client for eos needs an empty context and breaks if we reuse the incoming reva-created `ctx`. We suspect this is due to the eos protobuf and the reva protobuf being _different_ (in version?), as when we instantiate a GRPC client in reva to talk to another microservice we never have this issue.

pkg/eosclient/eosgrpc/eosgrpc.go
ffurano: Hi gmgigi,

 I don't really know the details of what you propose. Would you like to sit together and have a look at this?

glpatcern (dismissed):

Makefile
```diff
@@ -43,6 +43,10 @@ BUILD_FLAGS	= "`[[ -z "$(STATIC)" ]] && echo "" || echo "-extldflags=-static"` -
 revad:
 	go build -ldflags $(BUILD_FLAGS) -o ./cmd/revad/revad ./cmd/revad/main
 
+cernbox-revad:
+	go install github.com/cs3org/gaia@latest
+	gaia build -v --with github.com/cernbox/reva-plugins/group=github.com/cernbox/reva-plugins/group@38d2a9c --with github.com/cs3org/reva=/root/Park/reva/reva --with github.com/cernbox/reva-plugins -o ./cmd/revad/revad
+
```
gmgigi96, line 49:
```suggestion
cernbox-revad:
	go install github.com/cs3org/gaia@latest
	gaia build -v --with github.com/cernbox/reva-plugins/group=github.com/cernbox/reva-plugins/group@38d2a9c --with github.com/cs3org/reva=/root/Park/reva/reva --with github.com/cernbox/reva-plugins -o ./cmd/revad/revad

.PHONY: gaia
gaia:
	go install github.com/cs3org/gaia@latest

.PHONY: cernbox-revad
cernbox-revad: gaia
	gaia build --with github.com/cernbox/reva-plugins --with github.com/cs3org/reva=$(shell pwd) -o ./cmd/revad/revad
```

PR #4193, opened 2023-09-18
Fix the error when the user trying to overwrite a file with an empty file

2403905:


internal/http/services/owncloud/ocdav/put.go
```diff
@@ -172,32 +172,35 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+				Ref: ref,
+			})
+			if err != nil {
+				log.Error().Err(err).Msg("error sending grpc touch file request")
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			if sRes.Status.Code != rpc.Code_CODE_OK {
+				log.Error().Interface("status", sRes.Status).Msg("error touching file")
+				w.WriteHeader(http.StatusInternalServerError)
```
butonic, line 186: urgh ... there are so many places where we should return a proper error using `HandleErrorStatus` and not just dumbly fall back to internal server error and I hate to bother you with this, but can you try to use `HandleErrorStatus` here?

2403905: Thanks. Changed.

butonic (approved):

PR #4199, opened 2023-09-21
Dynamic storage router

javfg:
This PR adds a dynamic storage router. It uses a db table as routing table (with `path` - `mountid` pairs), and configuration to map the mountids to different provider addresses.

TODO: provide a config example

### DIFF
--- a/pkg/storage/registry/dynamic/dynamic.go
+++ b/pkg/storage/registry/dynamic/dynamic.go
@@ -89,13 +89,13 @@ func New(ctx context.Context, m map[string]interface{}) (storage.Registry, error
 	return d, nil
 }
 
-func initRoutingTree(dbusername, dbpassword, dbhost string, dbport int, dbname string) (*routingtree.RoutingTree, error) {
-	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s:%d)/%s", dbusername, dbpassword, dbhost, dbport, dbname))
+func initRoutingTree(dbUsername, dbPassword, dbHost string, dbPort int, dbName string) (*routingtree.RoutingTree, error) {
+	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s:%d)/%s", dbUsername, dbPassword, dbHost, dbPort, dbName))
 	if err != nil {
 		return nil, errors.Wrap(err, "error opening sql connection")
 	}
 
-	results, err := db.Query("SELECT path, mount_id, mount_type FROM routing")
+	results, err := db.Query("SELECT path, mount_id FROM routing")
 	if err != nil {
 		return nil, errors.Wrap(err, "error getting routing table from db")
 	}
@@ -104,7 +104,7 @@ func initRoutingTree(dbusername, dbpassword, dbhost string, dbport int, dbname s
 
 	for results.Next() {
 		var r routingtree.Route
-		err = results.Scan(&r.Path, &r.MountID, &r.MountType)
+		err = results.Scan(&r.Path, &r.MountID)
 		if err != nil {
 			return nil, errors.Wrap(err, "error scanning rows from db")
 		}
--- a/pkg/storage/registry/dynamic/dynamic.go
+++ b/pkg/storage/registry/dynamic/dynamic.go
@@ -1,0 +1,173 @@
+// Copyright 2018-2023 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+// Package dynamic contains the dynamic storage registry
+package dynamic
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	registrypb "github.com/cs3org/go-cs3apis/cs3/storage/registry/v1beta1"
+	"github.com/cs3org/reva/pkg/appctx"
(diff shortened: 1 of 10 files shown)


pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// AddRoute adds a new Route to the RoutingTree based on a path `p`.
+func (t *RoutingTree) AddRoute(r Route) {
+	parts := getAncestors(r.Path)
+	current := t
+
+	for _, path := range parts {
+		newNode := Route{
+			Path: path,
```
gmgigi96, line 94: Supposing that the `RoutingTree` is storing the paths `/eos/user/g/gdelmont`, `/eos/project/c/cernbox`, `/cephfs/project/c/cernbox`, I would expect it to be in memory:
```
/
├── cephfs
│   └── project
│       └── c
│           └── cernbox
└── eos
    ├── project
    │   └── c
    │       └── cernbox
    └── user
        └── g
            └── gdelmont
```

instead of:

```
/
├── /cephfs
│   └── /cephfs/project
│       └── /cephfs/project/c
│           └── /cephfs/project/c/cernbox
└── /eos
    ├── /eos/project
    │   └── /eos/project/c
    │       └── /eos/project/c/cernbox
    └── /eos/user
        └── /eos/user/g
            └── /eos/user/g/gdelmont
```

The first I think is easier and you don't need to compute each time the parts from the `getAncestors` function for the lookup.

pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,174 @@
...
+	leaves := subtree.getLeaves()
+	providers := []string{}
+	providerMap := make(map[string]bool)
+
+	for _, l := range leaves {
+		providerMap[l.route.MountID] = true
+	}
+
+	for p := range providerMap {
+		providers = append(providers, p)
```
gmgigi96, line 170: this path should be cleaned as well

pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,174 @@
...
+	return current, nil
+}
+
+func (t *RoutingTree) getLeaves() []*RoutingTree {
+	if len(t.nodes) == 0 {
+		return []*RoutingTree{t}
+	}
+
+	leafNodes := []*RoutingTree{}
+	queue := []*RoutingTree{t}
```
gmgigi96, line 136: you're re-implementing what the compiler is doing for a recursive function ;)
I would keep it recursive so that is clearer and you avoid to make some many allocation/deallocations for the heap.

pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// GetProviders returns a list of providers for a given path.
+func (t *RoutingTree) GetProviders(p string) ([]string, error) {
+	subtree, err := t.findNodeForPath(path.Clean(p))
+	if err != nil || subtree == nil {
+		return nil, err
+	}
+
+	leaves := subtree.getLeaves()
```
gmgigi96, line 161: Instead of getting all the leaves (these can be a lot, causing lot of allocations for the slice), you can get just the unique mount ids.

gmgigi96 (changes requested):

pkg/storage/registry/dynamic/dynamic.go
```diff
@@ -0,0 +1,174 @@
...
+			ProviderPath: p,
+			Address:      a,
+		})
+	}
+
+	return providers, nil
+}
+
+// GetHome returns the storage provider for the home path.
+func (d *dynamic) GetHome(ctx context.Context) (*registrypb.ProviderInfo, error) {
```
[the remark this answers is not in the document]
javfg: As discussed, you are right! This is already there (it already resolves aliases in `FindProviders`), but the rewrite was missing in `GetHome`. I've added it! :+1:

pkg/storage/registry/dynamic/dynamic.go
```diff
@@ -0,0 +1,174 @@
...
+			ProviderPath: d.c.HomePath,
+			Address:      a,
+		}, nil
+	}
+
+	return nil, errors.New("home not found")
+}
+
+// FindProviders returns the storage providers for a given ref.
+func (d *dynamic) FindProviders(ctx context.Context, ref *provider.Reference) ([]*registrypb.ProviderInfo, error) {
```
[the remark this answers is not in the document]
javfg: It should now check if the `storageID` is there and return the provider address from the config mapping. Going with the path resolution will only happen otherwise.

pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,150 @@
...
+			newNode.MountID = mountID
+			newNode.MountPath = route
+		}
+
+		current = current.addNode(newNode)
+	}
+}
+
+func (t *RoutingTree) findRoute(p string) (*RoutingTree, error) {
+	parts := strings.Split(path.Clean(p), "/")
```
gmgigi96, line 93: Use https://github.com/cs3org/reva/blob/d36bcd56dd658a9083495df6c3cae3a129812921/pkg/rhttp/router/router.go#L31 to navigate the path instead of creating a slice with all the tokens

gmgigi96 (changes requested):

pkg/storage/registry/dynamic/routingtree/routingtree.go
javfg: Done!

pkg/storage/registry/dynamic/routingtree/routingtree.go
```diff
@@ -0,0 +1,152 @@
...
+		if remainder == "/" {
+			newNode.MountID = mountID
+			newNode.MountPath = route
+		}
+
+		current = current.addNode(newNode)
+	}
+}
+
+func (t *RoutingTree) findRoute(p string) (*RoutingTree, error) {
```
gmgigi96, line 91:
```suggestion
func (t *RoutingTree) findRoute(p string) (*RoutingTree, error) {
    p = path.Clean(p)
```

gmgigi96 (changes requested):

pkg/storage/registry/dynamic/routingtree/routingtree.go
javfg: `ShiftPath` does a  `path.Clean` inside inconditionally (unless it is called with an empty string). Not sure we need this.

gmgigi96 (approved):

PR #4200, opened 2023-09-21
Ceph driver fixes

javfg:
This PR includes multiple fixes to bring the Ceph driver back to speed.


glpatcern (dismissed):
I think it's all right (the `MakeDir` without `Stat` was not really an issue IMHO but fine)

javfg:
> I think it's all right (the `MakeDir` without `Stat` was not really an issue IMHO but fine)

We're not sure why but it was causing the machine to freeze for some time retrying the mkdir over and over again.

pkg/storage/fs/cephfs/connections.go
```diff
@@ -63,7 +63,7 @@ func newCache() (c *connections, err error) {
 		MaxCost:     usrLimit,
 		BufferItems: 64,
 		OnEvict: func(item *ristretto.Item) {
-			v := item.Value.(cacheVal)
+			v := item.Value.(*cacheVal)
```
gmgigi96, line 66: I would still check that the type is correct with the notation `v, ok := item.Value.(*cacheVal)` otherwise in production it may panic.

gmgigi96 (changes requested):

PR #4214, opened 2023-09-25
pass default nats options

butonic:
The nats option passed completely replace the default nats options, so we need to pass in adjusted default options.


butonic:
fixes https://github.com/owncloud/ocis/issues/7056#issuecomment-1733650654

kobergj (approved):
Nice find :+1:

butonic:
verified this works in kubernetes as expected as per https://github.com/owncloud/ocis/issues/7056#issuecomment-1702490512

PR #4244, opened 2023-10-10
Allow listing received Shares by service accounts

kobergj:
Allows listing received shares by service account users

### DIFF
--- a/pkg/share/manager/memory/memory.go
+++ b/pkg/share/manager/memory/memory.go
@@ -20,6 +20,7 @@ package memory
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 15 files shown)


pkg/share/manager/memory/memory.go
```diff
@@ -282,11 +283,20 @@ func (m *manager) ListShares(ctx context.Context, filters []*collaboration.Filte
...
 // we list the shares that are targeted to the user in context or to the user groups.
-func (m *manager) ListReceivedShares(ctx context.Context, filters []*collaboration.Filter) ([]*collaboration.ReceivedShare, error) {
+func (m *manager) ListReceivedShares(ctx context.Context, filters []*collaboration.Filter, forUser *userv1beta1.UserId) ([]*collaboration.ReceivedShare, error) {
 	var rss []*collaboration.ReceivedShare
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	user := ctxpkg.ContextMustGetUser(ctx)
+	if user.GetId().GetType() == userv1beta1.UserType_USER_TYPE_SERVICE {
+		// TODO: gateway missing!
+		// u, err := utils.GetUser(forUser, nil)
```
aduffeck, line 293: Do you think the commented code is helpful? I think I would rather leave it out, it's not that hard to figure it out when actually doing the refactoring.

aduffeck (approved):
lgtm besides the one nit-pick.

PR #4267, opened 2023-10-18
Fix concurrency issue when listing received shares

aduffeck:
This PR fixes a concurrency issue when listing received shares.

### DIFF
--- a/pkg/share/manager/jsoncs3/receivedsharecache/receivedsharecache.go
+++ b/pkg/share/manager/jsoncs3/receivedsharecache/receivedsharecache.go
@@ -177,17 +177,6 @@ func (c *Cache) Get(ctx context.Context, userID, spaceID, shareID string) (*Stat
 	return c.ReceivedSpaces[userID].Spaces[spaceID].States[shareID], nil
 }
 
-// Sync updates the in-memory data with the data from the storage if it is outdated
-func (c *Cache) Sync(ctx context.Context, userID string) error {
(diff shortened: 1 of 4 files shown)


kobergj (approved):

pkg/share/manager/jsoncs3/receivedsharecache/receivedsharecache.go
```diff
@@ -177,15 +177,35 @@ func (c *Cache) Get(ctx context.Context, userID, spaceID, shareID string) (*Stat
...
 	defer unlock()
 
-	return c.syncWithLock(ctx, userID)
+	err := c.syncWithLock(ctx, userID)
+	if err != nil {
+		return nil, err
+	}
+
+	spaces := map[string]*Space{}
+	for spaceID, space := range c.ReceivedSpaces[userID].Spaces {
```
kobergj, line 195: What happens if `userID` is not in `c.ReceivedSpaces`? nilpointer dereference?

kobergj (changes requested):

kobergj (approved):

wkloucek:
@aduffeck is there a issue / other source that describes the symptoms? I know a few support cases that are related to share creation / listing and would like to know if one of them is fixed by this PR.

aduffeck:
@wkloucek no, I don't think there's an issue for that. The problem was that the CI pipelines started to flake after turning on the auto-acceptance of shares. Retrieving the list of received shares while also creating shares for that user could run into concurrency issues and panic. Not sure how likely it is to trigger that in real-world scenarios...

PR #4268, opened 2023-10-18
implement sharing roles

2403905:
## Description

Two endpoints that map roles have been added 
https://owncloud.dev/libre-graph-api/#/roleManagement

Redundant roles RoleUnknown,  RoleLegacy, RoleDenied, RoleNone that have no permissions have been skipped.

## Related Issue

- [https://github.com/owncloud/ocis/issues/7418](https://github.com/owncloud/ocis/issues/7418)
- spec here: https://github.com/owncloud/libre-graph-api/pull/117


rhafer:
Same as with the ocis pull request. Having some more context about what this actually is supposed to do and why, would be highly appreciated.

kobergj (approved):

PR #4271, opened 2023-10-20
enhancement: add search capability

fschade:
add the ability to define search specific capabilities needed for https://github.com/owncloud/ocis/pull/7455


2403905 (approved):

fschade:
not a fan of []map[string]interface{}, should we static type search properties? cc.: @2403905

2403905 (approved):

PR #4278, opened 2023-10-23
[full-ci] disable DEPTH infinity in PROPFIND

2403905:
Issue: [https://github.com/owncloud/ocis/issues/7359](https://github.com/owncloud/ocis/issues/7359)

Add support for disabling the DEPTH infinity in PROPFIND for
Personal /remote.php/dav/files/admin
Public link share /remote.php/dav/public-files/<token>
Trashbin /remote.php/dav/spaces/trash-bin/<personal-space-id>

PR #4279, opened 2023-10-24
disable DEPTH infinity in PROPFIND without bin

2403905:


2403905:
It is not relevant because of other one has been merged  https://github.com/cs3org/reva/pull/4278

PR #4284, opened 2023-10-25
Disallow moves

aduffeck:
We no longer allow moves between shares, even if they resolve to the same space.

### DIFF
--- a/internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
+++ b/internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
@@ -591,12 +591,6 @@ func (s *service) Move(ctx context.Context, req *provider.MoveRequest) (*provide
 		Interface("destination", req.Destination).
 		Msg("sharesstorageprovider: Got Move request")
 
-	if !utils.ResourceIDEqual(req.Source.ResourceId, req.Destination.ResourceId) {
-		return &provider.MoveResponse{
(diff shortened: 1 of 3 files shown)


internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
```diff
@@ -591,6 +591,12 @@ func (s *service) Move(ctx context.Context, req *provider.MoveRequest) (*provide
 		Interface("destination", req.Destination).
 		Msg("sharesstorageprovider: Got Move request")
 
+	if !utils.ResourceIDEqual(req.Source.ResourceId, req.Destination.ResourceId) {
```
kobergj, line 594: This implies that `Source` and `Destination` are relative references, right? Can't I move them by specific ids only? Wouldn't this fail then?

kobergj (changes requested):

internal/grpc/services/sharesstorageprovider/sharesstorageprovider.go
aduffeck: That's correct. The sharesstorageprovider works with references where the storage and space ids are hardcoded values and the opaque id is the encoded resource id of the shared resource (in the form `<storageid>:<spaceid>:<opaqueid>`. The actual referenced resource is selected by path, so it's always gonna be a relative reference.

kobergj (approved):

kobergj (approved):

PR #4287, opened 2023-10-26
Fixes registration and naming of services

gmgigi96:


glpatcern (commented):
Great for the `ocm` service. Otherwise, are we sure we want to make the share SQL driver a plugin, and leave in the core only the memory and JSON drivers, none of them good for a production service?

gmgigi96:
> Great for the `ocm` service. Otherwise, are we sure we want to make the share SQL driver a plugin, and leave in the core only the memory and JSON drivers, none of them good for a production service?

The `sql` plugin I have deleted is something called `oc10-sql`, and it's not the one we user for CERNBox. I wanted to include upstream the CERNBox one, but it's still to CERN specific, so for now better keep it as a plugin.

glpatcern (approved):

PR #4288, opened 2023-10-26
Print plugin version

gmgigi96:
Now on `revad -plugins`, the list reports the version of the plugins.
Example:

```
[grpc.services.groupprovider.drivers]
rest -> github.com/cernbox/reva-plugins/group (v0.0.4)
```


labkode:
@gmgigi96 can you add a sort before printing the plugins? So it's easy to compare different runs?
the map is not deterministic

labkode (approved):

PR #4291, opened 2023-10-27
Release expired locks when statting

kobergj:
When statting a locked resource, release the lock if it is expired. See details: https://github.com/owncloud/ocis/issues/7561


micbar (approved):
elegant 😄

PR #4292, opened 2023-10-27
Return 403 when user is not permitted to lock

kobergj:
Returns `403` instead `500` when a user tries to lock a file they have no write access to. See details: https://github.com/owncloud/ocis/issues/7600


micbar (approved):
Awesome!

PR #4296, opened 2023-10-27
Added listversions command, plus various git shuffling fun

ffurano:


gmgigi96:
Can you rebase on master?

pkg/eosclient/eosgrpc/eoshttp.go
```diff
@@ -390,9 +390,6 @@ func (c *EOSHTTPClient) PUTFile(ctx context.Context, remoteuser string, auth eos
 		log.Debug().Str("func", "PUTFile").Msg("sending req")
 
 		resp, err := c.doReq(req, remoteuser)
```
gmgigi96, line 392: You should check the error `err` here. If the error is nil, then the `resp.Body` is guaranteed by the stdlib to be non-nil, and you can defer the close of the body without checking for it to be non-nil. If you do this, you can remove all the manual body close below.

gmgigi96 (changes requested):

PR #4299, opened 2023-10-30
Fix http verb

aduffeck:
We changed the HTTP verb of the /generate-invite endpoint of the sciencemesh service to POST as it clearly has side effects for the system, it's not just a read-only call.

### DIFF
--- a/changelog/unreleased/fix-http-verb.md
+++ b/changelog/unreleased/fix-http-verb.md
@@ -1,0 +1,7 @@
+Bugfix: Fix HTTP verb of the generate-invite endpoint
+
+We changed the HTTP verb of the /generate-invite endpoint of the sciencemesh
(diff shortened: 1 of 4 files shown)


kobergj (approved):

glpatcern:
The change surely makes sense (the original action was meant to generate a token without storing it), yet this endpoint is consumed by the ScienceMesh web app, so such change needs to be coordinated with the frontend - and eventually propagated to Reva master.

What are your thoughts concerning the ScienceMesh web app?

gmgigi96:
Also, the Nextcloud/OC10 app needs to be update accordingly to use the new verb

aduffeck:
I will talk to the web team regarding their plans and how to adapt to this change. I'll go ahead and merge this nonetheless for now in order to get a clean start in edge if possible.

PR #4302, opened 2023-10-30
Fix filenamelength checking

kobergj:
Fixes a bug where the ocdav service would check for filepath length instead of filename length

Details: https://github.com/owncloud/ocis/issues/7610


changelog/unreleased/fix-filename-length-check.md
```diff
@@ -0,0 +1,5 @@
+Bugfix: Fix checking of filename length
+
+Instead of checking for length of the filename the ocs handler would sometimes check for complete file path.
```
micbar, line 3:
```suggestion
Instead of checking for length of the filename the ocdav handler would sometimes check for complete file path.
```

PR #4325, opened 2023-11-09
Enforce Favorite&Share Permissions

kobergj:
See https://github.com/owncloud/ocis/issues/7497 for details


internal/http/services/owncloud/ocdav/proppatch.go
```diff
@@ -217,6 +218,17 @@ func (s *svc) handleProppatch(ctx context.Context, w http.ResponseWriter, r *htt
 						return nil, nil, false
 					}
 					currentUser := ctxpkg.ContextMustGetUser(ctx)
+					ok, err := utils.CheckPermission(ctx, "Favorites.Write", client)
```
2403905, line 221: Can we use the constant WriteFavorites from a pkg/permission/permission.go instead of the "Favorites.Write"?

2403905: utils.CheckPermission(ctx, permission.WriteFavorites, client)

kobergj: Excellent idea :+1: I did it for other occurrences of `CheckPermission` too. Please double check.

2403905 (approved):

PR #4333, opened 2023-11-13
Ocm fixes

aduffeck:
Fix OCM issues that appeared while working with the web UI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -230,3 +230,5 @@ require (
 )
 
 replace github.com/go-micro/plugins/v4/store/nats-js-kv => github.com/kobergj/plugins/v4/store/nats-js-kv v0.0.0-20231207143248-4d424e3ae348
+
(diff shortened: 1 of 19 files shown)


go.mod
```diff
@@ -230,3 +230,5 @@ require (
 )
 
 replace github.com/go-micro/plugins/v4/store/nats-js-kv => github.com/kobergj/plugins/v4/store/nats-js-kv v0.0.0-20231207143248-4d424e3ae348
+
+replace github.com/studio-b12/gowebdav => github.com/aduffeck/gowebdav v0.0.0-20231123085457-ff658b6ea159
```
butonic, line 234: can you point to an upstream PR with the diff?

butonic (approved):

go.mod
aduffeck: As discussed in person we'll need to extend the changes to our fork, so no, there's no upstream PR yet.

PR #4336, opened 2023-11-14
bugfix: Role "file-editor" does not have resharing permissions

rhafer:
Similar to the generich "editor" role the "file-editor" should have resharing  permissions.
    
When creating an "editor" file share using `ocs`. `ocs` already adds the `AddGrant` permission. This commit just about fixing the mapping of resource permission back to the corresponding role.


micbar:
The bugfix is clear. Why do you remove the unified roles?

rhafer:
> Why do you remove the unified roles?

@micbar As mentioned in the commit message on the first commit, I think libregraph specific stuff should not be part of reva so they'd better live with the graph service. See https://github.com/owncloud/ocis/pull/7703/commits/9e50c42d83252115af6f1ac12f1577c9e97b6e45 . I am fine revisiting that if people disagree.

micbar (approved):

PR #4340, opened 2023-11-14
fix: role permissions and add listGrants to the viewer and editor roles

fschade:
Add listGrants permission to the viewer and editor roles


fschade:
on hold for now, needs clarification

fschade:
closed, the failure happens is in RoleFromOCSPermissions which currently has no knowledge if the role is for a project or personal space. Fix pending, cc.: @rhafer

PR #4342, opened 2023-11-15
Fix conversion of custom ocs permissions to roles

rhafer:
When creating shares with just `view` permission we wrongly converted that into the `SpacerViewer` sharing role. The correct role for that would be `legacy`.

Fixes: https://github.com/owncloud/enterprise/issues/6209


micbar (approved):

pkg/conversions/role.go
```diff
@@ -379,7 +379,7 @@ func NewManagerRole() *Role {
 // RoleFromOCSPermissions tries to map ocs permissions to a role
 // TODO: rethink using this. ocs permissions cannot be assigned 1:1 to roles
 // NOTE: If resharing=false in the system this function will return SpaceViewerRole instead ViewerRole
```
fschade, line 381: obsolete

fschade (approved):
LGTM 🎸

pkg/conversions/role.go
rhafer: fixed

PR #4343, opened 2023-11-15
Another fix for converting custom ocs permissions to roles

rhafer:
When creating (public) shares with the permissions read,update,create,delete we wrongly converted that into the `SpacerEditor` sharing role. The correct role for that would be `legacy`.

Fixes: https://github.com/owncloud/enterprise/issues/6209

### DIFF
--- a/changelog/unreleased/fix-roleconversion-custompermissions.md
+++ b/changelog/unreleased/fix-roleconversion-custompermissions.md
@@ -1,7 +1,8 @@
 Bugfix: Fix conversion of custom ocs permissions to roles
 
-When creating shares with just `view` permission we wrongly converted that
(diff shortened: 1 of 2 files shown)


changelog/unreleased/fix-roleconversion-custompermissions.md
```diff
@@ -4,4 +4,5 @@ When creating shares with just `view` permission we wrongly converted that
 into the `SpacerViewer` sharing role. The correct role for that would be `legacy`.
 
 https://github.com/cs3org/reva/pull/4342
+https://github.com/cs3org/reva/pull/4343
```
micbar, line 7: Changelog check fails because 4343 needs to be on top of the list.

micbar (approved):

kobergj (approved):

PR #4344, opened 2023-11-18
Add url extension to mime-type list

JanAckermann:
Adds url extension to the mime-type list.
This is necessary for owncloud/web to detect the correct mime type and handle url files correctly, e.G opens in a new browser.

We store url files in the following format: 
https://fileinfo.com/extension/url

Thus there is no real specification about it nor a well-fitting RFC specified mime type it's **up for discussion** which mime type fits the best, possible alternatives could be:
* text/url
* application/internet-shortcut


fschade (approved):
🤗 🚀

PR #4360, opened 2023-11-23
[full-ci] mountpoint spaces are always listed in the drives list

2403905:


2403905:
It is not relevant because the ticket has been closed

PR #4363, opened 2023-11-27
Add `nats-js-kv` store

kobergj:
Adds a new store type `nats-js-kv`. See respective go-micro PR for details: https://github.com/go-micro/plugins/pull/132 

Also fixes a panic when an error occurs during locking


pkg/store/store.go
```diff
@@ -126,6 +129,16 @@ func Create(opts ...microstore.Option) microstore.Store {
 				natsjs.NatsOptions(natsOptions), // always pass in properly initialized default nats options
 				natsjs.DefaultTTL(ttl))...,
 		) // TODO test with ocis nats
+	case TypeNatsJSKV:
+		// NOTE: nats needs a DefaultTTL option as it does not support per Write TTL ...
+		ttl, _ := options.Context.Value(ttlContextKey{}).(time.Duration)
+		natsOptions := nats.GetDefaultOptions()
+		natsOptions.Name = "TODO" // we can pass in the service name to allow identifying the client, but that requires adding a custom context option
```
butonic, line 136: TODO?

butonic (changes requested):
the go version bump requires touching more places and AFAICT there is a todo?

pkg/store/store.go
kobergj: copied from the natsjs store. The point is we would need another envvar `MICRO_STORE_NAME` or similar. Not sure if this is needed and/or wanted. Left it in so we do not forget. I can remove it if you want?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cs3org/reva/v2
 
-go 1.20
+go 1.21
```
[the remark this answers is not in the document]
kobergj: :+1: Correct. The ceph docker file needed the go bump also.

pkg/store/store.go
butonic: Nah, I'm ok with that ;-)

PR #4365, opened 2023-11-27
fix: public link creation err return if not ok

fschade:
If public link creation failed, it now returns an error instead of continuing the process execution.

### DIFF
--- a/internal/grpc/services/publicshareprovider/publicshareprovider.go
+++ b/internal/grpc/services/publicshareprovider/publicshareprovider.go
@@ -24,6 +24,10 @@ import (
 
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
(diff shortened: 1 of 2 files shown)


micbar (approved):

internal/grpc/services/publicshareprovider/publicshareprovider.go
```diff
@@ -166,6 +167,7 @@ func (s *service) CreatePublicShare(ctx context.Context, req *link.CreatePublicS
 	share, err := s.sm.CreatePublicShare(ctx, u, req.ResourceInfo, req.Grant)
 	if err != nil {
 		log.Debug().Err(err).Str("createShare", "shares").Msg("error connecting to storage provider")
+		return nil, err
```
rhafer, line 170: Hm, this looks fishy. Shouldn't the error be returned as a non-OK status (+Message) in a proper CreatePublicShareResponse, here? (See e.g. link 186)
 
Also the log message seems at least missleading. How can we be sure that this is a connection issue?

rhafer (changes requested):

PR #4369, opened 2023-11-28
eosfs: Workaround assumptions on stat/xattrs auth

ffurano:


pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1197,18 +1220,16 @@ func (fs *eosfs) GetMD(ctx context.Context, ref *provider.Reference, mdKeys []st
 		return nil, err
 	}
 
-	fn := ""
+	// TODO (lopresti) do we need this?
```
gmgigi96, line 1223: This block can be removed as the path is not needed anymore for the auth of lw accounts

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1254,13 +1275,12 @@ func (fs *eosfs) GetMD(ctx context.Context, ref *provider.Reference, mdKeys []st
 func (fs *eosfs) getMDShareFolder(ctx context.Context, p string, mdKeys []string) (*provider.ResourceInfo, error) {
 	fn := fs.wrapShadow(ctx, p)
 
-	u, err := getUser(ctx)
+	_, err := getUser(ctx)
```
gmgigi96, line 1278: remove this, as it's not needed anymore

gmgigi96 (changes requested):

pkg/storage/utils/eosfs/eosfs.go
glpatcern: This is what I thought too, and meant to ask you ;)

glpatcern (approved):

Issue #4334, opened 2023-11-13
dynamic routing issue when DB contains non-resolvable elements


labkode:
Example:

The storage provider was not configured in the gateway dynamic table.

By removing this, web worked:

```
MySQL [cernboxngcopy]> delete from routing where path="/eos/web";
```

PR #4370, opened 2023-11-28
Storage registry: fail at init if config is missing any providers

javfg:
This PR changes the behavior of the dynamic storage registry so it fails on initialization if the configuration is not complete (there are storage provider entries in the database with no entry in the config file)

### DIFF
--- a/changelog/unreleased/storage-registry-initfail.md
+++ b/changelog/unreleased/storage-registry-initfail.md
@@ -1,0 +1,8 @@
+Enhancement: Storage registry: fail at init if config is missing any providers
+
+This change makes the dynamic storage registry fail at startup if there are
+missing rules in the config file. That is, any `mount_id` in the routing table
+must have a corresponding `storage_id`/`address` pair in the config, otherwise
+the registry will fail to start.
+
+https://github.com/cs3org/reva/pull/4370
--- a/pkg/ocm/share/repository/sql/sql_test.go
+++ b/pkg/ocm/share/repository/sql/sql_test.go
@@ -47,7 +47,7 @@ import (
 var (
 	dbName               = "reva_tests"
 	address              = "localhost"
-	port                 = 3306
+	port                 = 33059
 	m                    sync.Mutex // for increasing the port
 	ocmShareTable        = "ocm_shares"
 	ocmAccessMethodTable = "ocm_shares_access_methods"
--- a/pkg/storage/registry/dynamic/dynamic.go
+++ b/pkg/storage/registry/dynamic/dynamic.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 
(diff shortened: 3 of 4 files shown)


glpatcern:
Thanks Javi, but I'd rather implement a more tolerant behavior, that is the missing route would be discarded (with a log warning) without preventing reva from starting.

The context I see here is at least for our own QA: we share the same database, but we may enable/disable some storage providers (e.g. the eosatlas one). Also, if a new route has to be added, we can easily decouple the db operation from the reconfiguration - until the storage provider appears, the new route remains hidden.

glpatcern:
This closes https://github.com/cs3org/reva/issues/4334

labkode:
@glpatcern I've discussed that option with @javfg, but that will not be robust enough. nd the fault tolerant one is just a side effect of our internal deployment.

This PR follows the same principles as other drivers, abort at init time if the runtime of the logic cannot be guaranteed at runtime.

javfg:
I understand flexibility is a good idea, in fact it was my first direction (as we discussed). But then we're presented with another problem:

The router will not differentiate in any way an incorrect provider from a provider which is not present in the config. Whether you try to resolve `eoshome-i04` (this being something which should exist but is missing because of a misconfiguration), or `somethingthatdoesntexist` (which will never be there), you will still get the same result: a nil storage provider.

This is bound to bring headaches in the future. :) Also, I think the correct behavior of a router would be to return errors when something is not routable.

But then again, this is very nuanced and I understand both sides. I'll agree with whatever you decide.

glpatcern:
I'm fine with the router not being able to distinguish between a genuine misconfiguration and a "deliberate" one (and temporary, typically - see my example of introducing a new route in production).

Anyway, what happens in the scenario above, i.e. entry in config but missing entry in db?

javfg:
Nothing, the rule just lays there and will never be used.

I imagined this part similar to a DNS, following the router idea. So while it is OK to have extra records there, it's not equally fine to have a missing record when you want to resolve a storage id into an address.

The downside is all entries in the table will need an entry in the config in all instances.

glpatcern:
Well, I understand the model but here we are deliberately saying that the DB is more important than the config, and that the system is not symmetric.

To me, this is as good (or as bad!) as having it symmetric and tolerant. Clearly the option of being symmetric and not tolerant is a no-go for operational reasons, but the other two options (not symmetric vs symmetric and tolerant) are at least equally legitimate. In particular I yet fail to see any real advantage of this non-symmetric approach.

labkode (approved):

labkode:
@glpatcern changes can always be introduced in the DB before they are in the config. Only on restart the changes need to be there, which is a good to catch for forgotten configs.

glpatcern:
> which is a good catch for forgotten configs

And it's an excellent time bomb should a daemon restart be kicked inadvertently...

OK, I won't complain further as you merged it, but IMHO this will hit us in the future.

PR #4384, opened 2023-12-04
Add outcome and event for retrying postprocessing steps

aduffeck:
The `retry` outcome can be used by postprocessing steps to reschedule themselves, e.g. after a temporary failures.

### DIFF
--- a/pkg/events/postprocessing.go
+++ b/pkg/events/postprocessing.go
@@ -49,6 +49,8 @@ var (
 	PPOutcomeAbort PostprocessingOutcome = "abort"
 	// PPOutcomeContinue means that the upload is moved to its final destination (eventually being marked with pp results)
 	PPOutcomeContinue PostprocessingOutcome = "continue"
+	// PPOutcomeRetry means that there was a temporary issue and the postprocessing should be retried at a later point in time
+	PPOutcomeRetry PostprocessingOutcome = "retry"
 )
 
 // BytesReceived is emitted by the server when it received all bytes of an upload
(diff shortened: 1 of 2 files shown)


pkg/events/postprocessing.go
```diff
@@ -153,6 +155,22 @@ func (PostprocessingFinished) Unmarshal(v []byte) (interface{}, error) {
 	return e, err
 }
 
+// PostprocessingRetry is emitted by *some* service which can decide that
+type PostprocessingRetry struct {
```
kobergj, line 159: Why not extend (and maybe rename) already existing `ResumePostprocessing` event?

kobergj (changes requested):

pkg/events/postprocessing.go
aduffeck: @kobergj  `PostprocessingRetry` is an informational event, emitted so that other services can learn about retries happening in the system. It is not used for control flow.
Instead the postprocessing service will, when it detects a `retry` outcome, emit the `PostprocessingRetry` right away and then emit a `StartPostprocessingStep` event after the according backoff period.

Does that make sense to you? If yes I'll try to adjust the event documentation to make that more clear.

kobergj: Ah I see. Who is listening to this event at the moment?

aduffeck: It's currently not being received by anyone :)

kobergj (approved):

PR #4391, opened 2023-12-07
[tests-only] CERNBox setup for ScienceMesh tests

glpatcern:


glpatcern:
Merging as this is for tests. An issue remains, to be sorted out at a later stage.

PR #4412, opened 2023-12-15
Allow authentication for nats connections

kobergj:
Allows authenticating nats connections with username/password


2403905 (approved):

internal/grpc/interceptors/eventsmiddleware/events.go
```diff
@@ -234,6 +234,8 @@ func publisherFromConfig(m map[string]interface{}) (events.Publisher, error) {
 			EnableTLS:            m["enable-tls"].(bool),
 			TLSInsecure:          m["tls-insecure"].(bool),
 			TLSRootCACertificate: tlsCert,
+			AuthUsername:         m["username"].(string),
+			AuthPassword:         m["password"].(string),
```
butonic, line 238: this is a little bit fragile ... if the config struct `m` does not have a key set this will panic with
```
"interface conversion: interface {} is nil, not string"
```
same for all the other keys, here ... still kinda meh.

butonic (approved):
on remark. I'd build the `stream.NatsConfig` beforhand and apply only the values that are set. But since this is just following the way the other properties are set I'm ok with merging this as is.

internal/grpc/interceptors/eventsmiddleware/events.go
kobergj: Good point :+1: I moved the definition to the source struct and use `mapstructure.Decode` to unmarshal it. Please double check.

butonic (approved):

PR #4422, opened 2023-12-19
Fix interceptor order to make sure tracing is set up first

aduffeck:
This PR fixes a problem where the appctx logger was using a new traceid instead of picking up the one from the trace parent.

Part of the fix of https://github.com/owncloud/ocis/issues/7928


butonic (approved):

aduffeck:
Streaming doesn't use the opentelemetry interceptor at all at that point. I'll take this to chat to discuss..

PR #4424, opened 2023-12-20
Avoid concurrent mapwrites in receivedsharecache

kobergj:
Avoid panics due to concurrent map writes in the receivedsharecache pkg.

References https://github.com/owncloud/ocis/issues/7697


rhafer (approved):
lgtm

PR #4429, opened 2023-12-21
publicsshareprovider: Allow creating/updating public links on received shares

rhafer:
The permission check when trying to create an internal link of a received share did not work correctly. Internal links have all Permissions set to false. The 'SufficientCS3Permissions' method treats an all-false PermissionSet as a deny grant, which a wrong in the public link case.
We can skip the 'SufficientCS3Permissions' check for internal links however, since the creation of a internal link should always be allowed as long as the user has the 'AddGrant' permission.

Related Issue: https://github.com/owncloud/ocis/issues/8039


micbar (approved):
Thanks! Missed that on my PR.

PR #4431, opened 2023-12-21
Revert "bugfix: Don't return disabled users on GetUser call"

rhafer:
If you think https://github.com/cs3org/reva/pull/4430 is too much for now. We can also revert the original PR

This reverts commit 9b7ec1e4afc754b9ce39cd053aa793febec478a7.

This needs some more work we need to allow admins to lookup disabled users.


rhafer:
Hm, adding a changelog for reverting an  unreleased bugfix doesn't really make sense, does it? It will just confuse people IMO.

kobergj (approved):

PR #4439, opened 2023-12-27
fix: correct status code when move is denied

micbar:
# Description

We fixed a bug when the status code was 502 instead of 403 when moving a file to a denied path to be compatible with oc10.

https://github.com/owncloud/ocis/issues/8063

### DIFF
--- a/changelog/unreleased/fix-statuscode-move-denied.md
+++ b/changelog/unreleased/fix-statuscode-move-denied.md
@@ -1,0 +1,5 @@
+Bugfix: Fixed wrong status code when moving a file to a denied path
+
+We fixed a bug when the status code was 403 instead of 502 when moving a file to a denied path to be compatible with oc10.
+
+https://github.com/cs3org/reva/pull/4439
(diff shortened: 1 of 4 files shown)


changelog/unreleased/fix-statuscode-move-denied.md
```diff
@@ -0,0 +1,5 @@
+Bugfix: Fixed wrong status code when moving a file to a denied path
+
+We fixed a bug when the status code was 403 instead of 502 when moving a file to a denied path to be compatible with oc10.
```
phil-davis, line 3: IMO you are correcting the status code that was 502, but now returns 403 (forbidden).
```suggestion
We fixed a bug when the status code 502 was returned when moving a file to a denied path. Status code 403 (forbidden) is now returned to be compatible with oc10.
```
Maybe the above text describes the bugfix better?

phil-davis (approved):

kobergj (approved):

fschade (approved):
🚀

PR #4445, opened 2024-01-03
drop unnecessary stat before delete

butonic:
We dropped an unnecessary stat before delete


butonic:
meh it is used for the ItemTrashed event in the events middleware. But why can we not use the reference in the request ...

butonic:
nothing for now ... I'm not sure if changing the interface te return the id as well is the right way to do this. maybe the event middleware is the wrong approach. we could use it if the stroage driver implementations themselves would have middlewares themselves. then we could first do a lookup id middleware (which is always the first step), then check permissions (which is always the second step), then an emit event middleware (not existing yet, but would then have the ids, paths, users and permissions) ...

PR #4446, opened 2024-01-04
disallow to delete a file during the processing

2403905:
We want to disallow deleting a file during the processing to prevent collecting the orphan uploads.

### DIFF
--- a/internal/grpc/services/storageprovider/storageprovider.go
+++ b/internal/grpc/services/storageprovider/storageprovider.go
@@ -702,10 +702,24 @@ func (s *service) Delete(ctx context.Context, req *provider.DeleteRequest) (*pro
 		}
 	}
 
-	md, err := s.storage.GetMD(ctx, req.Ref, []string{}, []string{"id"})
+	md, err := s.storage.GetMD(ctx, req.Ref, []string{}, []string{"id", "status"})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -709,6 +709,20 @@ func (s *service) Delete(ctx context.Context, req *provider.DeleteRequest) (*pro
 		}, nil
 	}
 
+	if utils.ReadPlainFromOpaque(md.GetOpaque(), "status") == "processing" {
```
butonic, line 712: The GetMD() call in line 705 only requests the `id`. Please add `status` so that if we implement the filter we can actually see that `session` is required.

2403905: The "status" is included in GetMD() implicitly.  Is it by design?

2403905: https://github.com/cs3org/reva/blob/edge/pkg/storage/utils/decomposedfs/node/node.go#L648C12-L648C12

2403905: The "status" is included if n.IsProcessing(ctx)

butonic (approved):

PR #4447, opened 2024-01-05
fix: correct statuscode and message when MOVEing between cross spaces

saw-jan:
This PR changes the `502` statuscode to `403` with correct message while MOVEing between cross spaces.

Fixes:
- https://github.com/owncloud/ocis/issues/8116
- https://github.com/owncloud/ocis/issues/7618
- https://github.com/owncloud/ocis/issues/8124
- https://github.com/owncloud/ocis/issues/8125

BUT there is this note in the code:
>WebDAV RFC https://www.rfc-editor.org/rfc/rfc4918#section-9.9.4 says to use
> 502 (Bad Gateway) - This may occur when the destination is on another
> server and the destination server refuses to accept the resource.
> This could also occur when the destination is on another sub-section
> of the same server namespace. but we only have a not supported error

Q: What should we expect in test? `502` or `403`?
CC @micbar @kobergj

### DIFF
--- a/pkg/storage/utils/decomposedfs/tree/tree.go
+++ b/pkg/storage/utils/decomposedfs/tree/tree.go
@@ -228,7 +228,7 @@ func (t *Tree) Move(ctx context.Context, oldNode *node.Node, newNode *node.Node)
 		// > This could also occur when the destination is on another sub-section
 		// > of the same server namespace.
 		// but we only have a not supported error
-		return errtypes.NotSupported("cannot move across spaces")
+		return errtypes.PermissionDenied("cross storage moves are not permitted, use copy and delete")
 	}
 	// if target exists delete it without trashing it
 	if newNode.Exists {


butonic:
The RFC is always right.

saw-jan:
> The RFC is always right.

Yeah,
But we recently changed 502 to 403 in most cases like MOVE between two shares, shares to personal and others

Scenarios matrix table is here: https://github.com/owncloud/ocis/issues/8121

micbar:
Needs to be 403 because of the oc10 compatibility.

saw-jan:
> Needs to be 403 because of the oc10 compatibility.

So does that mean this PR is valid? Should we change 502 to 403?

micbar:
@TheOneRing @dragotin 

We need to discuss. RFC is very clear.

saw-jan:
From https://github.com/owncloud/ocis/issues/4087#issuecomment-1177951438
>The return status will be 502 Bad Gateway according to WebDAV spec. This fix is already in reva.

saw-jan:
502 is the expected behavior
Ref: https://github.com/owncloud/ocis/issues/8063#issuecomment-1883281975

So closing this PR

PR #4454, opened 2024-01-09
Skip unnecessary share retrieval

butonic:
AFAICT we already got a list of all reveiced shares, including the Hidden flag. No need to make another GetReceivedShare request for every accepted share.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go
@@ -69,7 +69,11 @@ func (h *Handler) AcceptReceivedShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	sharedResource, ocsResponse := getSharedResource(ctx, client, rs.Share.Share.ResourceId)
+	rs.State = collaboration.ShareState_SHARE_STATE_ACCEPTED
+
+	// now we also need to determine the mount point by evaluating all other mounted shares
+
+	sharedResource, ocsResponse := getSharedResource(ctx, client, rs.Share.ResourceId)
 	if ocsResponse != nil {
 		response.WriteOCSResponse(w, r, *ocsResponse, nil)
 		return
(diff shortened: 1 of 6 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go
```diff
@@ -95,7 +95,6 @@ func (h *Handler) AcceptReceivedShare(w http.ResponseWriter, r *http.Request) {
 			}
 		} else {
 			if s.State == collaboration.ShareState_SHARE_STATE_ACCEPTED {
-				s.Hidden = h.getReceivedShareHideFlagFromShareID(r.Context(), shareID)
```
kobergj, line 98: We do not need the hidden flag any more?

Do we still need `getReceivedShareHideFlagFromShareID`?

kobergj (changes requested):

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go
kobergj: I mean: Don't we need to set the flag somewhere else?

butonic: there were two other places ... but both were actually unnecessary as well. we should have done a better review of the PRs ... 😞  https://github.com/cs3org/reva/pull/4454/commits/19b6b46043232bc0f0dd7b72c677fd1a45d35cfb

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go
```diff
@@ -166,15 +173,15 @@ func (h *Handler) RejectReceivedShare(w http.ResponseWriter, r *http.Request) {
 		Share: &collaboration.Share{
 			Id: &collaboration.ShareId{OpaqueId: shareID},
 		},
-		State:  collaboration.ShareState_SHARE_STATE_REJECTED,
-		Hidden: h.getReceivedShareHideFlagFromShareID(r.Context(), shareID),
+		State: collaboration.ShareState_SHARE_STATE_REJECTED,
 	}
-	updateMask := &fieldmaskpb.FieldMask{Paths: []string{"state", "hidden"}}
+	updateMask := &fieldmaskpb.FieldMask{Paths: []string{"state"}}
```
butonic, line 178: we need to make sure the share managers respect the field mask!

butonic: jsonsc3 handles it, memory handles it, owncloudsql handles it (but does not know anything about the hidden flag), json handles it (but does not know anything about the hidden flag), cs3 also has it. good to merge

kobergj (approved):

kobergj (approved):
Looks good to me, but please run a full-ci on ocis before merging. These sharing changes can have unforseen consequences.

dragonchaser (approved):

PR #4455, opened 2024-01-10
Limit max number of entries when listing recycle bin

glpatcern:
This PR introduces support for `from` and `to` dates when listing the recycle bin (by default no more than 14 days, configurable), as well as a (configurable) cap on the number of entries returned to the web UI.

The cap is obtained by `eos recycle ls max-entries` as well as via GRPC. This feature was released in eos 5.2.7.

We also update the clients in the test CI to eos `5.2.14`, as new clients are not compatible with old ones, and the listing is to be performed with `eos oldfind` as opposed to `eos find`.


labkode:
Goes into the right direction, we can add the data parameter to it as it was in the past:
https://github.com/cernbox/revaold/commit/607bfc11e8381b6aeb1c434b3382e77dced2f18f

labkode:
For record, while debugging the recycle issue on CERNBox, we saw that the limit of 14 days was not enforced, needs to be checked. We same hundreds of recycle ls calls being sent to EOS instead of only 14.

diocas:
This happened in the test instance, where the UI **was not** sending any `to` `from`.

glpatcern:
Yep, I saw that yesterday, but the backend must be resilient to that (default to last 14 days). Debugging it

glpatcern:
This now works, but requires eos 5.2.6+, therefore to be merged only after we have completed the current eos deployment campaign.

glpatcern:
Fully tested. The eosbinary works with eos 5.2.6+ on the server, the eosgrpc requires eos 5.2.13+ on the server. In all cases, eos 5.2.13+ is required as client where reva runs.

diocas (approved):

PR #4456, opened 2024-01-11
Fix lost revisions

aduffeck:
This fixes a problem where restoring a revision generated by `TouchFile` overwrites another version.

Fixes https://github.com/owncloud/enterprise/issues/6249

### DIFF
--- a/pkg/storage/utils/decomposedfs/revisions.go
+++ b/pkg/storage/utils/decomposedfs/revisions.go
@@ -263,10 +263,6 @@ func (fs *Decomposedfs) RestoreRevision(ctx context.Context, ref *provider.Refer
 	// copy blob metadata from restored revision to node
 	restoredRevisionPath := fs.lu.InternalPath(spaceID, revisionKey)
 	err = fs.lu.CopyMetadata(ctx, restoredRevisionPath, nodePath, func(attributeName string, value []byte) (newValue []byte, copy bool) {
-		if attributeName == prefixes.MTimeAttr {
-			// update mtime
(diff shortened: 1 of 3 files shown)


pkg/storage/utils/decomposedfs/revisions.go
```diff
@@ -275,6 +271,15 @@ func (fs *Decomposedfs) RestoreRevision(ctx context.Context, ref *provider.Refer
...
 		return errtypes.InternalError("failed to copy blob xattrs to old revision to node: " + err.Error())
 	}
+	// always set the node mtime to the current time
+	err = fs.lu.MetadataBackend().SetMultiple(ctx, nodePath,
+		map[string][]byte{
+			prefixes.MTimeAttr: []byte(time.Now().UTC().Format(time.RFC3339Nano)),
+		},
+		false)
+	if err != nil {
+		return errtypes.InternalError("failed to set mtime attribute on node: " + err.Error())
```
kobergj, line 281: Use `errors.Wrap`?

kobergj (approved):
Looks good, just a suggestion

pkg/storage/utils/decomposedfs/revisions.go
aduffeck: We just discussed this in person. Wrapping the error would no longer allow for type assertions on the accoring interface
```
type IsInternalError interface {
	IsInternalError()
}
```
which is currently the way to check for certain error types in reva. Thus we'll refrain from wrapping the error for now.

PR #4462, opened 2024-01-18
usershareprovider: Prevent setting container specific permissions on files

rhafer:
It was possible to set the 'CreateContainer', 'Move' or 'Delete' permissions on file resources with a CreateShare request. These permissions are meant to be only set on container resources. The UpdateShare request already has a similar check.

Parial Fix: https://github.com/owncloud/ocis/issues/8131

### DIFF
--- a/changelog/unreleased/fix-create-share-precheck.md
+++ b/changelog/unreleased/fix-create-share-precheck.md
@@ -1,0 +1,8 @@
+Bugfix: Prevent setting container specific permissions on files
+
+It was possible to set the 'CreateContainer', 'Move' or 'Delete' permissions on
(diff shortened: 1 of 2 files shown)


kobergj (approved):
I get that `CreateContainer` and `Move` make no sense on a share. But why are we blocking `Delete`?

Just out of curiosity. Code looks good :+1:

rhafer:
> But why are we blocking `Delete`? Just out of curiosity. Code looks good 👍

AFAIU the `Delete` permission is intended for Folder/Containers to indicate that you are allowed to delete content in them. (Similar to how on a posix file system to need `wx` permission on the folder to be able to delete files in there. But you don't need `write` permission on the file itself)

PR #4469, opened 2024-01-22
Handle partial uploads

aduffeck:
This RP fixes a problem where incomplete uploads were still finished as if they had been successful.

/cc @butonic 

Fixes https://github.com/owncloud/ocis/issues/8239

### DIFF
--- a/internal/http/services/owncloud/ocdav/copy.go
+++ b/internal/http/services/owncloud/ocdav/copy.go
@@ -303,6 +303,8 @@ func (s *svc) executePathCopy(ctx context.Context, selector pool.Selectable[gate
 			return err
 		}
 		httpUploadReq.Header.Set(datagateway.TokenTransportHeader, uploadToken)
+		httpUploadReq.Header.Set(net.HeaderUploadLength, strconv.FormatUint(cp.sourceInfo.GetSize(), 10))
+		httpUploadReq.ContentLength = int64(cp.sourceInfo.GetSize())
 
 		httpUploadRes, err := s.client.Do(httpUploadReq)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


internal/http/services/owncloud/ocdav/put.go
```diff
@@ -307,6 +307,8 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		}
 		Propagator.Inject(ctx, propagation.HeaderCarrier(httpReq.Header))
 		httpReq.Header.Set(datagateway.TokenTransportHeader, token)
+		httpReq.Header.Set(net.HeaderUploadLength, strconv.FormatInt(length, 10))
```
butonic, line 310: I think the httpReq.Header.Set(net.HeaderUploadLength line is unnecessary as we are using the `simple` PUT, not tus.
```suggestion
```

internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -303,6 +303,8 @@ func (s *svc) executePathCopy(ctx context.Context, selector pool.Selectable[gate
 			return err
 		}
 		httpUploadReq.Header.Set(datagateway.TokenTransportHeader, uploadToken)
+		httpUploadReq.Header.Set(net.HeaderUploadLength, strconv.FormatUint(cp.sourceInfo.GetSize(), 10))
```
butonic, line 306: I think the `httpUploadReq.Header.Set(net.HeaderUploadLength...` line is unnecessary as we are using the `simple` PUT, not tus.
```suggestion
```

internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -521,6 +523,8 @@ func (s *svc) executeSpacesCopy(ctx context.Context, w http.ResponseWriter, sele
...
```
butonic, line 526: I think the `httpUploadReq.Header.Set(net.HeaderUploadLength...` line is unnecessary as we are using the `simple` PUT, not tus.
```suggestion
```

butonic (approved):

PR #4472, opened 2024-01-23
fixed the race condition that led to concurrent map access in a publicshare manager

2403905:
Related Issue [https://github.com/owncloud/ocis/issues/8255](https://github.com/owncloud/ocis/issues/8255)

### DIFF
--- a/pkg/publicshare/manager/json/json.go
+++ b/pkg/publicshare/manager/json/json.go
@@ -219,6 +219,9 @@ func (m *manager) Dump(ctx context.Context, shareChan chan<- *publicshare.WithPa
 
 // Load imports public shares and received shares from channels (e.g. during migration)
 func (m *manager) Load(ctx context.Context, shareChan <-chan *publicshare.WithPassword) error {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
 	db, err := m.persistence.Read(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/publicshare/manager/json/json.go
```diff
@@ -414,6 +417,9 @@ func (m *manager) UpdatePublicShare(ctx context.Context, u *user.User, req *link
 
 // GetPublicShare gets a public share either by ID or Token.
 func (m *manager) GetPublicShare(ctx context.Context, u *user.User, ref *link.PublicShareReference, sign bool) (*link.PublicShare, error) {
+	m.mutex.Lock()
```
kobergj, line 420: Can we change this to use a `sync.RWMutex` instead? We could `RLock` here for improved performance.

kobergj (changes requested):
Just one question. Rest looks good :+1:

pkg/publicshare/manager/json/json.go
2403905: We can't, because we call `revokeExpiredPublicShare` inside and it lead to the `m.persistence.Write()` call

kobergj (approved):

butonic (approved):

butonic:
since this is only for the json manager I'm fine with all the mutexes ...

PR #4477, opened 2024-01-24
Extend ResumePostprocessingEvent

kobergj:
Instead of just sending an uploadID, one can set a postprocessing step now to restart all uploads in this step
Also adds a new postprocessing step - "finished" - which means that postprocessing is finished but the storage provider
hasn't acknowledged it yet.


2403905 (approved):

butonic:
How does the state transition to Finished happen? Needs oCIS changes?

kobergj:
> How does the state transition to Finished happen? Needs oCIS changes?

Yes, that is a task for the postprocessing service. Already on it.

PR #4478, opened 2024-01-25
add hellofs, cleanup and document storage driver interface

butonic:
A minimal hello world filesystem as an example for a read only storage driver.

I also documented the storage drive interface and deleted the unused GetHome and CreateHome functions.

To actually be able to use this in ocis I had to patch it as well. PRs incoming.


DeepDiver1975:
I am.missing any docs on how to actually use this when running reva/ocis.

Hints are welcome.

dragotin (approved):
Thank you, great!

There could always be more documentation, examples etc, but this is a great step.

PR #4497, opened 2024-02-05
Removed stat to all storage providers on a Depth:0 PROPFIND to "/"

glpatcern:


labkode:
Merging as this code is no longer necessary due to the lack of virtual share roots.

labkode (approved):

PR #4500, opened 2024-02-06
eosbinary: force find to query MGM and not QuarkDB

labkode:


pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -724,7 +724,7 @@ func (c *Client) Rename(ctx context.Context, auth eosclient.Authorization, oldPa
 
 // List the contents of the directory given by path.
 func (c *Client) List(ctx context.Context, auth eosclient.Authorization, path string) ([]*eosclient.FileInfo, error) {
-	args := []string{"find", "--fileinfo", "--maxdepth", "1", path}
+	args := []string{"find", "--cached --fileinfo", "--maxdepth", "1", path}
```
glpatcern, line 727:
```suggestion
	args := []string{"find", "--cached", "--fileinfo", "--maxdepth", "1", path}
```
Surely not the reason for the CI to fail, let's trigger it once again

gmgigi96: The client in the docker container might not be updated to have this flag

glpatcern:
With the understanding that eos will go via cache by default, and that for the time being we will be using the `oldfind` command, I think this can be closed

PR #4501, opened 2024-02-06
enhancement: allow sending multiple userIDs in one SSE event

fschade:
Sending multiple user ids in one sse event is now possible which reduces the number of sent events.

### DIFF
--- a/pkg/events/sse.go
+++ b/pkg/events/sse.go
@@ -4,9 +4,9 @@ import (
 	"encoding/json"
 )
 
-// SendSEE instructs the sse service to send a notification to a user
+// SendSSE instructs the sse service to send one or multiple notifications
 type SendSSE struct {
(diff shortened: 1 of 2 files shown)


pkg/events/sse.go
```diff
@@ -4,9 +4,9 @@ import (
 	"encoding/json"
 )
 
-// SendSEE instructs the sse service to send a notification to a user
+// SendSSE instructs the sse service to send a notification to a user
```
JanAckermann, line 7: Maybe adapt the comment (users) 🤷‍♀️

kobergj (approved):

JanAckermann (approved):

pkg/events/sse.go
fschade: THX

PR #4503, opened 2024-02-06
fix an error when move using destination id

2403905:


internal/http/services/owncloud/ocdav/move.go
```diff
@@ -266,7 +266,11 @@ func (s *svc) handleMove(ctx context.Context, w http.ResponseWriter, r *http.Req
 		}
 		// TODO what if intermediate is a file?
 	}
-
+	// resolve the destination path
+	if dst.Path == "." {
+		dst.Path = utils.MakeRelativePath(dstStatRes.GetInfo().GetName())
+		dst.ResourceId = dstStatRes.GetInfo().ParentId
```
kobergj, line 272:
```suggestion
		dst.ResourceId = dstStatRes.GetInfo().GetParentId()
```
My IDE gets angry when we don't use getters :(

kobergj (approved):
:+1: Point not critical.

kobergj (approved):

PR #4508, opened 2024-02-07
Add pprof http service and demote prom interceptor to not run by default

labkode:


glpatcern:
The prom service was temporarily disabled because of https://github.com/cs3org/reva/issues/4509

PR #4514, opened 2024-02-13
Pass lock metadata on uploads + use upstream EOS GRPC bindings

glpatcern:
We now pass relevant metadata (lock id and lock holder) downstream on uploads and on xattr operations, and handle the case of conflicts due to lock mismatch. The full implementation is only for eos (and requires eos 5.2.22+ on the server), cephfs is TODO.

Relates to https://github.com/cs3org/wopiserver/pull/139

Edit: the EOS GRPC bindings are taken from https://github.com/cern-eos/go-eosgrpc/


pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -720,19 +736,18 @@ func (c *Client) Write(ctx context.Context, auth eosclient.Authorization, path s
...
-func (c *Client) WriteFile(ctx context.Context, auth eosclient.Authorization, path, source string) error {
+func (c *Client) WriteFile(ctx context.Context, auth eosclient.Authorization, path, source, app string) error {
 	xrdPath := fmt.Sprintf("%s//%s", c.opt.URL, path)
 	args := []string{"--nopbar", "--silent", "-f", source, xrdPath}
 
 	if auth.Token != "" {
 		args[4] += "?authz=" + auth.Token
 	} else if auth.Role.UID != "" && auth.Role.GID != "" {
-		args = append(args, fmt.Sprintf("-ODeos.ruid=%s&eos.rgid=%s&eos.app=reva_eosclient::write", auth.Role.UID, auth.Role.GID))
+		args = append(args, fmt.Sprintf("-ODeos.ruid=%s&eos.rgid=%s&eos.app=%s", auth.Role.UID, auth.Role.GID, app))
```
labkode, line 750: The reva_eosclient is the appname used to distinguish traffic from Reva from other traffic (sync, fuse). To include the sub-app that is used to write or read a file we can include it such:

reva_eosclient::write::<app>>

if app is empty, the app should be:
`reva_eosclient::write` and not `reva_eosclient::write::`.

The `write or read` verb is left on purpose as the second argument and on not the third, to avoid having "apps" named "write" or "read" when the app is empty.

pkg/eosclient/eosgrpc/eoshttp.go
```diff
@@ -376,6 +378,9 @@ func (c *EOSHTTPClient) PUTFile(ctx context.Context, remoteuser string, auth eos
 		return err
 	}
 
+	if app != "" {
+		req.Header.Set("app", app)
```
labkode, line 382: Best practice would be to name this head: `X-App`

pkg/ocm/storage/outcoming/ocm.go
```diff
@@ -396,14 +397,16 @@ func getUploadProtocol(protocols []*gateway.FileUploadProtocol, protocol string)
 	return "", "", false
 }
 
-func (d *driver) Upload(ctx context.Context, ref *provider.Reference, content io.ReadCloser) error {
+func (d *driver) Upload(ctx context.Context, ref *provider.Reference, content io.ReadCloser, metadata map[string]string) error {
 	share, rel, err := d.shareAndRelativePathFromRef(ctx, ref)
 	if err != nil {
 		return err
 	}
```
labkode, line 404: A check is needed to make sure `metadata` is not nil to avoid runtime errors (nil pointers) down the code.

pkg/rhttp/datatx/manager/spaces/spaces.go
```diff
@@ -112,6 +121,8 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 				w.WriteHeader(http.StatusUnauthorized)
 			case errtypes.InsufficientStorage:
 				w.WriteHeader(http.StatusInsufficientStorage)
+			case errtypes.BadRequest:
```
labkode, line 124: Same comment as above.

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -2284,11 +2289,13 @@ func (fs *eosfs) convert(ctx context.Context, eosFileInfo *eosclient.FileInfo) (
...
 		},
-		ArbitraryMetadata: &provider.ArbitraryMetadata{
-			Metadata: filteredAttrs,
-		},
 	}
-
+	if eosFileInfo.Attrs[eosLockKey] != "" {
+		// populate the lock, fail silently if unable to decode
+		if l, _ := decodeLock(eosFileInfo.Attrs[lockPayloadKey], eosFileInfo.Attrs[eosLockKey]); l != nil {
+			info.Lock = l
```
labkode, line 2296: An error log would be useful for this silent fail.

pkg/storage/utils/eosfs/upload.go
```diff
@@ -75,7 +76,15 @@ func (fs *eosfs) Upload(ctx context.Context, ref *provider.Reference, r io.ReadC
 	if err != nil {
 		return err
```
labkode, line 77: Check for metadata not being nil

labkode (changes requested):
Few minor changes

pkg/eosclient/eosgrpc/eoshttp.go
glpatcern: Agreed, let's see if Andreas can change it

pkg/ocm/storage/outcoming/ocm.go
glpatcern: But `metadata` here is an implicit reference as it's a `map`, so cannot be nil right?

pkg/eosclient/eosbinary/eosbinary.go
glpatcern: This is tricky now, the issue being that EOS uses `app` both for traffic tagging and for locking. So on one side we could decide to go for some `reva_eosclient::<app>`  - currently we'd have `reva_<app>`, and the GRPC client does not tag the traffic because this `app` argument was not supported. But surely we can't have `write::<app>` and `read::<app>` in the context of locking.
That said I'd propose a generic `reva::write`, `reva::read` (in case of no locks) and `reva::app_<appname>` in case of locks for the tags to be used everywhere.

glpatcern: OK so the last commit does exactly that, by extracting an `EncodeAppName` function at the `eosfs` service level

labkode (approved):

labkode (approved):

PR #4518, opened 2024-02-15
[full-ci] fix an error when lock/unlock a file

2403905:
Description:
We fixed a bug when anonymous user with viewer role in public link of a folder can lock/unlock a file inside it

Related issue:
[https://github.com/owncloud/ocis/issues/7785](https://github.com/owncloud/ocis/issues/7785)
[https://github.com/owncloud/ocis/issues/7761](https://github.com/owncloud/ocis/issues/7761)

Related PR:
[https://github.com/owncloud/ocis/pull/8472](https://github.com/owncloud/ocis/pull/8472)

### DIFF
--- a/internal/http/services/owncloud/ocdav/locks.go
+++ b/internal/http/services/owncloud/ocdav/locks.go
@@ -260,11 +260,15 @@ func (cls *cs3LS) Unlock(ctx context.Context, now time.Time, ref *provider.Refer
 		return err
 	}
 
-	switch res.Status.Code {
+	switch res.GetStatus().GetCode() {
 	case rpc.Code_CODE_OK:
 		return nil
 	case rpc.Code_CODE_FAILED_PRECONDITION:
 		return errtypes.Aborted("file is not locked")
+	case rpc.Code_CODE_LOCKED:
+		appctx.GetLogger(ctx).Error().Str("token", token).Interface("unlock", ref).
+			Msg("could not unlock " + res.GetStatus().GetMessage())
+		return errtypes.Locked("another token")
 	default:
 		return errtypes.NewErrtypeFromStatus(res.Status)
 	}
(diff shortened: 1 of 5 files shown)


butonic:
shouldn't this be handled in the `internal/grpc/services/storageprovider/storageprovider.go`?
```

// SetLock puts a lock on the given reference
func (s *service) SetLock(ctx context.Context, req *provider.SetLockRequest) (*provider.SetLockResponse, error) {
	err := s.storage.SetLock(ctx, req.Ref, req.Lock)

	return &provider.SetLockResponse{
		Status: status.NewStatusFromErrType(ctx, "set lock", err),
	}, nil
}

// GetLock returns an existing lock on the given reference
func (s *service) GetLock(ctx context.Context, req *provider.GetLockRequest) (*provider.GetLockResponse, error) {
	lock, err := s.storage.GetLock(ctx, req.Ref)

	return &provider.GetLockResponse{
		Status: status.NewStatusFromErrType(ctx, "get lock", err),
		Lock:   lock,
	}, nil
}

// RefreshLock refreshes an existing lock on the given reference
func (s *service) RefreshLock(ctx context.Context, req *provider.RefreshLockRequest) (*provider.RefreshLockResponse, error) {
	err := s.storage.RefreshLock(ctx, req.Ref, req.Lock, req.ExistingLockId)

	return &provider.RefreshLockResponse{
		Status: status.NewStatusFromErrType(ctx, "refresh lock", err),
	}, nil
}

// Unlock removes an existing lock from the given reference
func (s *service) Unlock(ctx context.Context, req *provider.UnlockRequest) (*provider.UnlockResponse, error) {
	err := s.storage.Unlock(ctx, req.Ref, req.Lock)

	return &provider.UnlockResponse{
		Status: status.NewStatusFromErrType(ctx, "unlock", err),
	}, nil
}
```

2403905:
It's a fair point

internal/http/services/owncloud/ocdav/locks.go
```diff
@@ -639,16 +643,19 @@ func (s *svc) unlockReference(ctx context.Context, _ http.ResponseWriter, r *htt
...
+		switch err.(type) {
+		case errtypes.Aborted, errtypes.Locked:
+			return http.StatusLocked, err
+		case errtypes.PermissionDenied:
+			return http.StatusForbidden, err
+		case errtypes.PreconditionFailed:
+			return http.StatusConflict, err
+		default:
+			return http.StatusInternalServerError, err
+		}
```
butonic, line 659: why hide the switch in an if statement?
```go
switch err := s.LockSystem.Unlock(ctx, time.Now(), ref, t); err {
	case nil:
		return http.StatusNoContent, nil
	case errors.Aborted, errors.ErrLocked:
		return http.StatusLocked, err
	case errors.ErrForbidden:
		return http.StatusForbidden, err
	case errors.ErrNoSuchLock:
		return http.StatusConflict, err
	default:
		return http.StatusInternalServerError, err
}
```

Also, we need to make sure the LockSystem call only returns errors from the errors package, as that is what the interface description mentions. Can you ensure the implementation converts the errtypes errors to errors package errors? It is a necessary conversion.

butonic (changes requested):
The switch should be extended and the errors should all be from the errors package.

internal/http/services/owncloud/ocdav/locks.go
2403905: Added the Errtype mapping but kept `if err == nil` because it could be various codes 204, 200, etc.

butonic (approved):
now failing tests? wow! but we should do an ocis bump full test run to verify ;-)

PR #4534, opened 2024-02-21
fix remove/update share permissions when the file is locked

2403905:
This is a workaround that should prevent removing or changing the share permissions when the file is locked.
These limitations have to be removed after the wopi server will be able to unlock the file properly.
These limitations are not spread on the files inside the shared folder. 

Related issue [https://github.com/owncloud/ocis/issues/8273](https://github.com/owncloud/ocis/issues/8273)

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -106,6 +106,15 @@ func (s *svc) ListShares(ctx context.Context, req *collaboration.ListSharesReque
 }
 
 func (s *svc) updateShare(ctx context.Context, req *collaboration.UpdateShareRequest) (*collaboration.UpdateShareResponse, error) {
+	// TODO: update wopi server
+	// FIXME This is a workaround that should prevent removing or changing the share permissions when the file is locked.
+	// https://github.com/owncloud/ocis/issues/8474
+	if status, err := s.checkLock(ctx, req.GetShare().GetId()); err != nil {
+		return &collaboration.UpdateShareResponse{
+			Status: status,
+		}, nil
+	}
+
 	c, err := pool.GetUserShareProviderClient(s.c.UserShareProviderEndpoint)
 	if err != nil {
 		appctx.GetLogger(ctx).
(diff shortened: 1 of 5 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -712,6 +730,55 @@ func (s *svc) removeSpaceShare(ctx context.Context, ref *provider.ResourceId, gr
...
+		logger.Debug().Interface("share", shareId).Msg(msg)
+		if getShareRes.GetStatus().GetCode() != rpc.Code_CODE_NOT_FOUND {
+			return status.NewNotFound(ctx, msg), errors.New(msg)
+		}
+		return status.NewInternal(ctx, msg), errors.New(msg)
+	}
+	return s.checkShareLock(ctx, getShareRes.Share)
+}
+
+func (s *svc) checkShareLock(ctx context.Context, share *collaboration.Share) (*rpc.Status, error) {
```
butonic, line 756: I think this is tooo restrictive. AFAIU we can
A) allow adding new users to a file
B) increasing the permissions of a share, e.g. add edit permission for a viewer.

The current check disallows any sharing changes when a file is locked.

People often open a file in onlyoffice and then need to ADD new participants.

See https://github.com/owncloud/ocis/issues/8273#issuecomment-1951979302

butonic (changes requested):
We need to check the PermissionSet when updating a share.

butonic (approved):
Nevermind, I see now that share creation is unaffected. So new users can be added when a file is locked. That is good enough for me. IMO extending share permissions can be implemented in a followup PR.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -106,6 +106,15 @@ func (s *svc) ListShares(ctx context.Context, req *collaboration.ListSharesReque
...
 func (s *svc) updateShare(ctx context.Context, req *collaboration.UpdateShareRequest) (*collaboration.UpdateShareResponse, error) {
+	// TODO: update wopi server
+	// FIXME This is a workaround that should prevent removing or changing the share permissions when the file is locked.
+	// https://github.com/owncloud/ocis/issues/8474
+	if status, err := s.checkLock(ctx, req.GetShare().GetId()); err != nil {
+		return &collaboration.UpdateShareResponse{
+			Status: status,
+		}, nil
+	}
+
```
rhafer, line 117: I know I am a bit late to the party. But isn't this racy? The file  lock can  be created  after this check, before updating the share. Classical TOCTOU problem, isn't it?

internal/grpc/services/gateway/usershareprovider.go
butonic: True, we would have to send the lock in the creat/update share opaque... 🤔

internal/grpc/services/gateway/usershareprovider.go
butonic: Right, we would have to send the lock in the opaque of the sharing requests to make this (more) *atomic*.

PR #4557, opened 2024-03-05
fix ceph build

butonic:


glpatcern (approved):
I think I'll port this to `master` as well!

PR #4561, opened 2024-03-06
sharestorageprovider: Stat resource when resolving share by path

rhafer:
When resolving a received shared by mountpath, make sure that the shared resource actually exists. This avoids issues with dangling shares for already delete resources using the same mountpoint.
    
Fixes: https://github.com/owncloud/ocis/issues/7895

### DIFF
--- a/internal/grpc/services/sharesstorageprovider/sharesstorageprovider_test.go
+++ b/internal/grpc/services/sharesstorageprovider/sharesstorageprovider_test.go
@@ -261,7 +261,23 @@ var _ = Describe("Sharesstorageprovider", func() {
 						},
 					}
 				default:
-					if req.Ref.ResourceId.OpaqueId == "shareddir-merged" {
+					switch req.Ref.ResourceId.OpaqueId {
+					case "shareddir", "shareddir2":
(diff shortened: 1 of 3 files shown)


internal/grpc/services/sharesstorageprovider/sharesstorageprovider_test.go
```diff
@@ -261,7 +261,23 @@ var _ = Describe("Sharesstorageprovider", func() {
 						},
 					}
 				default:
-					if req.Ref.ResourceId.OpaqueId == "shareddir-merged" {
+					switch req.Ref.ResourceId.OpaqueId {
+					case "shareddir", "shareddir2":
```
2403905, line 265: Do we need to provide the constants for the `"shareddir", "shareddir2", "shareddir-merged"`?

rhafer: Looking at the surrounding code, especially the outer switch statement, I'd say: no :smiley: 

More seriously I guess the mocks could benefit from some refactoring, but I think the string constants are the least of on an issue.

2403905 (approved):

PR #4567, opened 2024-03-08
fix: authprovider error message when a user is not found in the auth.…

DeepDiver1975:
…Manager

### DIFF
--- a/internal/grpc/services/authprovider/authprovider.go
+++ b/internal/grpc/services/authprovider/authprovider.go
@@ -136,28 +136,16 @@ func (s *service) Authenticate(ctx context.Context, req *provider.AuthenticateRe
 	password := req.ClientSecret
 
 	u, scope, err := s.authmgr.Authenticate(ctx, username, password)
-	switch v := err.(type) {
-	case nil:
-		log.Info().Msgf("user %s authenticated", u.Id)
+	if err != nil {
+		log.Debug().Str("client_id", username).Err(err).Msg("authsvc: error in Authenticate")
 		return &provider.AuthenticateResponse{
-			Status:     status.NewOK(ctx),
-			User:       u,
-			TokenScope: scope,
-		}, nil
-	case errtypes.InvalidCredentials:
-		return &provider.AuthenticateResponse{
-			Status: status.NewPermissionDenied(ctx, v, "wrong password"),
-		}, nil
-	case errtypes.NotFound:
-		log.Debug().Str("client_id", username).Msg("unknown client id")
-		return &provider.AuthenticateResponse{
-			Status: status.NewNotFound(ctx, "unknown client id"),
-		}, nil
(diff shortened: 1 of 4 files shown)


2403905 (approved):

rhafer (commented):
Please note that not all authproviders treat the ClientID parameter as a username. For `auth-machine` e.g. the ClientID can be either a username or a `<attribute>:<value>` tuple,  `auth-service` expectes a "userid" and  `auth-oidc` doesn't look into it at all.

I guess that is why the it was called `client_id` and not `username`  in the first place.

DeepDiver1975:
When a users wants to login and this fails and the log holds `unknown client id` the last thing you assume is that the user is unknown. This is the intent of this change.

`unknown client id or user` might be better? And leaving the parameter unchanged in the logger ...

internal/grpc/services/authprovider/authprovider.go
```diff
@@ -149,9 +149,9 @@ func (s *service) Authenticate(ctx context.Context, req *provider.AuthenticateRe
 			Status: status.NewPermissionDenied(ctx, v, "wrong password"),
 		}, nil
 	case errtypes.NotFound:
-		log.Debug().Str("client_id", username).Msg("unknown client id")
+		log.Debug().Str("username", username).Msg("unknown user")
```
rhafer, line 152:
```suggestion
		log.Debug().Str("client_id", username).Err(err).Msg("unknown  client id or user")
```

internal/grpc/services/authprovider/authprovider.go
```diff
@@ -149,9 +149,9 @@ func (s *service) Authenticate(ctx context.Context, req *provider.AuthenticateRe
...
 		return &provider.AuthenticateResponse{
-			Status: status.NewNotFound(ctx, "unknown client id"),
+			Status: status.NewNotFound(ctx, "unknown user"),
```
rhafer, line 154: We could  just bubble up `err.Error()` here instead for `'unknown user"`.

rhafer (commented):
> When a users wants to login and this fails and the log holds unknown client id the last thing you assume is that the user is unknown. This is the intent of this change.

Yeah understood. Could we just add some more specific logging to the individual auth-providers? Also they might already return some specific error message. (See suggestion)

> unknown client id or user might be better? And leaving the parameter unchanged in the logger ...

Would the suggestion help you?

internal/grpc/services/authprovider/authprovider.go
DeepDiver1975: What I had in mind after reading your initial comment. :+1:

DeepDiver1975: (There is a double white space in the error message)

rhafer (approved):

PR #4583, opened 2024-03-18
Draft: clean more dead code

labkode:


pkg/events/server/nats.go
glpatcern, line 1: isn't this part of the current NATS-based notifications system?

PR #4584, opened 2024-03-18
allow copy on self (needed for cross space copy)

dragonchaser:
A minor change for https://github.com/owncloud/ocis/issues/1230 to re-enable copy on items onto themselves.


2403905:
Should we chek when we move?

PR #4585, opened 2024-03-18
gateway(sharing-ng): Check for remaing space manager before removing grant

rhafer:
This check is already done by the ocs sharing implementaion, but for the move to the graph API base sharing implementation we'd want to have it in a more central place.

### DIFF
--- a/changelog/unreleased/add-remaining-manager-check.md
+++ b/changelog/unreleased/add-remaining-manager-check.md
@@ -1,0 +1,8 @@
+Enhancement: Move more consistency checks to the usershare API
+
+The gateway now checks if there will be at least one space manager remaining before
+deleting a space member. The legacy ocs based sharing implementaion already does this
+on its own. But for the future graph based sharing implementation it is better to have
+the check in a more central place.
+
+https://github.com/cs3org/reva/pull/4585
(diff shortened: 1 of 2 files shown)


kobergj (changes requested):
The code looks fine, but are we really sure we want to do this? This is moving business logic to the gateway service. Previously that was a no-go as gateway should be dumb. Did this change?

cc @butonic

rhafer:
> The code looks fine, but are we really sure we want to do this? This is moving business logic to the gateway service. Previously that was a no-go as gateway should be dumb. Did this change?

I kind of discussed this already with @butonic . It's not optimal but it is for sure better than the checks duplicated at the higher levels (i.e. ocs and graph). Any idea what would be a better place for those checks. The usersharepovider in the gateway is already full of business logic because it needs to deal with the differences between space permissions and "real" shares. This change falls into the same category.

butonic:
Services infront of CS3 should just be protocol adapters. The business logic should be behind the CS3 gateway facade. It can be in the gateway or in a driver. From the outside that should be irrelevant.

If we actually split the CS3 API into two parts:
1. an actively managed space registry and
2. a storage provider
the gateway would be deployed twice. Actually, it might become obsolete and the functionality could move to middle wares and interceptors. 🤔 At least conceptually that should be the architecture. A real deployment would share the gateway and deploy it together with the space registry and storage provider.

Note that the CS3 API just defines Interfaces it does not say every API should be deployed in a dedicated service. It is meant to swap out implementations of clearly defined APIs.

kobergj (approved):

PR #4587, opened 2024-03-20
Introduce InitiatorIDs

kobergj:
Allows sending a header `Initiator-ID` on http requests. This will be passed down ocis and returned on sse events. Clients can use this to avoid unnecessary requests.

See ticket for more details: https://github.com/owncloud/ocis/issues/8677

### DIFF
--- a/internal/grpc/interceptors/token/token.go
+++ b/internal/grpc/interceptors/token/token.go
@@ -39,6 +39,14 @@ func NewUnary() grpc.UnaryServerInterceptor {
 					ctx = metadata.AppendToOutgoingContext(ctx, ctxpkg.TokenHeader, tkn)
 				}
 			}
+
+			if val, ok := md[ctxpkg.InitiatorHeader]; ok {
+				if len(val) > 0 && val[0] != "" {
+					initiatorID := val[0]
+					ctx = ctxpkg.ContextSetInitiator(ctx, initiatorID)
+					ctx = metadata.AppendToOutgoingContext(ctx, ctxpkg.InitiatorHeader, initiatorID)
+				}
+			}
 		}
 
 		return handler(ctx, req)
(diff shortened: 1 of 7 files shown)


kobergj:
Note: This hacks into the auth middleware because it is only adding 2 lines of code there. The alternative would be to add a lot more code for a dedicated middleware. However I am a bit reluctant to do that as it will bring more complexity to already complex part of the code

internal/grpc/interceptors/token/token.go
```diff
@@ -39,6 +39,14 @@ func NewUnary() grpc.UnaryServerInterceptor {
 					ctx = metadata.AppendToOutgoingContext(ctx, ctxpkg.TokenHeader, tkn)
 				}
 			}
+
```
butonic, line 42: please add this to the NewStream below as well

pkg/ctx/initiatorctx.go
```diff
@@ -0,0 +1,35 @@
...
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package ctx
+
+import "context"
+
+// InitiatorHeader is the header key used to pass the initiator id to http services and grpc calls.
+var InitiatorHeader = "x-initiator-id"
```
butonic, line 24: prefixing headers with `x-` is discouraged: https://stackoverflow.com/a/3561399

I still think a request-id header already captures the intention of finding out which request triggered an event. Inventing a new header IMO is unnecessary. The server does not care about client provided request or initiator ids. If clients need to keep track of their request ids per tab they can do that at their own peril. If we start adding a new header all clients may need to check it. To what end?

I really prefer to stay out of any implementation details of clients. Cementing a hack by raising it to a dedicated header is not the right solution IMO.

That being said, I'd go with `Initiator-ID`.

butonic:
Yeah ... not pretty ... the correct way would be a dedicated middleware. The current implementation only passes the initiator-id when the user is authenticated, but the initiator-id should not depend on that.

butonic:
I'm ok with this to get a PoC, but we should take the time an implement this properly.

kobergj:
> I'm ok with this to get a PoC, but we should take the time and implement this properly.

It is not really about time. I'm concerned additional middlewares make the logic even more complex. Also I am not really sure having a middleware that just extracts one parameter is worth it.  Maybe if we had more ids or other information we want to pass down to ocis then this would make more sense

> The current implementation only passes the initiator-id when the user is authenticated, but the initiator-id should not depend on that.

If you are not authenticated, you do not have an sse connection, hence you do not receive any clientlog events anyways. So there is no need for passing the initiatorid in that case

pkg/ctx/initiatorctx.go
kobergj: Consistency? We have `x-access-token` and `x-user-agent` already. But if you insist I can use `initiator-id` here.

butonic (approved):
YOLO

PR #4602, opened 2024-03-28
Rebuild the eos grpc proto

ffurano:
Rebuild the eos grpc proto


glpatcern:
Incorporated in https://github.com/cs3org/reva/pull/4514

PR #4615, opened 2024-04-05
write blob based on session id

butonic:
I'm trying to fix https://github.com/owncloud/ocis/issues/8606.

This PR will now always use the session blobid to determine the target destination.

I also changed the signature of the Blobstore interfaces to get rid of the *node.Node struct in the Parameters.

However, I no longer pass the expected blobsize to `Download()`. Only the s3 blobstore would use it to check that it got the correct amount of bytes. IMO that should be the responsibility of the caller.

And for that Blobstore interface change I had to first fix the mock generation: https://github.com/cs3org/reva/pull/4614

fun ... all day long ...

I tried to grasp what happens now when two uploads are finished in reverse order. AFAICT the size diff needs to be tested. maybe we can even determine the size change later ... some thoughts are currently in the code:

in pkg/storage/utils/decomposedfs/upload/session.go
```go
// NodeExists returns wether or not the node existed during InitiateUpload.
// FIXME If two requests try to write the same file they both will store a new
// random node id in the session and try to initialize a new node when
// finishing the upload. The second request will fail with an already exists
// error when trying to create the symlink for the node in the parent directory.
// A node should be created as part of InitiateUpload. When listing a directory
// we can decide if we want to skip the entry, or expose uploed progress
// information. But that is a bigger change and might involve client work.
func (s *OcisSession) NodeExists() bool {
	return s.info.Storage["NodeExists"] == "true"
}
```

and again in pkg/storage/utils/decomposedfs/upload/store.go
```go

	if session.NodeExists() { // TODO this is wrong. The node should be created when the upload starts, the revisions should be created independently of the node
		// we do not need to propagate a change when a node is created, only when the upload is ready.
		// that still creates problems for desktop clients because if another change causes propagation it will detects an empty file
		// so the first upload has to point to the first revision with the expected size. The file cannot be downloaded, but it can be overwritten (which will create a new revision and make the node reflect the latest revision)
		// any finished postprocessing will not affect the node metadata.
		// *thinking* but then initializing an upload will lock the file until the upload has finished. That sucks.
		// so we have to check if the node has been created meanwhile (well, only in case the upload does not know the nodeid ... or the NodeExists array that is checked by session.NodeExists())
		// FIXME look at the disk again to see if the file has been created in between, or just try initializing a new node and do the update existing node as a fallback. <- the latter!

		f, err = store.updateExistingNode(ctx, session, n, session.SpaceID(), uint64(session.Size()))
		if f != nil {
			appctx.GetLogger(ctx).Info().Str("lockfile", f.Name()).Interface("err", err).Msg("got lock file from updateExistingNode")
		}
	} else {
		f, err = store.initNewNode(ctx, session, n, uint64(session.Size()))
		if f != nil {
			appctx.GetLogger(ctx).Info().Str("lockfile", f.Name()).Interface("err", err).Msg("got lock file from initNewNode")
		}
	}
```

That is why this PR is in draft. I think it is less wrong, but I don't want to merge it yet, because I havent fully understood the implications of this change.

I'd like to better understand and actaully the corner cases when two uploads are trying to create the same new file and how they can be safely handled without causing data loss. That has a lot of variability, eg. when the request was made with or without preconditions ...

### DIFF
--- a/pkg/storage/fs/posix/blobstore/blobstore.go
+++ b/pkg/storage/fs/posix/blobstore/blobstore.go
@@ -26,7 +26,6 @@ import (
 	"path/filepath"
 
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/lookup"
-	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
 	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/pkg/errors"
 )
@@ -49,8 +48,8 @@ func New(root string) (*Blobstore, error) {
 }
 
 // Upload stores some data in the blobstore under the given key
-func (bs *Blobstore) Upload(node *node.Node, source string) error {
-	dest, err := bs.path(node)
+func (bs *Blobstore) Upload(spaceID, blobID string, blobSize int64, source string) error {
+	dest, err := bs.path(spaceID, blobID)
 	if err != nil {
 		return err
 	}
@@ -85,8 +84,8 @@ func (bs *Blobstore) Upload(node *node.Node, source string) error {
 }
 
 // Download retrieves a blob from the blobstore for reading
-func (bs *Blobstore) Download(node *node.Node) (io.ReadCloser, error) {
-	dest, err := bs.path(node)
+func (bs *Blobstore) Download(spaceID, blobID string, blobSize int64) (io.ReadCloser, error) {
+	dest, err := bs.path(spaceID, blobID)
 	if err != nil {
 		return nil, err
 	}
@@ -94,12 +93,20 @@ func (bs *Blobstore) Download(node *node.Node) (io.ReadCloser, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "could not read blob '%s'", dest)
 	}
+	fi, err := file.Stat()
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not stat blob '%s'", dest)
+	}
+	if fi.Size() != blobSize {
+		return nil, fmt.Errorf("blob has unexpected size. %d bytes expected, got %d bytes", blobSize, fi.Size())
+	}
+
 	return file, nil
 }
 
 // Delete deletes a blob from the blobstore
-func (bs *Blobstore) Delete(node *node.Node) error {
-	dest, err := bs.path(node)
+func (bs *Blobstore) Delete(spaceID, blobID string) error {
+	dest, err := bs.path(spaceID, blobID)
 	if err != nil {
 		return err
 	}
@@ -109,14 +116,17 @@ func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
 
-func (bs *Blobstore) path(node *node.Node) (string, error) {
-	if node.BlobID == "" {
-		return "", fmt.Errorf("blobstore: BlobID is empty")
+func (bs *Blobstore) path(spaceID, blobID string) (string, error) {
+	if spaceID == "" {
+		return "", fmt.Errorf("blobstore: spaceID is empty")
+	}
+	if blobID == "" {
+		return "", fmt.Errorf("blobstore: blobID is empty")
 	}
 	return filepath.Join(
 		bs.root,
 		filepath.Clean(filepath.Join(
-			"/", "spaces", lookup.Pathify(node.SpaceID, 1, 2), "blobs", lookup.Pathify(node.BlobID, 4, 2)),
+			"/", "spaces", lookup.Pathify(spaceID, 1, 2), "blobs", lookup.Pathify(blobID, 4, 2)),
 		),
 	), nil
 }
--- a/pkg/storage/fs/posix/blobstore/blobstore.go
+++ b/pkg/storage/fs/posix/blobstore/blobstore.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/lookup"
+	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
 	"github.com/cs3org/reva/v2/pkg/utils"
 	"github.com/pkg/errors"
 )
@@ -48,8 +49,8 @@ func New(root string) (*Blobstore, error) {
 }
 
 // Upload stores some data in the blobstore under the given key
-func (bs *Blobstore) Upload(spaceID, blobID string, blobSize int64, source string) error {
-	dest, err := bs.path(spaceID, blobID)
+func (bs *Blobstore) Upload(node *node.Node, source string) error {
+	dest, err := bs.path(node)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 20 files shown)


pkg/storage/fs/s3ng/blobstore/blobstore.go
```diff
@@ -94,40 +98,57 @@ func (bs *Blobstore) Upload(node *node.Node, source string) error {
...
-		return nil, errors.Wrapf(err, "blob path: %s", bs.path(node))
+		return nil, errors.Wrapf(err, "could not download object '%s' from bucket '%s'", dest, bs.bucket)
 	}
 
-	if node.Blobsize != stat.Size {
-		return nil, fmt.Errorf("blob has unexpected size. %d bytes expected, got %d bytes", node.Blobsize, stat.Size)
-	}
+	// FIXME move to caller
+	/*
+		stat, err := reader.Stat()
```
butonic, line 119: maybe a io.File is also used as a reader? then the calle rcould test if the returned reader implements Stat() ....

butonic: io.File implements a Stat() but it returns a different struct so this won't work

butonic: the size verification on download was introduced in https://github.com/cs3org/reva/pull/3341

I extended the check to the other blobstore implementations.

In the future we might want to introduce a new ReadBlob that can handle range requests ...

pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -278,14 +278,10 @@ func (session *OcisSession) ConcatUploads(_ context.Context, uploads []tusd.Uplo
...
 	defer span.End()
-	n, err := session.Node(ctx)
-	if err != nil {
-		return err
-	}
 
 	// upload the data to the blobstore
 	_, subspan := tracer.Start(ctx, "WriteBlob")
-	err = session.store.tp.WriteBlob(n, session.binPath())
+	err = session.store.tp.WriteBlob(session.SpaceID(), session.ID(), session.Size(), session.binPath())
```
butonic, line 284: @aduffeck this is the actual change ... but I need to check what happens after this ... or betdter yet: write a test ;-)

butonic:
When fixing this it opens up the possibility of metadata corruption. The following test will fail when listing the directory after the first upload has finished postprocessing with outcome delete (or abort) because the file is completely deleted.
```go
		It("correctly calculates the size when the second upload is finishes, even if first is deleted", func() {
			// finish postprocessing of second upload
			con <- events.PostprocessingFinished{
				UploadID: secondUploadID,
				Outcome:  events.PPOutcomeContinue,
			}
			// wait for upload to be ready
			ev, ok := (<-pub).(events.UploadReady)
			Expect(ok).To(BeTrue())
			Expect(ev.Failed).To(BeFalse())

			// check processing status
			resources, err := fs.ListFolder(ctx, rootRef, []string{}, []string{})
			Expect(len(resources)).To(Equal(1))

			item := resources[0]
			Expect(item.Path).To(Equal(ref.Path))
			Expect(utils.ReadPlainFromOpaque(item.Opaque, "status")).To(Equal(""))

			// size should match the second upload
			Expect(item.Size).To(Equal(uint64(len(file2Content))))

			// finish postprocessing of first upload
			con <- events.PostprocessingFinished{
				UploadID: uploadID,
				//				Outcome:  events.PPOutcomeDelete, // This will completely delete the file
				Outcome: events.PPOutcomeAbort, // This as well ... fck
			}
			// wait for upload to be ready
			ev, ok = (<-pub).(events.UploadReady)
			Expect(ok).To(BeTrue())
			Expect(ev.Failed).To(BeTrue())

			// check processing status
			resources, err = fs.ListFolder(ctx, rootRef, []string{}, []string{})
			Expect(len(resources)).To(Equal(1))

			// size should still match the second upload
			Expect(item.Size).To(Equal(uint64(len(file2Content))))
		})
```

I need to add logic that prevents that ... and correctly calculates the size. I'm still worried the sizeDiff is unapplied wrong in corner cases.

butonic:
@micbar we should backport this to ocis 5

pkg/storage/utils/decomposedfs/upload/upload.go
butonic: added an async test 💪

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -309,9 +309,16 @@ func (fs *Decomposedfs) Postprocessing(ch <-chan events.Event) {
...
 			if failed {
-				// propagate sizeDiff after failed postprocessing
-				if err := fs.tp.Propagate(ctx, n, -session.SizeDiff()); err != nil {
-					log.Error().Err(err).Str("uploadID", ev.UploadID).Msg("could not propagate tree size change")
+				// if no other upload session is in progress (processing id != session id) or has finished (processing id == "")
+				latestSession, err := n.ProcessingID(ctx)
+				if err != nil {
+					log.Error().Err(err).Str("node", n.ID).Str("uploadID", ev.UploadID).Msg("reading node for session failed")
+				}
+				if latestSession == session.ID() {
```
aduffeck, line 317: Hm, why do we only revert the size diff if the failed session is the latest session? At a first glance we seem to unconditionally apply the size diff for all uploads, no?

aduffeck: oh, I see. The size diff that is propagated before the postprocessing starts is calculated against the current size of the node.

hm... which could be either the original size or the size of the currently being processed upload, depending on the timing. Ugh...

butonic: When initiating an upload we always update the file size. If a new upload is initiated after postprocessing of the first started but before it ended we must not change the size back, or the size of the node will be wrong because the new upload session has a size diff that is based on the size of the node. so we must not revert it.

pkg/storage/fs/posix/blobstore/blobstore.go
```diff
@@ -109,14 +116,17 @@ func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
 
-func (bs *Blobstore) path(node *node.Node) (string, error) {
-	if node.BlobID == "" {
-		return "", fmt.Errorf("blobstore: BlobID is empty")
+func (bs *Blobstore) path(spaceID, blobID string) (string, error) {
```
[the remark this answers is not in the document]
butonic: IMO the blobstore interface is way to constrained. We already have a hack in the ocis filestore implementation that 'moves' the file. s3ng could use a reader. posix might also want to move .. well it has to copy the bytes because of the user id.

Taking downloads into account we may want to forward range request headers to the actual s3 backend.

So maybe a parameter struct that carries the different properties makes sense?

pkg/storage/utils/decomposedfs/upload/upload.go
butonic: This change only affects the positive case when an upload has to be transferred to the blobstore. This has no effect on failure cases.

aduffeck (approved):

micbar:
@butonic Please backport, that was not released yet.

PR #4622, opened 2024-04-10
fix: permission check in public share update

micbar:
# Description

We fixed the permission check for updating public shares. When updating the permissions of a public share while not providing a password, the check must be against the new permissions to take into account that users can opt out only for view permissions.

relates to https://github.com/owncloud/enterprise/issues/6558

### DIFF
--- a/changelog/unreleased/fix-public-share-update.md
+++ b/changelog/unreleased/fix-public-share-update.md
@@ -1,0 +1,5 @@
+Bugfix: Fix public share update
+
+We fixed the permission check for updating public shares. When updating the permissions of a public share while not providing a password, the check must be against the new permissions to take into account that users can opt out only for view permissions.
+
+https://github.com/cs3org/reva/pull/4622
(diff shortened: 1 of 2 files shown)


micbar:
ready for review.

2403905 (approved):

rhafer (approved):
LGTM

(it really sucks that the publicshareprovider doesn't have any unit tests)

micbar:
> (it really sucks that the publicshareprovider doesn't have any unit tests)
I am on it. We can also wait for them.

micbar:
Hm, @butonic The publicsharemanager is currently not testable, it creates its own gatewaySelector. We would need to refactor that to make it testable.

micbar:
> Cool 🎉 . I am fine to merge this as is and have another PR for the unit test.

Fine. I need to refactor the New() Method which creates a lot of noise. So i would merge that now here, backport to stable and then refactor it on `edge`.

PR #4625, opened 2024-04-11
test async processing cornercases

butonic:
We added tests to cover several bugs where file metadata or parent treesize might get corrupted when postprocessing errors occur in specific order.
For now, the added test cases test the current behavior but contain comments and FIXMEs for the expected behavior.

The PR adds eight new test cases:
- the first can succeed before the second succeeds - Works!
- the first can succeed after the second succeeds - Works!
- the first can succeed before the second fails - Works!
- the first can succeed after the second fails - the file should still be in "processing" after second upload is deleted
- the first can fail before the second succeeds - list of revisions should be 0
- the first can fail after the second succeeds - list of revisions should be 0
- the first can fail before the second fails - file should be deleted, size should be correctly reverted
- the first can fail after the second fails - file should be deleted, size should be correctly reverted

I'd like to get this in so it does not get lost in an spreadsheet on my desktop.

@aduffeck And I are entertaining the idea of creating the note on the initiateFile call and keep track of all ongoing uploads in the node metadata. It will allow resolving the correct size diff and processing status. Will create an issue to track that.

### DIFF
--- a/pkg/storage/utils/decomposedfs/upload_async_test.go
+++ b/pkg/storage/utils/decomposedfs/upload_async_test.go
@@ -62,8 +62,8 @@ var _ = Describe("Async file uploads", Ordered, func() {
 			Username: "username",
 		}
 
-		fileContent  = []byte("0123456789")
-		file2Content = []byte("01234567890123456789")
+		firstContent  = []byte("0123456789")
+		secondContent = []byte("01234567890123456789")
 
 		ctx = ruser.ContextSetUser(context.Background(), user)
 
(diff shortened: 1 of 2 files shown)


pkg/storage/utils/decomposedfs/upload_async_test.go
```diff
@@ -532,5 +532,472 @@ var _ = Describe("Async file uploads", Ordered, func() {
...
+			// wait for upload to be ready
+			ev, ok = (<-pub).(events.UploadReady)
+			Expect(ok).To(BeTrue())
+			Expect(ev.Failed).To(BeTrue())
+
+			// check file has been removed
+			// if all uploads have been processed with outcome delete -> delete the file
+			resources, err = fs.ListFolder(ctx, rootRef, []string{}, []string{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(resources)).To(Equal(1)) // FIXME this should be 0, but we are not deleting the resource
```
aduffeck, line 944: How about writing the correct expectations but marking the tests as pending (`PIt`) until the issues are fixed? I don't really feel comfortable expecting the wrong behavior.

pkg/storage/utils/decomposedfs/upload_async_test.go
```diff
@@ -532,5 +532,472 @@ var _ = Describe("Async file uploads", Ordered, func() {
...
+		It("the first can succeed before the second succeeds", func() {
+			// finish first postprocessing
+			con <- events.PostprocessingFinished{
+				UploadID: uploadID,
+				Outcome:  events.PPOutcomeContinue,
+			}
+			// wait for upload to be ready
+			ev, ok := (<-pub).(events.UploadReady)
+			Expect(ok).To(BeTrue())
+			Expect(ev.Failed).To(BeFalse())
```
aduffeck, line 545: The tests itself look good. But maybe we could DRY the tests up a bit using some helper methods, so this could turn into something like 
```suggestion
			// finish first postprocessing
			succeedPostprocessing(uploadID);
			
			// or
			failPostprocessing(uploadID);
```
?

pkg/storage/utils/decomposedfs/upload_async_test.go
```diff
@@ -532,5 +532,472 @@ var _ = Describe("Async file uploads", Ordered, func() {
...
+
+			// check processing status
+			resources, err := fs.ListFolder(ctx, rootRef, []string{}, []string{})
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(resources)).To(Equal(1))
+
+			item := resources[0]
```
aduffeck, line 552: DRY this up as well?

pkg/storage/utils/decomposedfs/upload_async_test.go
butonic: I expect the right values now, but commented thee lines and added a FIXME explaining what needs to be done

aduffeck (approved):
lgtm

PR #4630, opened 2024-04-15
use share id to identify ocm shares

butonic:
We now use the share id to correctly identify ocm shares.

### DIFF
--- a/changelog/unreleased/fix-ocm-share-id.md
+++ b/changelog/unreleased/fix-ocm-share-id.md
@@ -1,0 +1,5 @@
+Bugfix: fix ocm-share-id
+
+We now use the share id to correctly identify ocm shares.
+
+https://github.com/cs3org/reva/pull/4630
--- a/internal/grpc/interceptors/auth/scope.go
+++ b/internal/grpc/interceptors/auth/scope.go
@@ -89,7 +89,7 @@ func expandAndVerifyScope(ctx context.Context, req interface{}, tokenScope map[s
 					return nil
 				}
 			}
-			log.Err(err).Msgf("error resolving reference %s under scope %+v", ref.String(), k)
+			log.Err(err).Interface("ref", ref).Interface("scope", k).Msg("error resolving reference under scope")
 		}
 
 	} else if ref, ok := extractShareRef(req); ok {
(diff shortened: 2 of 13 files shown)


butonic:
I think some of the confusion comes from comments like https://github.com/cs3org/OCM-API/pull/57#issuecomment-1403557978 which do not have a path property in the url that can identify the ocm share. In fact the sharedSecret / token is required to determine what content is to be served under `/remote.php/webdav`, which violates URLs as resource identitifiers in the first place. I thought we had moved away from that with `/dav/files` and `/dav/public-files` in oc10 ... 

Well, with spaces at least we can finally provide actual URIs ... `/dav/ocm/{ocm-share-id}` is a proper URI ... we could even migrate that to `/dav/spaces/{ocmproviderid}!{ocmspaceid}${ocm-shareid}` ... maybe even use more than one ocmproviderid ... to directly route to the correct instance ... well ... maybe in the future. let us not go there now and stick to `/dav/ocm`.

Anyway, our `/dav/ocm/` endpoint needs to learn bearer auth.

butonic:
@aduffeck I think this clarifies some of the confusion around authenticating OCM webdav requests. Do you remember how you set this up to test this ... The unit tests don't fully cover this. Needs to be added ...

butonic:
@aduffeck shouldn't this condition be inverted:
```go
func (s *service) GetOCMShare(ctx context.Context, req *ocm.GetOCMShareRequest) (*ocm.GetOCMShareResponse, error) {
	// if the request is by token, the user does not need to be in the ctx
	var user *userpb.User
	if req.Ref.GetToken() == "" {
		user = ctxpkg.ContextMustGetUser(ctx)
	}
```

aduffeck:
> @aduffeck shouldn't this condition be inverted:
> 
> ```go
> func (s *service) GetOCMShare(ctx context.Context, req *ocm.GetOCMShareRequest) (*ocm.GetOCMShareResponse, error) {
> 	// if the request is by token, the user does not need to be in the ctx
> 	var user *userpb.User
> 	if req.Ref.GetToken() == "" {
> 		user = ctxpkg.ContextMustGetUser(ctx)
> 	}
> ```

I don't think so. If no token is set (`req.Ref.GetToken() == ""`) we require a user because we end up at https://github.com/cs3org/reva/blob/edge/pkg/ocm/share/repository/json/json.go#L309, otherwise we don't because we return at https://github.com/cs3org/reva/blob/edge/pkg/ocm/share/repository/json/json.go#L299 already.

aduffeck (approved):
:+1:

butonic:
To fix the ci I had to set up a debuggable environment ... and then I ran into https://github.com/owncloud/ocis/issues/8564: sharingNG does not support OCM, yet.

Some progress is in TODO, but it needs more work.

I'll commit the changes I made while dabbling with the setup to continue on it next week...

PR #4635, opened 2024-04-16
fix nil pointer when removing groups from space

2403905:
Related to an issue [#8768](https://github.com/owncloud/ocis/issues/8768)
Fixed the nil pointer when the function argument is incorrect.


kobergj (approved):
Nice :+1:

PR #4651, opened 2024-04-22
fix deleting space shares

butonic:
We no longer chech if a share is an ocm sharee if listng ocm shares has been disabled anyway. This allows unsharing space shares.


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -295,9 +295,14 @@ func (h *Handler) isPublicShare(r *http.Request, oid string) (*link.PublicShare,
...
-	if err != nil {
-		logger.Err(err)
+	switch {
+	case err != nil:
+		log.Err(err).Send()
 		return nil, false
+	case psRes.Status.Code == rpc.Code_CODE_OK:
+		fallthrough
+	case psRes.Status.Code == rpc.Code_CODE_NOT_FOUND:
+		log.Error().Str("message", psRes.GetStatus().GetMessage()).Str("code", psRes.GetStatus().GetCode().String()).Msg("isPublicShare received unexpected status")
```
butonic, line 305: Thasa two might be too restrictive but woe could change this to only log on internal errors ...

tbsbdr:
@butonic is this still needed?

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
kobergj: Good idea. Internal Errors should be logged on error level. We can log the others on info or debug level but I would like not to spam the error log with unnecessary messages.

So: Yes, please only log internal errors on error level

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
```diff
@@ -147,9 +147,14 @@ func (h *Handler) isUserShare(r *http.Request, oid string) (*collaboration.Share
...
 	})
-	if err != nil {
-		logger.Err(err)
+	switch {
+	case err != nil:
+		log.Err(err).Send()
 		return nil, false
+	case getShareRes.Status.Code == rpc.Code_CODE_OK:
+		fallthrough
+	case getShareRes.Status.Code == rpc.Code_CODE_NOT_FOUND:
```
kobergj, line 156: same here

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
```diff
@@ -172,9 +177,14 @@ func (h *Handler) isFederatedShare(r *http.Request, shareID string) bool {
...
 		},
 	})
-	if err != nil {
+	switch {
+	case err != nil:
 		log.Err(err).Send()
 		return false
+	case getShareRes.Status.Code == rpc.Code_CODE_OK:
+		fallthrough
+	case getShareRes.Status.Code == rpc.Code_CODE_NOT_FOUND:
```
kobergj, line 186: same here

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
```diff
@@ -247,9 +257,14 @@ func (h *Handler) isFederatedReceivedShare(r *http.Request, shareID string) bool
...
```
kobergj, line 266: same here

kobergj (changes requested):

kobergj (approved):

PR #4653, opened 2024-04-23
Replacement for TokenInfo endpoint

2403905:
Related issue [https://github.com/owncloud/ocis/issues/8858](https://github.com/owncloud/ocis/issues/8858)

### DIFF
--- a/changelog/unreleased/tokenInfo-endpoint-replacement.md
+++ b/changelog/unreleased/tokenInfo-endpoint-replacement.md
@@ -1,0 +1,12 @@
+Enhancement: Allow to resolve public shares without the ocs tokeninfo endpoint
+
+      
+Instead of querying the /v1.php/apps/files_sharing/api/v1/tokeninfo/ endpoint, a client can now resolve public and internal links by sending a PROPFIND request to /dav/public-files/{sharetoken}
+
+* authenticated clients accessing an internal link are redirected to the "real" resource (`/dav/spaces/{target-resource-id}
+* authenticated clients are able to resolve public links like before. For password protected links they need to supply the password even if they have access to the underlying resource by other means.
+* unauthenticated clients accessing an internal link get a 401 returned with  WWW-Authenticate set to Bearer (so that the client knows that it need to get a token via the IDP login page.
+* unauthenticated clients accessing a password protected link get a 401 returned with an error message to indicate the requirement for needing the link's password.
+
+https://github.com/cs3org/reva/pull/4653  
+https://github.com/owncloud/ocis/issues/8858
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -20,6 +20,7 @@ package ocdav
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"path"
 	"path/filepath"
@@ -28,6 +29,7 @@ import (
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	userv1beta1 "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/config"
 	"github.com/cs3org/reva/v2/internal/http/services/owncloud/ocdav/errors"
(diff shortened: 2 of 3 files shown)


internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -316,12 +347,20 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
-			ctx := ContextWithTokenStatInfo(ctx, sRes.Info)
-			r = r.WithContext(ctx)
-			if sRes.Info.Type != provider.ResourceType_RESOURCE_TYPE_CONTAINER {
-				h.PublicFileHandler.Handler(s).ServeHTTP(w, r)
+			if isInternal {
+				base := path.Join(ctx.Value(net.CtxKeyBaseURI).(string), "spaces")
+				ctx := context.WithValue(ctx, net.CtxKeyBaseURI, base)
+				r = r.WithContext(ctx)
+				r.URL.Path = "/" + storagespace.FormatResourceID(*sRes.GetInfo().GetId())
+				h.SpacesHandler.Handler(s, nil).ServeHTTP(w, r)
```
rhafer, line 355: Hm, I guess this is a bit of a misunderstanding. By:

"authenticated clients accessing a pubic link (password protected or not) for a resource they already have access to are also redirected to the "real" resource. (and always need to supply the password)"

With "redirected". We were indeed referring to a HTTP Redirect response (Status 30x) to  `/dav/files/<fileid>`. It wasn't the intention to just forward the request internally. (Unless something with that has changed.

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -286,6 +328,16 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				return
 			}
 
+			if userExists {
+				// Build new context without an authenticated user for the public link
+				baseURI := ctx.Value(net.CtxKeyBaseURI).(string)
+				logger := appctx.GetLogger(ctx)
+				span := trace.SpanFromContext(ctx)
```
2403905, line 335: @butonic What do you think about copying the trace context this way?

changelog/unreleased/tokenInfo-endpoint-replacement.md
```diff
@@ -0,0 +1,12 @@
+Enhancement: Replacement for TokenInfo Endpoint
+
+      
+The client should basically always send a PROPFIND to /dav/public-files/{sharetoken}
+
+* authenticated clients accessing an internal link are redirected to the "real" resource (`/dav/spaces/{target-resource-id}
+* authenticated clients accessing a public link (password protected or not) for a resource they already have access to are also redirected to the "real" resource. (and always need to supply the password)
```
rhafer, line 7: I think this is wrong. We're currently not creating redirect responses when an authenticate client access a public link. But that is ok. So please just remove the item from this list.

changelog/unreleased/tokenInfo-endpoint-replacement.md
```diff
@@ -0,0 +1,12 @@
+Enhancement: Replacement for TokenInfo Endpoint
```
rhafer, line 1: How about "Allow to resolve public shares without the ocs `tokeninfo` endpoint"

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -269,12 +310,13 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 			case res.Status.Code == rpc.Code_CODE_PERMISSION_DENIED:
 				fallthrough
 			case res.Status.Code == rpc.Code_CODE_UNAUTHENTICATED:
-				w.WriteHeader(http.StatusUnauthorized)
 				if hasValidBasicAuthHeader {
+					w.WriteHeader(http.StatusUnauthorized)
```
rhafer, line 314: Why not keep it outside the `if` like is was before?

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -286,6 +328,16 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				return
 			}
 
+			if userExists {
+				// Build new context without an authenticated user for the public link
```
rhafer, line 332: Please add some more detail here, why we need to rebuild the context without the user. Just to help other people looking at the code.

changelog/unreleased/tokenInfo-endpoint-replacement.md
2403905: How about this wording?
`authenticated clients accessing a public link (password-protected or not) able to resolve public shares like before (need to supply the password for password-protected public link).`

rhafer: Hm, I can't quite parse that sentence. :thinking: I think there is is verb missing somewhere.

How about:
"authenticated clients are able to resolve public links like before. For password protected links they need to supply the password even if they have access to the underlying resource by other means."

rhafer (approved):

PR #4656, opened 2024-04-23
fix space share update

2403905:
Related Issue [ocis#8905](https://github.com/owncloud/ocis/issues/8905)

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -20,6 +20,7 @@ package gateway
 
 import (
 	"context"
+	"slices"
 
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	collaboration "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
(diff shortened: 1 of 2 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -169,32 +170,43 @@ func (s *svc) updateSpaceShare(ctx context.Context, req *collaboration.UpdateSha
...
-		Creator:     creator.GetId(),
-	}
 	if grants.PermissionsEqual(req.Share.GetPermissions().GetPermissions(), &provider.ResourcePermissions{}) {
 		st, err = s.denyGrant(ctx, req.GetShare().GetResourceId(), req.GetShare().GetGrantee(), opaque)
 		if err != nil {
 			return nil, errors.Wrap(err, "gateway: error denying grant in storage")
 		}
 	} else {
-		if !grant.Permissions.RemoveGrant {
+		listGrantRes, err := s.listGrants(ctx, req.GetShare().GetResourceId())
```
rhafer, line 179: Can we skip the `listGrants` Call if the updatemask contains `permissions` and `expiration`? AFAIK `listGrants` can be somewhat expensive.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -169,32 +170,43 @@ func (s *svc) updateSpaceShare(ctx context.Context, req *collaboration.UpdateSha
...
+		if err != nil {
+			return nil, errors.Wrap(err, "gateway: error getting grant to remove from storage")
+		}
+		existsGrant := s.getGranteeGrant(listGrantRes.GetGrants(), req.GetShare().GetGrantee())
```
rhafer, line 183: Hm, do we need to check if the incoming request actually has a `grantee` set? Is it possible that we get a UpdateShare request without a `Grantee`?

rhafer (changes requested):
Looks fine to me. Just some small questions.

rhafer (approved):

PR #4657, opened 2024-04-24
Add Scandata to upload session

kobergj:
Adds virus scan data to upload sessions

### DIFF
--- a/pkg/storage/utils/decomposedfs/upload/session.go
+++ b/pkg/storage/utils/decomposedfs/upload/session.go
@@ -298,7 +298,8 @@ func (s *OcisSession) MTime() time.Time {
 
 // IsProcessing returns true if all bytes have been received. The session then has entered postprocessing state.
 func (s *OcisSession) IsProcessing() bool {
-	return s.info.Size == s.info.Offset
+	// We might need a more sophisticated way to determine processing status soon
+	return s.info.Size == s.info.Offset && s.info.MetaData["scanResult"] == ""
 }
 
 // binPath returns the path to the file storing the binary data.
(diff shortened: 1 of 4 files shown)


2403905 (approved):

pkg/storage/utils/decomposedfs/upload/session.go
```diff
@@ -298,7 +298,8 @@ func (s *OcisSession) MTime() time.Time {
 
 // IsProcessing returns true if all bytes have been received. The session then has entered postprocessing state.
 func (s *OcisSession) IsProcessing() bool {
-	return s.info.Size == s.info.Offset
+	// We might need a more sophisticated way to determine processing status soon
+	return s.info.Size == s.info.Offset && s.info.MetaData["scanResult"] == ""
```
butonic, line 302: when bytes are ready this will also return true because the scanResult has not been written yet. We should populate it when all bytes have been received. Maybe with `"processing"`?

kobergj: > when bytes are ready this will also return true because the scanResult has not been written yet.

This is what it is supposed to do. It should only return false when a virus was detected. But you are right, there is small time gap in which either
- virusscan data was written but postprocessing is not yet finished
- postprocessing is finished but virusscan data is not yet written
I'd consider that "eventual consistent", as this time frame *should* be very small.

> Maybe with "processing"

I don't see how this would change the outcome? Instead of checking for `""` we would check for `"postprocessing"`. Nothing else changes. 

In general I would consider an item `"processing"` from the moment all bytes are received until either the upload session gets the deleted  (success case) or a virus occurs (error case)

butonic: got it, I had the logic flipped in my head.

PR #4659, opened 2024-04-25
Add caching in Gateway.ListExistingReceivedShares

glpatcern:


glpatcern:
Superseded by a simpler implementation: the HTTP OCS cache remains as is, another one was implemented in the GRPC Gateway, which at runtime shares the same content as the OCS one. Eventually, the latter will be removed once the frontend does not use it any longer.

PR #4660, opened 2024-04-25
fix(auth/scope): fix nested resource check when creating a new resource

rhafer:
When creating a reource (e.g a document via the app/new endpoint) below a next folder structure of a public link, we can't stat the resource itself (it doesn't exit yet) for checking if it is a descendant of the share root. We now stat the resource's parent instead in that case.

Fixes: https://github.com/owncloud/ocis/issues/8957

### DIFF
--- a/internal/grpc/interceptors/auth/scope.go
+++ b/internal/grpc/interceptors/auth/scope.go
@@ -266,7 +266,7 @@ func checkIfNestedResource(ctx context.Context, ref *provider.Reference, parent 
 	if err != nil {
 		return false, err
 	}
-	if statResponse.Status.Code != rpc.Code_CODE_OK {
+	if statResponse.GetStatus().GetCode() != rpc.Code_CODE_OK {
 		return false, statuspkg.NewErrorFromCode(statResponse.Status.Code, "auth interceptor")
 	}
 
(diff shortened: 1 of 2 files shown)


internal/grpc/interceptors/auth/scope.go
```diff
@@ -313,6 +313,14 @@ func checkIfNestedResource(ctx context.Context, ref *provider.Reference, parent
 	if err != nil {
 		return false, err
 	}
+	if childStat.Status.Code == rpc.Code_CODE_NOT_FOUND && ref.GetPath() != "" && ref.GetPath() != "." {
```
2403905, line 316: could we use the getters? childStat.GetStatus().GetCode()

internal/grpc/interceptors/auth/scope.go
```diff
@@ -313,6 +313,14 @@ func checkIfNestedResource(ctx context.Context, ref *provider.Reference, parent
...
+		// The resource does not seem to exist (yet?). We might be part of an initiate upload request.
+		// Stat the parent to get its path and check that against the root path.
+		childStat, err = client.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{ResourceId: ref.GetResourceId()}})
+		if err != nil {
+			return false, err
+		}
+	}
 	if childStat.Status.Code != rpc.Code_CODE_OK {
```
2403905, line 324: could we use the getters? childStat.GetStatus().GetCode()

2403905 (approved):

kobergj (approved):

PR #4670, opened 2024-04-30
OCM: support bearer token access and other refactoring

glpatcern:
- Adding support for bearer token auth when exposing OCM shares to remote users, following latest OCM developments, and similarly access remote OCM shares via bearer token. OCM 1.0 is still supported when remote users access Reva, but we still don't implement accessing remote OCM 1.0 shares via basic auth.
- OCM client incorporated in the `ocmd` package
- The "open" invite manager is now truly open for use without ScienceMesh
- Some comments/logs added for clarity
- Introduced `/.well-known` endpoint for OCM


glpatcern:
Merging this to allow further testing by @MahdiBaghbani in the scope of the NLnet-funded OCM activities. None of the changes in this PR are yet used in production at CERN, therefore I consider the PR safe to merge.

PR #4682, opened 2024-05-14
[full-ci] [tests-only]Ocis bump

S-Panta:
Ocis bump of PR https://github.com/owncloud/ocis/pull/9200

PR #4683, opened 2024-05-15
[tests-only] [full-ci] Bump ocis commit id

Salipa-Gurung:
Bumping ocis latest commit Id


individual-it:
I don't think the fork is correct, that's why there are many conflicts in the clone phase
image

Salipa-Gurung:
The bumping commit id and the CI issue wil be addressed in different pr https://github.com/cs3org/reva/pull/4682, so closing this one.

PR #4685, opened 2024-05-15
Support t and x in ACEs

butonic:
To support view only shares (dowload forbidden) we added t (read attrs) and x (directory traversal) permissions to the decomposed FS ACEs. 

The change is backwards compatible. an ACE with `r` permission will still grant Stat, getPath, InitiateFileDownload and ListContainer permissions. Newly written grants will explicitly set `t` and `x` permissions. If we need to be more strict, we could increase the version byte to `x02` (`x01` was used in the very early days) which can than unparse the `r` to only grant the InitiateFileDownload permission.

### DIFF
--- a/pkg/conversions/role.go
+++ b/pkg/conversions/role.go
@@ -322,6 +322,7 @@ func NewUploaderRole() *Role {
 			GetPath:            true,
 			CreateContainer:    true,
 			InitiateFileUpload: true,
+			ListContainer:      true,
 		},
 		ocsPermissions: PermissionCreate,
 	}
--- a/pkg/conversions/role.go
+++ b/pkg/conversions/role.go
@@ -48,7 +48,9 @@ const (
 	RoleFileEditor = "file-editor"
 	// RoleCoowner grants co-owner permissions on a resource.
 	RoleCoowner = "coowner"
-	// RoleUploader grants uploader permission to upload onto a resource.
+	// RoleEditorLite grants permission to upload and download to a resource.
+	RoleEditorLite = "editor-lite"
+	// RoleUploader grants uploader permission to upload onto a resource (no download).
 	RoleUploader = "uploader"
 	// RoleManager grants manager permissions on a resource. Semantically equivalent to co-owner.
 	RoleManager = "manager"
@@ -313,10 +315,10 @@ func NewCoownerRole() *Role {
 	}
 }
 
-// NewUploaderRole creates an uploader role
-func NewUploaderRole() *Role {
+// NewEditorLiteRole creates an editor-lite role
+func NewEditorLiteRole() *Role {
 	return &Role{
-		Name: RoleUploader,
+		Name: RoleEditorLite,
 		cS3ResourcePermissions: &provider.ResourcePermissions{
 			Stat:                 true,
 			GetPath:              true,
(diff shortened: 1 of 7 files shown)


pkg/storage/utils/ace/ace.go
```diff
@@ -436,10 +440,17 @@ func unmarshalKV(s string) (*ACE, error) {
...
 	var b strings.Builder
 
-	if set.Stat || set.InitiateFileDownload || set.ListContainer || set.GetPath {
+	if set.Stat || set.GetPath {
+		b.WriteString("t")
+	}
+	if set.ListContainer { // we have no dedicated traversal permission, but to listing a container allows traversing it
+		b.WriteString("x")
+	}
+	if set.InitiateFileDownload || set.ListContainer {
```
rhafer, line 453: Hm, I don't think we can do this for `ListContainer`. This will result in download permissions for files in folder share with "secureview" permission.

rhafer (dismissed):

pkg/storage/utils/ace/ace.go
butonic: fixed

pkg/storage/utils/decomposedfs/grants_test.go
```diff
@@ -50,9 +50,10 @@ var _ = Describe("Grants", func() {
 				},
 			},
 			Permissions: &provider.ResourcePermissions{
-				Stat:   true,
-				Move:   true,
-				Delete: false,
+				Stat:                 true,
+				Move:                 true,
+				Delete:               false,
+				InitiateFileDownload: true,
```
butonic, line 56: file download is needed to produce the `r`

pkg/storage/utils/decomposedfs/grants_test.go
```diff
@@ -130,7 +131,7 @@ var _ = Describe("Grants", func() {
 				Expect(err).ToNot(HaveOccurred())
 				attr, err := n.XattrString(env.Ctx, prefixes.GrantUserAcePrefix+grant.Grantee.GetUserId().OpaqueId)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(attr).To(Equal(fmt.Sprintf("\x00t=A:f=:p=rw:c=%s:e=0\n", o.GetOpaqueId()))) // NOTE: this tests ace package
+				Expect(attr).To(Equal(fmt.Sprintf("\x00t=A:f=:p=trw:c=%s:e=0\n", o.GetOpaqueId()))) // NOTE: this tests ace package
```
butonic, line 134: `t` is now added to track the Stat permission

rhafer:
Created https://github.com/owncloud/ocis/pull/9188 to check this with ocis full-ci

rhafer:
@butonic I've now introduced the new role "UploaderOnly". This is needed by ocs to be able to create filedrop links.

I am still pondering if it might be better to keep the "Uploader" roles as it was before and introduce a new role instead for the "Upload/Download/Edit without Delete" thing that is supposed to be provided for user shares.

pkg/conversions/role.go
```diff
@@ -524,6 +543,9 @@ func RoleFromResourcePermissions(rp *provider.ResourcePermissions, islink bool)
 		}
 	}
 	if r.ocsPermissions == PermissionCreate {
+		if rp.GetPath && rp.InitiateFileDownload && rp.ListContainer && rp.Move {
+			r.Name = RoleEditorLite
+		}
```
kobergj, line 548:
```suggestion
		if rp.GetPath && rp.InitiateFileDownload && rp.ListContainer && rp.Move {
			r.Name = RoleEditorLite
			return r
		}
```
Otherwise `r.Name` is overwritten in the line below.

kobergj (changes requested):

pkg/conversions/role.go
butonic: hm ... I don't see a test covering RoleFromResourcePermissions ... should we add them?

kobergj: Feel free to add as many unit tests as you like :smiley:

rhafer: `RoleFromResourcePermissions` in general feels like it could use some overhaul. Why doing this via settting the ocsPermission first an then "guess" the role name from that. That procedure seems somewhat lossy.

I don't dare touching this currently. But we should re-work that thing once we can get rid of the ocsPermission stuff.

PR #4688, opened 2024-05-17
fix: convert dates for propfinds from RFC3339 (ocis) to RFC1123 (ocdav)

kulmann:
While oCIS uses RFC3339 time formats internally, we use RFC1123 time formats for webdav responses. Since oCIS doesn't know about the format, we now convert the time strings that arrive in reva from RFC3339 to RFC1123.

### DIFF
--- a/changelog/unreleased/add-photo-and-image-props.md
+++ b/changelog/unreleased/add-photo-and-image-props.md
@@ -3,3 +3,4 @@ Enhancement: Add photo and image props
 Add `oc:photo` and `oc:image` props to PROPFIND responses for propall requests or when they are explicitly requested.
 
 https://github.com/cs3org/reva/pull/4684
+https://github.com/cs3org/reva/pull/4688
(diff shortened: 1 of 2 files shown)


changelog/unreleased/add-photo-and-image-props.md
```diff
@@ -3,3 +3,4 @@ Enhancement: Add photo and image props
 Add `oc:photo` and `oc:image` props to PROPFIND responses for propall requests or when they are explicitly requested.
 
 https://github.com/cs3org/reva/pull/4684
+https://github.com/cs3org/reva/pull/4688
```
kobergj, line 6:
```suggestion
https://github.com/cs3org/reva/pull/4688
https://github.com/cs3org/reva/pull/4684
```
Not sure what causes this to fail, but maybe wrong order?

kobergj (approved):
Yes. Let's go with that option for now. We should think about how we want to handle this in the future.

PR #4695, opened 2024-05-23
refactor: changed from alllowercase to kebab-case for metadata keys

kulmann:
Changed from alllowercase to kebab-case for metadata keys in propfind responses.

Otherwise it's impossible to convert to a different case in an automated way.


changelog/unreleased/add-photo-and-image-props.md
```diff
@@ -3,3 +3,4 @@ Enhancement: Add photo and image props
 Add `oc:photo` and `oc:image` props to PROPFIND responses for propall requests or when they are explicitly requested.
 
 https://github.com/cs3org/reva/pull/4684
+https://github.com/cs3org/reva/pull/4695
```
kobergj, line 6:
```suggestion
https://github.com/cs3org/reva/pull/4695
https://github.com/cs3org/reva/pull/4684
```

I think it wants the current one up.

kobergj (changes requested):
Are we sure we can change that? Wouldn't that break backwards compatibility?

kulmann:
> Are we sure we can change that? Wouldn't that break backwards compatibility?

It's only for the metadata that @dschmidt and I introduced (audio, location, image, photo) and that is unused so far. But yes, it's a breaking change... don't know how to continue here. At least for `image` and `photo` the code was never released, so it should be safe to change that. `audio` and `location` have been in a release before though, so meh...

internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -1147,19 +1148,11 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
 	appendMetadataProp := func(metadata map[string]string, tagNamespace string, name string, metadataPrefix string, keys []string) {
 		content := strings.Builder{}
 		for _, key := range keys {
-			lowerCaseKey := strings.ToLower(key)
+			kebabCaseKey := strcase.ToKebab(key)
```
butonic, line 1151: The keys are hardcoded above. Why even lower or kebap case them? IMO `isVariableBitrate` should just remain camel case. Any reasons against that?

kulmann: @kobergj wanted to achieve a somewhat consistent format for keys in the propfind response. From a client perspective that is a good thing to achieve. A mix of kebab, camel and snake case in the keys is annoying as hell... 

However, my strongest need is, that I can convert the keys easily to a different case so that they match an interface of mine. Types/interfaces in web are all camelCase, so I need to be able to convert as needed.

internal/http/services/owncloud/ocdav/propfind/propfind.go
```diff
@@ -1147,19 +1148,11 @@ func mdToPropResponse(ctx context.Context, pf *XML, md *provider.ResourceInfo, p
...
 			if v, ok := metadata[fmt.Sprintf("%s.%s", metadataPrefix, key)]; ok {
-				content.WriteString("<")
-				content.WriteString(tagNamespace)
-				content.WriteString(":")
-				content.WriteString(lowerCaseKey)
-				content.WriteString(">")
+				content.WriteString(fmt.Sprintf("<%s:%s>", tagNamespace, kebabCaseKey))
```
dschmidt, line 1153: I haven't benchmarked it myself, but I assume this is slower than the previous code... it's not an incredible hot path but also not totally cold either

kulmann: meh, thanks for pointing that out... will change it back. I was going for better readability. ;-) But yes, that's less important of course.

kobergj (approved):
Only thing I'm concerned about is backwards compatibility. Let's maybe not backport it to `5.0` to be safe?

PR #4699, opened 2024-05-27
Export Path method in blobstores

kobergj:
Exports the `Path` method in ocis and s3ng blobstores


butonic:
um, to what end? you cannot download anything from the blobstore without going through the dataprovider ...

kobergj:
We need it for https://github.com/owncloud/ocis/pull/9238 
There is a requirement that we print the path of the blob in the blobstore.

butonic (approved):

PR #4702, opened 2024-05-28
Theme capability

fschade:
The ocs capabilities now contain a theme capability that exposes theme related configuration options to the clients.
As part of this change, the ocs capabilities are now exposed and can be used externally.

### DIFF
--- a/changelog/unreleased/add-theme-capability.md
+++ b/changelog/unreleased/add-theme-capability.md
@@ -1,0 +1,6 @@
+Enhancement: Added theme capability
+
+The ocs capabilities now contain a theme capability that exposes theme related configuration options to the clients.
+As part of this change, the ocs capabilities are now exposed and can be used externally.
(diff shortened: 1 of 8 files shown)


butonic:
oh well ... ocs ... 

in ms graph there is an [`organizationalBranding`](https://learn.microsoft.com/en-us/graph/api/resources/organizationalbranding?view=graph-rest-1.0) resource type to customizee the appearance of the login page. AFAICT getting it requires an [organization id](https://learn.microsoft.com/en-us/graph/api/resources/organization?view=graph-rest-1.0), which we currently do not have:
```http
GET /organization/{organizationId}/branding
```
It even is localized:
```http
GET /organization/{organizationId}/branding/localizations/{organizationalBrandingLocalizationId}
```

oh well ... someday ...

butonic (approved):

PR #4709, opened 2024-06-03
fixed the response code when the role/permission is empty on the share update

2403905:
We fixed the response code when the role/permission is empty on the share update

Related issue: https://github.com/owncloud/ocis/issues/8747


rhafer (approved):
Looks good to me. (Tiny nitpick: For the future it would be good if you linked the fixed issue in the commit message as well)

kobergj (approved):

PR #4710, opened 2024-06-04
fix: space listing when user has no personal space

micbar:
Bugfix: Tolerate missing user space index

We fixed a bug where the spaces for a user were not listed if the user had no space index by user. This happens when a user has the role "User Light" and has been invited to a project space via a group.


kobergj (approved):
Nice :+1: can confirm it works

PR #4711, opened 2024-06-05
fix: duplicate headers in DAV responses

micbar:
# Description

We fixed an issue where the DAV response headers were duplicated. This was caused by the WebDav handler which copied over all headers from the datagateways response. Now, only the relevant headers are copied over to the DAV response to prevent duplication.

Relates to https://github.com/owncloud/ocis/issues/9219

### DIFF
--- a/internal/http/services/owncloud/ocdav/get.go
+++ b/internal/http/services/owncloud/ocdav/get.go
@@ -131,7 +131,19 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	}
 	defer httpRes.Body.Close()
 
-	copyHeader(w.Header(), httpRes.Header)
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocdav/get.go
```diff
@@ -131,7 +131,16 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	}
 	defer httpRes.Body.Close()
 
-	copyHeader(w.Header(), httpRes.Header)
+	// copy only the headers relevant for the content served by the datagateway
```
fschade, line 134: can u add a check if the headers exist and only set the value if it exists?

micbar: done.

fschade (approved):
🎸

PR #4714, opened 2024-06-06
enhancement:  unique mount point path

fschade:
This PR introduces a new feature that ensures all mounted shares have a unique path. This is achieved by checking for potential name collisions against existing mount points and adjusting the mount point name if necessary.

Overview:

- Check if the requested mount point is available and if not, finds a suitable one.
- If no mount point is provided, it finds a it generated a suitable mount point based on the name
- If the mount point already exists, a number is inserted into the filename to make it unique.

This enhances the robustness of the share mounting process and prevents potential issues caused by duplicate mounts.

## before
Screenshot 2024-06-10 at 12 57 53

## after
Screenshot 2024-06-10 at 12 59 41


internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
+		})
+		switch {
+		case err != nil:
+			fallthrough
+		case resourceStat.GetStatus().GetCode() != rpc.Code_CODE_OK:
+			return &collaboration.UpdateReceivedShareResponse{
+				Status: resourceStat.GetStatus(),
+			}, err
+		}
+
```
kobergj, line 569: You could use `utils.GetResource` instead of all these lines.

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
+	if err != nil {
+		return "", errtypes.InternalError("grpc list received shares request failed")
 	}
-	return res, nil
+
+	if err := errtypes.NewErrtypeFromStatus(listReceivedSharesRes.GetStatus()); err != nil {
+		return "", err
+	}
+
+	// we need to sort the received shares by mount point in order to make things easier to evaluate.
```
kobergj, line 630: Where do we do this? Can we put this comment where the actual sort is made?

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
+		if s.State == collaboration.ShareState_SHARE_STATE_ACCEPTED {
+			// collect all accepted mount points
+			mountedShares = append(mountedShares, s.GetMountPoint().GetPath())
+			if s.GetMountPoint().GetPath() == mount {
+				// does the shared resource still exist?
+				_, err := utils.GetResourceByID(ctx, s.GetShare().GetResourceId(), gwc)
+				if err == nil {
+					pathExists = true
+				}
+				// TODO we could delete shares here if the stat returns code NOT FOUND ... but listening for file deletes would be better
```
kobergj, line 657: This `TODO` sounds urgent. Is it not?

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
+	if pathExists {
+		// now we have a list of shares, we want to iterate over all of them and check for name collisions agents a mount points list
+		for i := 1; i <= len(mountedShares)+1; i++ {
+			ext := filepath.Ext(base)
+			name := strings.TrimSuffix(base, ext)
+			// be smart about .tar.(gz|bz) files
+			if strings.HasSuffix(name, ".tar") {
+				name = strings.TrimSuffix(name, ".tar")
+				ext = ".tar" + ext
+			}
```
kobergj, line 677: I don't quite understand this. What is so special with `.tar` files? Could you brievly explain?

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/export_test.go
```diff
@@ -0,0 +1,3 @@
+package shares
+
+var GetUnmountedShares = getUnmountedShares
```
kobergj, line 3: What kind of black magic is this? What are we trying to achieve here?

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
-	if req.Share.Share.Id == nil {
+
+	if req.GetShare().GetShare().GetId() == nil {
 		return &collaboration.UpdateReceivedShareResponse{
 			Status: status.NewInvalid(ctx, "share id missing"),
 		}, nil
 	}
-	if req.Share.Share.Id.OpaqueId == "" {
+
+	if req.GetShare().GetShare().GetId().GetOpaqueId() == "" {
```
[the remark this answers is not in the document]
fschade: i wanted to not touch the old implementation, but ur right, gone

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: utils.GetResourceByID returns an error instead of the status...

```go
...
if err := checkStatusCode("getting resource", res.GetStatus().GetCode()); err != nil {
  return nil, err
}
...
```

i dont like that approach, since in my opinion the rpc endpoint should only if something network related or similar is off... what do you think?

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: hehe, i have no idea, looks like its a leftover:
https://github.com/cs3org/reva/blob/40abb344337b95e2af8f3384fbd73c87f4ffdb9a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go#L130

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: same, its something from the old code which is already in use, i would like to keep it!
https://github.com/cs3org/reva/blob/40abb344337b95e2af8f3384fbd73c87f4ffdb9a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go#L162

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: resources with a suffix needs the counter before the suffix, but true, it is not tar related, same is for zip, .psd, ...

https://github.com/cs3org/reva/blob/40abb344337b95e2af8f3384fbd73c87f4ffdb9a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/pending.go#L173

needs a followup

internal/grpc/services/usershareprovider/usershareprovider.go
kobergj: I prefer simplicity. We don't need to check explicitly for a network error here. Both is just an (unrecoverable) error for us. So I would opt for using `utils.GetResource`. 

But not critical of course if you prefer the other way.

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/export_test.go
fschade: i dont want to make the getUnmountedShares public, so... just expose it for the test which is not part of the artifact later on, similar to the go core:

https://github.com/golang/go/blob/master/src/fmt/export_test.go

internal/grpc/services/usershareprovider/usershareprovider.go
kobergj: I dont see the point. I would just remove it.

If you want to keep it, please put a `TODO` in front.

internal/grpc/services/usershareprovider/usershareprovider.go
kobergj: How can we make sure this isn't forgotten? This will be a nasty bug (should it ever occur).

internal/grpc/services/usershareprovider/usershareprovider.go
rhafer: I think that (deleting shares for no longer existing resources) is a topic for a different PR. I'd rather prefer not to open that can of worms here. :worm: :worm:

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: i start working on it right after that one is merged... but to be honest, this sits here since ages which makes me unhappy because your 100% right!

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: nah, its gone, i see no reason to have it sorted...

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,181 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
 	}
-	return res, nil
+
+	if err := errtypes.NewErrtypeFromStatus(listReceivedSharesRes.GetStatus()); err != nil {
+		return "", err
+	}
+
+	// we need to sort the received shares by mount point in order to make things easier to evaluate.
+	base := filepath.Clean(name)
+	mount := base
```
rhafer, line 614: why not rename `base` to `mount` in the first place?

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: i would like to keep it as it is... since a status that is off is not an error in my opinion

internal/grpc/services/usershareprovider/usershareprovider.go
rhafer: I guess this is because it has 2 file extension? `tar.<dot-something>` ?

Still, I don't think we should do this. I know it has been this way in ocs but is wrong in so many ways ..

rhafer (dismissed):

internal/grpc/services/usershareprovider/usershareprovider.go
fschade: good point done

fschade:
@rhafer is right with the strange tar handling, i removed it, lets see what the ci is saying, but i want to revisit this in a sec... i have a idea

fschade:
> @rhafer is right with the strange tar handling, i removed it, lets see what the ci is saying, but i want to revisit this in a sec... i have a idea

Ok, let’s move on and remove it if you’re fine with it too,
foo.tar.gz, basename is foo.tar and extension is gz, I think that’s valid!

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -497,40 +506,156 @@ func (s *service) GetReceivedShare(ctx context.Context, req *collaboration.GetRe
...
+			}, err
+		}
+
+		// handle mount point related updates
+		{
+			// find a suitable mount point
+			var requestedMountpoint string
+			switch {
+			case slices.Contains(req.GetUpdateMask().GetPaths(), _fieldMaskPathMountPoint) && req.GetShare().GetMountPoint().GetPath() != "":
+				requestedMountpoint = req.GetShare().GetMountPoint().GetPath()
```
rhafer, line 558: I am still not completely sure about the semantics when the incoming request actually contains a MountPoint. I think in that case we should not do the game with the prefixed number. But rather be explicit. I.e. if the mountpoint is already taken, make the request fail and return an appropriate Status.

fschade: I’m with you, but that means clients have to implement that a user can change the mount point name (I would prefer that too).

but that needs some time, maybe we should merge it that way, talk to the clients and then change it to error in the case of explicit conflicts

kobergj (approved):
Looks good to me for now :+1:

internal/grpc/services/usershareprovider/usershareprovider.go
rhafer: > I’m with you, but that means clients have to implement that a user can change the mount point name (I would prefer that too).

Hm, I am not sure I get that. Clients (as in CS3 clients) are already able to change the mount point name, aren't they?

The fact that our currently used clients at the higher-level (every thing that comes in via libregraph e.g.) don't make use of that does not change that. But in that case they shouldn't send a mountpoint at all.

rhafer: Alternatively I'd be find if we'd just not allow UpdateReceivedShare Request with a `_fieldMaskPathMountPoint` at all for now.

rhafer (approved):

PR #4722, opened 2024-06-11
The libregraph public link type representation added to the PROPFIND response

2403905:
The libregraph public link type representation added to the PROPFIND response
Related issue [#8740](https://github.com/owncloud/ocis/issues/8740)

### DIFF
--- a/pkg/conversions/role.go
+++ b/pkg/conversions/role.go
@@ -149,6 +149,29 @@ func (r *Role) WebDAVPermissions(isDir, isShared, isMountpoint, isPublic bool) s
 	return b.String()
 }
 
+// OCSPermissionsToPublicLinkType converts the public link OCSPermission to the sharingLinkType representation
+//
+//	VIEW            SharingLinkType = "view"
+//	UPLOAD          SharingLinkType = "upload"
+//	EDIT            SharingLinkType = "edit"
+//	CREATE_ONLY     SharingLinkType = "createOnly"
+func (r *Role) OCSPermissionsToPublicLinkType(rt provider.ResourceType) string {
+	p := r.OCSPermissions()
+	switch {
+	case p == PermissionRead:
+		return "view"
+	case p == PermissionRead|PermissionWrite && rt == provider.ResourceType_RESOURCE_TYPE_FILE:
+		return "edit"
+	case p == PermissionRead|PermissionCreate && rt == provider.ResourceType_RESOURCE_TYPE_CONTAINER:
+		return "upload"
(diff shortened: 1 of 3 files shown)


pkg/conversions/role.go
```diff
@@ -149,6 +149,29 @@ func (r *Role) WebDAVPermissions(isDir, isShared, isMountpoint, isPublic bool) s
...
 }
 
+// OCSPermissionsToPublicLinkType converts the public link OCSPermission to the sharingLinkType representation
+//
+//	VIEW            SharingLinkType = "view"
+//	UPLOAD          SharingLinkType = "upload"
+//	EDIT            SharingLinkType = "edit"
+//	CREATE_ONLY     SharingLinkType = "createOnly"
+func (r *Role) OCSPermissionsToPublicLinkType(rt provider.ResourceType) string {
+	p := r.OCSPermissions()
```
rhafer, line 159: Hm, I don't think we should converting this based on the OCS permissions here. OCS is going to go a away at some point (and `r.OCSPermissions()` will hopefully go with it. IMO we should do the conversion based on the CS3ResourcePermissions.

In ocis (https://github.com/owncloud/ocis/blob/fc2318a5ef419c559a46fad41fbc22f6211f50ec/services/graph/pkg/linktype/linktype.go#L33) we already have function for converting ResourcePermissions to libregraph link types. Maybe we should move that into reva to make it reusable in graph and reva?

BTW, libregraph also defines a `blocksDownload` type. Do need that here as well?

rhafer (changes requested):

pkg/conversions/role.go
2403905: This seems like an obvious solution but it will lead to coping more code. Since LinkType is similar to UnifiedRoleDefinition and UnifiedRoleDefinition was moved from reva to graph, it seems to me there is no point in bringing back LinkType to reva.

The `blocksDownload` is not used.

2403905: @rhafer We could try to move the Link permission sets to the `reva/pkg/convertions` and keep the `LinkType` in a graph.  Then  we can use those sets for the permission conversations.

2403905: @micbar Could we separate the `LinkType` and move the permission sets `provider.ResourcePermissions` from the graph to the [reva/pkg/conversions](https://github.com/cs3org/reva/tree/edge/pkg/conversions) that allows converting based on the OCS permissions?

micbar: In would suggest the opposite. Move the ocdav service to ocis.

PR #4725, opened 2024-06-14
fix: ocs handles the mount_point on its own

fschade:
ocs updates unmountedShares at the same time as a received share is accepted, therefore the mount_point handling is entirely done there too.


fschade:
@nirajacharya2 can you check the ci, looks like https://github.com/cs3org/reva/pull/4724 introduced some unexpected passes

PR #4732, opened 2024-06-18
move user data to avoid issues with open files

individual-it:


individual-it:
closing in favor of #4731

PR #4740, opened 2024-06-25
disallow reserved names

butonic:
We now disallow the reserved `..` and `.` filenames. They are only allowed as destinations of move or copy operations.

supersedes https://github.com/cs3org/reva/pull/4721

### DIFF
--- a/internal/http/services/owncloud/ocdav/validation.go
+++ b/internal/http/services/owncloud/ocdav/validation.go
@@ -27,12 +27,26 @@ func ValidatorsFromConfig(c *config.Config) []Validator {
 
 // ValidateName will validate a file or folder name, returning an error when it is not accepted
 func ValidateName(name string, validators []Validator) error {
+	return ValidateDestination(name, append(validators, notReserved()))
+}
+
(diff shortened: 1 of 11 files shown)


internal/http/services/owncloud/ocdav/validation.go
```diff
@@ -27,6 +27,9 @@ func ValidatorsFromConfig(c *config.Config) []Validator {
 
 // ValidateName will validate a file or folder name, returning an error when it is not accepted
 func ValidateName(name string, validators []Validator) error {
+	if err := notReserved()(name); err != nil {
```
kobergj, line 30: This is mostly the same code as in `ValidateDestination`. Why not just
```golang
return ValidateDestination(name, append(validators, notReserved())
```

kobergj (changes requested):

saw-jan:
```feature
      | dav-path-version | file-name | http-status-code |
      | old              | /.        | 500              |
        Failed step: Then the HTTP status code should be "500"
        HTTP status code 400 is not the expected value 500
        Failed asserting that 400 matches expected '500'.
```
need test adjustment in ocis. and for that we need this to be merged first and then reva bump in ocis

kobergj (approved):

PR #4746, opened 2024-07-01
Fix same file upload

kobergj:
Fixes a bug where uploading the same file multiple times leads to orphaned blobs in the blobstore.

Reva part of https://github.com/owncloud/ocis/issues/9498

### DIFF
--- a/pkg/storage/utils/decomposedfs/upload/store.go
+++ b/pkg/storage/utils/decomposedfs/upload/store.go
@@ -354,8 +354,46 @@ func (store OcisStore) updateExistingNode(ctx context.Context, session *OcisSess
 		versionPath := session.store.lu.InternalPath(spaceID, n.ID+node.RevisionIDDelimiter+oldNodeMtime.UTC().Format(time.RFC3339Nano))
 
 		// create version node
-		if _, err := os.Create(versionPath); err != nil {
-			return unlock, err
+		_, err := os.OpenFile(versionPath, os.O_CREATE|os.O_EXCL, 0600)
+		if err != nil {
+			if !errors.Is(err, os.ErrExist) {
+				return unlock, err
+			}
+
+			// a revision with this mtime does already exist.
+			// If the blobs are the same we can just delete the old one
+
+			checksum, err := old.Xattr(ctx, prefixes.ChecksumPrefix+"sha1")
+			if err != nil {
+				return unlock, err
+			}
+
+			revisionChecksum, err := session.store.lu.MetadataBackend().Get(ctx, versionPath, prefixes.ChecksumPrefix+"sha1")
+			if err != nil {
+				return unlock, err
+			}
+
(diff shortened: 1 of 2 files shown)


pkg/storage/utils/decomposedfs/upload/store.go
```diff
@@ -354,8 +354,41 @@ func (store OcisStore) updateExistingNode(ctx context.Context, session *OcisSess
...
-		if _, err := os.Create(versionPath); err != nil {
-			return unlock, err
+		_, err := os.OpenFile(versionPath, os.O_CREATE|os.O_EXCL, 0600)
+		if err != nil {
+			if !errors.Is(err, os.ErrExist) {
+				return unlock, err
+			}
+
+			// a revision with this mtime does already exist.
+			// If the blobs are the same we can just delete the old one
```
rhafer, line 364: Disclaimer: I am not very familiar with the upload code, so I might be missing something obvious.

Why can we be sure here that this is really the same blob, when we only look at the mtimes  and blobsizes?

kobergj: Good question. I asked myself the same. But what else CAN we check? The `etag` is just a hash over id and mtime, which are both the same here. 

At the moment we are just silently discarding the old blob without any further checks. So this is at least a improvement. However if you have any good idea how we can savely say that this is the same blob I am very happy to check it here.

aduffeck: We already generate checksums for the content of the upload at https://github.com/cs3org/reva/blob/67b3f12080874a366ffc2222bd18f3738658b21a/pkg/storage/utils/decomposedfs/upload/upload.go#L118-L145. Could we just compare the new checksums with the previous ones?

rhafer: Can't we use the checksums? For the upload I think the caller of `updateExistingNode` (`CreateNodeForUpdate`) already has the checksums. For the old I guess we can get them as well somehow?

kobergj: Yes. Good idea :+1: I managed to get the checksums from the xattrs. Also got rid of the blobsize check as it isn't needed any more.

BUT: At the moment I am only checking `md5` sum. Should I check another one? Should I check all three?

rhafer: If you only want to check one I think I'd choose `sha1`.  :man_shrugging: 

I guess at some point we need to switch to some less out-dated hashing algorithm, though.

kobergj: I can check two or all three if that makes sense. Just not sure if it makes sense :smiley: 

I'll switch to `sha1` meanwhile

aduffeck (approved):
LGTM :+1:

PR #4759, opened 2024-07-09
bump go-cs3apis to latest master

rhafer:
This needed to touch quite a few unit test since the tooling to generate the go bindings was updated to a more recent version. Which causes issues with e.g. using reflect.DeepEqual on proto messages. We're basically running into the changes described here: https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.20.0#v1.20-generated-code

### DIFF
--- a/changelog/unreleased/bump-go-cs3apis.md
+++ b/changelog/unreleased/bump-go-cs3apis.md
@@ -1,0 +1,6 @@
+Change: Updated to the latest version of the go-cs3apis
+
+The go-cs3apis dependency was updated to the latest version
+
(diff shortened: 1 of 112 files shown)


rhafer:
Keeping this as a draft for now as I need to test it against ocis.

rhafer:
There's a draft  PR testing this in ocis: https://github.com/owncloud/ocis/pull/9587

micbar (approved):
Heroic! 🎉 

@glpatcern next time it would be helpful to upgrade the proto tooling in a separate change because that caused a lot of fallout.

PR #4766, opened 2024-07-15
Set output format via query param 'output-format'

AlexAndBear:
Fixes:  https://github.com/owncloud/ocis/issues/9399


AlexAndBear:
This is a breaking change, should we change the title subject to breaking ?

kobergj (approved):
Welcome to backend team :heart:

kobergj:
> This is a breaking change, should we change the title subject to breaking ?

No need to change it, we will have a breaking release anyways!

PR #4767, opened 2024-07-17
feat: add folder id to dav mkcol responses

JammingBen:
Adds the `OC-FileId` header to dav `mkcol` responses, containing the id of the newly created folder.

This is necessary for clients that work with ids and want to fetch a folder immediately after creating.

fixes https://github.com/owncloud/ocis/issues/9618


kobergj:
I'm not sure this is the correct approach. We would need another stat call, which will impact performance. We should pass the id with the grpc responses so we don't have to call it again. This needs some refactoring, but I still prefer it over the extra stat call.

JammingBen:
> I'm not sure this is the correct approach. We would need another stat call, which will impact performance. We should pass the id with the grpc responses so we don't have to call it again. This needs some refactoring, but I still prefer it over the extra stat call.

Sounds reasonable, could you give me a hint where to start for this? :) Or will this be a more complicated task...?

micbar:
We can use an opaque property for now

kobergj:
Unfortunately not so easy :cry: 

The decomposedfs method `CreateDir` does not return anything, hence it cannot return the id: https://github.com/cs3org/reva/blob/edge/pkg/storage/utils/decomposedfs/decomposedfs.go#L682

We would need to change the interface and adjust all other implementations.

JammingBen:
Argh ok. Then I'm afraid this is a bit too much for me currently, especially since I don't know how to test all these storage drivers.

kobergj:
Just for the records: We have the same issue with `Move`, `Delete` and `TouchFile`. All of them do not return anything and therefore need extra calls to get their id (or similar values)

kobergj:
@micbar will you do your IDE magic for the refactor? I volunteer to review it. (Though I probably regret that :wink: )

PR #4771, opened 2024-07-19
fix: dav delete via resource id

JammingBen:
Fixes an issue where deleting a resource via its id, using the `/dav/spaces/{fileId}` endpoint, wouldn't work.

The issue was happening because the code was checking for an empty path to prevent spaces from being deleted this way (because spaces have an empty path). When only providing a resource id however, the path will be empty for regular resources as well. For this reason we now additionally check for the resource's `opaqueId`. If the resource doesn't have such and the path is empty, we can assume that it's a space and prevent deletion via the dav endpoint.

fixes https://github.com/owncloud/ocis/issues/9619

### DIFF
--- a/internal/http/services/owncloud/ocdav/delete.go
+++ b/internal/http/services/owncloud/ocdav/delete.go
@@ -137,8 +137,11 @@ func (s *svc) handleSpacesDelete(w http.ResponseWriter, r *http.Request, spaceID
 
 	// do not allow deleting spaces via dav endpoint - use graph endpoint instead
 	// we get a relative reference coming from the space root
-	// so if the path is "empty" we a referencing the space
-	if ref.GetPath() == "." {
+	// so if the path is "empty" and no opaque id is present or the opaque id equals
+	// the space id, we are referencing the space
+	rid := ref.GetResourceId()
+	if ref.GetPath() == "." &&
+		(rid.GetOpaqueId() == "" || rid.GetOpaqueId() == rid.GetSpaceId()) {
 		return http.StatusMethodNotAllowed, errors.New("deleting spaces via dav is not allowed")
 	}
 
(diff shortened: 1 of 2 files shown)


JammingBen:
@kobergj Pinging you specifically for a review because I saw that you touched the initial code a while ago :)

internal/http/services/owncloud/ocdav/delete.go
```diff
@@ -137,8 +137,8 @@ func (s *svc) handleSpacesDelete(w http.ResponseWriter, r *http.Request, spaceID
 
 	// do not allow deleting spaces via dav endpoint - use graph endpoint instead
 	// we get a relative reference coming from the space root
-	// so if the path is "empty" we a referencing the space
-	if ref.GetPath() == "." {
+	// so if the path is "empty" and no opaque id is present, we are referencing the space
+	if ref.GetPath() == "." && ref.GetResourceId().GetOpaqueId() == "" {
```
kobergj, line 141: The check is not correct. It could also be a space root if
```
ref.GetResourceId().GetOpaqueId() == ref.GetResourceId().GetSpaceId() && ref.GetPath() == "."
```

kobergj (approved):

internal/http/services/owncloud/ocdav/delete.go
JammingBen: True! But we still need to check for `ref.GetResourceId().GetOpaqueId() == ""`, no? Because we can't say for sure that the opaqueId is given?

kobergj: Yes exactly. We need to check both. Good luck with making the condition readable :smile:

JammingBen: Gave it my best shot 😄

kobergj (approved):
Nice! :+1:

butonic (approved):

PR #4773, opened 2024-07-22
ocis bumped. the expected failures removed.

2403905:


phil-davis (approved):
LGTM if CI passes.

kobergj:
Please don't merge red PRs. The changelog pipeline was failing, which is not acceptable. The only step which is allowed to be red is the `license` step.

We had to fix it on our feature branch: https://github.com/cs3org/reva/pull/4774

phil-davis:
> Please don't merge red PRs. 

Sorry, I saw that the drone CI had passed, and I knew that "other stuff is known to fail" and didn't look closely. I will be more careful in future. 

(this is a problem with having "known red" in CI - is there something that we can do to fix that? Or maybe we should remove the License Compliance workflow, if we are ignoring it anyway)

kobergj:
> (this is a problem with having "known red" in CI - is there something that we can do to fix that? Or maybe we should remove the License Compliance workflow, if we are ignoring it anyway)

I 100% agree. But as far as I know we cannot remove this step.

@micbar is there any chance we can disable this step? Accepted red pipelines are bad.

PR #4774, opened 2024-07-22
fix micro ocdav service

butonic:
We no longer call Init to configure default options because it was replacing the existing options.

### DIFF
--- a/changelog/unreleased/bump-ocis.md
+++ b/changelog/unreleased/bump-ocis.md
@@ -1,4 +1,4 @@
-Change: Ocis bumped.
+Change: Ocis bumped
 
 Ocis bumped. The expected failures removed.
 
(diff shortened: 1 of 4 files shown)


kobergj (approved):
Nice :+1:

kobergj (approved):

phil-davis:
Is this fix related to https://github.com/owncloud/ocis/issues/9659 ?
(where test pipelines had a problem related to ocdav)

kobergj:
Yes. This should fix all pipelines that failed with
```
{
  "level": "error",
  "service": "com.owncloud.web.ocdav",
  "error": "not found",
  "time": "2024-07-22T00:51:06Z",
  "message": "could not select service from the registry"
}
```
(presumably)

PR #4784, opened 2024-07-30
Bump tusd to v2.4.0

kobergj:
Bumps tusd to v2.4.0 to enable tackling https://github.com/owncloud/ocis/issues/1755


kobergj:
Full-ci ocis pipeline is green: https://github.com/owncloud/ocis/pull/9714

2403905:
If I understand correctly we also have to implement the  Unwrap interface otherwise we getting warnings regarding NetworkControlError/NetworkTimeoutError and "feature not supported"

https://tus.github.io/tusd/advanced-topics/usage-package/#i-am-getting-warnings-regarding-networkcontrolerrornetworktimeouterror-and-feature-not-supported-why

pkg/rhttp/datatx/manager/tus/tus.go
```diff
@@ -99,7 +98,8 @@ func (m *manager) Handler(fs storage.FS) (http.Handler, error) {
 	config := tusd.Config{
 		StoreComposer:         composer,
 		NotifyCompleteUploads: true,
-		Logger:                log.New(appctx.GetLogger(context.Background()), "", 0),
+		// TODO: bring back!
```
dragonchaser, line 101: Is that on purpose?

dragonchaser (dismissed):

pkg/rhttp/datatx/manager/tus/tus.go
kobergj: Good Question. The problem is that the logger interface changed in v2. It now uses https://pkg.go.dev/golang.org/x/exp/slog Afaikt there a two option how to solve it

A) We adjust our logger to follow new interface. I think we'd need to implement https://pkg.go.dev/golang.org/x/exp/slog#Handler

B) We use their default logger.

Is it worth going for A?

kobergj: Brought back the logger - but it is still noop.

kobergj:
> If I understand correctly we also have to implement the Unwrap interface otherwise we getting warnings regarding NetworkControlError/NetworkTimeoutError and "feature not supported"

Correct, but we are passing a noop logger (with did that before too). Hence we will not see those error messages.

fschade (approved):

PR #4785, opened 2024-07-31
Block overwriting mountpoints through COPY

kobergj:
reva part of https://github.com/owncloud/ocis/issues/6999

### DIFF
--- a/internal/http/services/owncloud/ocdav/copy.go
+++ b/internal/http/services/owncloud/ocdav/copy.go
@@ -127,7 +127,7 @@ func (s *svc) handlePathCopy(w http.ResponseWriter, r *http.Request, ns string) 
 		return
 	}
 
-	cp := s.prepareCopy(ctx, w, r, spacelookup.MakeRelativeReference(srcSpace, src, false), spacelookup.MakeRelativeReference(dstSpace, dst, false), &sublog)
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocdav/copy.go
```diff
@@ -688,6 +683,38 @@ func (s *svc) prepareCopy(ctx context.Context, w http.ResponseWriter, r *http.Re
...
 			return nil
 		}
+
+		if utils.IsSpaceRoot(dstStatRes.GetInfo()) {
+			log.Error().Msg("overwriting is not allowed")
+			w.WriteHeader(http.StatusBadRequest)
+			return nil
+		}
+
+		// we must not allow to override mountpoints - so we check if we have write access to the parent. If not this is a mountpoint
```
butonic, line 693: hm, what if we are in a readonly space where we were granted write permissions on a subfolder?

butonic (changes requested):

internal/http/services/owncloud/ocdav/copy.go
kobergj: the comment is wrong - I only stat it, so view permissions is enough.

kobergj:
> Should you prevent the overwriting via the MOVE?

I don't think so. Cross-storage MOVEs are not supported as far as I know.

butonic (approved):

butonic (approved):

PR #4790, opened 2024-08-04
OCM: Fix open driver

MahdiBaghbani:
The error happens because of an error probing OCM services at the remote server:
`error doing request: Get \"revaowncloud1.docker/ocm-provider\": unsupported protocol scheme \"\"`
since the provider domain doesn't have `http://` or `https://`.

The changes in this PR add a scheme to the domain before the probing.

Fixes: #4791

### DIFF
--- a/pkg/ocm/provider/authorizer/open/open.go
+++ b/pkg/ocm/provider/authorizer/open/open.go
@@ -72,7 +72,7 @@ func (a *authorizer) GetInfoByDomain(ctx context.Context, domain string) (*ocmpr
 	// to ocmClient.Discover would fail.
 
 	// convert domain into lowercase to avoid adding scheme if the domain
-	//looks like this: HtTpS://domain.docker
+	// looks like this: HtTpS://domain.docker
 	domain = strings.ToLower(domain)
 
 	// add a scheme to domain if it doesn't have one.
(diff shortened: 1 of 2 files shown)


pkg/ocm/provider/authorizer/open/open.go
```diff
@@ -67,6 +67,19 @@ func (a *authorizer) GetInfoByDomain(ctx context.Context, domain string) (*ocmpr
...
+	domain = strings.ToLower(domain)
+
+	// add a scheme to domain if it doesn't have one.
+	if !strings.HasPrefix(domain, "http://") && !strings.HasPrefix(domain, "https://") {
+		domain = "https://" + domain
+	}
+
 	// not yet known: try to discover the remote OCM endpoint
 	ocmClient := client.NewClient(time.Duration(10)*time.Second, true)
 	ocmCaps, err := ocmClient.Discover(ctx, domain)
```
glpatcern, line 85: I think I see where the issue is: `Discover()` requires an endpoint, whereas `domain` is what it is - a domain to be used as such in the `ProviderInfo` structure below.

So what I would do here is something like:
```code
var host string
if !HasPrefix(...) {
   host := "https://" + domain
} else {
   host := domain
}
ocmClient.Discover(ctx, host)
```
(also, no need to deal with `ToLower`)

glpatcern (changes requested):
Thanks for this Mahdi! But see below for where to put the patch...

pkg/ocm/provider/authorizer/open/open.go
MahdiBaghbani: Thanks! 

I removed `toLower` and also couldn't name the new variable `host` since it would conflict with the line:

`host, _ := url.Parse(ocmCaps.Endpoint)`

so I have changed it to `endpoint`.

PR #4796, opened 2024-08-05
Fix Copy III

kobergj:
Mountpoint overwriting is still broken. This fixes it.


phil-davis:
It complains about changelog - should there be one?

kobergj:
>  It complains about changelog - should there be one?

Yes, of course. Thanks!

fschade (approved):

PR #4801, opened 2024-08-07
tests: bump openldap and deprecate HOSTNAME variable

labkode:
This PR unblocks the recurrent errors on the CI tests where the ldap container is not available.


glpatcern (approved):
Thanks Hugo!

PR #4805, opened 2024-08-12
Return proper error when trying to read attrs from a non-existing node

aduffeck:
Do not try to read xattrs from non-existing nodes but return a proper error instead.

Fixes https://github.com/owncloud/ocis/issues/9734


rhafer (approved):
lgtm

PR #4807, opened 2024-08-13
Tusd PATCH Checksums

kobergj:
Compares checksums in PATCH requests during chunked uploads


pkg/storage/utils/decomposedfs/upload/upload.go
```diff
@@ -376,3 +375,20 @@ func joinurl(paths ...string) string {
...
 }
+
+func verifyChecksum(checksum string, sha1h hash.Hash, md5h hash.Hash, adler32h hash.Hash32) error {
+	parts := strings.SplitN(checksum, " ", 2)
+	if len(parts) != 2 {
+		return errtypes.BadRequest("invalid checksum format. must be '[algorithm] [checksum]'")
+	}
+	switch strings.ToLower(parts[0]) {
+	case "sha1":
+		return checkHash(parts[1], sha1h)
```
butonic, line 386: afaict checkHash still needs to base64 decode the checksum

kobergj: Won't work. We also use this function to compare checksums from upload sessions (which are not base64 encoded). I added a bool parameter to reflect that.

butonic:
running into https://github.com/ankitpokhrel/tus-php/issues/422 :(

PR #4808, opened 2024-08-13
owncloudsql: fix path, naming and listing spaces

butonic:
@dragotin owncloudsql has bitrotted. uploads are still broken, even with this PR. oCIS seems to equire the storage driver to implement listing upload sessions and using the event handler to integrate with postprocessing.

This PR is just the basic fixes to get listing spaces and basic file navigation working again.

why do I still do this? I think being able to switch from oc10 / nc to ocis effortless is a worthwile thing. It would be great if clients could deal with multiple 'personal' spaces, then we could just deploy an owncloudsql storage alongside decomposedfs and then users could migrate at will. 

For that a readonly owncloudsql might make sense ... hm ... we could list the owncloudsql spaces as projects ... that only the one user has access to ... 🤔 

it any case pls give me a +1 so I can start working on the upload sessions ...

### DIFF
--- a/pkg/storage/fs/owncloudsql/spaces.go
+++ b/pkg/storage/fs/owncloudsql/spaces.go
@@ -31,12 +31,14 @@ import (
 	"github.com/cs3org/reva/v2/pkg/errtypes"
 	"github.com/cs3org/reva/v2/pkg/storage/fs/owncloudsql/filecache"
 	"github.com/cs3org/reva/v2/pkg/storagespace"
+	"github.com/cs3org/reva/v2/pkg/utils"
 )
 
 // ListStorageSpaces lists storage spaces according to the provided filters
 func (fs *owncloudsqlfs) ListStorageSpaces(ctx context.Context, filter []*provider.ListStorageSpacesRequest_Filter, unrestricted bool) ([]*provider.StorageSpace, error) {
 	var (
 		spaceID = "*"
+		// uid     *userpb.UserId
 	)
 
 	filteringUnsupportedSpaceTypes := false
@@ -49,7 +51,7 @@ func (fs *owncloudsqlfs) ListStorageSpaces(ctx context.Context, filter []*provid
 		case provider.ListStorageSpacesRequest_Filter_TYPE_ID:
 			_, spaceID, _, _ = storagespace.SplitID(filter[i].GetId().OpaqueId)
 		case provider.ListStorageSpacesRequest_Filter_TYPE_USER:
-			_, spaceID, _, _ = storagespace.SplitID(filter[i].GetId().OpaqueId)
+			// uid = filter[i].GetUser()
 		}
 	}
 	if filteringUnsupportedSpaceTypes {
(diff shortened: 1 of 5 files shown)


dragotin (approved):
doesn't look harmful ;-)

pkg/rgrpc/todo/pool/selector.go
```diff
@@ -130,9 +132,15 @@ func (s *Selector[T]) Next(opts ...Option) (T, error) {
 
 	existingClient, ok := s.clientMap.Load(target)
 	if ok {
+		if options.logger != nil {
```
kobergj, line 135: We should not allow `nil` loggers. The `logger` field should default to a noop logger.

pkg/rgrpc/todo/pool/selector.go
```diff
@@ -130,9 +132,15 @@ func (s *Selector[T]) Next(opts ...Option) (T, error) {
...
+			options.logger.Debug().Str("address", target).Msg("reusing existing client")
+		}
 		return existingClient.(T), nil
 	}
 
+	if options.logger != nil {
```
kobergj, line 141: Same here. Expect `logger` is not nil

pkg/storage/fs/owncloudsql/spaces.go
```diff
@@ -31,12 +31,14 @@ import (
...
 	"github.com/cs3org/reva/v2/pkg/storage/fs/owncloudsql/filecache"
 	"github.com/cs3org/reva/v2/pkg/storagespace"
+	"github.com/cs3org/reva/v2/pkg/utils"
 )
 
 // ListStorageSpaces lists storage spaces according to the provided filters
 func (fs *owncloudsqlfs) ListStorageSpaces(ctx context.Context, filter []*provider.ListStorageSpacesRequest_Filter, unrestricted bool) ([]*provider.StorageSpace, error) {
 	var (
 		spaceID = "*"
+		// uid     *userpb.UserId
```
kobergj, line 41: Remove comment

pkg/storage/fs/owncloudsql/spaces.go
```diff
@@ -49,7 +51,7 @@ func (fs *owncloudsqlfs) ListStorageSpaces(ctx context.Context, filter []*provid
 		case provider.ListStorageSpacesRequest_Filter_TYPE_ID:
 			_, spaceID, _, _ = storagespace.SplitID(filter[i].GetId().OpaqueId)
 		case provider.ListStorageSpacesRequest_Filter_TYPE_USER:
-			_, spaceID, _, _ = storagespace.SplitID(filter[i].GetId().OpaqueId)
+			// uid = filter[i].GetUser()
```
kobergj, line 54: This one also

kobergj (changes requested):

pkg/storage/fs/owncloudsql/spaces.go
butonic: not yet. we need to implement the user filter in a subsequent pr. this PR just makes listing spaces and navigating files work again. I'm trying to make things work gradually...

kobergj (approved):

PR #4809, opened 2024-08-14
implement OCM wellknown for discovery

butonic:
The `wellknown` service now implements the `/.well-known/ocm` endpoint for OCM discovery. The unused endpoints for openid connect and webfinger have been removed. This aligns the wellknown implementation with the master branch.

Categorizing as 'Enhancement' because the existing service was unused in ocis.


glpatcern:
Good you're looking after this @butonic ! but don't forget the legacy endpoint must remain.

And more in general: what about having a checkpoint meeting on what is pure OCM vs what is ScienceMesh etc.? I didn't receive any reply to my email in July but people should be back in the coming days. Let's sort out the details by email?

butonic:
so ... discovery is only part of sciencemesh?

glpatcern:
> so ... discovery is only part of sciencemesh?

how is that related to the wellknown endpoint? Anyway, discovery is part of OCM.

dragonchaser (approved):

PR #4815, opened 2024-08-19
Fix small difference with the OCM specification

MahdiBaghbani:


glpatcern:
Interesting finding, thanks. I realize the OCM specs have been written as `produces: application/hal+json` as opposed to plain JSON, though they never seem to "exploit" such augmentation.

MahdiBaghbani:
> If this is indeed required, then we should also fix all other OCM payloads in Reva.

Yes! Exactly, I missed this the first time.

As I was implementing an OCM-compliant API server from scratch in [Aver](https://github.com/MahdiBaghbani/aver) I did notice the `application/hal+json` for the first endpoint in the spec.

Looking closely at the specification again, it seems that everything returned by any OCM endpoint except the `/mfa-capable` should be `application/hal+json`.

glpatcern:
> Looking closely at the specification again, it seems that everything returned by any OCM endpoint except the `/mfa-capable` should be `application/hal+json`.

Right, can you please propagate this change in all relevant places?

MahdiBaghbani:
I'm waiting for this https://github.com/cs3org/OCM-API/pull/123 to be merged.

Then we can remove the `hal+json` and return `json` instead.

glpatcern:
Well in fact because of https://github.com/cs3org/OCM-API/pull/123 we can close this PR and keep things as they are - i.e. return `json`. Do you agree?

PR #4818, opened 2024-08-21
allow listing directory trash items by key

butonic:
The storageprovider now passes on the key without inventing a `/` as the relative path when it was not present at the end of the key. This allows differentiating requests that want to get the trash item of a folder itself (where the relative path is empty) or listing the children of a folder in the trash (where the relative path at least starts with a `/`).

We also fixed the `/dav/spaces` endpoint to not invent a `/` at the end of URLs to allow clients to actually make these different requests.

As a byproduct we now return the size of trashed items.

required by https://github.com/owncloud/ocis/pull/9879

### DIFF
--- a/pkg/storage/utils/decomposedfs/recycle.go
+++ b/pkg/storage/utils/decomposedfs/recycle.go
@@ -23,7 +23,6 @@ import (
 	iofs "io/fs"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/storage/utils/decomposedfs/recycle.go
```diff
@@ -196,14 +201,24 @@ func (fs *Decomposedfs) ListRecycle(ctx context.Context, ref *provider.Reference
 
 // readTrashLink returns path, nodeID and timestamp
 func readTrashLink(path string) (string, string, string, error) {
+	start := time.Now()
```
2403905, line 204: Shouldn't we clean up the `println` execution time logging before merging?

2403905 (approved):

PR #4825, opened 2024-08-27
bump dependencies

butonic:
we need to do some housekeeping. note that the cs3 apis also updated, but it requires code changes to bump it, so they are left out.

also bumps the linter to 1.60.3 and fixes new linter errors

also bumps go to 1.22


rhafer (approved):

glpatcern:
Yep, for the record the changes in the CS3 APIs revolve around the concept discussed a few months ago: the Gateway is where the logic resides, the OCS / LibreGraph / HTTP layers should be thin layers with no extra logic.

In particular, listing shares to feed the `shared by me` | `shared with others` | `shared publicly` views is done by the Gateway, which is expected to return only the existing shares with their associated `resourceInfo` (i.e. execute all the required stat calls).

micbar:
@butonic has that been superseded by  another PR?

butonic:
superseded by https://github.com/cs3org/reva/pull/4893

PR #4835, opened 2024-09-04
Fix sharejail stat

butonic:
Stating a share jail mountpoint now returns the same resourceid as in the directory listing of the share jail root.

This needs to be tested with WOPI, as I fear the web client might now try to use the sharejail id to access a resource ...

see https://github.com/owncloud/ocis/issues/9933

### DIFF
--- a/changelog/unreleased/fix-sharejail-stat-id.md
+++ b/changelog/unreleased/fix-sharejail-stat-id.md
@@ -1,0 +1,6 @@
+Bugfix: Fix sharejail stat id
+
+Stating a share jail mountpoint now returns the same resourceid as in the directory listing of the share jail root.
+
+https://github.com/cs3org/reva/pull/4835
+https://github.com/owncloud/ocis/issues/9933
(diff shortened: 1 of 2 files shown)


butonic:
@micbar since you are knee deep in wopi work can you try this out with a working wopi setup?

felix-schwarz:
@butonic Is there any way (test instance or similar) that we could use to verify if it fixes the issue for the iOS app and/or if there are any other side effects?

dragonchaser (approved):

micbar:
You could use the ocis_full docker compose locally.

1. make -C ocis dev-docker
2. export OCIS_DOCKER_TAG=dev
3. export OCIS_DOCKER_IMAGE=owncloud/ocis
4. docker compose up -d
5. use ocis.owncloud.test as url (needs etc/hosts entry

butonic:
while the implementation is correct, the ocis testsuite needs a fix: https://github.com/owncloud/ocis/pull/10172#issuecomment-2377283073

butonic:
test got added to expected failures, merging

PR #4840, opened 2024-09-09
allow updating of ocm shares

dragonchaser:
Bugfix: Allow update of ocm shares

We fixed a bug that prevented ocm shares to be updated or removed.

https://github.com/owncloud/ocis/issues/9926

### DIFF
--- a/changelog/unreleased/allow-update-ocm-shares.md
+++ b/changelog/unreleased/allow-update-ocm-shares.md
@@ -1,0 +1,6 @@
+Bugfix: Allow update of ocm shares
+
+We fixed a bug that prevented ocm shares to be updated or removed.
+
+https://github.com/cs3org/reva/pull/4840/
+https://github.com/owncloud/ocis/issues/9926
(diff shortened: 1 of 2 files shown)


pkg/ocm/provider/authorizer/json/json.go
```diff
@@ -104,7 +104,11 @@ func normalizeDomain(d string) (string, error) {
 		return "", err
 	}
 
-	return u.Host, nil
+	normalizedDomain := u.Hostname()
+	if port := u.Port(); port != "" {
+		normalizedDomain += ":" + port
+	}
+	return normalizedDomain, nil
```
butonic, line 111: @fschade is that not part of any other PR? Related to the current breaking ocm setup in web?

butonic: ok, dropped this  and combinde the commits to glue the changelog to the change

butonic (approved):

PR #4842, opened 2024-09-10
backport: fix micro ocdav service and nats key encoding

butonic:
backporting to stable5
- https://github.com/cs3org/reva/pull/4774
- https://github.com/cs3org/reva/pull/4678/commits/df0ba6d4af0689c005940f197b2244b60cac07a3

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -152,6 +152,7 @@ func Create(opts ...microstore.Option) microstore.Store {
 		return natsjskv.NewStore(
 			append(opts,
 				natsjskv.NatsOptions(natsOptions), // always pass in properly initialized default nats options
+				natsjskv.EncodeKeys(),
 				natsjskv.DefaultTTL(ttl))...,
 		)
 	case TypeMemory, "mem", "": // allow existing short form and use as default
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -152,7 +152,6 @@ func Create(opts ...microstore.Option) microstore.Store {
 		return natsjskv.NewStore(
 			append(opts,
 				natsjskv.NatsOptions(natsOptions), // always pass in properly initialized default nats options
-				natsjskv.EncodeKeys(),
 				natsjskv.DefaultTTL(ttl))...,
 		)
 	case TypeMemory, "mem", "": // allow existing short form and use as default
--- a/changelog/unreleased/fix-micro-ocdav-registry.md
+++ b/changelog/unreleased/fix-micro-ocdav-registry.md
@@ -1,0 +1,6 @@
+Bugfix: Fix micro ocdav service init and registration
+
+We no longer call Init to configure default options because it was replacing the existing options.
+
+https://github.com/cs3org/reva/pull/4842
+https://github.com/cs3org/reva/pull/4774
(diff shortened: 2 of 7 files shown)


pkg/store/store.go
```diff
@@ -152,6 +152,7 @@ func Create(opts ...microstore.Option) microstore.Store {
 		return natsjskv.NewStore(
 			append(opts,
 				natsjskv.NatsOptions(natsOptions), // always pass in properly initialized default nats options
+				natsjskv.EncodeKeys(),
```
kobergj, line 155: Isn't this a breaking change? Existing stores will not be able to read their keys any more?

kobergj (changes requested):

pkg/store/store.go
kobergj: Maybe not relevant if nats-js-kv anyways didn't work in `5.x`. Degrading to comment.

kobergj: Please note that stores are also used as key-value stores. For example in the event history or userlog service. If you encode the keys these stores can no longer read old keys

butonic: damn, right. I was confusing this with the registry, but it instantiates its own store and usese the EncodeKeys option ...

unfortunately, something else changed. without this option we can no longer store anything because nats does not like our keys anymore:
```
2024-09-17T14:09:40+02:00 ERR error uploading file error="Failed to store data in bucket 'spaces/f1/bdd61a-da7c-49fc-8203-0558109d1b4f/nodes/64/bb/19/df/-1598-4691-a07a-d2c2f5714815.REV.2024-09-17T08:56:50.792574043Z': nats: invalid key" datatx=simple line=/home/jfd/Repositories/reva/pkg/rhttp/datatx/manager/simple/simple.go:147 pkg=rhttp service=storage-system traceid=49388ac282e8ffca3a5165cb484669b1
```

but ... why? did nats change? did we change the keys we store in nats? I can reproduce this when running ocis with `"OCIS_CACHE_STORE": "nats-js-kv","STORAGE_USERS_ID_CACHE_STORE":"nats-js-kv"`

kobergj: This has always been like this. This is why we invented the `EncodeKeys` option anyways. So caches and the registry can savely use the Encoded keys. Just not sure about the stores. As said, if they are not used in `5.x` we are fine.

butonic: yeah, we changed this in https://github.com/cs3org/reva/blame/774910f82ee098c8b5ceb3457d1678245d20b137/pkg/store/store.go#L155

@micbar we might need a migration step from 5 to 6

butonic: tracked in https://github.com/owncloud/ocis/issues/10098

butonic: while STORAGE_USERS_FILEMETADATA_CACHE_STORE is not used in the helm charts OCIS_CACHE_STORE is.

butonic: so thekey encoding changd for v6: 
grafik

butonic: hm so for the backport we could leave out this line.

kobergj (approved):

PR #4846, opened 2024-09-16
Add specific application/x-binary mime type for bin extension files to mime type mapping

AlexAndBear:
Enhancement: Add specific mime type for bin files

We've added the specific mime type 'application/x-binary' for bin files to the list of mime types.

### DIFF
--- a/changelog/unreleased/enhancement-add-specific-mime-type-for-bin-files.md
+++ b/changelog/unreleased/enhancement-add-specific-mime-type-for-bin-files.md
@@ -1,0 +1,5 @@
+Enhancement: Add specific mime type for bin files
+
+We've added the specific mime type 'application/x-binary' for bin files to the list of mime types.
+
+https://github.com/cs3org/reva/pull/4846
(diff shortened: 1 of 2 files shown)


kobergj (dismissed):

kobergj (commented):
@2403905 will take the review over. We are not sure this fixes it without side effects.

2403905 (changes requested):
We map the mime type to files by extension and not only `bin` has type `application/octet-stream`, there are `bpk`, buffer` and etc. in a `mimeTypes` map. We also set the `application/octet-stream` to any files w/o extension.

I think we should take a look at the  `ListSupportedMimeTypes`

2403905:
I think the problem is here https://github.com/owncloud/ocis/blob/4d5168f6c3ab182db596c996e8b9090f4dedc55e/services/collaboration/pkg/helpers/registration.go#L41

```
	mimeTypesMap := make(map[string]bool)
	for _, extensions := range appUrls {
		for ext := range extensions {
			m := mime.Detect(false, ext)
			mimeTypesMap[m] = true
		}
	}
```

The `mime.Detect` returns the "application/octet-stream" as default. So, we could skip there the "application/octet-stream" mime type.

AlexAndBear:
And then the files have no mime type anymore? I think the default is there on purpose.
 
I am wondering why OnlyOffice and Collabora are set to be able to be able to open especially the application/ocet-stream mime type. 

Shouldn't we just avoid that ?

2403905:
Fixed in a ocis

PR #4848, opened 2024-09-17
Fix bsd builds

aduffeck:
Fix compilation on darwin and other bsds.


kobergj (changes requested):
Please add a changelog. Rest looks :+1:

aduffeck:
I'm lazy so I extended the existing changelog instead, I think that does the trick as well ;)

kobergj (approved):

PR #4852, opened 2024-09-18
Indicate publiclink access

kobergj:
Indicates if a user in the context is coming through a public link

### DIFF
--- a/pkg/auth/manager/ocmshares/ocmshares.go
+++ b/pkg/auth/manager/ocmshares/ocmshares.go
@@ -159,6 +159,8 @@ func (m *manager) Authenticate(ctx context.Context, ocmshare, sharedSecret strin
 		},
 	}
 
+	user.Opaque = utils.AppendJSONToOpaque(user.Opaque, "impersonating-user", userRes.RemoteUser)
+
 	return user, scope, nil
 }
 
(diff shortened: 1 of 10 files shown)


pkg/auth/manager/ocmshares/ocmshares.go
```diff
@@ -159,6 +159,8 @@ func (m *manager) Authenticate(ctx context.Context, ocmshare, sharedSecret strin
 		},
 	}
 
+	user.Opaque = utils.AppendJSONToOpaque(user.Opaque, "impersonating-user", userRes.RemoteUser)
```
2403905, line 162: is the "impersonating-user" suitable for an anonymous user?

kobergj: Not sure what you mean.

"impersonating-user" can be 
* a hardcoded struct when this is a public link
* a remote user when this is a federate share

2403905: Nevermind. This is just a naming nitpick.

2403905 (approved):

kobergj:
I'll wait for green ocis pipeline before merge https://github.com/owncloud/ocis/pull/10116  But no more changes expected

PR #4863, opened 2024-09-23
WIP - Implement favourites for eos/grpc

ffurano:
Implement favourites for eos/grpc, following the logic of the eosbinary plugin

### DIFF
--- a/pkg/eosclient/eosgrpc/eosgrpc.go
+++ b/pkg/eosclient/eosgrpc/eosgrpc.go
@@ -47,7 +47,9 @@ import (
 )
 
 const (
-	versionPrefix = ".sys.v#."
+	versionPrefix  = ".sys.v#."
+	userACLEvalKey = "eval.useracl"
+	favoritesKey   = "http://owncloud.org/ns/favorite"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -47,7 +47,9 @@ import (
 )
 
 const (
-	versionPrefix = ".sys.v#."
+	versionPrefix  = ".sys.v#."
+	userACLEvalKey = "eval.useracl"
```
glpatcern, line 51: Looks good, but why do you need this key? We shouldn't use it any longer (and I have a PR to remove it from eosbinary as well)

ffurano: Well, I had copied that part from the current master of eosbinary

glpatcern: Ah ok! then you can remove it ;)

glpatcern, line 51:
```suggestion
```

glpatcern (approved):
This looks good, we need QA validation prior to merging

glpatcern (approved):

PR #4864, opened 2024-09-24
Add option to disable versioning in EOS upon PUT request

### LINKED ISSUE #4855: eosfs: allow for a way to disable versioning upon saving a file
Some applications have autosave features that should be treated differently than a normal save (initiated by the user). One example of this is Jupyter, which only expects a new version if the user manually clicks on the save button. We could apply a similar logic to the office apps as well.

We should extend the save mechanism in Reva to allow us to disable versioning in such cases. 

This could mean a change in the CS3 API's to make it a parameter (as reported in [cs3apis#97](https://github.com/cs3org/cs3apis/issues/97)), but for now, we agreed to use an opaque field for simplicity.

jessegeens:
Fixes #4855 

This PR introduces a new HTTP header for PUT requests, `X-Disable-Versioning`. When this header is added to a PUT request, the EOS storage driver will turn off versioning for this particular save. This is particularly useful for applications that perform automatic saves every x minutes: disabling versioning on these automatic saves ensures that the version history of these files is not polluted

### DIFF
--- a/changelog/unreleased/disable-versioning.md
+++ b/changelog/unreleased/disable-versioning.md
@@ -1,0 +1,4 @@
+Enhancement: Add HTTP header to disable versioning on EOS
+
+This enhancement introduces a new header, `X-Disable-Versioning`, on PUT requests. EOS will not version this file save whenever this header is set with a truthy value.
+See also: https://github.com/cs3org/reva/pull/4864
--- a/internal/http/services/owncloud/ocdav/put.go
+++ b/internal/http/services/owncloud/ocdav/put.go
@@ -280,6 +280,14 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		httpReq.Header.Set(HeaderLockHolder, lockholder)
 	}
 
+	// Propagate X-Disable-Versioning header
+	// Used to disable versioning for applications that do not expect this behaviour
+	// See reva#4855 for more info
+	disableVersioning, err := strconv.ParseBool(r.Header.Get(HeaderDisableVersioning))
+	if err == nil && disableVersioning {
+		httpReq.Header.Set(HeaderDisableVersioning, strconv.FormatBool(true))
+	}
+
 	httpRes, err := s.client.Do(httpReq)
 	if err != nil {
 		log.Error().Err(err).Msg("error doing PUT request to data service")
--- a/internal/http/services/owncloud/ocdav/put.go
+++ b/internal/http/services/owncloud/ocdav/put.go
@@ -280,6 +280,14 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		httpReq.Header.Set(HeaderLockHolder, lockholder)
 	}
 
+	// We need to pass Content-Length to the storage backend (e.g. EOS).
+	// However, the Go HTTP Client library may arbitrarily modify or drop
+	// the Content-Length header, for example when it compresses the data
+	// See: https://pkg.go.dev/net/http#Request
+	// Therefore, we use another header to pass it through the internal
+	// data server
+	httpReq.Header.Set(HeaderUploadLength, strconv.FormatInt(length, 10))
+
 	// Propagate X-Disable-Versioning header
 	// Used to disable versioning for applications that do not expect this behaviour
 	// See reva#4855 for more info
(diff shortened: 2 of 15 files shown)


pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -736,18 +736,23 @@ func (c *Client) Write(ctx context.Context, auth eosclient.Authorization, path s
...
 	xrdPath := fmt.Sprintf("%s//%s", c.opt.URL, path)
 	args := []string{"--nopbar", "--silent", "-f", source, xrdPath}
 
 	if auth.Token != "" {
 		args[4] += "?authz=" + auth.Token
 	} else if auth.Role.UID != "" && auth.Role.GID != "" {
-		args = append(args, fmt.Sprintf("-ODeos.ruid=%s&eos.rgid=%s&eos.app=%s", auth.Role.UID, auth.Role.GID, app))
+		queryParams := fmt.Sprintf("-ODeos.ruid=%s&eos.rgid=%s&eos.app=%s", auth.Role.UID, auth.Role.GID, app)
+
+		if disableVersioning {
```
diocas, line 752: Why is this dependent on the auth type? We will switch from gateway-based auth to token-based, so I imagine that would fall under the previous if, and therefore there would be no version disabling possibility?

pkg/eosclient/eosgrpc/eoshttp.go
```diff
@@ -373,12 +373,20 @@ func (c *EOSHTTPClient) GETFile(ctx context.Context, remoteuser string, auth eos
...
 	// Now send the req and see what happens
-	finalurl, err := c.buildFullURL(urlpath, auth)
+	tempUrl, err := c.buildFullURL(urlpath, auth)
+	base, err := url.Parse(tempUrl)
+	queryValues := base.Query()
+
+	if disableVersioning {
+		queryValues.Add("eos.versioning", strconv.Itoa(0))
+	}
+	finalurl := base.String() + queryValues.Encode()
```
diocas, line 388: If `base` contains query parameters, you are duplicating them.

You might want to `base.RawQuery = queryValues.Encode()` and then just `finalurl := base.String()`

diocas: Also, you don't seem to treat the `err` that `buildFullURL` might give.

pkg/rgrpc/todo/pool/pool.go
```diff
@@ -97,6 +97,13 @@ func NewConn(options Options) (*grpc.ClientConn, error) {
 	return conn, nil
 }
 
+func RegisterGatewayServiceClient(client gateway.GatewayAPIClient, endpoint string) {
```
diocas, line 100: Explain in the commit message why you had to do this.

diocas (changes requested):
First commit! Thanks.

I've made some comments. Please clean up the commit history to split the feature (and its tests) and all the other things you had to do to be able to do the tests (like updates and registering the gw?).
I'm not able to evaluate if the tests are ok like this. I'll wait for more knowledgeable people.

changelog/unreleased/disable-versioning.md
```diff
@@ -0,0 +1,4 @@
+Enhancement: Add HTTP header to disable versioning on EOS
+
+This enhancement introduces a new header, `X-Disable-Versioning`, on PUT requests. Whenever this header is set, with any value (but `true` is recommended), EOS will not version this file save.
```
diocas, line 3: Is it really any value? Isn't any value convertible by `ParseBool`? I would rephrase for something like

```suggestion
This enhancement introduces a new header, `X-Disable-Versioning`, on PUT requests. EOS will not version this file save whenever this header is set with a truthy value.
```

diocas (approved):

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -278,6 +278,9 @@ func (s *svc) handlePut(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		httpReq.Header.Set(HeaderLockHolder, lockholder)
 	}
 
+	// Set Content-Length
```
diocas, line 281: I wonder if we should add a few more comments here. The bug in EOS is fixed, but it's still recommended to always pass the length.

pkg/storage/utils/eosfs/upload.go
```diff
@@ -26,6 +26,7 @@ import (
 	"strconv"
 
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/internal/http/services/owncloud/ocdav"
```
diocas, line 29: Conceptually, do you thing ocdav is the correct place to have the content length header? It seems to be that it should be in some less specific place, used both inside the ocdav code and here.

diocas (approved):

PR #4867, opened 2024-09-25
fix(decomposedfs): available space calculation

rhafer:
Move the calculation of the available size to the blobstore interface. We were just returning the available size of the local disk, which is wrong when using the S3 blobstore.

Also, as the S3 blobstore doesn't have a concept of available size, we don't return a 'remaining' for S3 backed spaces with unlimited quota.

Fixes: https://github.com/owncloud/ocis/issues/9245

### DIFF
--- a/pkg/storage/fs/ocis/blobstore/blobstore.go
+++ b/pkg/storage/fs/ocis/blobstore/blobstore.go
@@ -116,6 +116,11 @@ func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
 
+// GetAvailableSize returns the available size in the blobstore in bytes
+func (bs *Blobstore) GetAvailableSize(n *node.Node) (uint64, error) {
+	return node.GetAvailableSize(n.InternalPath())
+}
+
 // List lists all blobs in the Blobstore
 func (bs *Blobstore) List() ([]*node.Node, error) {
 	dirs, err := filepath.Glob(filepath.Join(bs.root, "spaces", "*", "*", "blobs", "*", "*", "*", "*", "*"))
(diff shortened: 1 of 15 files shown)


pkg/storage/fs/ocis/blobstore/blobstore.go
```diff
@@ -116,6 +116,10 @@ func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
 
+func (bs *Blobstore) GetAvailableSize(n *node.Node) (uint64, error) {
```
kobergj, line 119: exported func `GetAvailableSize` should have comment or be unexported.

pkg/storage/fs/posix/blobstore/blobstore.go
```diff
@@ -88,3 +88,7 @@ func (bs *Blobstore) Download(node *node.Node) (io.ReadCloser, error) {
 func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
+
+func (bs *Blobstore) GetAvailableSize(n *node.Node) (uint64, error) {
```
kobergj, line 92: exported func `GetAvailableSize` should have comment or be unexported.

pkg/storage/fs/s3ng/blobstore/blobstore.go
```diff
@@ -128,6 +129,11 @@ func (bs *Blobstore) Delete(node *node.Node) error {
 	return nil
 }
 
+func (bs *Blobstore) GetAvailableSize(n *node.Node) (uint64, error) {
```
kobergj, line 132: exported func `GetAvailableSize` should have comment or be unexported.

pkg/storage/utils/decomposedfs/tree/tree.go
```diff
@@ -48,7 +48,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-var tracer trace.Tracer
+var (
+	tracer           trace.Tracer
+	ErrSizeUnlimited = errors.New("unlimited")
```
kobergj, line 53: I know it's catched shortly after, but can't still have a more describing error like "quota unlimited"?

kobergj (changes requested):
Usual nitpicking :+1:

kobergj (approved):

PR #4874, opened 2024-10-02
fixed the activity when file with file-id gives move activity instead of rename

2403905:
Related issue [#9744](https://github.com/owncloud/ocis/issues/9744)

### DIFF
--- a/pkg/storage/utils/decomposedfs/decomposedfs.go
+++ b/pkg/storage/utils/decomposedfs/decomposedfs.go
@@ -836,13 +836,13 @@ func (fs *Decomposedfs) Move(ctx context.Context, oldRef, newRef *provider.Refer
 		return
 	}
 
-	rp, err := fs.p.AssemblePermissions(ctx, oldNode)
+	orp, err := fs.p.AssemblePermissions(ctx, oldNode)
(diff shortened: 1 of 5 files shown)


pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -1241,3 +1247,35 @@ func (fs *Decomposedfs) PurgeRecycleItem(ctx context.Context, ref *provider.Refe
 func (fs *Decomposedfs) EmptyRecycle(ctx context.Context, ref *provider.Reference) error {
 	return fs.trashbin.EmptyRecycle(ctx, ref)
 }
+
+func (fs *Decomposedfs) getNodePath(ctx context.Context, n *node.Node, perms *provider.ResourcePermissions) (string, error) {
+	hp := func(n *node.Node) bool {
+		if perms == nil {
+			return false
+		}
+		return perms.GetPath
```
kobergj, line 1256: Why not
```suggestion
		return perms.GetGetPath()
```

kobergj (changes requested):

kobergj (approved):

PR #4876, opened 2024-10-02
fix share jail mountpoint parent id

butonic:
Stating a share jail mountpoint now returns the share jail root as the parent id.

ref https://github.com/owncloud/ocis/issues/9933


butonic:
@aduffeck posix failed again: https://drone.owncloud.com/cs3org/reva/5349

PR #4882, opened 2024-10-09
feat: indicate templates for weboffice

micbar:
# Templates for WebOffice

needed for https://github.com/owncloud/ocis/pull/10276

### DIFF
--- a/internal/http/services/appprovider/templates.go
+++ b/internal/http/services/appprovider/templates.go
@@ -1,0 +1,90 @@
+package appprovider
+
+import (
+	"strings"
+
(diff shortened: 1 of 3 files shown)


internal/http/services/appprovider/templates.go
```diff
@@ -0,0 +1,90 @@
...
+
+type Template struct {
+	Extension       string `json:"extension"`
+	MimeType        string `json:"mime_type"`
+	TargetExtension string `json:"target_extension"`
+}
+
+var tl = TemplateList{
+	Templates: map[string][]Template{
+		"collabora": {
```
butonic, line 21: we need to introduce a product property. the app name can be customized ...

butonic (approved):

PR #4897, opened 2024-10-22
Calculate remaining Quota correctly

kobergj:
Calculates the remaining quota as `total-used` and does not take free disc space into account.


rhafer (changes requested):
Can we remove the `GetAvailableSize` from the `Blobstore` interface again (basically revert b5c3c6781528bbc218e8f34a2bb0b9518d7de83c) ? Or is that still used anywhere?

rhafer (approved):

PR #4898, opened 2024-10-23
Fix broken EOS ACL operations

jessegeens:
This PR solves two issues: 
* `AddACL`  fails, because the current implementation of AddACL in the EOS gRPC client always sets `msg.Recursive = true`. This causes issues on the EOS side, because it will try running a recursive find on a file, which fails. 
* `RemoveACL`  fails, because it tries matching ACL rules with a uid to ACL rules with a username. This PR changes this approach to use an approach similar to what is used in the binary client: just set the rule that you want to have deleted with no permissions.


glpatcern (approved):
Looks good!

PR #4901, opened 2024-10-25
Fix broken PROPFIND permissions in projects over gRPC

jessegeens:
The permissions returned by `PROPFIND` on a folder in a project were erroneous because ACL permissions were being ignored. This stems from a bug in `grpcMDResponseToFileInfo`, where the `SysACL` attribute of the `FileInfo` struct was not being populated.

### DIFF
--- a/pkg/eosclient/eosgrpc/eosgrpc.go
+++ b/pkg/eosclient/eosgrpc/eosgrpc.go
@@ -1242,7 +1242,9 @@ func (c *Client) List(ctx context.Context, auth eosclient.Authorization, dpath s
 			if parent != nil && parent.SysACL != nil {
 				if fi.SysACL == nil {
 					log.Warn().Str("func", "List").Str("path", dpath).Str("SysACL is nil, taking parent", "").Msg("grpc response")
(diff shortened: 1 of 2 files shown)


pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -1629,6 +1631,10 @@ func (c *Client) grpcMDResponseToFileInfo(ctx context.Context, st *erpc.MDRespon
 			fi.Attrs[strings.TrimPrefix(k, "user.")] = string(v)
 		}
 
+		if fi.Attrs["sys.acl"] != "" {
```
diocas, line 1634: Duplicate of the code bellow?

diocas (changes requested):

pkg/eosclient/eosgrpc/eosgrpc.go
jessegeens: Different if statement. I can also not really pull it out because `fi.Attrs` must be populated first, which happens differently in each branch

diocas (approved):

PR #4905, opened 2024-10-30
Fix bug in ListDeletedEntries where the PurgeDate was not set correctly

jessegeens:
The date range that can be passed to `ListDeletedEntries` was not taken into account due to a bug in reva: the `Purgedate` argument was set, which only works for `PURGE` requests, and not for `LIST` requests. Instead, the `Listflag` argument must be used.  Additionally, there was a bug in the loop that is used to iterate over all days in the date range.


glpatcern (approved):
Thanks, looks good

PR #4908, opened 2024-11-04
ocm: extract file checksum from owncloud specific properties

rhafer:
When the remote instance of the OCM storage provider returns file checksums in
its PROPFIND responses we're now passing them through to in Stat responses.
This allows e.g. the oCIS thumbnailer to work with ocm shares.

See: https://github.com/owncloud/ocis/issues/10272


kobergj (approved):

micbar (approved):
Nice 🎉

PR #4923, opened 2024-11-08
fix download consistency

butonic:
fixes https://github.com/owncloud/ocis/issues/10526

- [x] refactor code to reuse sections
- [x] changelog


butonic:
How would changing the signature of `Download()` differ from using the `ConsistentDownloader` interface? `ConsistentDownload()` already returns both: the metadata (so we can omit the stat, which is used to return the content length header ... and we should return the checksum IIRC) as well as a callback to get a `ReadCloser`. Why a callback? Because for s3 getting the reader might be expensive and unnecessary for a HEAD operation.

🤔 

Posixps could already open the file but then how do we close it if it is never read?

aduffeck:
Hm. No, we [return early](https://github.com/cs3org/reva/blob/edge/pkg/rhttp/datatx/utils/download/download.go#L113) when the etag didn't change, without calling `getContent`, no?  `Download` already handles the `if-match` header case. If we also teach it the `if-none-match` case we wouldn't have to open the reader for the cases where it won't be needed.

butonic:
Yeah, I'd really like to be efficient in that case and not even open the file then.

pkg/storage/utils/decomposedfs/decomposedfs.go
```diff
@@ -1054,54 +1054,49 @@ func (fs *Decomposedfs) Delete(ctx context.Context, ref *provider.Reference) (er
...
-		return nil, errors.Wrap(err, "Decomposedfs: error download blob '"+n.ID+"'")
+	var reader io.ReadCloser
+	if openReaderFunc(ri) {
+		reader, err = fs.tp.ReadBlob(n)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "Decomposedfs: error download blob '"+n.ID+"'")
+		}
 	}
-	return reader, nil
+	return ri, reader, nil
```
butonic, line 1099: hm, we are returning the ResourceInfo here ... even though the user may not have stat permissions ... I'd say that is ok for now, because Download() is only calvled by the deteprovider. still meh.

butonic:
urgh the ceph fs build fails ... we should have used an additional interface ...

pkg/storage/utils/decomposedfs/decomposedfs.go
kobergj: We are also returning a reader to the file. So I see no additional security implications.

kobergj (approved):

PR #4928, opened 2024-11-12
Propagate lock in PROPPATCH in WebDAV

roman-vanicek:
## Description
Windows built-in WebDAV client in default configuration cannot copy/create files after mounting owncloud as a disk drive using net use D: ... This is one of two necessary changes to make it work.

## Related Issue

- another issue that otherwise blocks Windows WebDAV client is a name of the HTTP header that contains the lock value. In previous version of reva it used the HTTP header "If". The current version uses header "Lock-Token". I do not know the reason for this change so it is not part of this PR. Moreover this issue can be worked around on reverse proxy by duplicating the HTTP header value.

## Motivation and Context

No Windows client can use WebDAV in the default configuration on ocis and Windows.

## How Has This Been Tested?

Tested on my own ocis build that contains just this fix. No test case added.


kobergj (approved):
:+1: Code looks good. But it needs a changelog in `changelog/unreleased` folder to get a green pipeline

kobergj (approved):

kobergj (approved):
Thanks @roman-vanicek :+1:

roman-vanicek:
not much better now unfortunately

micbar:
Thanks!

PR #4932, opened 2024-11-13
fix: make TUS POST more robust

micbar:
Bugfix: Make TUS POST more robust

We improved the handling of TUS POST uploads when the metadata can not be read directly after sucessfully uploading the file to the storage.


kobergj (approved):

micbar:
This doesn't solve the issue we found, closing.

PR #4924, opened 2024-11-11
Fix SyncPropagation Defer

kobergj:
Fixes the defer in the sync propagation. Removes the unnecessary `Close` call.

PR #4936, opened 2024-11-14
do not delete mlock files after update

butonic:
To prevent stale NFS file handles we no longer delete empty mlock files after updating the metadata.

related https://github.com/cs3org/reva/pull/4924


micbar:
override merging this, POSIXFS tests are unrelated.

PR #4937, opened 2024-11-14
@labkode steps down as owner of the Reva project.

labkode:
Title says it all + make sure Jesse exists :)


glpatcern (approved):
If there's one PR I don't like to approve, this is it :)

You won't escape the debugging sessions though ;)

Issue #601, opened 2020-03-26
Discuss home logic


labkode:
To simplify our lives:

1) We need to include #578, that will mean that the SP responsibility is to always handle paths that start with root (/), only for forever, mount paths are left to the gw+registry.

2) Assuming 1, we can configure a SP to be provider of a `home`. That means that this SP will need to map root(/) to a user home directory. Internal business is not exposed.

Having 1 and 2, we can enforce namespace in the gateway the following convention:

- `/home` to always map to the home SP (this means GetHome()  on the GW always returns `/home`)

Still leaving the possibility to expose the home SP under other paths:

- `/users/` can also be mapped to the home SP (this means GetHome() on the GW needs to take into consideration the user in context and reply with `/users/hugo` and use it for wrap/unwrap paths instead of just `home`. This model is particularly useful when exposing the home directory of a user in a global view.

With these changes there is not need anymore of having the GetHome method implemented in the SP and in the storage drivers, simplifying its implementation.

madsi1m:
Good idea

PR #4911, opened 2024-11-05
cephfs refactoring + make home layout configurable

glpatcern:
Authored by @labkode, being exercised in QA

PR #4940, opened 2024-11-15
Refactoring of /home

glpatcern:
This is work in progress to restore #4911 in full.

- [x] Check why the eos storage provider returns 404 on PROPFIND
- [ ] Remove eoshomewrapper (from https://github.com/cernbox/reva-plugins/tree/master/storage) and consolidate to a single eos storage provider -> we'll address that in a different PR.


internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -924,17 +928,6 @@ func (s *service) ListContainerStream(req *provider.ListContainerStreamRequest,
 
 func (s *service) ListContainer(ctx context.Context, req *provider.ListContainerRequest) (*provider.ListContainerResponse, error) {
 	newRef, err := s.unwrap(ctx, req.Ref)
-	if err != nil {
```
jessegeens, line 927: Why is the error handling gone here? `unwrap` can still return an error right?

jouvin:
Fixes https://github.com/cs3org/reva/issues/5069

glpatcern:
Fixes #601

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -107,171 +103,30 @@ func (s *svc) GetHome(ctx context.Context, _ *provider.GetHomeRequest) (*provide
...
-	}
-
-	if s.isSharedFolder(ctx, p) {
-		log.Debug().Str("path", p).Msg("path points to shared folder")
-		err := errtypes.PermissionDenied("gateway: cannot download share folder: path=" + p)
-		log.Err(err).Msg("gateway: error downloading")
+	statReq := &provider.StatRequest{Ref: req.Ref}
+	statRes, err := s.stat(ctx, statReq)
+	if err != nil {
 		return &gateway.InitiateFileDownloadResponse{
```
jessegeens, line 119: Let's maybe also log this?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -676,137 +326,8 @@ func (s *svc) TouchFile(ctx context.Context, req *provider.TouchFileRequest) (*p
...
-			}, nil
-		}
-
-		// append child to target
-		req.Ref.Path = path.Join(ri.Path, shareChild)
-		return s.delete(ctx, req)
-	}
-
-	panic("gateway: delete called on unknown path:" + p)
+	return s.delete(ctx, req)
```
jessegeens, line 330: Can we not directly embed this in here?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -567,75 +281,11 @@ func (s *svc) GetPath(ctx context.Context, req *provider.GetPathRequest) (*provi
 }
 
 func (s *svc) CreateContainer(ctx context.Context, req *provider.CreateContainerRequest) (*provider.CreateContainerResponse, error) {
-	log := appctx.GetLogger(ctx)
-
 	if utils.IsRelativeReference(req.Ref) {
```
jessegeens, line 284: why the check if you do the same thing in both cases?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -341,148 +196,7 @@ func (s *svc) initiateFileDownload(ctx context.Context, req *provider.InitiateFi
...
-			}, nil
-		}
-
-		// append child to target
-		req.Ref.Path = path.Join(ri.Path, shareChild)
-		return s.initiateFileUpload(ctx, req)
-	}
-
-	panic("gateway: upload: unknown path:" + p)
+	return s.initiateFileUpload(ctx, req)
```
jessegeens, line 199: Can we not directly embed this?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -830,118 +351,7 @@ func (s *svc) delete(ctx context.Context, req *provider.DeleteRequest) (*provide
...
-		req.Source = src
-		req.Destination = dst
-
-		return s.move(ctx, req)
-	}
-
-	return &provider.MoveResponse{
-		Status: status.NewStatusFromErrType(ctx, "move", errtypes.BadRequest("gateway: move called on unknown path: "+p)),
-	}, nil
+	return s.move(ctx, req)
```
jessegeens, line 354: Can we not directly embed this in here?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1248,147 +567,7 @@ func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.St
...
-		}
-
-		// we need to make sure we don't expose the reference target in the resource
-		// information.
-		res.Info.Path = p
-		return res, nil
-	}
-
-	panic("gateway: stating an unknown path:" + p)
+	return s.stat(ctx, req)
```
jessegeens, line 570: Same comment here:
a) Why the check for relative reference if we don't use it?
b) it's just a call to `stat`, so why even split into `Stat` and `stat`?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -1664,183 +765,7 @@ func (s *svc) listContainerAcrossProviders(ctx context.Context, req *provider.Li
...
-		for _, info := range newRes.Infos {
-			base := path.Base(info.Path)
-			info.Path = path.Join(shareName, shareChild, base)
-		}
-
-		return newRes, nil
-	}
-
-	panic("gateway: stating an unknown path:" + p)
+	return s.listContainer(ctx, req)
```
jessegeens, line 768: No need to embed

internal/grpc/services/storageprovider/storageprovider.go
```diff
@@ -172,6 +172,10 @@ func New(ctx context.Context, m map[string]interface{}) (rgrpc.Service, error) {
 		return nil, err
 	}
 
+	if fs == nil {
```
jessegeens, line 175: Should `getFS` not return an error if `fs` is `nil`?

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/uploads.go
```diff
@@ -49,32 +49,41 @@ func (h *Handler) getCapabilitiesForUserAgent(_ context.Context, userAgent strin
...
-	case chunkV1:
-		// 2.7+ will use Chunking V1 if "capabilities > files > bigfilechunking" is "true" AND "capabilities > dav > chunking" is not there
-		c.Files.BigFileChunking = true
-		c.Dav = nil
-		c.Files.TusSupport = nil
+	// 2.7+ will use Chunking V1 if "capabilities > files > bigfilechunking" is "true" AND "capabilities > dav > chunking" is not there
+	c.Files.BigFileChunking = true
+	c.Dav = nil
+	c.Files.TusSupport = nil
+	/*
```
jessegeens, line 56: Do we need to keep the comment?

jessegeens (changes requested):

internal/grpc/services/gateway/storageprovider.go
glpatcern: Actually this Stat is redundant, because in `initiateFileDownload` we already call `s.find()`!

internal/grpc/services/storageprovider/storageprovider.go
glpatcern: `getFS` returns `f(ctx, c.Drivers[c.Driver])`, which itself may return `nil`

internal/http/services/owncloud/ocs/handlers/cloud/capabilities/uploads.go
glpatcern: Good question, I'd say yes for now if we have to resurrect support for TUS.

glpatcern:
@rawe0 if you commit your fix it would be great, and then @jessegeens please have a look again - there was indeed quite more stuff to consolidate/drop!

jessegeens (approved):

PR #4951, opened 2024-11-18
Pass logger

aduffeck:
Make the initialized logger available further down the stack (grpc services, storage drivers).


micbar (approved):
Awesome!

kobergj (approved):

PR #4954, opened 2024-11-18
prevent panic when logging error

butonic:
We fixed a panic when constructing a path failed to get the parent for a node.

### DIFF
--- a/pkg/storage/utils/decomposedfs/lookup/lookup.go
+++ b/pkg/storage/utils/decomposedfs/lookup/lookup.go
@@ -235,14 +235,17 @@ func (lu *Lookup) GenerateSpaceID(spaceType string, owner *user.User) (string, e
 // Path returns the path for node
 func (lu *Lookup) Path(ctx context.Context, n *node.Node, hasPermission node.PermissionFunc) (p string, err error) {
 	root := n.SpaceRoot
+	var child *node.Node
 	for n.ID != root.ID {
 		p = filepath.Join(n.Name, p)
+		child = n
 		if n, err = n.Parent(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/storage/utils/decomposedfs/lookup/lookup.go
```diff
@@ -27,17 +27,18 @@ import (
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/cs3org/reva/v2/pkg/appctx"
 	"github.com/cs3org/reva/v2/pkg/errtypes"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/metadata/prefixes"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/node"
 	"github.com/cs3org/reva/v2/pkg/storage/utils/decomposedfs/options"
-	"github.com/google/uuid"
```
micbar, line 36: 1. Why do you sort it differently?
2. Why do you sort it in a wrong way?

butonic: it is not a *wrong* way: https://github.com/golang/go/issues/32049

I don't know who introduced the grouping by std, local, external imports, but I wanted to play nice end enabled it in my VS Code. 

I'm happy to revert to stock go fmt ... but expect annoyed PR reviews by me in the future, then.

rhafer (approved):
lgtm

micbar:
merging, failures are unrelated.

PR #4959, opened 2024-11-19
attempt to fix sse

dragonchaser:
Bugfix: Fix missing file touched event

We have fixed an issue where the `file touched` event was not being triggered when an
office document was created.

refs https://github.com/owncloud/ocis/issues/8950


butonic (approved):

butonic:
@aduffeck I think then we can revert https://github.com/cs3org/reva/pull/4479/files#diff-bd2b21d1a9573e95f2ce195dc75f4a759d15940ab7b641a1f147cfa613e74a51R96-R102 because tha simple upload manager should absolutely be able to handle 0 byte uploads

PR #4963, opened 2024-11-19
Fix name and displayName in an ocm PROPFIND response

2403905:
Related issue [#10582](https://github.com/owncloud/ocis/issues/10582)

### DIFF
--- a/changelog/unreleased/fix-ocm-propfind.md
+++ b/changelog/unreleased/fix-ocm-propfind.md
@@ -1,0 +1,6 @@
+Bugfix: Fix name and displayName in an ocm 
+
+Fixed name and displayName in an ocm PROPFIND response
+
+https://github.com/cs3org/reva/pull/4963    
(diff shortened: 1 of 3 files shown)


pkg/ocm/storage/received/ocm.go
```diff
@@ -284,7 +284,7 @@ func convertStatToResourceInfo(ref *provider.Reference, f fs.FileInfo, share *oc
 		Id:       id,
 		MimeType: mime.Detect(f.IsDir(), f.Name()),
 		Path:     name,
-		Name:     name,
+		Name:     filepath.Base(name),
```
rhafer, line 287: Isn't this a bug in the webdav module? I guess we should fix it there. Shouldn't we?

2403905: Moved to the gowebdav

rhafer (approved):

PR #4967, opened 2024-11-21
[tests-only] Fix posixfs flakiness

aduffeck:
Fix posixfs flakiness


micbar:
@aduffeck Interesting. How did that fix it?

aduffeck:
@micbar the problem was that the integration tests create new users for each test, with different ids but the same username. That in return lead to the space roots having the same path according to the `"users/{{.User.Username}}"` path template, leading to clashes and confusion in the cache. With the `"users/{{.User.Id.OpaqueId}}"` template it is guaranteed that there's a clean state for each test run now.

micbar (approved):

PR #4969, opened 2024-11-22
Fixed uploads that fail due to removing cbox account from ACL

jessegeens:
Uploads would return a 500 error, even though they succeeded, due to a failed `Stat` after an upload. The `Stat` failed because it was executed as `cbox`, which no longer has access to user files. This is fixed by using the user's auth now instead.


glpatcern:
For the record, this was reverted with commit `4820f8ca4d95bee8` and a different solution is being developed.

PR #4971, opened 2024-11-25
Use manager to list shares

butonic:
When updating a received share the usershareprovider now uses the share manager directly to list received shares instead of going through the gateway again.

fixes https://github.com/owncloud/ocis/issues/10646


butonic:
hm ... the whole ocs share manager is actually dead unused code .... but we cannot kill it because the reva tests still use them...

micbar (approved):

PR #4984, opened 2024-12-03
eosfs: use external script for home creation + drop support for shadow namespace

glpatcern:
This comes as part of the effort to operate EOS without being root, see https://github.com/cs3org/reva/pull/4977/

In this PR the post-home creation hook (and corresponding flag) is replaced by a `create_home_hook`, and the following configuration parameters are suppressed:
* `shadow_namespace`
* `share_folder`
* `default_quota_bytes`
* `default_secondary_quota_bytes`
* `default_quota_files`
* `uploads_namespace` (unused)

### DIFF
--- a/pkg/storage/utils/eosfs/eosfs.go
+++ b/pkg/storage/utils/eosfs/eosfs.go
@@ -289,50 +289,27 @@ func getUser(ctx context.Context) (*userpb.User, error) {
 
 func (fs *eosfs) getLayout(ctx context.Context) (layout string) {
 	if fs.conf.EnableHome {
-		u, err := getUser(ctx)
-		if err != nil {
-			panic(err)
-		}
+		u := ctx.ContextMustGetUser(ctx)
 		layout = templates.WithUser(u, fs.conf.UserLayout)
 	}
 	return
 }
 
-func (fs *eosfs) getInternalHome(ctx context.Context) (string, error) {
+func (fs *eosfs) getInternalHome(ctx context.Context) (string) {
 	if !fs.conf.EnableHome {
-		return "", errtypes.NotSupported("eos: get home not supported")
+		// TODO(lopresti): this is to be removed as we always want to support home,
+		// cf. https://github.com/cs3org/reva/pull/4940
+		return "/"
 	}
 
-	u, err := getUser(ctx)
-	if err != nil {
-		err = errors.Wrap(err, "eosfs: wrap: no user in ctx and home is enabled")
-		return "", err
-	}
-
+	u := ctx.ContextMustGetUser(ctx)
 	relativeHome := templates.WithUser(u, fs.conf.UserLayout)
-	return relativeHome, nil
-}
-
-func (fs *eosfs) wrapShadow(ctx context.Context, fn string) (internal string) {
-	if fs.conf.EnableHome {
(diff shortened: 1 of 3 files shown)


pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1450,40 +1270,23 @@ func (fs *eosfs) createNominalHome(ctx context.Context) error {
...
-
-	err = fs.c.SetQuota(ctx, rootAuth, quotaInfo)
-	if err != nil {
-		err := errors.Wrap(err, "eosfs: error setting quota")
-		return err
-	}
 
-	if fs.conf.EnablePostCreateHomeHook {
-		if err := fs.runPostCreateHomeHook(ctx); err != nil {
+	if fs.conf.CreateHomeHook != "" {
```
jessegeens, line 1279: And the CreateHomeHook will be a local call to Manos' script then?

glpatcern: Yes! Eventually to be replaced by a call to an HTTP endpoint.

jessegeens (approved):

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1450,40 +1270,23 @@ func (fs *eosfs) createNominalHome(ctx context.Context) error {
...
 	}
 
-	err = fs.createUserDir(ctx, u, home, false)
-	if err != nil {
-		err := errors.Wrap(err, "eosfs: error creating user dir")
-		return err
-	}
-
 	// set quota for user, depending on its type
 	quotaBytes := fs.conf.DefaultQuotaBytes
```
diocas, line 1274: If we no longer handle the home creation here, maybe it would make more sense to pass the type of account to the script, and it's there that we decide the quota we give and not here?

diocas: Then we don't need to update reva to change the quota logic.

diocas (commented):
What was the reason for the shadow ns? Maybe it's worth to explain in the PR why it's ok to remove it.

glpatcern:
> What was the reason for the shadow ns? Maybe it's worth to explain in the PR why it's ok to remove it.

I actually put that in a [ADR](https://gitlab.cern.ch/cernbox/adr/-/merge_requests/8). Hugo just confirmed this is a legacy attempt to implement the "share jail", but it has never been used.

pkg/storage/utils/eosfs/eosfs.go
glpatcern: I had exactly the same thought, but yet preferred to have the "policy brain" in Reva so the script is a simple "agent".

Thinking at this again now, and with the idea that possibly the lifecycle action is called outside of Reva, it actually makes sense to externalize the policy as well, otherwise we may have duplication of policy configuration. The script will include and "own" the quota policy then.

diocas: I just realised that we might not even want to pass the type of account and let the account be resolved in the script

diocas: Agreed to keep it, as it's harder to get the information in the script.

diocas (approved):

PR #4986, opened 2024-12-03
Fix ListStorageSpaces for ocm

rhafer:
Currently there are no mountpoints for federated shares. We just return an empty list on the ListStorageSpaces request.

Related: https://github.com/owncloud/ocis/issues/10689


rhafer:
Now I am confused. So we have tests that expect the ocm shares to be accessible via  `/ocm/{id}`. For that to work the ListStorageSpaces request acutally needs to return them as mountpoints, I think.

I guess this needs some discussion.

PR #5022, opened 2024-12-30
[tests-only][full-ci] update drone config to run specific parts in pipeline

prashant-gurung899:
Related issue:  https://github.com/owncloud/QA/issues/871

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -117,7 +117,7 @@ def main(ctx):
         cs3ApiValidatorS3NG(),
         # virtual views don't work on edge at the moment
         #virtualViews(),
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
This should be OK. It will run the integration tests in less parallel pipelines. On average each pipeline will take longer, because it will have more test scenarios.

But, over time, we have skipped various test scenarios on reva, so some of the previous "many" pipelines ended up having no scenarios in them (all the scenarios happened to be skipped). So, actually, it is good to reduce the number of separate pipelines that we run. That will reduce the "flood" of pipelines that the drone CI server has to try and start.

PR #5028, opened 2025-01-07
eosclient: handle empty traces

labkode:
Empty traces (i.e not defined in `ctx`) cause the eosclient to fail commands that set the `--comment` flag.


glpatcern (approved):
Interesting finding, thanks!

PR #5029, opened 2025-01-07
Pseudo-transactionalize sharing

jessegeens:
Currently, sharing is not transactionalized: setting ACLs and writing the share to the db is completely independent. In the current situation, shares are written to the db before setting the ACL, letting users falsely believe that they successfully shared a resource, even if setting the ACL afterwards fails. his enhancement improves the situation by doing the least reliable (setting ACLs on EOS) first:
a) first pinging the db
b) writing the ACLs
c) writing to the db

### DIFF
--- a/changelog/unreleased/pseudo-tx-share.md
+++ b/changelog/unreleased/pseudo-tx-share.md
@@ -1,0 +1,8 @@
+Enhancement: pseudo-transactionalize sharing
+
+Currently, sharing is not transactionalized: setting ACLs and writing the share to the db is completely independent. In the current situation, shares are written to the db before setting the ACL, letting users falsely believe that they successfully shared a resource, even if setting the ACL afterwards fails. his enhancement improves the situation by doing the least reliable (setting ACLs on EOS) first:
+a) first pinging the db
+b) writing the ACLs
+c) writing to the db
+
+https://github.com/cs3org/reva/pull/5029
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"strings"
 
 	userpb "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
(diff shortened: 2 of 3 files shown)


internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -38,33 +39,43 @@ import (
 )
 
 // TODO(labkode): add multi-phase commit logic when commit share or commit ref is enabled.
+// See reva ADR 003
```
glpatcern, line 42: I'd drop this here and also the TODO above

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -78,18 +89,33 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 					Status: denyGrantStatus,
 				}, err
 			}
-			return res, nil
+			// return res, nil
```
glpatcern, line 92:
```suggestion
```

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -78,18 +89,33 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
-		}
-		if addGrantStatus.Code != rpc.Code_CODE_OK {
-			return &collaboration.CreateShareResponse{
-				Status: addGrantStatus,
-			}, err
-		}
+	// Then we commit to the db
+	// ------------------------
+
+	// TODO the user share manager needs to be able to decide if the current user is allowed to create that share (and not eg. incerase permissions)
```
glpatcern, line 109: This comment comes from before and would apply to the storage provider rather than the DB as "share manager". But as a matter of fact, it does not apply to us in the sense that our storage drivers - EOS in particular here, and CephFS in the future - are doing that check. So it can be removed.

internal/grpc/services/usershareprovider/usershareprovider.go
```diff
@@ -172,7 +172,7 @@ func (s *service) GetShare(ctx context.Context, req *collaboration.GetShareReque
 	if err != nil {
 		return &collaboration.GetShareResponse{
 			Status: status.NewInternal(ctx, err, "error getting share"),
-		}, nil
+		}, err
```
glpatcern, line 175: I believe this was one of those "tricks" to convert an error to a proper response (but with `Internal Error`) that does not carry an `err` on top.
So returning `nil` ought to be correct, and ideally the `NewInternal` could be changed.

glpatcern (approved):
Looks good, just a few minor comments

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -78,18 +87,29 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
-			return res, nil
+		} else {
+			addGrantStatus, err := s.addGrant(ctx, req.ResourceInfo.Id, req.Grant.Grantee, req.Grant.Permissions.Permissions)
+			if err != nil {
+				return nil, errors.Wrap(err, "gateway: error adding grant to storage")
+			}
+			if addGrantStatus.Code != rpc.Code_CODE_OK {
+				return &collaboration.CreateShareResponse{
+					Status: addGrantStatus,
+				}, err
```
diocas, line 98: I know it's the same, but `err` is (and can only be) `nil` so I would return `nil` here for clarity.

changelog/unreleased/pseudo-tx-share.md
```diff
@@ -0,0 +1,8 @@
+Enhancement: pseudo-transactionalize sharing
+
+Currently, sharing is not transactionalized: setting ACLs and writing the share to the db is completely independent. Because of this, sometimes a share is written to the database even though the ACLs have not been set. This enhancement fixes this by
```
diocas, line 3: You do not fix it; you make the situation better. Just explain that the DB falsely led users to believe it was shared and is more reliable than the storage, so we improve the situation by doing the least reliable first.

diocas (dismissed):

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -76,20 +85,31 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
 					Status: denyGrantStatus,
-				}, err
+				}, nil
+			}
+		} else {
+			addGrantStatus, err := s.addGrant(ctx, req.ResourceInfo.Id, req.Grant.Grantee, req.Grant.Permissions.Permissions)
+			if err != nil {
+				return nil, errors.Wrap(err, "gateway: error adding grant to storage")
+			}
+			if addGrantStatus.Code != rpc.Code_CODE_OK {
```
glpatcern, line 95: So following the discussion today, we could log here the `addGrantStatus.Message` before returning (or return it as part of the response), hoping to learn why the storage fails at times to create the share.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -76,20 +85,31 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
+	// Then we commit to the db
+	// ------------------------
+	res, err := shareClient.CreateShare(ctx, req)
+
+	if err != nil {
+		log.Error().Msg("Failed to Create Share but ACLs are already set")
+		return nil, errors.Wrap(err, "gateway: error calling CreateShare")
+	}
+	if res.Status.Code != rpc.Code_CODE_OK {
+		return res, nil
```
glpatcern, line 112: This seems some left-over copy/paste, probably to be dropped (we already return `res, nil` below anyway)

jessegeens: Whoops, I missed this comment. If the RPC returns an error through `res` I think we should also return an error, so updated the code to do this

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -76,20 +85,32 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
-			return &collaboration.CreateShareResponse{
-				Status: addGrantStatus,
-			}, err
-		}
+	// Then we commit to the db
+	// ------------------------
+	res, err := shareClient.CreateShare(ctx, req)
+
+	if err != nil {
+		log.Error().Msg("Failed to Create Share but ACLs are already set")
```
glpatcern, line 109: Also here we should log some more context, i.e. `req.ResourceInfo` and `req.Grant.Grantee` so we can track things.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -76,20 +85,32 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 			if denyGrantStatus.Code != rpc.Code_CODE_OK {
 				return &collaboration.CreateShareResponse{
 					Status: denyGrantStatus,
-				}, err
+				}, nil
+			}
+		} else {
+			addGrantStatus, err := s.addGrant(ctx, req.ResourceInfo.Id, req.Grant.Grantee, req.Grant.Permissions.Permissions)
+			if err != nil {
+				log.Error().Err(err).Str("Message", addGrantStatus.Message).Msg("Failed to Create Share: error during addGrant")
```
glpatcern, line 93: Good for the log, let's add also `req.ResourceInfo` and `req.Grant.Grantee` to the message

glpatcern (approved):
I think this is ready to be merged now, just two more additions to the log

glpatcern (approved):

PR #5060, opened 2025-02-03
[tests-only][full-ci]Bump latest ocis commit id

prashant-gurung899:
Part of: https://github.com/owncloud/QA/issues/876

### DIFF
--- a/.drone.env
+++ b/.drone.env
@@ -1,4 +1,4 @@
 # The test runner source for API tests
-APITESTS_COMMITID=d7156bce943299aa117521eab2a42c82585fb0a5
+APITESTS_COMMITID=527cddbc6402a68bcb8447dd52ca503ac221e491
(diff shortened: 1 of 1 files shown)


phil-davis:
I don't see any drone CI result here.
Has drone been disabled on the `edge` branch?

prashant-gurung899:
> I don't see any drone CI result here. Has drone been disabled on the `edge` branch?

May be it's disabled, as there's no drone CI running in the previous PR as well. https://github.com/cs3org/reva/pull/5052
@phil-davis

phil-davis:
> May be it's disabled, as there's no drone CI running in the previous PR as well.

Please find out what is the situation with running acceptance tests in CI for cs3org/reva `edge` branch.

There is no point moving to newer versions of test code, if the tests are not going to be run in CI.

PR #5061, opened 2025-02-03
Fix OCM Wildcards

kobergj:
Allows sharing with more than one instance using wildcards


2403905 (approved):

saw-jan:
@kobergj @2403905 Is it something that we can test in API?

saw-jan:
NOTE:
It was an issue with config containing wildcards
https://github.com/owncloud/ocis/blob/7146fefb3414ff1b8769517159c0d4527111b9ff/deployments/examples/ocis_ocm/config/ocis/ocm.providers.json#L29
Testing directly with API might not be possible as the config is static

PR #5071, opened 2025-02-05
Fix role conversion

bastianbeier:
Fix role from resource permission conversion


kobergj (commented):
@2403905 you mentioned we don't need this at all. Should we put it in anyways and clean it up later? I'm not entirely sure it won't break something....

2403905:
lets keep it.

PR #5072, opened 2025-02-05
Fix for (sometimes) broken Download / Restore revisions

jessegeens:
Extension of #5064 for `DownloadRevision` and `RestoreRevision`


changelog/unreleased/revisions-impersonate
```diff
@@ -0,0 +1,5 @@
+Bugfix: impersonate owner on Revisions
```
diocas, line 1: Maybe it's worth to have a single changelog entry. Or at least don't change the previous entry.

diocas (dismissed):

changelog/unreleased/revisions-impersonate
jessegeens: Oops, yeah I tried just changing it first but the changelog checker didn't like that

diocas (approved):

PR #5076, opened 2025-02-12
WIP for OCM 1.2

glpatcern:


pkg/ocm/storage/received/ocm.go
```diff
@@ -414,3 +445,26 @@ func (d *driver) CreateStorageSpace(ctx context.Context, req *provider.CreateSto
 func (d *driver) UpdateStorageSpace(ctx context.Context, req *provider.UpdateStorageSpaceRequest) (*provider.UpdateStorageSpaceResponse, error) {
 	return nil, errtypes.NotSupported("operation not supported")
 }
+
+// Cleanup function to remove expired cache entries
+func (d *driver) cleanupCache() {
```
jessegeens, line 450: Why not use `ttlcache.Cache` here? Then you don't have to manually do cache cleanups like this, and we already use it in several places in Reva

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -263,42 +249,82 @@ func getOCMShareType(t string) ocm.ShareType {
...
+func discoverOcmRoot(r *http.Request, proto string) (string, error) {
+	// implements the OCM discovery logic to fetch the root at the remote host that sent the share for the given proto, see
 	// https://cs3org.github.io/OCM-API/docs.html?branch=v1.1.0&repo=OCM-API&user=cs3org#/paths/~1ocm-provider/get
 	ctx := r.Context()
 	log := appctx.GetLogger(ctx)
-	log.Debug().Str("sender", r.Host).Msg("received OCM 1.0 share, attempting to discover sender endpoint")
+
+	// assume the sender host is either given in the usual reverse proxy headers or as RemoteAddr, and that the
+	// remote end listens on https regardless if the incoming connection got its TLS terminated upstream of us
+	senderURL := r.Header.Get("X-Real-Ip")
```
jessegeens, line 300: I'm still not sure why we need this instead of just reading it from the request body (eventually signed). Because you can fake headers as easily as you can fake the body data, so I would think you may just as well just take whatever the initiator declares, which makes the behaviour clear, instead of relying on a bunch of headers and as a fallback the real remote addr

glpatcern: You know what? I completely agree! And for two reasons:
1) Security: I thought about it on Friday after talking to you, after all the headers or the `RemoteAddr` are not more trustworthy than the payload, and eventually a signed payload will be trustworthy.
2) Functionality: for deployments such as CERNBox, with reverse proxies in between, both the `RemoteAddr` and any eventual header are simply wrong.
Therefore, I'll change the implementation to use the given `URI`, and fall back to the `RemoteAddr` (and the `X-Real-IP` header as a 2nd fall back?) only if missing (case of Nextcloud). Plus will submit a PR to the OCM specs, to "un-deprecate" the full URI in the payload.

pkg/ocm/storage/received/ocm.go
glpatcern: Good point, will do!

glpatcern (commented):
OK, this is good enough for now: we discover remote endpoints according to specs, but we still send full URIs to remotes. Also the integration tests assume full URIs, so the discovery part is not covered (but was tested independently).
In the future we may reconfigure the test to set the owner's address to `localhost:<port>` and pass relative URIs so to trigger discovery (but `https` is assumed when contacting the remote, so we'd have to configure TLS on the test as well).

jessegeens (approved):

PR #5081, opened 2025-02-20
ORMify spaces

jessegeens:


glpatcern (approved):
This is very good in first approximation, we may want to further extend the schema as part of the review of Spaces but that can happen later

PR #5093, opened 2025-02-24
Fix OCM create share

2403905:
Related issue [ocis/#11046](https://github.com/owncloud/ocis/issues/11046)


internal/grpc/services/gateway/ocmshareprovider.go
```diff
@@ -56,6 +56,9 @@ func (s *svc) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareRequest
 	if err != nil {
 		return nil, errors.Wrap(err, "gateway: error calling CreateOCMShare")
 	}
+	if res.Status.Code != rpc.Code_CODE_OK {
```
kobergj, line 59: Can't we use getters?
```suggestion
	if res.GetStatus().GetCode() != rpc.Code_CODE_OK {
```

kobergj (changes requested):

PR #5099, opened 2025-02-25
[dirty] Fix Password Protected Folders

kobergj:
Just to be crystal clear: I don't like the fix very much. But it could help avoiding the password policies on Password Protected Folders.

@2403905  what do you think? How much do you NOT like it?

### DIFF
--- a/internal/grpc/services/publicshareprovider/publicshareprovider.go
+++ b/internal/grpc/services/publicshareprovider/publicshareprovider.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


2403905:
@kobergj We could use It but:
1. Should we check if the ".PasswordProtectedFolders" is in a Personal space root to avoid some manipulation going around the password policies for the regular public links like `<roo>/<subfolder>/.PasswordProtectedFolders/`?
2. If user able to update the password we should fix the `UpdatePublicShare` to https://github.com/cs3org/reva/blob/150e0960954e7ac0c423d8173e1360772b6b6dba/internal/grpc/services/publicshareprovider/publicshareprovider.go#L592
3. Is these changes relevant for the osc ?https://github.com/cs3org/reva/blob/edge/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go

PR #5123, opened 2025-03-13
Use binary client for Set / UnsetAttr

jessegeens:
EOS < 5.3 has a couple of bugs related to attributes:
* Attributes can only be removed as root or the owner, but over gRPC we cannot become root
* The recursive property is ignored on set attributes over gRPC

We circumvent these two issues by calling the binary client until we have deployed EOS 5.3\

That means that **this PR has to be reverted once EOS 5.3 is live**


glpatcern:
This is ugly, as discussed, but we need it. At this point I'd add the same for the `SetAttr`, so that we fix the other issue with recursivity, and once we deploy EOS 5.3 this whole PR can be reverted.

diocas (approved):

PR #5127, opened 2025-03-24
ListMyOfficeFiles

jessegeens:
This PR implements the `ListMyOfficeFiles` functionality.

**This PR is to be reverted in the future**

Since this is a temporary feature, there are some hacky things in here (such as making `eosfs` public, to circumvent having to update the storage interface)

Note that for this PR to work, changes have to also be merged into reva-plugins: see https://github.com/cernbox/reva-plugins/pull/25


internal/http/services/owncloud/ocdav/report.go
```diff
@@ -117,21 +119,31 @@ func (s *svc) doFilterFiles(w http.ResponseWriter, r *http.Request, ff *reportFi
...
 		}
-		w.Header().Set(HeaderDav, "1, 3, extended-mkcol")
-		w.Header().Set(HeaderContentType, "application/xml; charset=utf-8")
-		w.WriteHeader(http.StatusMultiStatus)
-		if _, err := w.Write([]byte(responsesXML)); err != nil {
-			log.Err(err).Msg("error writing response")
-		}
+	}
+
+	responsesXML, err := s.multistatusResponse(ctx, &propfindXML{Prop: ff.Prop}, resourceInfos, namespace, nil, nil)
```
diocas, line 136: I understand that spliting this code out of the if is much better, since it's used in both cases. On the other hand, it makes it harded to revert, which is what we need. If we duplicate the code, then the next time we revert, it's just a matter of removing the if, even if smth changed in the meantime with the favs code. Other option would be to split this in a separate commit, in which you move this code outside of the if, and the other commit just adds the new else if.

pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -703,6 +703,17 @@ func (c *Client) Rename(ctx context.Context, auth eosclient.Authorization, oldPa
 	return err
 }
 
+func (c *Client) ListWithRegex(ctx context.Context, auth eosclient.Authorization, path string, depth uint, regex string) ([]*eosclient.FileInfo, error) {
```
diocas, line 706: ListFilesWithRegex?

pkg/eosclient/eosgrpc/eosgrpc.go
```diff
@@ -1220,7 +1220,6 @@ func (c *Client) List(ctx context.Context, auth eosclient.Authorization, dpath s
 	fdrq.Id.Path = []byte(dpath)
 
 	fdrq.Role = new(erpc.RoleId)
```
diocas, line 1222: Unecessary change.

pkg/myofficefiles/manager.go
```diff
@@ -0,0 +1,99 @@
...
+type Manager interface {
+	// ListMyOfficeFiles returns all recent Office files of a user.
+	ListMyOfficeFiles(ctx context.Context, user *userpb.User) ([]*provider.ResourceInfo, error)
+}
+
+// This feature is only enabled for users that are in the
+const (
+	targetGroup      = "cernbox-office-view"
+	officeFilesRegex = "(.*?)(.xls|.doc|.ppt|.XLS|.DOC|.PPT)[x|X]?"
+	depth            = 5
```
diocas, line 30: Let's put at least 10

pkg/myofficefiles/manager.go
```diff
@@ -0,0 +1,99 @@
...
+// Manager defines an interface for a MyOfficeFiles manager.
+type Manager interface {
+	// ListMyOfficeFiles returns all recent Office files of a user.
+	ListMyOfficeFiles(ctx context.Context, user *userpb.User) ([]*provider.ResourceInfo, error)
+}
+
+// This feature is only enabled for users that are in the
+const (
+	targetGroup      = "cernbox-office-view"
+	officeFilesRegex = "(.*?)(.xls|.doc|.ppt|.XLS|.DOC|.PPT)[x|X]?"
```
diocas, line 29: The feature ticket requested that the filtering is done by type of office file. So our search should not return all office files combined in a single list, but instead by powerpoint OR excel OR word files.

diocas: Also, the regex is not matching the extension with the end of the filename.

diocas (changes requested):

pkg/eosclient/eosbinary/eosbinary.go
jessegeens: I can rename it if you want, chose this name because it is basically the same as `List`, just with a regex parameter added

internal/http/services/owncloud/ocdav/report.go
jessegeens: Ah yes, good point. I'm a fan of the second option :) I'll tackle the other issues first, and once everything works I'll split this out into a second commit that we can keep

diocas (approved):

PR #5129, opened 2025-03-26
added new Make target for local revaplugins

jessegeens:


glpatcern (approved):

glpatcern:
Yes, very welcome! We could even make it the default, and fall back to GitHub when the folder is missing, but that's already good!

PR #5132, opened 2025-03-31
Add expiration dates reinforcement

rodcoffani:


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/public.go
```diff
@@ -631,6 +692,17 @@ func isPermissionUploader(permissions *provider.ResourcePermissions) bool {
 	return conversions.RoleFromResourcePermissions(publicSharePermissions.Permissions).Name == conversions.RoleUploader
 }
 
+func isPermissionEditor(permissions *provider.ResourcePermissions) bool {
```
glpatcern, line 695: I think we have functions in some utils package that do that, no need to reimplement

rodcoffani: I found some similar functions, but none with the same behavior in order to replace this. There is [HasPublicShareRole](https://github.com/cs3org/reva/blob/eb3722dffe9f0ddbba04628ff5a7d3382ec9b800/pkg/utils/utils.go#L395), which returns the role name if exists, but it wouldn't replace the whole function, just the comparison line. There is also [hasRoleEditor](https://github.com/cs3org/reva/blob/eb3722dffe9f0ddbba04628ff5a7d3382ec9b800/pkg/auth/scope/scope.go#L61), but it returns true with the `Uploader` role as well.

As this function follows the same pattern as the one before it, `isPermissionUploader`, I think it should be ok to implement it here

changelog/unreleased/expiration-dates-reinforcement.md
```diff
@@ -0,0 +1,5 @@
+Feat: add expiration dates reinforcement
```
glpatcern, line 1:
```suggestion
Enhancement: add expiration dates reinforcement
```

changelog/unreleased/expiration-dates-reinforcement.md
```diff
@@ -0,0 +1,5 @@
+Enhancement: add expiration dates reinforcement
+
+Public links of folders with RW permissions now have a configurable expiration date.
+
+<https://github.com/cs3org/reva/pull/5132>
```
glpatcern, line 5:
```suggestion
https://github.com/cs3org/reva/pull/5132
```

rodcoffani: Sorry, auto-formatting :P

PR #5151, opened 2025-04-24
Fix for spaces bug related to opening apps

jessegeens:
There were still some places where Reva assumed that we are running spaces; while not verifying this. This caused non-spaces WebUIs to break. This PR fixes this.


pkg/spaces/utils.go
```diff
@@ -83,6 +83,9 @@ func EncodeResourceID(r *provider.ResourceId) string {
 // in the format <storage_id>$base32(<path>).
 func EncodeSpaceID(storageID, path string) string {
 	encodedPath := base32.StdEncoding.EncodeToString([]byte(path))
+	if encodedPath == "" {
```
glpatcern, line 86: isn't this equivalent to:
`if path == "" { ... }` ?

jessegeens: yes :)

glpatcern (approved):

PR #5174, opened 2025-05-21
Upgrade jwt library to v5

jessegeens:


glpatcern (approved):
All good, and this is now a breaking change

PR #5185, opened 2025-06-02
Refactor docs and tutorials

glpatcern:
With this PR we move all example configurations out to a dedicated repository, https://github.com/cs3org/reva-configs, and the tutorials to the Wiki. We also remove all obsoleted content, and keep the auto-generated doc on a dedicated folder.

This is to help the community further develop the documentation and configuration on separate repositories.

The Hugo site that hosts the auto-generated doc does not build but the issue is there since months in master. We will have to fix it at a later stage.


jessegeens (approved):
Approved, I like the idea of moving to the wiki and a separate repo :)

PR #5189, opened 2025-06-06
fix: shares parentReference

rodcoffani:
- rollback change: replace `md.Id.SpaceID` with the `storageSpaceID`.
  - instead of `<space-id>`, it should use `<storage-id>$<space-id>`
- fix: parentReference
  - add space info to `id`
  - removes double encoding of `driveId`
  - new function to return relative path inside a space root

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -518,21 +518,19 @@ func spaceHref(ctx context.Context, baseURI string, md *provider.ResourceInfo) (
 	if md.Id == nil || md.Id.SpaceId == "" {
 		return "", errors.New("Space ID must be set to calculate Href")
 	}
+	storageSpaceID := spaces.ConcatStorageSpaceID(md.Id.StorageId, md.Id.SpaceId)
 
-	spacePath, ok := ctx.Value(ctxSpacePath).(string)
+	_, spacePath, ok := spaces.DecodeStorageSpaceID(storageSpaceID)
 	if !ok {
-		// If no space path is set in the context, we calculate it
-		_, spacePath, ok = spaces.DecodeStorageSpaceID(fmt.Sprintf("%s$%s", md.Id.StorageId, md.Id.SpaceId))
-		if !ok {
-			return "", errors.New("Failed to decode space ID")
-		}
+		return "", errors.New("Failed to decode space ID")
 	}
 
 	relativePath, err := filepath.Rel(spacePath, md.Path)
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to calculate path relative to space root: %v", spacePath)
 	}
-	return path.Join(baseURI, md.Id.SpaceId, relativePath), nil
+
+	return path.Join(baseURI, storageSpaceID, relativePath), nil
 }
 
 func appendSlash(path string) string {
--- a/internal/http/services/owncloud/ocgraph/shares.go
+++ b/internal/http/services/owncloud/ocgraph/shares.go
@@ -219,7 +219,6 @@ func (s *svc) share(w http.ResponseWriter, r *http.Request) {
 	_ = json.NewEncoder(w).Encode(&ListResponse{
 		Value: lgPerm,
 	})
-
 }
 
 func (s *svc) createLink(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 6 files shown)


pkg/spaces/utils.go
```diff
@@ -154,10 +155,24 @@ func spacesLevel(path string) int {
 	}
 }
 
+// Removes the Space ID from the path of a resource info, preserving the leading slash.
 func RelativePathToSpaceID(info *provider.ResourceInfo) string {
 	return strings.TrimPrefix(info.Path, info.Id.SpaceId)
 }
 
+// Returns the path relative to the space root.
+func RelativePathToSpaceRoot(info *provider.ResourceInfo) (relativePath string, err error) {
```
jessegeens: I would rename to `PathRelativeToSpaceRoot`. This makes it seem like you turn a relative path into a space root.

rodcoffani: Didn't think of it that way, thanks!

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -532,7 +532,13 @@ func spaceHref(ctx context.Context, baseURI string, md *provider.ResourceInfo) (
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to calculate path relative to space root: %v", spacePath)
 	}
-	return path.Join(baseURI, md.Id.SpaceId, relativePath), nil
+
+	spaceID, ok := ctx.Value(ctxSpaceID).(string)
```
jessegeens: I don't see why we need to add a dependency on having the space id in the context here. You say 
>  instead of <space-id>, it should use <storage-id>$<space-id>, which is defined in the ctxSpaceID.

but if we already have the storage id and space id set in `md`, why not use that immediately?

Otherwise, if we really need to rely on this being in the context, let's also add that expectation to the comment of this function then :)

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -327,12 +326,11 @@ func (s *svc) createLink(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	_ = json.NewEncoder(w).Encode(lgPerm)
-
 }
 
 func encodeSpaceIDForShareJail(res *provider.ResourceInfo) string {
 	return spaces.EncodeResourceID(res.Id)
-	//return spaces.EncodeStorageSpaceID(res.Id.StorageId, res.Path)
+	// return spaces.EncodeStorageSpaceID(res.Id.StorageId, res.Path)
```
jessegeens: Let's maybe just drop this

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -555,24 +553,37 @@ func (s *svc) getSharedByMe(w http.ResponseWriter, r *http.Request) {
...
+	}
+
 	d := &libregraph.DriveItem{
 		ETag:                 libregraph.PtrString(info.Etag),
 		Id:                   libregraph.PtrString(spaces.EncodeResourceID(info.Id)),
 		LastModifiedDateTime: libregraph.PtrTime(utils.TSToTime(info.Mtime)),
 		Name:                 libregraph.PtrString(info.Name),
 		ParentReference: &libregraph.ItemReference{
-			DriveId: libregraph.PtrString(spaces.EncodeStorageSpaceID(info.Id.StorageId, info.Id.SpaceId)),
+			DriveId: libregraph.PtrString(fmt.Sprintf("%s$%s", info.ParentId.StorageId, info.ParentId.SpaceId)),
```
jessegeens: Don't we have a method to do the `fmt.Sprintf("%s$%s")` somewhere in the spaces utils? If not I would create this and use this instead, so we have a single place where we define that `$` is the demarcator for storage -> space

pkg/spaces/utils.go
```diff
@@ -154,10 +155,24 @@ func spacesLevel(path string) int {
 	}
 }
 
+// Removes the Space ID from the path of a resource info, preserving the leading slash.
```
jessegeens: Why do we keep the leading slash? Then it's not clear this path is relative ...
In any case I think the name of this function is also really bad. Do we use it anywhere still, or can we replace it with `PathRelativeToSpaceRoot`?

rodcoffani: This comment is related to `RelativePathToSpaceID`, to give more context to it.

`PathRelativeToSpaceRoot` is the next function, but I agree the name is a bit confusing between the two.

internal/http/services/owncloud/ocgraph/shares.go
rodcoffani: Added a new function `ConcatStorageSpaceID`.
About having a single place to define the separator between storage/space IDs, should we do the same thing even inside the space utils? i.e., calling this new from other functions like `EncodeResourceID` or `PathRelativeToSpaceRoot`?

pkg/spaces/utils.go
jessegeens: So let's maybe try to replace the usages of `RelativePathToSpaceID` with `PathRelativeToSpaceRoot`

jessegeens (approved):

PR #5190, opened 2025-06-06
fix: file versions

rodcoffani:
Current version listing for spaces crashes when building the file href.

Restoring and downloading: expecting resource id, web-ui sending parent folder id. 

Changes:
- new `spaceHref` function for versions
- fix `ParseResourceID` return, using the `SpaceID` instead of `path`

### DIFF
--- a/pkg/eosclient/eosbinary/eosbinary.go
+++ b/pkg/eosclient/eosbinary/eosbinary.go
@@ -43,11 +43,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-const (
-	versionPrefix = ".sys.v#."
-	favoritesKey  = "http://owncloud.org/ns/favorite"
-)
-
 func serializeAttribute(a *eosclient.Attribute) string {
 	return fmt.Sprintf("%s.%s=%s", attrTypeToString(a.Type), a.Key, a.Val)
 }
@@ -73,7 +68,6 @@ func isValidAttribute(a *eosclient.Attribute) bool {
 
 // Options to configure the Client.
 type Options struct {
-
 	// ForceSingleUserMode forces all connections to use only one user.
 	// This is the case when access to EOS is done from FUSE under apache or www-data.
 	ForceSingleUserMode bool
@@ -392,7 +386,7 @@ func (c *Client) GetFileInfoByInode(ctx context.Context, auth eosclient.Authoriz
 		return nil, err
 	}
 
-	if c.opt.VersionInvariant && isVersionFolder(info.File) {
+	if c.opt.VersionInvariant && eosclient.IsVersionFolder(info.File) {
 		info, err = c.getFileInfoFromVersion(ctx, auth, info.File)
 		if err != nil {
 			return nil, err
@@ -431,7 +425,7 @@ func (c *Client) GetFileInfoByPath(ctx context.Context, auth eosclient.Authoriza
 		return nil, err
 	}
 
-	if c.opt.VersionInvariant && !isVersionFolder(path) && !info.IsDir {
+	if c.opt.VersionInvariant && !eosclient.IsVersionFolder(path) && !info.IsDir {
 		ownerAuth := eosclient.Authorization{Role: eosclient.Role{
 			UID: strconv.FormatUint(info.UID, 10),
 			GID: strconv.FormatUint(info.GID, 10),
@@ -473,7 +467,7 @@ func (c *Client) SetAttr(ctx context.Context, auth eosclient.Authorization, attr
 	}
 
 	// Favorites need to be stored per user so handle these separately
-	if attr.Type == eosclient.UserAttr && attr.Key == favoritesKey {
+	if attr.Type == eosclient.UserAttr && attr.Key == eosclient.FavoritesKey {
 		info, err := c.getRawFileInfoByPath(ctx, auth, path)
 		if err != nil {
 			return err
@@ -521,7 +515,7 @@ func (c *Client) handleFavAttr(ctx context.Context, auth eosclient.Authorization
 			return err
 		}
 	}
-	favStr := info.Attrs[favoritesKey]
+	favStr := info.Attrs[eosclient.FavoritesKey]
 	favs, err := acl.Parse(favStr, acl.ShortTextForm)
 	if err != nil {
 		return err
@@ -557,7 +551,7 @@ func (c *Client) unsetEOSAttr(ctx context.Context, auth eosclient.Authorization,
 
 	var err error
 	// Favorites need to be stored per user so handle these separately
-	if !deleteFavs && attr.Type == eosclient.UserAttr && attr.Key == favoritesKey {
+	if !deleteFavs && attr.Type == eosclient.UserAttr && attr.Key == eosclient.FavoritesKey {
 		info, err := c.getRawFileInfoByPath(ctx, auth, path)
 		if err != nil {
 			return err
@@ -831,7 +825,7 @@ func (c *Client) PurgeDeletedEntries(ctx context.Context, auth eosclient.Authori
 
 // ListVersions list all the versions for a given file.
 func (c *Client) ListVersions(ctx context.Context, auth eosclient.Authorization, p string) ([]*eosclient.FileInfo, error) {
-	versionFolder := getVersionFolder(p)
+	versionFolder := eosclient.GetVersionFolder(p)
 	finfos, err := c.List(ctx, auth, versionFolder)
 	if err != nil {
 		// we send back an empty list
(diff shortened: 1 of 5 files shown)


internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -535,6 +539,24 @@ func spaceHref(ctx context.Context, baseURI string, md *provider.ResourceInfo) (
 	return path.Join(baseURI, md.Id.SpaceId, relativePath), nil
 }
 
+func versionSpaceHref(ctx context.Context, baseURI string, md *provider.ResourceInfo) (string, error) {
+	spacePath, ok := ctx.Value(ctxSpacePath).(string)
+	if !ok {
+		// If no space path is set in the context, we calculate it
+		_, spacePath, ok = spaces.DecodeStorageSpaceID(fmt.Sprintf("%s$%s", md.Id.StorageId, md.Id.SpaceId))
+		if !ok {
+			return "", errors.New("Failed to decode space ID")
```
jessegeens: Maybe specify in the error that this is in the `versionSpaceHref`?

internal/http/services/owncloud/ocdav/versions.go
```diff
@@ -142,9 +141,10 @@ func (h *VersionsHandler) doListVersions(w http.ResponseWriter, r *http.Request,
 	versions := lvRes.GetVersions()
 	infos := make([]*provider.ResourceInfo, 0, len(versions)+1)
 	// add version dir . entry, derived from file info
-	infos = append(infos, &provider.ResourceInfo{
-		Type: provider.ResourceType_RESOURCE_TYPE_CONTAINER,
-	})
+	_, spacePath, ok := spaces.DecodeStorageSpaceID(fmt.Sprintf("%s$%s", rid.StorageId, rid.SpaceId))
+	if !ok {
+		sublog.Error().Msg("error decoding storage space id")
```
jessegeens: should this not return?

internal/http/services/owncloud/ocdav/propfind.go
rodcoffani: Yes, also included a hint in the `spaceHref`

internal/http/services/owncloud/ocdav/versions.go
rodcoffani: it should! return and status error added.

internal/http/services/owncloud/ocdav/get.go
```diff
@@ -122,9 +121,9 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 
 	w.Header().Set(HeaderContentType, info.MimeType)
 	w.Header().Set(HeaderContentDisposistion, "attachment; filename*=UTF-8''"+
-		url.PathEscape(path.Base(r.URL.Path))+"; filename=\""+path.Base(r.URL.Path)+"\"")
```
diocas, line 125: Why no longer the escape?

internal/http/services/owncloud/ocdav/get.go
```diff
@@ -122,9 +121,9 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+		path.Base(info.Path)+"; filename=\""+path.Base(info.Path)+"\"")
 	w.Header().Set(HeaderETag, info.Etag)
-	w.Header().Set(HeaderOCFileID, resourceid.OwnCloudResourceIDWrap(info.Id))
+	w.Header().Set(HeaderOCFileID, spaces.EncodeResourceID(info.Id))
```
diocas, line 127: Does this break the current web? In principle no, because if spaces are disabled `SpaceId == ""` is always true, right?

diocas: Idem for others

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -519,12 +519,16 @@ func spaceHref(ctx context.Context, baseURI string, md *provider.ResourceInfo) (
 		return "", errors.New("Space ID must be set to calculate Href")
 	}
 
+	if md.Id.StorageId == "versions" {
```
diocas: Am I missing something, or this could be simplified to the following code at the end?

```go
if md.Id.StorageId == "versions" {
    return path.Join(baseURI, relativePath), nil
}
return path.Join(baseURI, md.Id.SpaceId, relativePath), nil
```

So, no extra `versionSpaceHref` func.

diocas: And can you add a comment explaining?

changelog/unreleased/ocdav-versions-api.md
```diff
@@ -0,0 +1,6 @@
+Enhancement: Add download endpoint to ocdav versions API
+
+Added missing endpoints to the ocdav versions API. This enables downloads of previous file versions. 
+
+https://github.com/owncloud/ocis/issues/3741
+https://github.com/cs3org/reva/pull/2855
```
diocas: You can reference their PR if there was smth taken from it, but this changelog is part of this PR so it needs to be identified.

internal/http/services/owncloud/ocdav/versions.go
```diff
@@ -142,9 +168,12 @@ func (h *VersionsHandler) doListVersions(w http.ResponseWriter, r *http.Request,
 	versions := lvRes.GetVersions()
 	infos := make([]*provider.ResourceInfo, 0, len(versions)+1)
 	// add version dir . entry, derived from file info
-	infos = append(infos, &provider.ResourceInfo{
-		Type: provider.ResourceType_RESOURCE_TYPE_CONTAINER,
-	})
+	_, spacePath, ok := spaces.DecodeStorageSpaceID(fmt.Sprintf("%s$%s", rid.StorageId, rid.SpaceId))
```
diocas: Isn't this code used with spaces disabled as well? Shouldn't you check that `SpaceId` is defined? Smth like:

```go
vPath = "v"
if rid.SpaceId != "" {
(...)
    if ok {
        vPath = spacePath
    }
}
```

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -2201,6 +2220,11 @@ func (fs *Eosfs) getEosMetadata(finfo *eosclient.FileInfo) []byte {
 	return v
 }
 
+func (fs *Eosfs) getVersionFolder(p string) string {
+	versionPrefix := ".sys.v#."
```
diocas: Why do we have this hardcoded here and have this method, that is also repeated in the eosbinary and eosgrpc files?

Let's move this (and the other that are repeated) to the eosclient utils?

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1203,6 +1201,27 @@ func (fs *Eosfs) GetMD(ctx context.Context, ref *provider.Reference, mdKeys []st
 	}
 
 	if ref.ResourceId != nil {
+		// Check if it's a version
+		if strings.Contains(ref.ResourceId.OpaqueId, "@") {
```
diocas, line 1208: `StorageId == "versions"`?

diocas (changes requested):

internal/http/services/owncloud/ocdav/propfind.go
rodcoffani: What changes is related to the request url.
When requesting for versions, the baseURI in this point is: 
`baseURI=/remote.php/dav/meta/<resource-id>`, with `relative_path=v/.sys.v#.<version-timestamp>`

When listing other resources, the values are: `baseURI=/remote.php/dav/spaces` and `relative_path=<file-name>`.

I'll put a similar explanation in the code!

internal/http/services/owncloud/ocdav/get.go
rodcoffani: It came together when cherry-picking the "ocdav versions" commit, I'll leave as it was before :)

internal/http/services/owncloud/ocdav/get.go
rodcoffani: Exactly, it shouldn't break because spaces is not enabled in the current version, thus the `SpaceId` is never set.

pkg/storage/utils/eosfs/eosfs.go
rodcoffani: Here, that is not possible. The real `StorageId` is needed in order to get the right storage provider.

diocas: This still seems a bit hardcoded, but I'll accept it. Was wondering if there would be a way to pass that info in the ctx or as a param.

pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -44,8 +44,7 @@ import (
 )
 
 const (
-	versionPrefix = ".sys.v#."
-	favoritesKey  = "http://owncloud.org/ns/favorite"
+	favoritesKey = "http://owncloud.org/ns/favorite"
```
diocas: Since we did a cleanup, can we move this as well? It's repeated in both clients.

PR #5197, opened 2025-06-11
pprof improvements

jessegeens:


internal/http/services/pprof/pprof.go
```diff
@@ -40,6 +41,8 @@ func New(ctx context.Context, m map[string]interface{}) (global.Service, error)
 
 	c.ApplyDefaults()
 
+	fmt.Println("pprof service created")
```
glpatcern: Leftover or intentional?

jessegeens: leftover :sweat_smile: will remove after meeting with Elvin

glpatcern (approved):

PR #5205, opened 2025-06-19
Stop gateway from crashing on unknown routes in the routing table

jessegeens:


glpatcern (approved):
Just for the record, this is a direct revert of https://github.com/cs3org/reva/pull/4370. And at the time I was very worried about such behavior, that eventually hit us enough.

PR #5210, opened 2025-06-25
Libregraph API: sharing permissions support

jessegeens:
Extension of the libregraph API to fix the following issues:
* Creating links / shares now gets a proper response
* Support for updating links / shares
* Support for deleting links / shares
* Removal of unsupported roles from `/roleDefinitions` endpoint

### DIFF
--- a/internal/http/services/owncloud/ocgraph/drive_permissions.go
+++ b/internal/http/services/owncloud/ocgraph/drive_permissions.go
@@ -211,7 +211,7 @@ func (s *svc) updateLinkPermissions(ctx context.Context, w http.ResponseWriter, 
 		return
 	}
 
-	update, err := getLinkUpdate(permission, &statRes.Info.Type)
+	update, err := s.getLinkUpdate(ctx, permission, statRes.Info.Type)
 	if err != nil {
 		log.Error().Err(err).Msg("nothing provided to update")
 		w.WriteHeader(http.StatusBadRequest)
@@ -273,42 +273,20 @@ func (s *svc) updateSharePermissions(ctx context.Context, w http.ResponseWriter,
 		return
 	}
 
-	perms, err := s.lgPermToCS3Perm(ctx, lgPerm, statRes.Info.Type)
+	update, err := s.getShareUpdate(ctx, lgPerm, statRes.Info.Type)
 	if err != nil {
-		log.Error().Err(err).Msg("nothing provided to update")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
-	var res *collaborationv1beta1.UpdateShareResponse
-	if lgPerm.ExpirationDateTime.IsSet() {
-		res, err = gw.UpdateShare(ctx, &collaborationv1beta1.UpdateShareRequest{
-			Ref: &collaborationv1beta1.ShareReference{
-				Spec: &collaborationv1beta1.ShareReference_Id{
-					Id: share.Id,
-				},
+
+	res, err := gw.UpdateShare(ctx, &collaborationv1beta1.UpdateShareRequest{
+		Ref: &collaborationv1beta1.ShareReference{
+			Spec: &collaborationv1beta1.ShareReference_Id{
+				Id: share.Id,
 			},
-			Field: &collaborationv1beta1.UpdateShareRequest_UpdateField{
-				Field: &collaborationv1beta1.UpdateShareRequest_UpdateField_Expiration{
-					Expiration: nullableTimeToCs3Timestamp(lgPerm.ExpirationDateTime),
-				},
-			},
-		})
-	} else {
-		res, err = gw.UpdateShare(ctx, &collaborationv1beta1.UpdateShareRequest{
-			Ref: &collaborationv1beta1.ShareReference{
-				Spec: &collaborationv1beta1.ShareReference_Id{
-					Id: share.Id,
-				},
-			},
-			Field: &collaborationv1beta1.UpdateShareRequest_UpdateField{
-				Field: &collaborationv1beta1.UpdateShareRequest_UpdateField_Permissions{
-					Permissions: &collaborationv1beta1.SharePermissions{
-						Permissions: perms,
-					},
-				},
-			},
-		})
-	}
+		},
+		Field: update,
+	})
 
 	if err != nil {
 		log.Error().Err(err).Msg("error updating public share")
@@ -587,7 +565,7 @@ func (s *svc) writePermissions(ctx context.Context, w http.ResponseWriter, actio
 	}
 }
 
-func getLinkUpdate(permission *libregraph.Permission, resourceType *providerpb.ResourceType) (*linkv1beta1.UpdatePublicShareRequest_Update, error) {
+func (s *svc) getLinkUpdate(ctx context.Context, permission *libregraph.Permission, resourceType providerpb.ResourceType) (*linkv1beta1.UpdatePublicShareRequest_Update, error) {
 	if permission.ExpirationDateTime.IsSet() {
 		return &linkv1beta1.UpdatePublicShareRequest_Update{
 			Type: linkv1beta1.UpdatePublicShareRequest_Update_TYPE_EXPIRATION,
@@ -604,7 +582,7 @@ func getLinkUpdate(permission *libregraph.Permission, resourceType *providerpb.R
 			DisplayName: *permission.Link.LibreGraphDisplayName,
 		}, nil
 	} else if permission.Link != nil && permission.Link.Type != nil {
-		permissions, err := CS3ResourcePermissionsFromSharingLink(permission.Link.GetType(), *resourceType)
+		permissions, err := CS3ResourcePermissionsFromSharingLink(permission.Link.GetType(), resourceType)
 		if err != nil {
 			return nil, errors.Wrap(err, "error converting link type to permissions")
 		}
@@ -619,4 +597,26 @@ func getLinkUpdate(permission *libregraph.Permission, resourceType *providerpb.R
 	} else {
 		return nil, errors.New("body contained nothing to update")
 	}
+}
+
+func (s *svc) getShareUpdate(ctx context.Context, permission *libregraph.Permission, resourceType providerpb.ResourceType) (*collaborationv1beta1.UpdateShareRequest_UpdateField, error) {
+
+	if permission.ExpirationDateTime.IsSet() {
+		return &collaborationv1beta1.UpdateShareRequest_UpdateField{
+			Field: &collaborationv1beta1.UpdateShareRequest_UpdateField_Expiration{
+				Expiration: nullableTimeToCs3Timestamp(permission.ExpirationDateTime),
+			},
+		}, nil
+	}
+	perms, err := s.lgPermToCS3Perm(ctx, permission, resourceType)
+	if err != nil || perms == nil {
+		return nil, errors.New("Failed to extract permissions")
+	}
+	return &collaborationv1beta1.UpdateShareRequest_UpdateField{
+		Field: &collaborationv1beta1.UpdateShareRequest_UpdateField_Permissions{
+			Permissions: &collaborationv1beta1.SharePermissions{
+				Permissions: perms,
+			},
+		},
+	}, nil
 }


internal/http/services/owncloud/ocgraph/groups.go
```diff
@@ -45,6 +46,9 @@ const (
 	propGroupMail        GroupSelectableProperty = "mail"
 	propGroupDescription GroupSelectableProperty = "description"
 	propGroupMembers     GroupSelectableProperty = "members"
+
+	// Max number of groups to return in a ListUsers query
+	maxGroupResponseLength int = 30
```
glpatcern, line 51: Why do we have that? And what if I have 500 groups?

internal/http/services/owncloud/ocgraph/ocgraph.go
```diff
@@ -111,10 +111,21 @@ func (s *svc) initRouter() {
 			r.Get("/sharedByMe", s.getSharedByMe)
 		})
 		r.Get("/roleManagement/permissions/roleDefinitions", s.getRoleDefinitions)
-		r.Get("/drives/{space-id}/root/permissions", s.getRootDrivePermissions)
-		r.Get("/drives/{space-id}/items/{resource-id}/permissions", s.getDrivePermissions)
-		r.Post("/drives/{space-id}/items/{resource-id}/invite", s.share)
-		r.Post("/drives/{space-id}/items/{resource-id}/createLink", s.createLink)
+		r.Route("/drives/{space-id}", func(r chi.Router) {
```
glpatcern, line 114: I wonder if we could somehow extract the API from the code!

internal/http/services/owncloud/ocgraph/users.go
```diff
@@ -46,6 +47,9 @@ const (
 	propUserMailDesc                     UserSelectableProperty = "mail desc"
 	propUserOnPremisesSamAccountName     UserSelectableProperty = "onPremisesSamAccountName"
 	propUserOnPremisesSamAccountNameDesc UserSelectableProperty = "onPremisesSamAccountName desc"
+
+	// Max number of users to return in a ListUsers query
+	maxUserResponseLength int = 30
```
glpatcern, line 52: Same as above. What is the scope of a `ListUsers` called by a user?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,620 @@
...
+
+	shareOrLink, err := s.getShareOrLink(ctx, shareID, resourceID)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	// Decode the requested permissions
+
+	// Unfortunately, we need to do something sketchy here: we can only read `r.Body` once
```
glpatcern: As discussed over chat, this seems bad design of the API and might be worth reporting to OpenCloud. E.g. ensuring that `expiration = -1` for a removal, and equivalently for other properties.

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -1128,11 +1128,16 @@ func quoteEtag(etag string) string {
...
-	if ok && contextUser.Id != nil && owner != nil &&
-		contextUser.Id.Idp == owner.Idp &&
+	if !ok || contextUser.Id == nil || owner == nil {
+		return false
+	}
+	if contextUser.Id.Idp == owner.Idp &&
 		contextUser.Id.OpaqueId == owner.OpaqueId {
 		return true
 	}
+	if owner.Idp == "" && contextUser.Id.OpaqueId == owner.OpaqueId {
```
diocas, line 1138: shouldn't a user be uniquely identified by the pair idp, id? If we allow empty idp, isn't this a risk? (I suppose is not a risk for our deployment, but this is being merged in upstream reva...)

changelog/unreleased/libregraph-permissins.md
```diff
@@ -0,0 +1,9 @@
+Enhancement: libregraph permission support
```
diocas: filename: "permissiOns.md"

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,620 @@
...
+			shareType: "link",
+			ID:        shareID,
+			link:      link.Share,
+		}, nil
+	}
+
+	return nil, err
+}
+
+func (s *svc) updateDrivePermissions(w http.ResponseWriter, r *http.Request) {
```
diocas, line 106: Calling these methods "XXXXPermissions" (the others below as well) is confusing, since they can update more things than the permissions, right? This is "DriveMetadata", maybe? Or Attributes?

diocas: And it's not for "Drives" right? "ShareOrLink"?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,620 @@
...
+
+	perms, err := s.lgPermToCS3Perm(ctx, lgPerm, statRes.Info.Type)
+	if err != nil {
+		log.Error().Err(err).Msg("nothing provided to update")
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	if strings.Contains(requestBody, "expiration") {
+		// TODO: implement support for updating expiration
```
diocas: when support added in the cs3apis, right? make the comment more explicit and please create a Jira task for this :) thank youu!

internal/http/services/owncloud/ocgraph/drive_permissions.go
diocas: Uhm, just realised they called the endpoint "permissions", so your naming convention makes sense with the (non)sense of the api definition.

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,620 @@
...
+		"value": perms,
+	}); err != nil {
+		log := appctx.GetLogger(ctx)
+		log.Error().Err(err).Msg("error marshalling permissions as json")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
+
+func getLinkUpdate(permission *libregraph.Permission, resourceType *providerpb.ResourceType, body string) (*linkv1beta1.UpdatePublicShareRequest_Update, error) {
```
diocas: There are missing update types. For example, changing a link name.

internal/http/services/owncloud/ocgraph/groups.go
diocas: I suppose we don't want to return a huge list to break the browser. I would expect users to ignore such a huge list and continue adding chars to narrow down the search.

internal/http/services/owncloud/ocgraph/unifiedrole.go
```diff
@@ -379,8 +381,10 @@ func GetLegacyName(role libregraph.UnifiedRoleDefinition) string {
 
 // CS3ResourcePermissionsToUnifiedRole tries to find the UnifiedRoleDefinition that matches the supplied
 // CS3 ResourcePermissions.
-func CS3ResourcePermissionsToUnifiedRole(p *provider.ResourcePermissions) *libregraph.UnifiedRoleDefinition {
+func CS3ResourcePermissionsToUnifiedRole(ctx context.Context, p *provider.ResourcePermissions) *libregraph.UnifiedRoleDefinition {
+	log := appctx.GetLogger(ctx)
 	role := conversions.RoleFromResourcePermissions(p)
+	log.Info().Interface("role", role).Interface("perms", p).Msg("Converting cs3 resource permissions to unified role")
```
diocas: Maybe this gets too verbose? Debug instead?

internal/http/services/owncloud/ocgraph/unifiedrole.go
```diff
@@ -429,8 +433,8 @@ func GetBuiltinRoleDefinitionList() []*libregraph.UnifiedRoleDefinition {
 	return []*libregraph.UnifiedRoleDefinition{
 		NewViewerUnifiedRole(),
 		NewEditorUnifiedRole(),
-		NewFileEditorUnifiedRole(),
-		NewManagerUnifiedRole(),
+		//NewFileEditorUnifiedRole(),
```
diocas, line 443: Put a comment explaining why we comment these out as opposed to deleting them.

internal/http/services/owncloud/ocgraph/drive_permissions.go
diocas: Using `-1` seems incorrect to me. It would not be possible to use it for any value we want to unset. Maybe something like https://jsonpatch.com ?
Or, maybe simpler, https://github.com/oapi-codegen/nullable/blob/main/nullable.go ?

internal/http/services/owncloud/ocgraph/groups.go
jessegeens: Exactly; in this case it's for a already-filtered query. If we don't limit this you may get a huge response which slows down the browser

internal/http/services/owncloud/ocdav/propfind.go
jessegeens: The problem is that we don't store the idp in the shares db (only the user id), so there is no way to "compare" this here

internal/http/services/owncloud/ocgraph/unifiedrole.go
jessegeens: Yes, good point

internal/http/services/owncloud/ocgraph/drive_permissions.go
jessegeens: The issue is that it's already nullable though, but there is no way to discern between "explicit null" and "implicit null" (because the value was missing)

jessegeens: Though maybe they always pass the expiration when it is set? And then it would suffice to only pass the update if it is different from the currently set expiration? This I could check

internal/http/services/owncloud/ocdav/propfind.go
diocas: I understand that our deployment asks for this, the question is if this makes sense to be merged in the general Reva*. If you don't think this will be an issue, it's fine, I just want to be sure.

* Knowing that we can always change how we do things at CERN. In practice, it's our implementation that ignores the fact that a user is uniquely identified by such a pair and, hence, is "wrong".

internal/http/services/owncloud/ocgraph/ocgraph.go
jessegeens: What do you mean?

internal/http/services/owncloud/ocdav/propfind.go
jessegeens: One thing we could do is put the `isShareOwner(ctx context.Context, owner *userv1beta1.UserId) bool` in the plugins and use this instead? Though we did want to merge these back into upstream reva, but in any case this is a direct consequence of the db schema so I think it's OK (and in the future we could also update the schema to include this)

diocas: What I meant was precisely that the db schema can be extended to include the idp :) But fine, let's merge as is.

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,622 @@
...
+			ResourceId: resourceId,
+		},
+	})
+	if err != nil || statRes.Status.Code != rpcv1beta1.Code_CODE_OK {
+		log.Error().Err(err).Msg("Failed to stat resource")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	perms, err := s.lgPermToCS3Perm(ctx, lgPerm, statRes.Info.Type)
```
diocas: Wouldn't a request to update the expiration return err here, hence the if below would not be reached?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -0,0 +1,622 @@
...
+					Id: share.Id,
+				},
+			},
+			Field: &collaborationv1beta1.UpdateShareRequest_UpdateField{
+				Field: &collaborationv1beta1.UpdateShareRequest_UpdateField_Expiration{
+					Expiration: nullableTimeToCs3Timestamp(lgPerm.ExpirationDateTime),
+				},
+			},
+		})
+	} else {
```
diocas: I have the impression this should be an if-else and not just an else.. "else" means the request sent contains nothing to update, and maybe that would be a separate case (where we bail).

diocas: Maybe similar to what we have for the pl's, where we get what needs to be updated in `getLinkUpdate`.

internal/http/services/owncloud/ocgraph/drive_permissions.go
jessegeens: No, in this case it just returns `nil, nil`. But you have a good point that we only have to call this in the case that we want to update the permissions themselves, so I can move this block into the `else` block.

internal/http/services/owncloud/ocgraph/drive_permissions.go
jessegeens: Can do that yes. And then throw an error when there is nothing found to update?

jessegeens: Implemented `getShareUpdate` in the latest commit!

PR #5211, opened 2025-06-26
feat: add error codes to DAV responses

rodcoffani:
- adapt code from edge branch (#4749)


diocas:
@rodcoffani can you add an explanation in this PR as to why this is needed? (some example where is used suffices)

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -52,6 +52,14 @@ type DavHandler struct {
 	OCMSharesHandler    *WebDavHandler
 }
 
+const (
```
diocas, line 58: Some of these errors are not being used... Wether we add them somewhere where they make sense (the best) or we remove them (not so "best").

diocas (changes requested):

internal/http/services/owncloud/ocdav/dav.go
rodcoffani: The way they're used in the `edge` branch and on the reva forks happens to also be like this, I tried to minimize the differences.
Also, the same is declared in the web repo:
https://github.com/cernbox/web/blob/dev/packages/web-client/src/webdav/constants/dav.ts#L17-L22

what would be the best approach?

diocas: let's remove them. if we need them in the future then we add them.

PR #5212, opened 2025-06-27
Fix creating home directories

jessegeens:


jessegeens:
This was fixed in https://github.com/cs3org/reva/pull/5213

PR #5215, opened 2025-06-30
feat: resource private link

rodcoffani:
- add private link property to propfind response
- not configurable to choose between id or path based links (path only)

in order to `s.c.PublicURL` have a value, a config in puppet is needed.


diocas (dismissed):

diocas:
Are we changing the way we config this, then, @rodcoffani @jessegeens ? Or ready to go?

rodcoffani:
@diocas related to the config, I think it's okay to leave like this, the same value already exists in other places in the same way.
But in general, I would check with @jessegeens to see if his commit should be part of this PR or dropped.

PR #5216, opened 2025-07-02
Spaces: make downloading and restoring revisions work

jessegeens:
* Some extra logging
* Fixed a bug in `IsVersionFolder`
* Fixed a bug in handling `GET` requests on the `VersionsHandler`

### DIFF
--- a/changelog/unreleased/spaces-revisions.md
+++ b/changelog/unreleased/spaces-revisions.md
@@ -1,0 +1,7 @@
+Bugfix: download / restore versions in spaces
+
+*  Some extra logging
+*  Fixed a bug in IsVersionFolder
+*  Fixed a bug in handling GET requests on the VersionsHandler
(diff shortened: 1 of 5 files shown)


internal/http/services/owncloud/ocdav/get.go
```diff
@@ -73,6 +73,8 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	log.Info().Any("ref", ref).Any("statRes", sRes.Info).Msg("handleGet")
```
diocas: Debug?

jessegeens: Whoops, this one I missed indeed

jessegeens:
Should only be debug logs left now

PR #5222, opened 2025-07-04
feat: inode to path reverse resolution in the cephfs driver

rawe0:
A proof of concept for the inode to path reverse resolution in the cephfs driver.

The main function is `GetPathByID` which does the following:
1. Query the Ceph monitor and parse the active mds (`getFSStatus()` & `parseActiveMDS()`)
2. Convert the `opaque_id` to an `integer` (to be seen if needed in the future, for testing it was required)
3. Query the Ceph MDS (Metadata Server) to "dump inode" to get the path (`dumpInode()` & `extractPathFromInodeOutput()`)
4. Trim the Reva mount path from the full path.

The path we get from the mds is the full path on the ceph cluster e.g. `/volumes/_nogroup/rasmus/welander`,  however Reva mounts at the root `/volumes/_nogroup/rasmus`, so the corresponding cephfs driver path is `/rasmus/welander`. 

**_Note:_** _this proof of concept uses a powerful admin key to the ceph cluster, which will need to be generated for the production ceph cluster with the required permissions, and has only been tested on a test ceph cluster._ 

**Line 215 -  224 should be removed since it's only there for testing purposes.**

### DIFF
--- a/pkg/storage/fs/cephfs/cephfs.go
+++ b/pkg/storage/fs/cephfs/cephfs.go
@@ -306,8 +306,136 @@ func (fs *cephfs) RestoreRevision(ctx context.Context, ref *provider.Reference, 
 	return errtypes.NotSupported("cephfs:  RestoreRevision not supported")
 }
 
+func (fs *cephfs) getFSStatus() ([]byte, error) {
+
+	cmdObj := map[string]interface{}{
+		"prefix": "fs status",
+		"format": "json",
+	}
+
+	jsonCmd, err := json.Marshal(cmdObj)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error marshalling fs status command")
+	}
+
+	command := [][]byte{jsonCmd}
+	output, status, err := fs.adminConn.radosConn.MgrCommand(command)
+	if err != nil {
+		if strings.Contains(status, "does your client key have mgr caps?") {
+			return nil, errtypes.NotSupported("cephfs: permission denied when executing 'MgrCommand', " +
+				"check if the admin key has permission to execute the command")
+		}
+		return nil, errors.Wrap(err, "Error executing 'MgrCommand'")
+	}
+
+	return output, nil
+}
+
+func (fs *cephfs) parseActiveMDS(output []byte) (string, error) {
+
+	var mgrJson struct {
+		MDSMap []struct {
+			Name  string `json:"name"`
+			State string `json:"state"`
+		} `json:"mdsmap"`
+	}
+
+	err := json.Unmarshal(output, &mgrJson)
+	if err != nil {
+		return "", errors.Wrap(err, "Error unmarshalling MDS map")
+	}
+
+	// Filter for state == "active"
+	var active []string
+	for _, m := range mgrJson.MDSMap {
+		if m.State == "active" {
+			active = append(active, m.Name)
+		}
+	}
+
+	if len(active) == 0 {
+		return "", errors.Wrap(err, "No active MDS found")
+	}
+
+	return active[0], nil
+}
+
+func (fs *cephfs) dumpInode(mdsSpec string, inode int) ([]byte, error) {
+
+	cmdObj := map[string]interface{}{
+		"prefix": "dump inode",
+		"number": inode,
+	}
+
+	jsonCmd, err := json.Marshal(cmdObj)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error marshalling dump inode command")
+	}
+
+	command := [][]byte{jsonCmd}
+	output, status, err := fs.adminConn.adminMount.MdsCommand(mdsSpec, command)
+	if err != nil {
+		if strings.Contains(status, "does your client key have mgr caps?") {
+			return nil, errtypes.NotSupported("cephfs: permission denied when executing 'MdsCommand', " +
+				"check if the admin key has permission to execute the command")
+		}
+		return nil, errors.Wrap(err, "Error executing 'MdsCommand'")
+	}
+
+	return output, nil
+}
+
+func (fs *cephfs) extractPathFromInodeOutput(ctx context.Context, output []byte) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/storage/fs/cephfs/cephfs.go
```diff
@@ -306,8 +307,134 @@ func (fs *cephfs) RestoreRevision(ctx context.Context, ref *provider.Reference,
...
+	}
+
+	// Extract path from inode output
+	path, err := fs.extractPathFromInodeOutput(inodeOutput)
+	if err != nil {
+		return "", err
+	}
+
+	// Extract the path without the root
+	path = strings.TrimPrefix(path, fs.conf.Root+"/")
```
glpatcern: I think we need to keep the leading `/` as we want to return an absolute path (yet relative to the mountpoint)

pkg/storage/fs/cephfs/user.go
```diff
@@ -200,12 +199,30 @@ func (user *User) fileAsResourceInfo(cv *cacheVal, path string, stat *goceph.Cep
...
-		return "", errtypes.NotSupported("cephfs: path not provided, id based refs are not supported")
+		path, err := user.fs.GetPathByID(user.ctx, ref.ResourceId)
+		if err != nil {
+			log.Debug().Any("OpaqueId", ref.ResourceId.OpaqueId).Err(err).Msg("No matching path found for inode")
+			return "", err
+		}
+		return path, nil
+		// For testing purposes, remove later
+		// Test via curl -u <user> -O https://oc.cernbox.cern.ch/webdav/cephfs/1099828998403
+	} else if ref.GetPath() == "/1099828998403" {
```
glpatcern: I guess time to remove this? :)

pkg/storage/fs/cephfs/cephfs.go
```diff
@@ -306,8 +307,134 @@ func (fs *cephfs) RestoreRevision(ctx context.Context, ref *provider.Reference,
...
+	err := json.Unmarshal(output, &mdsJson)
+	if err != nil {
+		log.Debug().Err(err).Msg("Error unmarshalling MDS inode output")
+		return "", err
+	}
+
+	path, ok := mdsJson["path"].(string)
+	if !ok {
+		log.Debug().Any("mds_output", mdsJson).Msg("Path not found in MDS output")
+		return "", fmt.Errorf("path not found in MDS output")
```
glpatcern: Strings without `%` formatters don't need `fmt`, you can simply do `errors.New("error string")

pkg/storage/fs/cephfs/cephfs.go
```diff
@@ -45,6 +45,7 @@ import (
 	"github.com/cs3org/reva/pkg/utils"
 	"github.com/cs3org/reva/pkg/utils/cfg"
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
```
glpatcern: This might have been included by VSCode, a priori it's not needed

pkg/storage/fs/cephfs/cephfs.go
```diff
@@ -306,8 +307,134 @@ func (fs *cephfs) RestoreRevision(ctx context.Context, ref *provider.Reference,
...
 
+func (fs *cephfs) getFSStatus() ([]byte, error) {
+	cmdObj := map[string]interface{}{
+		"prefix": "fs status",
+		"format": "json",
+	}
+
+	jsonCmd, err := json.Marshal(cmdObj)
+	if err != nil {
+		log.Debug().Err(err).Msg("Error marshalling fs status command")
```
glpatcern: Instead, to be able to log anywhere, one has to do (Go is a bit cumbersome I should say):
* pass `ctx Context.context` as first argument
* `log := appctx.GetLogger(ctx)`

Or alternatively, even better for such helper functions, you could return the `error` value as `errors.Wrap(err, "your message") without logging, and then in the caller you log the `err`.

pkg/storage/fs/cephfs/cephfs.go
glpatcern: Now I also see your comment - I guess it kept being added by VS Code because of the call in the comment below

pkg/storage/fs/cephfs/cephfs.go
rawe0: This is now fixed, I tested it because I seemed to remember that it did not work without this, however it does:

*config*
[http.services.dataprovider.drivers.cephfs]
root = "/volumes/_nogroup/rasmus"

***with*** the `+'/'`

```
2025-07-17 16:06:53.508 DBG root/reva/pkg/storage/fs/cephfs/cephfs.go:432 > GetPathByID fs.conf.root=/volumes/_nogroup/rasmus path=/volumes/_nogroup/rasmus/welander pid=826638 traceid=c3973b5b-6e97-4307-9356-f6305dcbaa05

2025-07-17 16:06:53.508 DBG root/reva/pkg/storage/fs/cephfs/cephfs.go:434 > GetPathByID resolved path path=welander pid=826638 traceid=c3973b5b-6e97-4307-9356-f6305dcbaa05
```

***without*** `+'/'`
```
2025-07-17 16:10:42.942 DBG root/reva/pkg/storage/fs/cephfs/cephfs.go:432 > GetPathByID fs.conf.root=/volumes/_nogroup/rasmus path=/volumes/_nogroup/rasmus/welander pid=827740 traceid=6804c16c-c898-4bbf-9027-1ae4dc37e2e2
2025-07-17 16:10:42.942 DBG root/reva/pkg/storage/fs/cephfs/cephfs.go:434 > GetPathByID resolved path path=/welander pid=827740 traceid=6804c16c-c898-4bbf-9027-1ae4dc37e2e2
```

glpatcern (approved):

glpatcern:
As discussed, let's just include some logic to return `Not Supported` when the Ceph key is not "powerful enough" to make the command succeed. Then, this can be merged 👍

PR #5225, opened 2025-07-11
feat: add drives create permission

rodcoffani:


internal/http/services/owncloud/ocapi/permissions.json
```diff
@@ -9,6 +9,7 @@
         "Language.ReadWrite.all",
         "Favorites.List.own",
         "Drives.ReadWrite.all",
-        "Shares.Write.all"
+        "Shares.Write.all",
+        "Drives.Create.all"
```
diocas, line 13: The problem with this is that it is going to show to people with lightweight accounts as well, right?

diocas: I see that you check in web if the user also has a personal space. I would say our json should be dynamic instead of static.. It's fine to merge like this _now_, but please open a Jira ticket to improve this.

rodcoffani: I agree!

PR #5232, opened 2025-07-15
fix: valid download url in public links

rodcoffani:
Downloading files doesn't work with
`GET https://cbox-ocisdev-rodrigo.cern.ch/remote.php/dav/public-files/public/<public-link-token>/<filename>`
instead, it should be:
`GET https://cbox-ocisdev-rodrigo.cern.ch/remote.php/dav/public-files/<public-link-token>/<filename>`

The other properties, as `href` and `privatelink` for example, do work with the resource path starting with `/public`, so there seems to be no reason to remove this from the path itself. The exception is only for `downloadURL`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,16 +53,16 @@ gaia:
 
 .PHONY: cernbox-revad
 cernbox-revad: gaia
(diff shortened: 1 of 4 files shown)


diocas (approved):
Indeed this is the behaviour of the current production. Other option would be to change our endpoint, but would break backwards compatibility (or we maintained both).

I'm approving!

PR #5235, opened 2025-07-23
Allow POST to /shares endpoint to send multiple users at once

jessegeens:
Sending multiple POST requests for multiple users leads to parallel calls to EOS, which suffers from a critical race condition when setting ACLs. So, now the reva OCS endpoint supports sending multiple comma-seperated users.

### DIFF
--- a/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
+++ b/internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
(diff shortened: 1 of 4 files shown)


internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
```diff
@@ -52,51 +53,55 @@ func (h *Handler) createUserShare(w http.ResponseWriter, r *http.Request, statIn
...
+			response.WriteOCSError(w, r, response.MetaServerError.StatusCode, "error searching recipient", err)
+			return
+		}
 
-	if userRes.Status.Code != rpc.Code_CODE_OK {
-		response.WriteOCSError(w, r, response.MetaNotFound.StatusCode, "user not found", err)
-		return
-	}
+		if userRes.Status.Code != rpc.Code_CODE_OK {
+			response.WriteOCSError(w, r, response.MetaNotFound.StatusCode, "user not found", err)
```
glpatcern, line 71: What if a user is not resolvable but others are? Do we fail the entire request?

glpatcern (dismissed):
Given that this is a stop-gap solution it can go as is, regardless the answer to my question below. It's more important to define the behavior in the endpoint we use with spaces!

internal/http/services/owncloud/ocs/handlers/apps/sharing/shares/user.go
jessegeens: I would say so yes. In theory every user should be resolvable, otherwise you wouldn't have found him in the search box.

glpatcern: Yeah the normal UI usage is protected by the search box, but "you never know" (anyway the GRPC API does not use this code). I'd advocate (in the ocgraph implementation) for skipping the problematic user and do the rest + report the error, otherwise the behavior depends on the order of requests.

But OK this PR is good to go, will merge it

glpatcern (approved):

PR #5252, opened 2025-08-05
Don't set a resource id on the virtual folder root of a public-link-shared file

jessegeens:
The front-end expects to not have a file ID on the root of a public link when it is a virtual folder around a single file share, for it to automatically open in the default app. The file id of this virtual folder has now been removed.

Additionally, this also fixes the `OC-Checksum: Invalid:` header on downloads of public link shared files

### DIFF
--- a/internal/http/services/owncloud/ocdav/get.go
+++ b/internal/http/services/owncloud/ocdav/get.go
@@ -137,7 +137,7 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	} else {
 		w.Header().Set(HeaderContentLength, strconv.FormatUint(info.Size, 10))
 	}
-	if info.Checksum != nil {
+	if info.Checksum != nil && info.Checksum.Type != provider.ResourceChecksumType_RESOURCE_CHECKSUM_TYPE_INVALID && info.Checksum.Type != provider.ResourceChecksumType_RESOURCE_CHECKSUM_TYPE_UNSET {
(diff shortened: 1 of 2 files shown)


glpatcern (approved):

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -775,7 +775,7 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
 				// I tested the desktop client and phoenix to annotate which properties are requestted, see below cases
 				case "fileid": // phoenix only
 					if md.Id == nil {
-						propstatNotFound.Prop = append(propstatNotFound.Prop, s.newProp("oc:fileid", ""))
+						continue
```
diocas: Do we really need to ignore it completely, even not putting it in the not found section?

internal/http/services/owncloud/ocdav/get.go
```diff
@@ -137,7 +137,7 @@ func (s *svc) handleGet(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	} else {
 		w.Header().Set(HeaderContentLength, strconv.FormatUint(info.Size, 10))
 	}
-	if info.Checksum != nil {
+	if info.Checksum != nil && info.Checksum.Type != provider.ResourceChecksumType_RESOURCE_CHECKSUM_TYPE_INVALID && info.Checksum.Type != provider.ResourceChecksumType_RESOURCE_CHECKSUM_TYPE_UNSET {
```
diocas, line 140: This fix is not documented in the PR nor in the changelog. Shouls also be a separate commit.

PR #5260, opened 2025-08-11
Add support for updating readme, thumbnail in spaces

jessegeens:
This PR adds support for:
* displaying description (called readme in the backend), subtitle (called description in the backend) and thumbnail of a space.
*  updating the thumbnail and readme (support for the description has also been added to the CS3API, but not used at the moment as we don't support users updating it)

This required:
* An update of the libregraph API
* An update of the CS3APIs (https://github.com/cs3org/cs3apis/pull/245, https://github.com/cs3org/cs3apis/pull/246)
* Hooking up the project catalogue to the spaces registry

Additionally, the projects catalogue now directly implements the methods of the spaces registry.

### DIFF
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -194,21 +194,27 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				r = r.WithContext(ctx)
 				h.TrashbinHandler.Handler(s).ServeHTTP(w, r)
 			default:
-				// path is of type: space_id/relative/path/from/space
-				// the space_id is the base64 encode of the path where
-				// the space is located
+				// there are two possible types of path
+				// 1) path is of type: space_id/relative/path/from/space
+				//    the space_id is the base64 encode of the path where
+				//    the space is located
+				// 2) path is of type: resource_id/relative/path
+				//    here, resource_id is the encoded resource id of a folder
+				//    i.e. in the form of storage$space_id!inode
+				//    and the relative path is relative to this folder
 
 				_, base, ok := spaces.DecodeStorageSpaceID(head)
 				if ok {
-					fullPath := filepath.Join(base, r.URL.Path)
-					r.URL.Path = fullPath
-					ctx = context.WithValue(ctx, ctxSpaceFullPath, fullPath)
-					ctx = context.WithValue(ctx, ctxSpaceRelativePath, r.URL.Path)
+					// this is case (1)
 					ctx = context.WithValue(ctx, ctxSpaceID, head)
+					fullPath := filepath.Join(base, r.URL.Path)
+					// like this, we can use the existing DAV handler
+					// we replace the space id with it's actual path in the URL
+					r.URL.Path = fullPath
 
 					// We support doing a PUT and a PROPFIND on a resource ID (eg eos$space!inode/file.txt)
 				} else if r.Method == http.MethodPut || r.Method == MethodPropfind {
-					// If it's not a space ID, we try to parse it as a resource ID
+					// If it's not a space ID, we try to parse it as a resource ID, i.e. case (2)
 					var storageId, itemId string
 					storageId, base, itemId, ok = spaces.DecodeResourceID(head)
 					if !ok {
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -199,18 +199,30 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				// the space is located
 
 				_, base, ok := spaces.DecodeStorageSpaceID(head)
-				if !ok {
-					w.WriteHeader(http.StatusBadRequest)
-					return
+				if ok {
+					fullPath := filepath.Join(base, r.URL.Path)
+					r.URL.Path = fullPath
+					ctx = context.WithValue(ctx, ctxSpaceFullPath, fullPath)
+					ctx = context.WithValue(ctx, ctxSpaceRelativePath, r.URL.Path)
+					ctx = context.WithValue(ctx, ctxSpaceID, head)
+
+					// We support doing a PUT and a PROPFIND on a resource ID (eg eos$space!inode/file.txt)
+				} else if r.Method == http.MethodPut || r.Method == MethodPropfind {
+					// If it's not a space ID, we try to parse it as a resource ID
+					var storageId, itemId string
+					storageId, base, itemId, ok = spaces.DecodeResourceID(head)
+					if !ok {
+						w.WriteHeader(http.StatusBadRequest)
+						return
+					}
+
+					spaceId := spaces.EncodeSpaceID(base)
+					ctx = context.WithValue(ctx, ctxSpaceID, spaceId)
+					ctx = context.WithValue(ctx, ctxStorageId, storageId)
+					ctx = context.WithValue(ctx, ctxResourceOpaqueId, itemId)
 				}
 
-				fullPath := filepath.Join(base, r.URL.Path)
-				r.URL.Path = fullPath
-
-				ctx = context.WithValue(ctx, ctxSpaceID, head)
-				ctx = context.WithValue(ctx, ctxSpaceFullPath, fullPath)
 				ctx = context.WithValue(ctx, ctxSpacePath, base)
-				ctx = context.WithValue(ctx, ctxSpaceRelativePath, r.URL.Path)
 				r = r.WithContext(ctx)
 				h.SpacesHandler.Handler(s).ServeHTTP(w, r)
 			}
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -72,30 +72,24 @@ const (
 // ns is the namespace that is prefixed to the path in the cs3 namespace.
 func (s *svc) handlePathPropfind(w http.ResponseWriter, r *http.Request, ns string) {
 	ctx := r.Context()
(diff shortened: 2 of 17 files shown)


rodcoffani:
Related fix in front-end: github.com/cernbox/web/pull/206

internal/http/services/owncloud/ocdav/put.go
diocas, line 1: Unecessary

internal/http/services/owncloud/ocgraph/users.go
```diff
@@ -64,9 +64,9 @@ func (s UserSelectableProperty) Valid() bool {
 func (s *svc) getMe(w http.ResponseWriter, r *http.Request) {
 	user := appctx.ContextMustGetUser(r.Context())
 	me := &libregraph.User{
```
diocas, line 66: Are these changes relevant to this PR?

pkg/projects/manager/sql/sql.go
```diff
@@ -71,9 +73,19 @@ type Project struct {
...
+	Readers string
+	// Writers e-group ID
+	Writers string
+	// Admins e-group ID
+	Admins string
+	// Called description in libregraph API
+	// Called subtitle in front-end
+	Description string
+	// Inode of .space/readme.md
+	ReadmeInode string `gorm:"size:32"`
```
diocas: To be discussed tomorrow. For now I would keep only the relevant changes to this PR and leave the others to a separate PR adding all agreed projects/spaces fields.

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -199,18 +199,30 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
 
 				_, base, ok := spaces.DecodeStorageSpaceID(head)
-				if !ok {
-					w.WriteHeader(http.StatusBadRequest)
-					return
+				if ok {
+					fullPath := filepath.Join(base, r.URL.Path)
+					r.URL.Path = fullPath
+					ctx = context.WithValue(ctx, ctxSpaceFullPath, fullPath)
+					ctx = context.WithValue(ctx, ctxSpaceRelativePath, r.URL.Path)
```
diocas: I'm confused about this. Are full path and relative the same? If so, I would use fullPath as the variable in both context assignments to make it clear.

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -199,18 +199,30 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
 				// the space is located
 
 				_, base, ok := spaces.DecodeStorageSpaceID(head)
-				if !ok {
-					w.WriteHeader(http.StatusBadRequest)
-					return
+				if ok {
```
diocas, line 207: Add a comment explaining the 2 different types of head that might appear here.

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -199,18 +199,30 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
+					fullPath := filepath.Join(base, r.URL.Path)
+					r.URL.Path = fullPath
+					ctx = context.WithValue(ctx, ctxSpaceFullPath, fullPath)
+					ctx = context.WithValue(ctx, ctxSpaceRelativePath, r.URL.Path)
+					ctx = context.WithValue(ctx, ctxSpaceID, head)
+
+					// We support doing a PUT and a PROPFIND on a resource ID (eg eos$space!inode/file.txt)
```
diocas, line 215: Why only those 2?

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -83,7 +85,18 @@ func (s *svc) handlePathPropfind(w http.ResponseWriter, r *http.Request, ns stri
 		return
 	}
 
-	ref := &provider.Reference{Path: fn}
+	// We check if the PROPFIND was made to a resource ID instead of a path
+	if opaqueId := ctx.Value(ctxResourceOpaqueId); opaqueId != nil {
```
diocas: Can we make this an aux function and reuse it in both cases? (or potentially more)?

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -111,13 +111,29 @@ func isContentRange(r *http.Request) bool {
...
+	if opaqueId := ctx.Value(ctxResourceOpaqueId); opaqueId != nil {
+		storageId := ctx.Value(ctxStorageId)
+		if storageId != nil {
+			// We make the path relative
+			ref.Path = path.Join(".", fn)
+			ref.ResourceId = &provider.ResourceId{
+				StorageId: storageId.(string),
+				OpaqueId:  opaqueId.(string),
+			}
+			s.handlePut(ctx, w, r, ref, sublog)
```
diocas: Why?

internal/http/services/owncloud/ocgraph/users.go
jessegeens: Yes. I needed to update the libregraph api package for the Update struct, and they changed the type of these fields...

pkg/projects/manager/sql/sql.go
```diff
@@ -135,30 +157,121 @@ func (m *mgr) ListProjects(ctx context.Context, user *userpb.User) ([]*provider.
...
-func projectBelongToUser(user *userpb.User, p *Project) (*provider.ResourcePermissions, bool) {
+func (m *mgr) GetStorageSpaces(ctx context.Context, name string) (*provider.StorageSpace, error) {
+	var fetchedProjects []*Project
+
+	user, ok := appctx.ContextGetUser(ctx)
+	if !ok {
+		return nil, errors.New("must provide a user for fetching storage spaces")
+	}
+
+	if res, err := m.cache.Get(cacheKey); err == nil && res != nil {
```
diocas: Need some explanation here to understand what is happening...

diocas (changes requested):

internal/http/services/owncloud/ocdav/dav.go
jessegeens: ~~No, `fullPath := filepath.Join(base, r.URL.Path)` is the key here ;)~~

internal/http/services/owncloud/ocdav/dav.go
jessegeens: Because those are the ones needed for the thumbnail to work. In theory it might work with others too, but then we would have to test that it actually works with those methods imo

pkg/projects/manager/sql/sql.go
jessegeens: Nothing good was happening there, good catch :)

internal/http/services/owncloud/ocdav/dav.go
diocas: Yes, I know, which we could do and have it consistent for all :) It's fine to leave it to a separate PR, but I would create a ticket to keep track.

internal/http/services/owncloud/ocdav/dav.go
jessegeens: (removed the unused entries)

internal/http/services/owncloud/ocdav/dav.go
jessegeens: OK!

jessegeens: Added https://its.cern.ch/jira/browse/CERNBOX-4013

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -83,7 +85,10 @@ func (s *svc) handlePathPropfind(w http.ResponseWriter, r *http.Request, ns stri
 		return
 	}
 
-	ref := &provider.Reference{Path: fn}
+	// We check if the PROPFIND was made to a resource ID instead of a path
```
diocas: Why isn't this done before the readpropfind, together with the rest of the ref checks?

internal/http/services/owncloud/ocdav/put.go
```diff
@@ -111,13 +111,20 @@ func isContentRange(r *http.Request) bool {
...
-	s.handlePut(ctx, w, r, ref, sublog)
+	// We check if the PROPFIND was made to a resource ID instead of a path
+	if ref, ok := requestWasMadeToResourceId(ctx, fn); ok {
+		s.handlePut(ctx, w, r, ref, sublog)
+	} else {
+		// Request was made to an actual path
+		ref := &provider.Reference{
+			Path: fn,
+		}
+		s.handlePut(ctx, w, r, ref, sublog)
```
diocas: Why is this code different then the propfind if they have the same logic? Just put the if..else to get the ref (in both cases), and then call a single handleput.

pkg/projects/manager/sql/sql.go
```diff
@@ -135,30 +157,116 @@ func (m *mgr) ListProjects(ctx context.Context, user *userpb.User) ([]*provider.
...
+		return nil, errors.New("must provide a user for fetching storage spaces")
+	}
+
+	query := m.db.Model(&Project{}).Where("name = ?", name)
+	res := query.Find(&fetchedProjects)
+	if res.Error != nil {
+		return nil, res.Error
+	}
+
+	projects := []*provider.StorageSpace{}
```
diocas: Don't understand why we use an array if we only need one. 

`name` is unique, query returned smth and it belongs to the user, therefore we can return immediately, or not, and we return an error.

internal/http/services/owncloud/ocdav/propfind.go
jessegeens: I've moved it up

PR #5273, opened 2025-08-19
cephmount: new hybrid driver to connect to CephFS clusters with go 1.25

labkode:
Read the accompanying README.md for a comprehensive summary of this driver.
This change requires go1.25 to benefit from the os.Root methods.

- [x] use chroot for all local filesytem operations
- [x] use librados and libcephfs for obtaining path from device inode through MDS
- [x] create gorouting pool and map users to dedicated physical threads to set effective user id on the mount
- [x] adapt test/CI to go 1.25
- [x] validate in integration scenario with upstream services (ocdav, etc ...): :warning: refer to https://github.com/cs3org/reva/issues/5300

This PR introduces also some collateral changes:
- Bumps go to version 1.25 (needed to benefit from os.Root guarantees on most of the os ops)
- Replaces `golangci-lint` with standards `go vet` (there wasn't much added value on running it)
- Fixes other tests to avoid using `reflect.DeepEqual` on protobuf's structs.


glpatcern (approved):
Let's go ahead with this important change so that we can start field-testing it and adjusting it as necessary. A full code review remains due, during the field tests.

Issue #5269, opened 2025-08-18
Use stime in EOSgrpc driver


labkode:
The EOS binary driver exposes sync time `stime` rather than mtime for folders. We should bring back that behaviour to the grpc driver.

eos binary: https://github.com/cs3org/reva/blob/master/pkg/eosclient/eosbinary/eosbinary.go#L1213

PR #5274, opened 2025-08-19
Use stime instead of mtime for directories

jessegeens:
This fixes https://github.com/cs3org/reva/issues/5269

### DIFF
--- a/pkg/eosclient/eosgrpc/file_info.go
+++ b/pkg/eosclient/eosgrpc/file_info.go
@@ -178,6 +178,10 @@ func (c *Client) grpcMDResponseToFileInfo(ctx context.Context, st *erpc.MDRespon
 		fi.UID = st.Cmd.Uid
 		fi.GID = st.Cmd.Gid
 		fi.MTimeSec = st.Cmd.Mtime.Sec
+		// For directories, we prefer stime over mtime
+		if st.Cmd.Stime != nil {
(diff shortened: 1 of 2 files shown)


pkg/eosclient/eosgrpc/file_info.go
```diff
@@ -178,6 +178,10 @@ func (c *Client) grpcMDResponseToFileInfo(ctx context.Context, st *erpc.MDRespon
 		fi.UID = st.Cmd.Uid
 		fi.GID = st.Cmd.Gid
 		fi.MTimeSec = st.Cmd.Mtime.Sec
+		// For directories, we prefer stime over mtime
+		if st.Cmd.Stime != nil {
```
glpatcern, line 182: That means that `stime` is nil for files?

labkode: stime and mtime for files should be the same.

diocas: If they are the same for files (which also means both always exist), then we don't need the if. I understand this is not the case, @jessegeens ?

jessegeens: For files, `stime` is not even exposed over gRPC (but it does exist, although it is not necessarily set). But in principe at least `stime` should never be newer than `mtime`

PR #5283, opened 2025-08-29
add support for public spaces to spacesregistry

jessegeens:


changelog/unreleased/public-spaces.md
```diff
@@ -0,0 +1,5 @@
+Enhancement: public spaces
+
+We now support exposing other EOS areas as "public" spaces (with spaceType `public`)
```
diocas: "explorer"

internal/grpc/services/spacesregistry/spacesregistry.go
```diff
@@ -315,6 +339,51 @@ func (s *service) userSpace(ctx context.Context, user *userpb.User) (*provider.S
...
+	log := appctx.GetLogger(ctx)
+	publicSpaces := make([]*provider.StorageSpace, 0)
+	for spaceName, content := range s.c.PublicSpaces {
+		path, ok := content["path"]
+		if !ok {
+			log.Error().Msgf("No `path` found for public space %s, ignoring this space", spaceName)
+			continue
+		}
+
+		statRes, err := s.gw.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{
```
diocas: The web asks for all the spaces when it starts. This means that an issue with a public space would delay access to the UI. Do we really need to stat or we can assume that the config is always correct? Any access to the spaces will then check if they exist or not.

glpatcern: The thing is we need the fileid of the root, but I agree we should not be blocked here - and actually show those spaces in all cases as opposed to hiding them on stat failure, delaying actual failures to when users really navigate them.
We could fake the `mtime` and the `spaceID` with the config information we have, but not really the root fileid: shall we put `0` and see how the frontend would navigate through?

jessegeens: I would maybe be more a fan of putting a ttlcache here then ... in the end the public spaces are going to be the same for everyone so we don't need to do the actual stat all the time

glpatcern: Right, that works well too. And then the chance that those external eos instances are down precisely when we restart Reva (e.g. following interventions) is very slim (and we can control as part of the intervention).

diocas (approved):

PR #5286, opened 2025-09-01
OCM: Fixes to enable OCM sharing in the LibreGraph implementation

rawe0:
This enables for creation of OCM shares through the LibreGraph implementation instead of using the old /sciencemesh endpoint that is gone in the new frontend. 

- LinkOrShare has been changed to include OCMShares and renamed
- Web app now sends a POST request instead of a GET for OCM invite generation - this has been reflected in the backend.
- Adds the IDP to the user id when returned if user is of type federated.
- Cleaned up dead code since the /sciencemesh/create-share is no longer used.
- Conversions now take OCM shares into account as well. 

Note: Frontend needs to be updated to include a RecipientType 'Remote' if user is federated.


internal/http/services/owncloud/ocgraph/users.go
```diff
@@ -170,9 +170,15 @@ func mapToLibregraphUsers(users []*userpb.User, selection []UserSelectableProper
 			continue
 		}
 		lgUser := libregraph.User{}
+		var id string = u.Id.OpaqueId
+
+		// If the user is federated, include the IdP in the ID
```
glpatcern:
```suggestion
		// for federated users, form their OCM address by appending @IdP to the OpaqueId
```

internal/http/services/owncloud/ocgraph/conversions.go
```diff
@@ -5,94 +5,175 @@ import (
...
+		if err != nil {
+			return nil, errors.New("Failed to fetch user info")
 		}
-		return perm, nil
+		grantedTo.SetUser(libregraph.Identity{
+			Id:          libregraph.PtrString(grantee.GetUserId().OpaqueId),
+			DisplayName: u.DisplayName,
+		})
+	case provider.GranteeType_GRANTEE_TYPE_GROUP:
+		return nil, errors.New("Groups are not supported in OCM shares")
```
glpatcern:
```suggestion
		return nil, errors.New("Groups are currently not supported in OCM shares")
```

internal/http/services/owncloud/ocgraph/conversions.go
```diff
@@ -5,94 +5,175 @@ import (
...
 	}
+
+	return grantedTo, nil
+}
+
+func (s *svc) buildInvitation(ctx context.Context, creator *user.UserId) (*libregraph.SharingInvitation, error) {
```
glpatcern, line 172: True that this logic existed before, but maybe add a comment to state what is an "invitation" in this context

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -83,10 +87,97 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
...
+			share.NewWebappAccessMethod(viewMode),
+		},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.Status.Code != rpcv1beta1.Code_CODE_OK {
+		return nil, errors.New("failed to create share: " + resp.Status.Message)
```
```suggestion
		return nil, errors.New("failed to create remote share: " + resp.Status.Message)
```

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -171,57 +263,56 @@ func (s *svc) share(w http.ResponseWriter, r *http.Request) {
 		if recipient.ObjectId == nil {
 			handleError(ctx, errors.New("missing recipient data"), http.StatusBadRequest, w)
 		}
+		// If the recipient is a user or a group, we create a local share
+		switch *recipient.LibreGraphRecipientType {
+		case "user", "group":
+			resp, err := s.createLocalShare(ctx, gw, storageID, itemID, path, owner, statRes.Info.Type, *recipient.LibreGraphRecipientType, *recipient.ObjectId, exp, requestedPerms)
+			if err != nil {
+				log.Error().Err(err).Msg("failed to create share")
```
glpatcern: Let's simplify, `err` already contains all we need
```suggestion
				log.Error().Err(err).Msg("")
```

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -171,57 +263,56 @@ func (s *svc) share(w http.ResponseWriter, r *http.Request) {
...
+				log.Error().Err(err).Any("share", share).Err(err).Any("lgPerm", lgPerm).Msg("error converting created share to permissions")
+				handleError(ctx, err, http.StatusInternalServerError, w)
+				return
+			}
+			response = append(response, lgPerm)
+		case "remote":
+			username, idp := s.extractUsernameAndIdp(*recipient.ObjectId)
+			resp, err := s.createOCMShare(ctx, gw, resourceId, path, owner, statRes.Info.Type, *recipient.LibreGraphRecipientType, username, idp, exp, roles[0])
+			if err != nil {
+				log.Error().Err(err).Msg("failed to create remote share")
```
glpatcern: As above
```suggestion
				log.Error().Err(err).Msg("")
```

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -83,10 +87,97 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
...
+		return nil, err
+	}
+
+	if resp.Status.Code != rpcv1beta1.Code_CODE_OK {
+		return nil, errors.New("failed to create share: " + resp.Status.Message)
+	}
+	return resp, nil
+}
+
+func (s *svc) extractUsernameAndIdp(recipientID string) (string, string) {
```
glpatcern: We know what this does, but let's make it explicit that we want to use it for OCM addresses **only** (yet some usernames may have a `@`). So I'd rename it `decomposeOcmAddress` and fail when the `if` returns false

rawe0: Done!

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -174,57 +267,60 @@ func (s *svc) share(w http.ResponseWriter, r *http.Request) {
...
+			if err != nil || lgPerm == nil {
+				log.Error().Err(err).Any("share", share).Err(err).Any("lgPerm", lgPerm).Msg("error converting created share to permissions")
+				handleError(ctx, err, http.StatusInternalServerError, w)
+				return
+			}
+			response = append(response, lgPerm)
+		case "remote":
+			username, idp := s.decomposeOCMAddress(*recipient.ObjectId)
+			if username == "" || idp == "" {
+				handleError(ctx, errors.New("invalid remote recipient address, must be user@idp"), http.StatusBadRequest, w)
```
```suggestion
				handleError(ctx, errors.New("invalid remote recipient address, must be remote_user_id@remote_ocm_fqdn"), http.StatusBadRequest, w)
```

internal/http/services/owncloud/ocgraph/conversions.go
```diff
@@ -22,86 +20,169 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	invitation, err := s.buildInvitation(ctx, share.ocmshare.Creator)
+	if err != nil {
+		return nil, err
+	}
+
+	// OCM shares only has one access method implemented, so we can safely use the first one here
```
glpatcern:
```suggestion
	// Here we assume that the OCM shares offered by us always contain both `webdav` and `webapp` access methods with
	// identical permissions, therefore we use the first one here (`webdav`) to generate the libregraph representation
```

glpatcern (commented):
A couple of trivial changes for clarity, otherwise we can merge and QA this

internal/http/services/owncloud/ocgraph/shares.go
rawe0: This is correct and indeed what I meant! (Will change to what you wrote)

internal/http/services/owncloud/ocgraph/conversions.go
rawe0: I realized this later but forgot to go back and change it, and indeed both access methods are there.

PR #5287, opened 2025-09-02
feat: new rules for public links in graphAPI

rodcoffani:
- update is now able to handle multiple fields changes in the same request
- new rules for expiration dates for public links with Editor permissions

### DIFF
--- a/internal/http/services/owncloud/ocgraph/drive_permissions.go
+++ b/internal/http/services/owncloud/ocgraph/drive_permissions.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"time"
 
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	collaborationv1beta1 "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	linkv1beta1 "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	providerpb "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/cs3org/reva/v3/internal/http/services/owncloud/ocs/conversions"
 	"github.com/cs3org/reva/v3/pkg/appctx"
 	"github.com/cs3org/reva/v3/pkg/errtypes"
 	"github.com/cs3org/reva/v3/pkg/spaces"
@@ -140,7 +142,7 @@ func (s *svc) updateDrivePermissions(w http.ResponseWriter, r *http.Request) {
 	if shareOrLink.shareType == "share" {
 		s.updateSharePermissions(ctx, w, shareOrLink.share, permission, resourceID)
 	} else {
-		s.updateLinkPermissions(ctx, w, &linkv1beta1.PublicShareId{OpaqueId: shareOrLink.ID}, permission, resourceID)
+		s.updateLinkPermissions(ctx, w, shareOrLink.link, permission, resourceID)
 	}
 }
 
@@ -190,7 +192,7 @@ func (s *svc) deleteDrivePermissions(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *svc) updateLinkPermissions(ctx context.Context, w http.ResponseWriter, linkId *linkv1beta1.PublicShareId, permission *libregraph.Permission, resourceId *providerpb.ResourceId) {
+func (s *svc) updateLinkPermissions(ctx context.Context, w http.ResponseWriter, link *linkv1beta1.PublicShare, permission *libregraph.Permission, resourceId *providerpb.ResourceId) {
 	log := appctx.GetLogger(ctx)
 
 	gw, err := s.getClient()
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -211,42 +213,45 @@ func (s *svc) updateLinkPermissions(ctx context.Context, w http.ResponseWriter,
...
+			log.Error().Err(err).Msg("error updating public share")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		if uRes.Status.Code != rpcv1beta1.Code_CODE_OK {
+			log.Error().Interface("response", uRes).Msg("error updating public share")
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		lgPerm, err = s.shareToLibregraphPerm(ctx, &ShareOrLink{
```
diocas, line 245: Why is this inside the loop? Don't we care only about the last one?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -562,38 +567,90 @@ func (s *svc) writePermissions(ctx context.Context, w http.ResponseWriter, actio
...
+
+	defaultExpirationDefined := time.Second * time.Duration(s.c.PubRWLinkDefaultExpiration)
+	maxExpirationDefined := time.Second * time.Duration(s.c.PubRWLinkMaxExpiration)
+	endOfDay := time.Date(time.Now().Year(), time.Now().Month(), time.Now().Day(), 23, 59, 59, 0, time.Now().Location())
+	defaultExpiration := endOfDay.Add(defaultExpirationDefined)
+	maxExpiration := endOfDay.Add(maxExpirationDefined)
+
+	isExpirationEnforced := maxExpirationDefined > 0
+	isEditorLink := false
+	if permission.Link != nil && permission.Link.Type != nil {
```
jessegeens, line 581: was already the case, but you might have a nilpointer here if `permission` is nil

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -562,38 +567,90 @@ func (s *svc) writePermissions(ctx context.Context, w http.ResponseWriter, actio
...
 			Grant: &linkv1beta1.Grant{
-				Expiration: nullableTimeToCs3Timestamp(permission.ExpirationDateTime),
+				Expiration: nullableTimeToCs3Timestamp(finalExpiration),
 			},
-		}, nil
-	} else if permission.Link != nil && permission.Link.LibreGraphDisplayName != nil {
-		if *permission.Link.LibreGraphDisplayName == "" {
+		})
+	}
+	if permission.Link != nil && permission.Link.LibreGraphDisplayName != nil {
```
diocas, line 603: You don't have `else if` any longer because we cannot enter in more than one if, is this correct and certain?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -562,38 +567,90 @@ func (s *svc) writePermissions(ctx context.Context, w http.ResponseWriter, actio
...
+	}
+	if permission.Link != nil && permission.Link.Type != nil {
 		permissions, err := CS3ResourcePermissionsFromSharingLink(permission.Link.GetType(), resourceType)
 		if err != nil {
 			return nil, errors.Wrap(err, "error converting link type to permissions")
 		}
-		return &linkv1beta1.UpdatePublicShareRequest_Update{
+
+		finalExpiration := libregraph.NullableTime{}
+		if isEditorLink && isExpirationEnforced {
```
diocas, line 619: This code seems repeated. Can we externalise this?

diocas: Although, I'm trying to think of a way to do it and it's not straightforward...

internal/http/services/owncloud/ocgraph/drive_permissions.go
diocas: Nvm.. we _want_ to have more than one.. that is the whole purpose...

internal/http/services/owncloud/ocgraph/drive_permissions.go
diocas: I would fix this

internal/http/services/owncloud/ocgraph/drive_permissions.go
rodcoffani: as the update itself is an atomic operation, this way it's possible to update one or more fields in the same request by running though the fields that where updated.

as iterates through each update, `lgPerm` is set to the last data of the share (which is later returned)

internal/http/services/owncloud/ocgraph/drive_permissions.go
rodcoffani: It's a possibility to combine `isEditorLink` and `isExpirationEnforced` in the same var, but I would keep like this for the sake of readability.

internal/http/services/owncloud/ocgraph/drive_permissions.go
rodcoffani: Is it possible? I guess that is already taken care of.

Around line 131~138 the variable is declared as 
```
	permission := &libregraph.Permission{}
	dec := json.NewDecoder(r.Body)
	dec.DisallowUnknownFields()
	if err := dec.Decode(permission); err != nil {
		log.Error().Err(err).Interface("Body", r.Body).Msg("failed unmarshalling request body")
		w.WriteHeader(http.StatusBadRequest)
		return
	}
```
and only then is sent to the other functions.

PR #5289, opened 2025-09-03
Better logging in GraphAPI

jessegeens:
* Improved logging in graphAPI and `GetQuota`
* Moved one conversion function in graph api from `drives` to `conversions`

### DIFF
--- a/internal/http/services/owncloud/ocgraph/conversions.go
+++ b/internal/http/services/owncloud/ocgraph/conversions.go
@@ -2,6 +2,7 @@ package ocgraph
 
 import (
 	"context"
+	"fmt"
 	"path"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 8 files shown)


diocas (approved):

internal/http/services/owncloud/ocgraph/conversions.go
```diff
@@ -162,3 +166,43 @@ func LinkTypeToPermissions(lt libregraph.SharingLinkType, resourceType provider.
...
+	resourceRelativePath, _ := spaces.PathRelativeToSpaceRoot(rsi.ResourceInfo)
+
+	return &libregraph.Drive{
+		Id:         libregraph.PtrString(libregraphShareID(rsi.ReceivedShare.Share.Id)),
+		DriveType:  libregraph.PtrString("mountpoint"),
+		DriveAlias: libregraph.PtrString(rsi.ReceivedShare.Share.Id.OpaqueId), // this is not used, but must not be the same alias as the drive item
+		Name:       filepath.Base(rsi.ResourceInfo.Path),
+		WebUrl:     libregraph.PtrString(fullURL(s.c.WebBase, rsi.ResourceInfo.Path)),
+		Quota: &libregraph.Quota{
+			Total:     libregraph.PtrInt64(24154390300000),
```
glpatcern, line 457: I know this was already there, but why do we have those hardcoded numbers?

jessegeens: Good question. We don't really have quota's on spaces at the moment, but for `used` at least we could populate this with the size of the resource

diocas: mountpoints are other users homes. not sure we want to get their quotas.

diocas: Also, for a PR just for cleanup, I would not put logic changes.

diocas: I'm resolving for now, we can revisit later.

PR #5290, opened 2025-09-03
Fix nilpointers in spaces shares

jessegeens:
* Set `SkipFetchingGroupMembers` and `SkipFetchingUserGroups` to `true` in libregraph API; we don't need those there and fetching group members of `cern-all-users` crashes the daemon
* Better nil handling
* Move some conversion methods from `shares.go` to `conversions.go`

### DIFF
--- a/internal/http/services/owncloud/ocgraph/shares.go
+++ b/internal/http/services/owncloud/ocgraph/shares.go
@@ -23,15 +23,14 @@ package ocgraph
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
-	"path"
-	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -64,6 +62,12 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if resShares.Status == nil || resShares.Status.Code != rpc.Code_CODE_OK {
+		handleRpcStatus(ctx, resShares.Status, w)
+	}
+
+	log.Error().Msgf("FETCHED %d SHARES", len(resShares.ShareInfos))
```
diocas: clean?

jessegeens: whoops, fixed!

PR #5291, opened 2025-09-04
add support for filtering users on type

jessegeens:


internal/http/services/owncloud/ocgraph/users.go
```diff
@@ -94,13 +94,22 @@ func (s *svc) listUsers(w http.ResponseWriter, r *http.Request) {
...
+		Query:                  queryVal,
+	}
+
+	filters, err := generateUserFilters(req)
+	if err != nil {
+		handleError(ctx, err, http.StatusBadRequest, w)
+		return
+	}
+
+	request.Filter = filters
```
diocas:
```suggestion
	request := &userpb.FindUsersRequest{
		SkipFetchingUserGroups: true,
		Query:                  queryVal,
		Filters:                filters,
	}
```
?

PR #5293, opened 2025-09-05
Make preferences work for LW accounts

jessegeens:
* Make preferences work for LW accounts
* Fix nilpointer panic when fetching preference fails


glpatcern (approved):
So with this we're able to persist LW accounts' preferences as well after all

PR #5301, opened 2025-09-09
WIP: show shares to project admins

jessegeens:


internal/grpc/services/spacesregistry/spacesregistry.go
```diff
@@ -235,7 +221,19 @@ func (s *service) listSpacesByType(ctx context.Context, user *userpb.User, space
...
-		fallthrough
+		// but only if the request was not made with an ID-filter,
+		// as that would mean the requestor is looking for a specific space
+		//
+		// Having a `fallthrough` here would've been nice, but Go does
+		// not allow conditional fallthroughs
+		if _, isFilterById := isFilterByID(req.Filters); !isFilterById {
+			publicSpaces, err := s.getPublicSpaces(ctx)
+			if err != nil {
+				return nil, err
```
glpatcern: Shouldn't we just log and ignore, so that the other spaces are still returned?

pkg/projects/manager/sql/sql.go
```diff
@@ -70,6 +70,7 @@ const cacheKey = "projects/projectsListCache"
 // Project represents a project in the DB.
 type Project struct {
 	gorm.Model
+	SpaceID   string `gorm:"size:255;uniqueIndex:i_space_id"`
```
glpatcern, line 73: how are we going to populate the existing entries?

jessegeens: We'll have to use a script :)

internal/grpc/services/spacesregistry/spacesregistry.go
jessegeens: this we could also do

internal/grpc/services/spacesregistry/spacesregistry.go
```diff
@@ -162,7 +162,7 @@ func (s *service) ListStorageSpaces(ctx context.Context, req *provider.ListStora
 
 		projects, err := s.listSpacesByType(ctx, req, user, spaces.SpaceTypeProject)
 		if err != nil {
-			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error())}, nil
+			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error()), StorageSpaces: projects}, nil
 		}
 		sp = append(sp, projects...)
```
glpatcern: shouldn't we do this?
```suggestion
		if projects != nil {
			sp = append(sp, projects...)
		}
		if err != nil {
			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error()), StorageSpaces: sp}, nil
		}
```

internal/grpc/services/spacesregistry/spacesregistry.go
```diff
@@ -154,58 +154,44 @@ func (s *service) ListStorageSpaces(ctx context.Context, req *provider.ListStora
...
 		sp = append(sp, homes...)
 
-		projects, err := s.listSpacesByType(ctx, user, spaces.SpaceTypeProject)
+		projects, err := s.listSpacesByType(ctx, req, user, spaces.SpaceTypeProject)
+		if projects != nil {
+			sp = append(sp, projects...)
+		}
 		if err != nil {
-			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error())}, nil
+			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error()), StorageSpaces: projects}, nil
```
glpatcern:
```suggestion
			return &provider.ListStorageSpacesResponse{Status: status.NewInternal(ctx, err, err.Error()), StorageSpaces: sp}, nil
```

PR #5318, opened 2025-09-17
dynamic router: do not fail if a mount configuration is missing

glpatcern:
Continuation of #5314

### DIFF
--- a/pkg/storage/registry/dynamic/dynamic.go
+++ b/pkg/storage/registry/dynamic/dynamic.go
@@ -217,13 +217,10 @@ func (d *dynamic) FindProviders(ctx context.Context, ref *provider.Reference) ([
 				Address:      address,
 			})
 		} else {
-			err := errtypes.InternalError("storage provider address not configured for mountID " + p.ProviderId)
-			l.Error().Err(err).Send()
-			return nil, err
+			l.Error().Str("mount", p.ProviderId).Msg("Storage provider address not configured for mount, skipping")
 		}
 	}
 
-	l.Debug().Msgf("resolved storage providers %+v", providers)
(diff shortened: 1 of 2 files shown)


pkg/storage/registry/dynamic/dynamic.go
```diff
@@ -217,13 +217,10 @@ func (d *dynamic) FindProviders(ctx context.Context, ref *provider.Reference) ([
...
-			err := errtypes.InternalError("storage provider address not configured for mountID " + p.ProviderId)
-			l.Error().Err(err).Send()
-			return nil, err
+			l.Error().Str("mount", p.ProviderId).Msg("Storage provider address not configured for mount, skipping")
 		}
 	}
 
-	l.Debug().Msgf("resolved storage providers %+v", providers)
-
+	l.Info().Msgf("resolved storage providers %+v", providers)
```
diocas: Maybe info is too much for this?
```suggestion
	l.Debug().Msgf("resolved storage providers %+v", providers)
```

jessegeens: and while we're at it, I think this log line doesn't provide much value if we don't also log the original reference we found the providers for

glpatcern: good point about logging the original reference. I thought to promote it to `Info` to see it in prod, but it's true that we call this method quite often (at least for every personal space...), so after all I can leave it `Debug`.

glpatcern: (in fact the `ref` is already logged by how the logger was initialized, so I've just reverted to Debug)

PR #5319, opened 2025-09-18
Modify SetQuota to consider New Project Level Quotas

pmedinar01:


pkg/eosclient/eosgrpc/quota.go
```diff
@@ -112,7 +112,13 @@ func (c *Client) SetQuota(ctx context.Context, rootAuth eosclient.Authorization,
 
 	// We set a quota for a user, not a group!
 	msg.Id.Uid = uidInt
```
glpatcern: This comment and this line now only applies on the `else` below, whereas in case of Gid = 99, you need to set `msg.id.Uid = 0`.

glpatcern (changes requested):
The copilot-generated comment actually makes sense: this magic number `99` is very EOS-specific, so it's preferable to define it as a constant once, in the eosclient, and use it both in eosbinary and eosgrpc. Could you please do that?

pkg/eosclient/eosgrpc/quota.go
jessegeens: Why? `uid = 0` means root, which won't work over gRPC right?

jessegeens: (though I'm not sure what the comment means - probably that we kept the original uid and only modified the gid so that we set it as the correct user? )

pmedinar01: Added ProjectQuotaGID as constant.

glpatcern: I guess that's what the comment means - so far we only provided the Uid and set the Gid to `0`, so symmetrically we should do the opposite I think (which should similarly work over gRPC? To be tested!)

pmedinar01: Fixed.

pkg/eosclient/eosbinary/eosbinary.go
```diff
@@ -649,10 +649,16 @@ func (c *Client) GetQuota(ctx context.Context, user eosclient.Authorization, roo
...
 
 // SetQuota sets the quota of a user on the quota node defined by path.
-func (c *Client) SetQuota(ctx context.Context, rootAuth eosclient.Authorization, info *eosclient.SetQuotaInfo) error {
+func (c *Client) SetQuota(ctx context.Context, user eosclient.Authorization, rootAuth eosclient.Authorization, info *eosclient.SetQuotaInfo) error {
 	maxBytes := fmt.Sprintf("%d", info.MaxBytes)
 	maxFiles := fmt.Sprintf("%d", info.MaxFiles)
-	args := []string{"quota", "set", "-u", info.Username, "-p", info.QuotaNode, "-v", maxBytes, "-i", maxFiles}
+	var args []string
+	if user.Role.GID == eosclient.ProjectQuotaGID {
+		// NewStyle project quota
```
glpatcern:
```suggestion
		// new style project quota: we set a group quota on the space itself
```

pkg/eosclient/eosgrpc/quota.go
```diff
@@ -110,9 +110,16 @@ func (c *Client) SetQuota(ctx context.Context, rootAuth eosclient.Authorization,
 		return err
 	}
 
-	// We set a quota for a user, not a group!
-	msg.Id.Uid = uidInt
-	msg.Id.Gid = 0
+	if user.Role.GID == eosclient.ProjectQuotaGID {
+		// We set a quota for a group, not a user!
```
glpatcern:
```suggestion
		// new style project quota: we set a group quota on the space itself
```

glpatcern (approved):

PR #5316, opened 2025-09-16
OCM: List OCM shares in LibreGraph implementation

rawe0:
This PR includes the following:

- Adapating the LibreGraph implementation to also list OCM shares
- Bugfix for uploading (e.g. editing files) through OCM, headers had to be updated
- Bugfix where stating a non existing file did not return a `not found` error - this made CreateDir and Touch fail since reva could not determine if the file exists. 
- Bugfix for creating files and directories through OCM, essentially the old logic was to try and stat the file to test different the different authentication methods basic and bearer, however this doesn't work for `Touch` and `CreateDir` since the stat is meant the fail here, the logic needs to be looked over. Further the OCM basic auth over webdav is broken and this causes these two operations to fail - since the stat fails and it reverts to basic auth which fails since it is currently broken.

PR #5320, opened 2025-09-18
OCM: Update, Delete and Show OCM in SharedWith

rawe0:
This PR adds support for displaying OCM shares in the SharedWith view when a file is selected. It also enables updating the permissions of those shares and deleting them directly from the view.


rawe0:
> I guess this one depends on #5316 ?

In fact they're kind of separate now that I think of it but it makes sense to merge them in this order I think

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -181,16 +205,18 @@ func (s *svc) deleteDrivePermissions(w http.ResponseWriter, r *http.Request) {
...
-	shareOrLink, err := s.getShareOrLink(ctx, shareID, resourceID)
+	genericShare, err := s.getGenericShare(ctx, shareID, resourceID)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
 
-	if shareOrLink.shareType == "share" {
-		s.deleteSharePermissions(ctx, w, r, &collaborationv1beta1.ShareId{OpaqueId: shareOrLink.ID})
+	if genericShare.shareType == "share" {
```
jessegeens: maybe a `switch` here would be nicer? and turning the `shareType` into an enum?

internal/http/services/owncloud/ocgraph/drive_permissions.go
```diff
@@ -318,6 +344,69 @@ func (s *svc) updateSharePermissions(ctx context.Context, w http.ResponseWriter,
 	_ = json.NewEncoder(w).Encode(lgPerm)
 }
 
+func (s *svc) updateOCMSharePermissions(ctx context.Context, w http.ResponseWriter, ocmshare *ocm.Share, lgPerm *libregraph.Permission, resourceId *providerpb.ResourceId) {
```
jessegeens: I wonder if at this point we can pull out some of this code since we have now at least a part of this three times ... but not sure how easy that would be

glpatcern (approved):

PR #5323, opened 2025-09-23
Config cleanup: add database config to sharedconf

jessegeens:
Add database configuration to `sharedconf`, so that it doesn't have to be repeated for every driver

Depends on https://github.com/cs3org/reva/pull/5322 being merged

### DIFF
--- a/pkg/share/manager/sql/public_link.go
+++ b/pkg/share/manager/sql/public_link.go
@@ -45,7 +45,7 @@ import (
 )
 
 type PublicShareMgr struct {
-	c  *config
+	c  *Config
 	db *gorm.DB
 }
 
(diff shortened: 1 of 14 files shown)


pkg/share/manager/sql/public_link.go
```diff
@@ -54,11 +55,19 @@ type ExpiryRange struct {
 	To   time.Time
 }
 
-func NewPublicShareManager(ctx context.Context, m map[string]interface{}) (publicshare.Manager, error) {
-	var c config
+func (PublicShareMgr) RevaPlugin() reva.PluginInfo {
```
glpatcern: Why do we want to expose the PublicShareManager as a Reva plugin?

jessegeens: Good question, to be honest I don't remember. So I removed it now

glpatcern: Maybe a leftover from when it was a plugin!

PR #5326, opened 2025-09-25
check status of ListReceivedOCMShares before reading its values

jessegeens:


internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -86,14 +80,28 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
...
-		drive, err := s.OCMReceivedShareToDriveItem(ctx, share)
+	if s.c.OCMEnabled {
+		// include ocm shares in the response
+		ocmShareResp, err := gw.ListReceivedOCMShares(ctx, &ocm.ListReceivedOCMSharesRequest{})
 		if err != nil {
-			log.Error().Err(err).Any("share", share).Msg("error parsing received share, ignoring")
-		} else {
-			shares = append(shares, drive)
+			//handleError(ctx, err, http.StatusInternalServerError, w)
+			log.Fatal().Err(err).Msg("Cannot list ReceivedOCMShares")
```
```suggestion
			log.Fatal().Err(err).Msg("ListReceivedOCMShares returned error despite OCM is enabled, users won't be able to see their OCM shares")
```

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -86,14 +80,28 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
...
+		}
+		if ocmShareResp != nil {
+			if ocmShareResp.Status == nil || ocmShareResp.Status.Code != rpc.Code_CODE_OK {
+				handleRpcStatus(ctx, ocmShareResp.Status, "ocgraph: failed to perform ListReceivedOCMShares ", w)
```
glpatcern, line 92: Don't we want to log the same Fatal error here?

internal/http/services/owncloud/ocgraph/shares.go
jessegeens: This is not true - *this specific user* won't be able to see their OCM shares ;)

internal/http/services/owncloud/ocgraph/shares.go
jessegeens: Hmm; I would tackle this later. We cannot just log here because this block needs to finish; we cannot loop over `ocmShareResp` later otherwise

internal/http/services/owncloud/ocgraph/shares.go
glpatcern: Not sure, the error here would suggest that the OCM provider didn't respond. But OK let's tackle this at a later stage where decide how to handle errors - and have a clear separation between critical system-level errors vs non-critical system-level errors vs user-level errors

glpatcern (approved):

PR #5327, opened 2025-09-25
WIP

rawe0:
Draft pull request for gormifying the ocm share related tables and some of the functions.

- `model.go` be compared to the `init.sql` (`pkg/ocm/share/repository/sql/init.sql`).
- `ocm_shares.go` should be compared `sql.go` (`pkg/ocm/share/repository/sql/`). 
- `conversions.go` has been moved as well and there are some minor changes there for now, since the data structures changed to the ones from `model.go`.

The tests are yet to be converted but will come as well.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	golang.org/x/sys v0.36.0
 	golang.org/x/term v0.35.0
 	google.golang.org/genproto v0.0.0-20250603155806-513f23925822
-	google.golang.org/grpc v1.75.1
+	google.golang.org/grpc v1.76.0
 	google.golang.org/protobuf v1.36.9
 	gorm.io/datatypes v1.2.7
 	gorm.io/driver/mysql v1.6.0
@@ -96,7 +96,7 @@ require (
 	github.com/fsnotify/fsnotify v1.4.9 // indirect
 	github.com/gabriel-vasile/mimetype v1.4.8 // indirect
 	github.com/go-asn1-ber/asn1-ber v1.5.8-0.20250403174932-29230038a667 // indirect
-	github.com/go-jose/go-jose/v4 v4.1.1 // indirect
+	github.com/go-jose/go-jose/v4 v4.1.2 // indirect
 	github.com/go-kit/kit v0.10.0 // indirect
 	github.com/go-openapi/errors v0.22.1 // indirect
 	github.com/go-openapi/strfmt v0.23.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -969,8 +969,8 @@ github.com/go-fonts/stix v0.1.0/go.mod h1:w/c1f0ldAUlJmLBvlbkvVXLAD+tAMqobIIQpmn
 github.com/go-gl/glfw v0.0.0-20190409004039-e6da0acd62b1/go.mod h1:vR7hzQXu2zJy9AVAgeJqvqgH9Q5CA+iKCZ2gyEVpxRU=
 github.com/go-gl/glfw/v3.3/glfw v0.0.0-20191125211704-12ad95a8df72/go.mod h1:tQ2UAYgL5IevRw8kRxooKSPJfGvJ9fJQFa0TUsXzTg8=
 github.com/go-gl/glfw/v3.3/glfw v0.0.0-20200222043503-6f7a984d4dc4/go.mod h1:tQ2UAYgL5IevRw8kRxooKSPJfGvJ9fJQFa0TUsXzTg8=
-github.com/go-jose/go-jose/v4 v4.1.1 h1:JYhSgy4mXXzAdF3nUx3ygx347LRXJRrpgyU3adRmkAI=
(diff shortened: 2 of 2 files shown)


pkg/share/manager/sql/model/model.go
```diff
@@ -99,6 +99,95 @@ type ShareState struct {
 	Alias  string `gorm:"size:64"`
 }
 
+// OCM Shares - uses BaseModel to share common ShareID with other share types
+type OcmShares struct {
```
jessegeens: Should be `OCMShare` (since this is a single share). GORM will append the `s` to the table name :)

pkg/share/manager/sql/model/model.go
```diff
@@ -99,6 +99,95 @@ type ShareState struct {
...
+	Owner        string             `gorm:"size:255;not null;uniqueIndex:idx_fileid_source_share_with"`
+	Initiator    string             `gorm:"type:text;not null"`
+	Ctime        time.Time          `gorm:"not null"`
+	Mtime        time.Time          `gorm:"not null"`
+	Expiration   datatypes.NullTime `gorm:"default:null"`
+	Type         uint8              `gorm:"not null"`
+}
+
+// OCM Shares Access Methods
+type OcmSharesAccessMethods struct {
```
jessegeens: `OCMShareAccessMethod`? (once again, GORM will add the last `s`, and these are access methods for a single share right?)

pkg/share/manager/sql/model/model.go
```diff
@@ -99,6 +99,95 @@ type ShareState struct {
...
+	gorm.Model
+	OcmShareID uint      `gorm:"not null;uniqueIndex:idx_ocm_share_method"`
+	OcmShare   OcmShares `gorm:"constraint:OnDelete:CASCADE;foreignKey:OcmShareID;references:Id"`
+	Type       uint8     `gorm:"not null;uniqueIndex:idx_ocm_share_method"`
+}
+
+// WebDAV Access Method
+type OcmAccessMethodWebdav struct {
```
jessegeens: Since we are redesigning the tables anyway, can we not just put this info in `OCMShareAccessMethods` anyway? Since `OcmAccessMethodWebapp` and `OcmAccessMethodWebdav` basically have the same columns?

pkg/share/manager/sql/model/model.go
```diff
@@ -99,6 +99,95 @@ type ShareState struct {
...
+
+// WebApp Access Method
+type OcmAccessMethodWebapp struct {
+	OcmAccessMethodID uint                   `gorm:"primaryKey;not null"`
+	AccessMethod      OcmSharesAccessMethods `gorm:"constraint:OnDelete:CASCADE;foreignKey:OcmAccessMethodID;references:ID"`
+	ViewMode          uint32                 `gorm:"not null"`
+}
+
+// OCM Received Shares
+type OcmReceivedShares struct {
```
jessegeens: If you want to have this one have unique IDs (across shares), you should embed `BaseModel`, not `gorm.Model`

jessegeens: Though since it's the Received Shares I guess that's not necessary, so feel free to discard my comment in that case :)

glpatcern: That depends on the behavior of the `Shared with Me` view, as that's the only place where they will appear

pkg/share/manager/sql/model/model.go
rawe0: Done!

pkg/share/manager/sql/model/model.go
rawe0: Done!

pkg/share/manager/sql/model/model.go
rawe0: I've combined the tables for the protocols and the access methods now.

pkg/share/manager/sql/model/model.go
rawe0: > Though since it's the Received Shares I guess that's not necessary, so feel free to discard my comment in that case :)

Yep it's not really relevant here, they can have their own IDs

pkg/share/manager/sql/model/model.go
```diff
@@ -99,6 +159,73 @@ type ShareState struct {
...
+	Uri          string `gorm:"size:255;default:null"`
+	SharedSecret string `gorm:"type:text;default:null"`
+	Permissions  int    `gorm:"default:null"`
+
+	// WebApp Protocol fields
+	UriTemplate string `gorm:"size:255;default:null"`
+	ViewMode    int    `gorm:"default:null"`
+
+	// Transfer Protocol fields
+	SourceUri string `gorm:"size:255;default:null"`
```
glpatcern: Here the idea would be to only use `Uri` (`not null`) for all protocols, and `Permissions` to also represent the `ViewMode` values given that we have a 1-to-1 mapping. So those fields can be dropped altogether.

glpatcern (commented):
Haven't reviewed the rest of the code yet, this is for now just to align on the model

pkg/share/manager/sql/model/model.go
rawe0: done!

PR #5335, opened 2025-09-30
Fix versions and lightweight accounts for EOS 5.2.22

jessegeens:
* Remove "x" bit from permissions in EOS token
* assume role instead ofusing token when restoring / listing / downloading revisions


glpatcern (approved):

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -2126,7 +2104,9 @@ func (fs *Eosfs) getEOSToken(ctx context.Context, u *userpb.User, fn string) (eo
 	if info.IsDir {
 		// EOS expects directories to have a trailing slash when generating tokens
 		fn = path.Clean(fn) + "/"
+		perm = perm + "x"
 	}
+	log.Info().Msgf("FindMe - Generating a token for %s with perms %s", fn, perm)
```
glpatcern: I guess you want to drop `FindMe` and downgrade to `Debug`?

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -2106,7 +2082,9 @@ func (fs *Eosfs) getEOSToken(ctx context.Context, u *userpb.User, fn string) (eo
 		},
 	}
 
-	perm := "rwx"
+	// EOS tokens can only go to files
+	// for files, the "x" bit cannot be set
```
glpatcern:
```suggestion
	// initialize with the EOS token permissions for files, which cannot have the "x" bit set
```

changelog/unreleased/fix-eostokens.md
```diff
@@ -0,0 +1,7 @@
+Bugfix: fix EOS tokens + EOS 5.3.22
+
+Since EOS 5.3.22 does more strict checks on tokens, we:
+*  Remove "x" bit from permissions in EOS token
+*  assume role instead ofusing token when restoring / listing / downloading revisions
```
diocas:
```suggestion
*  assume role instead of using token when restoring / listing / downloading revisions
```

pkg/eosclient/eosgrpc/versions.go
```diff
@@ -59,7 +59,7 @@ func (c *Client) RollbackToVersion(ctx context.Context, auth eosclient.Authoriza
 	if resp.GetError() != nil {
 		log.Info().Str("func", "RollbackToVersion").Int64("errcode", resp.GetError().Code).Str("errmsg", resp.GetError().Msg).Msg("grpc response")
 	}
-	return err
+	return nil
```
diocas, line 62: What is the implication of this?

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -2138,6 +2118,27 @@ func (fs *Eosfs) getEOSToken(ctx context.Context, u *userpb.User, fn string) (eo
 	return eosclient.Authorization{Token: tkn}, nil
 }
 
+// // Returns an Authorization wich assumes the role of the owner of the file `fn`
+func (fs *Eosfs) getOwnerAuth(ctx context.Context, u *userpb.User, fn string) (eosclient.Authorization, error) {
```
diocas: `user` is not used for anything, so it shouldn't be here (it would be clearer as well not to have it).  Removing it here also removes the need to get it from ctx.

But btw... Q: doesn't the metadata of the file (that you get before calling this) contain the owner?

diocas (changes requested):

pkg/eosclient/eosgrpc/versions.go
jessegeens: nothing, because we already return if we have an error earlier :)
now it also always returned nil, just implicitly

pkg/storage/utils/eosfs/eosfs.go
jessegeens: No. Because for example in `DownloadRevision` and `RestoreRevision`, a `Reference` is passed (so without any info)

diocas (approved):

PR #5336, opened 2025-10-01
sharedByMe works with deleted accounts

jessegeens:
The `/sharedByMe` endpoint now works, even if the user has shares with users that no longer exist


internal/http/services/owncloud/ocgraph/conversions.go
```diff
@@ -544,12 +544,16 @@ func (s *svc) cs3sharesToPermissions(ctx context.Context, shares []*GenericShare
 
 			creator, err := s.getUserByID(ctx, e.share.Creator)
 			if err != nil {
-				return nil, err
+				log.Error().Err(err).Msg("Failed to convert cs3 share to permission in ocgraph - ignoring this share")
+				continue
+				//return nil, err
```
diocas: clear comment?

PR #5339, opened 2025-10-02
NotifyUploads on dropzones in libregraph

jessegeens:


glpatcern (commented):
I think it makes perfect sense to remove the `CreateContainer`, especially as we flatten folders when people upload them (otherwise it would defeat the purpose of not knowing what is in the storage). You can remove it "for good" by dropping the comments IMO.

glpatcern (approved):

PR #5340, opened 2025-10-03
Support requests to /app/new on space root without container ID

jessegeens:


glpatcern (approved):
Good, after all it may turn useful in other places

PR #5345, opened 2025-10-06
Fix 502 on /shared{By,With}Me for LW accs when OCM is enabled 

jessegeens:
When OCM is enabled, going to `/sharedWithMe` or `/sharedByMe` results in a 502.
We now skip checking for OCM shares on LW accounts. Additionally, error handling in the libregraph layer has been improved

### DIFF
--- a/internal/http/services/owncloud/ocgraph/shares.go
+++ b/internal/http/services/owncloud/ocgraph/shares.go
@@ -46,6 +46,7 @@ import (
 	"github.com/cs3org/reva/v3/pkg/appctx"
 	"github.com/cs3org/reva/v3/pkg/ocm/share"
 	"github.com/cs3org/reva/v3/pkg/spaces"
+	"github.com/cs3org/reva/v3/pkg/utils"
 	libregraph "github.com/owncloud/libre-graph-api-go"
 )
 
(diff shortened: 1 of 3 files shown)


internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -81,14 +89,13 @@ func (s *svc) getSharedWithMe(w http.ResponseWriter, r *http.Request) {
...
+	if s.c.OCMEnabled && !utils.IsLightweightUser(u) {
 		// include ocm shares in the response
 		ocmShareResp, err := gw.ListReceivedOCMShares(ctx, &ocm.ListReceivedOCMSharesRequest{})
 		if err != nil {
-			//handleError(ctx, err, http.StatusInternalServerError, w)
+			handleError(ctx, err, w)
 			log.Fatal().Err(err).Msg("ListReceivedOCMShares returned error - user will not be able to see their OCM shares")
-		}
-		if ocmShareResp != nil {
+		} else if ocmShareResp != nil {
```
glpatcern, line 98: It's all fine and the error handling refactoring is welcome. Here I guess the function may return both values != nil and that's why it was handled with two separate `if`s?

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -549,16 +556,20 @@ func (s *svc) getSharedByMe(w http.ResponseWriter, r *http.Request) {
...
 
 	user, ok := appctx.ContextGetUser(ctx)
 	if !ok {
-		handleError(ctx, fmt.Errorf("No user in context"), http.StatusUnauthorized, w)
+		handleCustomError(ctx, fmt.Errorf("No user in context"), http.StatusUnauthorized, w)
 		return
 	}
 
+	if utils.IsLightweightUser(user) {
+		handleCustomError(ctx, errors.New("Ligthweight accounts cannot share"), http.StatusUnauthorized, w)
```
```suggestion
		handleCustomError(ctx, errors.New("External accounts do not have permission to share"), http.StatusUnauthorized, w)
```

internal/http/services/owncloud/ocgraph/shares.go
jessegeens: in theory no, but I figured the else wouldn't hurt just in case :)

glpatcern (approved):

PR #5347, opened 2025-10-07
temp fix to make eosatlas and eosmedia work

jessegeens:
We have added a special case to the `spacesLevel` function
to make eosmedia and eosatlas's spaces work. This is temporary, since we
plan to get rid of this function and just use the SpaceID that
comes from the projects database.


glpatcern (dismissed):
Yes, to be removed as soon as possible, but for now this is perfectly adequate.

glpatcern (approved):

PR #5363, opened 2025-10-14
OCM: include OCM shares in SharedByMe view

### LINKED ISSUE #5383: Parsing of OCM Address in case of more than one "@" present
If we look at this code:
https://github.com/cs3org/reva/blob/e7ca6b20cb2611f404e25cc11c419e7ba51cabb0/internal/http/services/owncloud/ocgraph/users.go#L264-L274

and if we assume that the `id = "mahdi@it-dept"`, `idp = "azadehafzar.io"`, then it would become `mahdi@it-dept@azadehafzar.io`

then in the share creation we would decompose it like:
https://github.com/cs3org/reva/blob/e7ca6b20cb2611f404e25cc11c419e7ba51cabb0/internal/http/services/owncloud/ocgraph/shares.go#L330-L342

which is done by:
https://github.com/cs3org/reva/blob/e7ca6b20cb2611f404e25cc11c419e7ba51cabb0/internal/http/services/owncloud/ocgraph/shares.go#L208-L218

SplitN with n = 2 splits at most once, producing two substrings:
- Before the first @
- Everything after the first @

If I'm not mistaken it would result into this bug:
- `username = "mahdi"`
- `idp = "it-dept@azadehafzar.io"`

Similar to #5375

rawe0:
Needs to be tested once the changes to the CS3APIs have been merged, https://github.com/cs3org/cs3apis/pull/255.

- The CS3APis verison has been updated to include "ListExistingOcmShares".
- The OCM shares are now included in the getSharedByMe call.
- The filters have been updated to adapt to changes from the updated CS3APIs.
- Fixed bug where only ocm users were queried if it was enabled.
- Consolidated OCM Address resolutions in a single function, fixes #5383

### DIFF
--- a/internal/http/services/owncloud/ocgraph/conversions.go
+++ b/internal/http/services/owncloud/ocgraph/conversions.go
@@ -552,15 +552,32 @@ func (s *svc) getDisplayNameForOCMUser(ctx context.Context, userId *userpb.UserI
 		return "Federated User"
 	}
 
+	// Pass the current user as opaque parameter. It seems the getUserFilter in internal/grpc/services/ocminvitemanager/ocminvitemanager.og
+	// is not able to return the current user and this way we can force it.
+	// TODO(lopresti) here we want a GetRemoteUser CS3 API that only requires the invitee's id
+	user := appctx.ContextMustGetUser(ctx)
+	userFilter, err := utils.MarshalProtoV1ToJSON(user.Id)
+	if err != nil {
+		log.Error().Err(err).Msg("getDisplayNameForOCMUser: failed to marshal current user id to json")
+		return "Federated User"
+	}
+
 	remoteUserRes, err := gw.GetAcceptedUser(ctx, &invitepb.GetAcceptedUserRequest{
 		RemoteUserId: userId,
+		Opaque: &types.Opaque{
+			Map: map[string]*types.OpaqueEntry{
+				"user-filter": {
+					Decoder: "json",
+					Value:   userFilter,
+				},
+			},
+		},
 	})
 	if err != nil || remoteUserRes.Status.Code != rpcv1beta1.Code_CODE_OK {
 		log.Error().Err(err).Any("response", remoteUserRes).Any("remoteUser", userId).Msg("failed to fetch OCM user")
-		return "Federated User at " + userId.Idp
+		return "Federated User"
 	}
 
-	log.Debug().Any("remoteUser", remoteUserRes.RemoteUser).Msg("found OCM user")
 	return remoteUserRes.RemoteUser.DisplayName
 }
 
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -525,9 +525,14 @@ func (s *svc) getGroupByID(ctx context.Context, g *grouppb.GroupId) (*grouppb.Gr
...
-func groupByResourceID(shares []*gateway.ShareResourceInfo, publicShares []*gateway.PublicShareResourceInfo) (map[string][]*GenericShare, map[string]*provider.ResourceInfo) {
-	grouped := make(map[string][]*GenericShare, len(shares)+len(publicShares)) // at most we have the sum of both lists
-	infos := make(map[string]*provider.ResourceInfo, len(shares)+len(publicShares))
+func groupByResourceID(shares []*gateway.ShareResourceInfo, publicShares []*gateway.PublicShareResourceInfo, ocmShares []*gateway.OCMShareResourceInfo) (map[string][]*GenericShare, map[string]*provider.ResourceInfo) {
+	var ocmSharesLen int = 0
+	if ocmShares != nil {
+		ocmSharesLen = len(ocmShares)
+	}
+
+	grouped := make(map[string][]*GenericShare, len(shares)+len(publicShares)+ocmSharesLen) // at most we have the sum of all three lists
```
glpatcern: we could put a comment for this `ocmSharesLen` use, mentioning that the OCM share provider is optional. Though we could ensure that the list is always allocated - see below

internal/http/services/owncloud/ocgraph/shares.go
```diff
@@ -601,8 +618,38 @@ func (s *svc) getSharedByMe(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
 
-	grouped, infos := groupByResourceID(shares.ShareInfos, publicShares.ShareInfos)
+	grouped, infos := groupByResourceID(
+		shares.ShareInfos,
+		publicShares.ShareInfos,
+		func() []*gateway.OCMShareResourceInfo {
+			if OCMShares == nil {
+				return nil
```
glpatcern: what about returning here an empty array? this will simplify the handling above

glpatcern (commented):
Some preliminary comments, and for now let's ignore the CI - apparently docker is suffering some outages because of a cascade effect

internal/http/services/owncloud/ocgraph/shares.go
rawe0: This is a great idea, will remove the whole need for the check, done!

glpatcern (approved):
This now works entirely

PR #5365, opened 2025-10-16
spaces: skip spaces not available

labkode:
adds new configuration option:
```
space_resolution_timeout = 4
```

### DIFF
--- a/internal/grpc/services/spacesregistry/spacesregistry.go
+++ b/internal/grpc/services/spacesregistry/spacesregistry.go
@@ -329,7 +329,7 @@ func (s *service) decorateProject(ctx context.Context, proj *provider.StorageSpa
 	token := authRes.Token
 	owner := authRes.User
 
-	ownerCtx := appctx.ContextSetToken(context.Background(), token)
+	ownerCtx := appctx.ContextSetToken(ctx, token)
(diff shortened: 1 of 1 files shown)


internal/grpc/services/spacesregistry/spacesregistry.go
glpatcern: Here the request is made with a context generated from `Background()`, so as far as I understand, the decorated `ctx` with timeout won't be taken into account, right?

internal/grpc/services/spacesregistry/spacesregistry.go
```diff
@@ -73,6 +73,7 @@ type config struct {
 	ResourceInfoCacheDrivers map[string]map[string]interface{} `mapstructure:"resource_info_caches"`
 	ResourceInfoCacheDriver  string                            `mapstructure:"resource_info_cache_type"`
 	ResourceInfoCacheTTL     int                               `mapstructure:"resource_info_cache_ttl"`
+	TimeoutSkipSpaces        int                               `mapstructure:"timeout_skip_spaces"`
```
glpatcern: May I suggest a better config name?
```suggestion
	TimeoutSkipSpaces        int                               `mapstructure:"space_resolution_timeout" docs:"nil;Timeout to resolve a space with stat: if it does not respond within the given time (defaults to 3 secs), it is skipped"`
```

PR #5375, opened 2025-10-21
Fix parsing of OCM Address in case of more than one "@" present

MahdiBaghbani:


changelog/unreleased/fix-ocm-address-parsing.md
```diff
@@ -0,0 +1,6 @@
+Bugfix: Fix parsing of OCM Address in case of more than one "@" present
+
+I've fixed the behavior for parsing , a long-standing annoyance for users who had
+OCM Address like "mahdi-baghbani@it-department@azadehafzar.io".
```
glpatcern:
```suggestion
I've fixed the behavior for parsing a long-standing annoyance for users who had
OCM Address like "mahdi-baghbani@it-department@azadehafzar.io".
```

glpatcern:
> oh well, I made a mistake, the "elsa@" doesn't throw an error

Right, and for completeness you could throw error on `id == ""` as well

PR #5385, opened 2025-10-26
feat(ocm): add wayf specific /discover and /federations endpoints to sciencemesh package

MahdiBaghbani:
This PR is a part of https://github.com/cs3org/OCM-STA/issues/1

## Description
This PR implements WAYF (Where Are You From) specific discovery endpoints for the ScienceMesh package, enabling dynamic OCM provider discovery and federation management.

## Technical
Things I've added to Reva
- `/sciencemesh/federations` endpoint: Lists available federations and their associated servers
- `/sciencemesh/federations` endpoint: Enables active discovery of OCM providers and their capabilities
- Automatic `InviteAcceptDialog` discovery: Dynamically retrieves invitation handling URLs from remote providers
- Federation configuration support: Loads federation data from configurable JSON files

### Code changes
- WAYF Handler `internal/http/services/sciencemesh/wayf.go`: Core logic for federation management and provider discovery
- Updated routing `internal/http/services/sciencemesh/sciencemesh.go`: Added new endpoints and configuration options
- OCM Discovery 1.2 compliance `internal/http/services/wellknown/ocm.go`: Changed the default path of `InviteAcceptDialog` for the new GUI chnages

### DIFF
--- a/internal/http/services/sciencemesh/wayf.go
+++ b/internal/http/services/sciencemesh/wayf.go
@@ -45,6 +45,22 @@ type DiscoverResponse struct {
 	InviteAcceptDialog string `json:"inviteAcceptDialog"`
 }
 
+// makeAbsoluteURL takes a base URL and a path/URL and returns an absolute URL.
+// If dialogURL is already absolute (has scheme and host), it returns it as-is.
+// Otherwise, it joins the dialogURL with the baseURL to create an absolute URL.
+func makeAbsoluteURL(baseURL, dialogURL string) (string, error) {
+	if dialogURL == "" {
+		return "", nil
+	}
+
+	parsed, err := url.Parse(dialogURL)
+	if err == nil && parsed.Scheme != "" && parsed.Host != "" {
+		return dialogURL, nil
+	}
+
+	return url.JoinPath(baseURL, dialogURL)
+}
+
 func (h *wayfHandler) init(c *config) error {
 	log := appctx.GetLogger(context.Background())
 
@@ -110,19 +126,21 @@ func (h *wayfHandler) init(c *config) error {
 
 			inviteDialog := disco.InviteAcceptDialog
 
-			// If it's a relative path (starts with /), make it absolute
-			if inviteDialog != "" && inviteDialog[0] == '/' {
-				baseURL, parseErr := url.Parse(srv.URL)
-				if parseErr == nil {
-					inviteDialog = baseURL.Scheme + "://" + baseURL.Host + inviteDialog
-					log.Debug().Str("original", disco.InviteAcceptDialog).Str("converted", inviteDialog).Msg("Converted relative path to absolute")
-				} else {
-					log.Debug().Err(parseErr).
+			if inviteDialog != "" {
+				absoluteURL, err := makeAbsoluteURL(srv.URL, inviteDialog)
+				if err != nil {
+					log.Debug().Err(err).
+						Str("federation", directoryService.Federation).
+						Str("server", srv.DisplayName).
 						Str("url", srv.URL).
 						Str("inviteDialog", disco.InviteAcceptDialog).
-						Msg("Failed to parse server URL for relative path conversion")
+						Msg("Failed to construct absolute URL, skipping server")
 					continue
 				}
+				if absoluteURL != inviteDialog {
+					log.Debug().Str("original", inviteDialog).Str("absolute", absoluteURL).Msg("Converted to absolute URL")
+				}
+				inviteDialog = absoluteURL
 			}
 
 			validServers = append(validServers, ocmd.DirectoryServiceServer{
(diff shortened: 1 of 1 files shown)


internal/http/services/sciencemesh/sciencemesh.go
```diff
@@ -69,6 +69,7 @@ type config struct {
 	SubjectTemplate  string                      `mapstructure:"subject_template"`
 	BodyTemplatePath string                      `mapstructure:"body_template_path"`
 	OCMMountPoint    string                      `mapstructure:"ocm_mount_point"`
+	FederationsFile  string                      `mapstructure:"federations_file"`
```
glpatcern: A mini-RFC: could you make this a `FederationsURL`, to be read via HTTP GET as opposed to be a file?

internal/http/services/sciencemesh/sciencemesh.go
```diff
@@ -102,13 +106,20 @@ func (s *svc) routerInit() error {
...
+
 	s.router.Post("/generate-invite", tokenHandler.Generate)
 	s.router.Get("/list-invite", tokenHandler.ListInvite)
 	s.router.Post("/accept-invite", tokenHandler.AcceptInvite)
 	s.router.Get("/find-accepted-users", tokenHandler.FindAccepted)
 	s.router.Delete("/delete-accepted-user", tokenHandler.DeleteAccepted)
 	s.router.Get("/list-providers", providersHandler.ListProviders)
 	s.router.Post("/open-in-app", appsHandler.OpenInApp)
+	s.router.Get("/federations", wayfHandler.GetFederations)
+	s.router.Post("/discover", wayfHandler.DiscoverProvider)
```
glpatcern, line 124: I'm missing the frontend still, but what is this endpoint for?

internal/http/services/sciencemesh/wayf.go
```diff
@@ -0,0 +1,256 @@
...
+	Federation string             `json:"federation"`
+	Servers    []FederationServer `json:"servers"`
+}
+
+// FederationServer represents a single provider with discovery info
+type FederationServer struct {
+	DisplayName        string `json:"displayName"`
+	URL                string `json:"url"`
+	InviteAcceptDialog string `json:"inviteAcceptDialog,omitempty"`
+}
```
glpatcern: As this is in the official OCM specs, I think it would be better to move those definitions in `internal/http/service/sopencloudmesh/ocmd/spec.go`, even if they are "consumed" here.

internal/http/services/sciencemesh/sciencemesh.go
MahdiBaghbani: The federations url returns a list of federations with their multiple servers and  accept invite dialog urls for each of them so the UI can show on the WAYF page and redirect user when clicked on any of them


The discover endpoint is used for manual entry in the WAYF page, user inserts his home EFSS, ui asks backend to query that EFSS and return its accept invite url to ui in order to redirect user to their home EFSS

internal/http/services/sciencemesh/sciencemesh.go
MahdiBaghbani: IMO this needs a bit more design discussion, if we are planning not to rely on any centralized mechanism then everyone should host their own list of federations and each federation contains multiple providers.

In case of a URL, you are restricted to only one federation, Imagine EOSC federation hosts their list of providers in eosc.tld, and SUNET for example is a part of that. 

However SUNET also wants to be on a Swedish universities federation as well, how can we achieve that?

Since I tried to keep it similar too the Nextcloud implementation, @mickenordin did I get it right?

mickenordin: IN Nextcloud we set a configuration option that is a space delimited list of url:s that we then loop through. We also cache the results, and only get the url:s in a scheduled job every five minutes.

MahdiBaghbani: Now I understand! Thanks!

internal/http/services/sciencemesh/wayf.go
MahdiBaghbani: Yes, We'll do it soon

internal/http/services/sciencemesh/sciencemesh.go
glpatcern: That's indeed a good idea, a list of URLs would cover all cases. For now you can leave the caching and refreshing outside - I'd do that every ~24h or so anyway, not very often.

internal/http/services/wellknown/ocm.go
```diff
@@ -128,13 +128,13 @@ func (h *wkocmHandler) init(c *OcmProviderConfig) {
 		ShareTypes: []string{"user"}, // so far we only support `user`
 		Protocols:  rtProtos,         // expose the protocols as per configuration
 	}}
-	// for now we hardcode the capabilities, as this is currently only advisory
+	// for now, we hardcoded the capabilities, as this is currently only advisory
 	d.Capabilities = []string{"invites", "webdav-uri", "protocol-object"}
```
glpatcern:
```suggestion
	d.Capabilities = []string{"invites", "webdav-uri", "protocol-object", "invite-wayf"}
```

internal/http/services/wellknown/ocm.go
```diff
@@ -128,13 +128,13 @@ func (h *wkocmHandler) init(c *OcmProviderConfig) {
...
 	d.InviteAcceptDialog, _ = url.JoinPath(c.Endpoint, c.InviteAcceptDialog)
 	h.data = d
 }
 
-// This handler implements the OCM discovery endpoint specified in
+// Ocm This handler implements the OCM discovery endpoint specified in
```
glpatcern:
```suggestion
// This handler implements the OCM discovery endpoint specified in
```

internal/http/services/sciencemesh/wayf.go
```diff
@@ -0,0 +1,231 @@
...
+		reqres.WriteError(w, r, reqres.APIErrorNotFound,
+			fmt.Sprintf("Provider at '%s' does not support OCM discovery", req.Domain), err)
+		return
+	}
+
+	inviteDialog := disco.InviteAcceptDialog
+	if inviteDialog != "" && inviteDialog[0] == '/' {
+		baseURL, _ := url.Parse(domain)
+		inviteDialog = baseURL.Scheme + "://" + baseURL.Host + inviteDialog
+	}
```
glpatcern, line 241: Here you can use standard libraries, for this case it would be `inviteDialog, err := url.JoinPath(domain, disco.InviteAcceptDialog)`

MahdiBaghbani: Nice catch, I completely missed that!

I'll fix it now.

MahdiBaghbani: I've fixed it in two locations.

glpatcern (approved):
Looks good, looking forward to the corresponding web frontend code!

PR #5402, opened 2025-11-13
Refactoring of the GORM model for shares

glpatcern:
With this PR we introduce new constraints and rename some fields for better consistency:

* Types used by OCM structures only are prefixed with
  `Ocm`, and `AccessMethod` and `Protocol` were
  consolidated into `OcmProtocol`
* ItemType is used in OCM shares as well
* The `(FileIdPrefix, ItemSource)` tuple is now
  `(Instance, Inode)` in `OcmShare`, and it was
  removed from `OcmReceivedShare` as unused
* Unique index constraints have been created for regular `Shares`
and for `OcmShares` on `(instance, inode, shareWith, deletedAt)`
* The unique indexes have been renamed with a `u_`
  prefix for consistency: this affected `u_shareid_user`,
  `u_link_token`. The `i_share_with` was dropped
  as redundant.
* `Alias` and `Hidden` were added in `OcmReceivedShare`

### DIFF
--- a/pkg/share/manager/sql/model/model.go
+++ b/pkg/share/manager/sql/model/model.go
@@ -118,7 +118,7 @@ type ProtoShare struct {
 }
 
 // Share is a regular share between users or groups. The unique index ensures that there
-// can only be one share per (inode, instance, permissions, recipient) tuple, unless the share is deleted.
+// can only be one share per (inode, instance, recipient) tuple, unless the share is deleted.
 type Share struct {
 	ProtoShare
 	DeletedAt         gorm.DeletedAt `gorm:"uniqueIndex:u_share"`


glpatcern:
As we've seen with Rasmus, everything works but the indexes are not created the way we want, so this requires a bit more work before going ahead.

pkg/share/manager/sql/model/model.go
```diff
@@ -130,111 +108,108 @@ type BaseModel struct {
...
+// ProtoShare contains fields that are common between PublicLinks and Shares.
 type ProtoShare struct {
-	// Including gorm.Model will embed a number of gorm-default fields
 	BaseModel
 	UIDOwner     string   `gorm:"size:64"`
-	UIDInitiator string   `gorm:"size:64;index"`
+	UIDInitiator string   `gorm:"size:64"`
 	ItemType     ItemType `gorm:"size:16;index"` // file | folder | reference | symlink
 	InitialPath  string
-	Inode        string `gorm:"size:32;index"`
```
jessegeens, line 143: I would still leave the index here, purely for performance reasons. Now you overwrite these fields in the `Share` model, but not in the `PublicLink` model, which means that public links will not be able to use indices for queries on these fields anymore!

pkg/share/manager/sql/model/model.go
```diff
@@ -130,111 +108,108 @@ type BaseModel struct {
...
 }
 
+// Share is a regular share between users or groups. The unique index ensures that there
+// can only be one share per (inode, instance, permissions, recipient) tuple, unless the share is deleted:
+// for that, we redeclare the corresponding fields for GORM to define the unique index.
 type Share struct {
 	ProtoShare
-	ShareWith         string `gorm:"size:255;index:i_share_with"` // 255 because this can be a lw account, which are mapped from email addresses / ...
+	DeletedAt         gorm.DeletedAt `gorm:"uniqueIndex:u_share"`
+	Inode             string         `gorm:"size:32;uniqueIndex:u_share"`
```
jessegeens, line 125: In general I'm also maybe not the biggest fan of duplicating all of this again. 
Perhaps we can use something like this: https://gorm.io/docs/indexes.html#Shared-composite-indexes ?

pkg/share/manager/sql/model/model.go
glpatcern: Agreed!

pkg/share/manager/sql/model/model.go
glpatcern: Now that's the pain point - I had tried the other approach, with all indexes declared in the `BaseModel`, but eventually GORM generates all indexes in all tables. The composite indexes also do not seem to help because we never "instantiate" the `BaseModel` as a table. Checking for alternatives it really seems you have to repeat the columns you want to put an index on, as in the comment above.

changelog/unreleased/ocm-model-refactor.md
```diff
@@ -0,0 +1,24 @@
...
+some fields for better consistency:
+
+* Types used by OCM structures only are prefixed with
+  `Ocm`, and `AccessMethod` and `Protocol` were
+  consolidated into `OcmProtocol`
+* ItemType is used in OCM shares as well
+* `Type` has been renamed to `RecipientType` with
+  type `OcmShareType`, and converted both ways
+* Unique index constraints have been created as follows:
+  * For `Shares` on `(inode, instance, permissions, recipient, deletedAt)`
```
diocas: Why are permissions part of the index? One recipient should have at most 1 share, so it's not dependent on the permissions.

pkg/share/manager/sql/model/model.go
diocas: Why don't we remove the duplications from the base class?

pkg/share/manager/sql/model/model.go
```diff
@@ -144,97 +120,99 @@ type ProtoShare struct {
...
+	Instance          string         `gorm:"size:32;uniqueIndex:u_share"`
+	Permissions       uint8          `gorm:"uniqueIndex:u_share"`
```
diocas: As mentioned above, don't understand why we add permissions to the index.

pkg/share/manager/sql/model/model.go
```diff
@@ -144,97 +120,99 @@ type ProtoShare struct {
...
+	ShareWith         string         `gorm:"size:255;uniqueIndex:u_share"` // 255 because this can be an external account, which has a long representation
 	SharedWithIsGroup bool
 	Description       string `gorm:"size:1024"`
 }
 
+// PublicLink is a public link share.
+// TODO(lopresti) We could enforce a unique index on (UIDInitiator, Inode, Permissions, DeleteAt) but for now web allows
```
diocas: This is a feature we want (I might want to create different links to share with different people, or being able to switch permissions without worrying about another link already created with the same). So, I would remove this todo...

changelog/unreleased/ocm-model-refactor.md
glpatcern: Right! Will fix

pkg/share/manager/sql/model/model.go
glpatcern: > Why don't we remove the duplications from the base class?

Because not everything is duplicated, and many fields are used elsewhere (e.g. `PublicLink`), so it makes sense to keep the common fields there.

pkg/share/manager/sql/model/model.go
glpatcern: Yes, agreed. In fact the tables were "good" but the structs had duplicated fields. Now there's a single ancestor and everything that requires a unique index is pushed down.

pkg/share/manager/sql/model/model.go
```diff
@@ -29,212 +28,184 @@ import (
...
-	Inode        string `gorm:"size:32;index"`
-	Instance     string `gorm:"size:32;index"`
 	Permissions  uint8
 	Orphan       bool
-	Expiration   datatypes.NullTime
+	Expiration   datatypes.NullTime `gorm:"index"`
 }
 
+// Share is a regular share between users or groups. The unique index ensures that there
+// can only be one share per (inode, instance, permissions, recipient) tuple, unless the share is deleted.
```
diocas:
```suggestion
// can only be one share per (inode, instance, recipient) tuple, unless the share is deleted.
```

diocas (dismissed):
Small typo and it's ready to merge.

pkg/share/manager/sql/model/model.go
```diff
@@ -29,212 +28,184 @@ import (
...
+	ProtoShare
+	DeletedAt     gorm.DeletedAt     `gorm:"uniqueIndex:u_ocmshare"`
+	Inode         string             `gorm:"size:64;not null;uniqueIndex:u_ocmshare"`
+	Instance      string             `gorm:"size:64;not null;uniqueIndex:u_ocmshare"`
+	Token         string             `gorm:"size:255;not null;uniqueIndex:u_ocmshare_token"`
+	Name          string             `gorm:"type:text;not null"`
+	ShareWith     string             `gorm:"size:255;not null;uniqueIndex:u_ocmshare"`
+	Owner         string             `gorm:"size:255;not null"`
+	Initiator     string             `gorm:"type:text;not null"`
+	Ctime         uint64             `gorm:"not null"`
```
jessegeens, line 171: What's the difference between `Ctime` / `Mtime` and `CreatedAt` / `UpdatedAt`? Are these the `Ctime` / `Mtime`s of the underlying resource? If so maybe put a comment there explaining this, otherwise remove

jessegeens (approved):

pkg/share/manager/sql/model/model.go
glpatcern: Yep, this is also to be fixed in a later PR, where we should drop those properties in favor of the GORM embedded ones. Likewise with the repeated `Initiator` and `Owner`.

PR #5404, opened 2025-11-18
add: LocalFS virtual namespace support

MahdiBaghbani:
Add optional VirtualHomeTemplate config to expose localhome paths in a virtual namespace (e.g., `/home/user/`) while storing files in the existing flat layout. When empty, preserves original behavior for EOS and other deployments.

Fixes `PathToSpaceID()` encoding issues where localhome returned paths like `/file.txt` instead of `/home/user/file.txt`, causing each file to be treated as its own space.

### DIFF
--- a/internal/grpc/services/gateway/storageprovider.go
+++ b/internal/grpc/services/gateway/storageprovider.go
@@ -479,12 +479,6 @@ func (s *svc) Unlock(ctx context.Context, req *provider.UnlockRequest) (*provide
 func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.StatResponse, error) {
 	log := appctx.GetLogger(ctx)
 	
-	log.Debug().
-		Str("path", req.Ref.GetPath()).
-		Str("storage_id", req.Ref.GetResourceId().GetStorageId()).
-		Str("opaque_id", req.Ref.GetResourceId().GetOpaqueId()).
-		Msg("gateway: Stat request")
-	
 	providers, err := s.findProviders(ctx, req.Ref)
 	if err != nil {
 		return &provider.StatResponse{
@@ -493,16 +487,8 @@ func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.St
 	}
 	providers = getUniqueProviders(providers)
 
-	log.Debug().
-		Int("num_providers", len(providers)).
-		Msg("gateway: found storage providers")
-
 	resPath := req.Ref.GetPath()
 	if len(providers) == 1 && (utils.IsRelativeReference(req.Ref) || resPath == "" || strings.HasPrefix(resPath, providers[0].ProviderPath)) {
-		log.Debug().
-			Str("provider_address", providers[0].Address).
-			Str("provider_path", providers[0].ProviderPath).
-			Msg("gateway: routing Stat to storage provider")
 		
 		c, err := s.getStorageProviderClient(ctx, providers[0])
 		if err != nil {
--- a/internal/http/services/owncloud/ocdav/meta.go
+++ b/internal/http/services/owncloud/ocdav/meta.go
@@ -49,11 +49,6 @@ func (h *MetaHandler) Handler(s *svc) http.Handler {
 			return
 		}
 
-		log.Debug().
-			Str("raw_id", id).
-			Str("url_path", r.URL.Path).
-			Msg("meta: received resource ID from Web UI")
-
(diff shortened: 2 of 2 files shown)


MahdiBaghbani:
I've also cherrypicked the two fixes into this since I need all of them to rebuild my docker image

MahdiBaghbani:
I just realized (during testing) that the user can create files and folder, cannot rename or delete

Doesn't affect the sharing tests, but better to be fixed later

glpatcern:
> I just realized (during testing) that the user can create files and folder, cannot rename or delete

Shouldn't this be fixed in this PR, so we can merge it without regressions?

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -478,6 +478,13 @@ func (s *svc) Unlock(ctx context.Context, req *provider.UnlockRequest) (*provide
 
 func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.StatResponse, error) {
 	log := appctx.GetLogger(ctx)
+
```
glpatcern, line 481: I'd prefer to remove this log, we have it in the storageprovider already and `Stat` is called way too often

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -486,8 +493,17 @@ func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.St
 	}
 	providers = getUniqueProviders(providers)
 
+	log.Debug().
+		Int("num_providers", len(providers)).
```
glpatcern: Same here

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -486,8 +493,17 @@ func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.St
...
+		Msg("gateway: found storage providers")
+
 	resPath := req.Ref.GetPath()
 	if len(providers) == 1 && (utils.IsRelativeReference(req.Ref) || resPath == "" || strings.HasPrefix(resPath, providers[0].ProviderPath)) {
+		log.Debug().
```
glpatcern: and here

internal/grpc/services/gateway/storageprovider.go
```diff
@@ -499,6 +515,14 @@ func (s *svc) Stat(ctx context.Context, req *provider.StatRequest) (*provider.St
 			log.Error().Err(err).Msg("Failed to stat " + resPath)
 			return rsp, err
 		}
+		
+		log.Debug().
```
glpatcern, line 505: maybe let's leave just this one only

internal/http/services/owncloud/ocdav/meta.go
```diff
@@ -38,13 +39,21 @@ func (h *MetaHandler) init(c *Config) error {
...
+		log := appctx.GetLogger(ctx)
+		
 		var id string
 		id, r.URL.Path = router.ShiftPath(r.URL.Path)
 		if id == "" {
 			http.Error(w, "400 Bad Request", http.StatusBadRequest)
 			return
 		}
 
+		log.Debug().
```
glpatcern: Remove this

internal/http/services/owncloud/ocdav/meta.go
```diff
@@ -56,10 +65,21 @@ func (h *MetaHandler) Handler(s *svc) http.Handler {
 			}
 		}
 
+		log.Debug().
```
glpatcern, line 63: And let's keep this

internal/http/services/owncloud/ocdav/meta.go
```diff
@@ -56,10 +65,21 @@ func (h *MetaHandler) Handler(s *svc) http.Handler {
...
+			Str("storage_id", rid.StorageId).
+			Str("space_id", rid.SpaceId).
+			Str("opaque_id", rid.OpaqueId).
+			Msg("meta: parsed resource ID")
+
 		var head string
 		head, r.URL.Path = router.ShiftPath(r.URL.Path)
 		switch head {
 		case "v":
+			log.Debug().
```
glpatcern: This is also a bit redundant, let's remove it

internal/grpc/services/gateway/storageprovider.go
MahdiBaghbani: removed

glpatcern (approved):

PR #5408, opened 2025-11-21
modernize codebase to follow latest go idioms

jessegeens:
This PR [modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize)s the codebase: it removes syntax that used to be idiomatic but now has better alternatives


pkg/auth/manager/machine/machine.go
```diff
@@ -112,12 +113,7 @@ func (m *manager) Authenticate(ctx context.Context, user, secret string) (*userp
 }
 
 func contains(lst []string, s string) bool {
```
diocas: Maybe we could just remove this and replace it directly with the slices.contains.

pkg/storage/registry/dynamic/routingtree/routingtree_test.go
```diff
@@ -108,7 +108,6 @@ var _ = Describe("Routing Tree", func() {
 
 	Context("resolving providers", func() {
 		for r, ps := range leaf {
-			r := r
 			ps := ps
```
diocas: Pretty sure these would've been found if we ran modernize again... Maybe we should run it more than once, actually.

pkg/utils/maps/maps.go
```diff
@@ -18,17 +18,15 @@
 
 package maps
 
+import "maps"
+
 // Merge returns a map containing the keys and values from both maps.
 // If the two maps share a set of keys, the result map will contain
 // only the value of the second map.
 func Merge[K comparable, T any](m, n map[K]T) map[K]T {
```
diocas, line 26: Should we also remove this one and use copy wherever we call it?

pkg/storage/registry/dynamic/routingtree/routingtree_test.go
glpatcern: We could actually inject it in the CI :-P

pkg/utils/maps/maps.go
jessegeens: Hmm but at least this one offers some extra functionality, so this one I would keep

glpatcern: As a matter of fact this is never used in the code base, maybe it was in the past. But OK I'm not against keeping it.

PR #5409, opened 2025-11-21
Bring back proper DAV support

jessegeens:
Spaces broke proper DAV support, because returned `href`s in the `PROPFIND` always contained space IDs, even if these were not present in the incoming request. This is fixed now, by writing the href based in the incoming URL

### DIFF
--- a/internal/http/services/owncloud/ocdav/propfind.go
+++ b/internal/http/services/owncloud/ocdav/propfind.go
@@ -43,6 +43,7 @@ import (
 	"github.com/cs3org/reva/v3/internal/http/services/owncloud/ocs/conversions"
 	"github.com/cs3org/reva/v3/pkg/appctx"
 	"github.com/cs3org/reva/v3/pkg/spaces"
+	"github.com/pkg/errors"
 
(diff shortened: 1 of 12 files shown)


diocas (approved):
No remarks, just happiness

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -553,43 +514,29 @@ func (s *svc) isOpenable(path string) bool {
...
+	href := ref
+	if parent != nil {
+		relativePath, err := filepath.Rel(parent.Path, md.Path)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to calculate path relative to parent: %v", md.Path)
 		}
-	} else {
-		md.Path = strings.TrimPrefix(md.Path, ns)
-
-		if ocm, _ := ctx.Value(ctxOCM).(bool); ocm {
```
glpatcern, line 581: Where do we do now the OCM-specific processing?

glpatcern: Confirmed that OCM works, so that special stuff is indeed not needed!

Issue #5417, opened 2025-11-30
OCM invite token expiration comparison is wrong


MahdiBaghbani:
https://github.com/cs3org/reva/blob/8855f78744fbabcb610a331517dcc3e5a957c5ba/pkg/ocm/invite/repository/json/json.go#L171-L173

The comparison is backwards! It says token is expired if Expiration > Now().
But that's the opposite, if expiration is in the future, the token is not expired.

PR #5418, opened 2025-11-30
Bugfix: Fix inverted expiry check in JSON invite repository

MahdiBaghbani:
Fixes: #5417


glpatcern (approved):
Well spotted Mahdi, I wonder how we could make invites work last week

PR #5420, opened 2025-12-01
User auth for user attrs

jessegeens:
Use user's own auth for setting arbitrary metadata, instead of relyin…g on the cbox user


jessegeens:
Unfortunately, doesn't work for projects: EOS does not allow setting user attributes for files not owned by you

PR #5428, opened 2025-12-09
refactor permissions into a first-class citizen in Reva

jessegeens:
Permissions are now, at least partially, handled and exposed within a single package (which was important for cernboxcop), with conversions between the different types of permissions

More improvements to follow later hopefully


pkg/share/manager/sql/ocm_shares.go
```diff
@@ -140,8 +139,8 @@ func (m *mgr) StoreShare(ctx context.Context, s *ocm.Share) (*ocm.Share, error)
 func storeWebDAVAccessMethod(tx *gorm.DB, shareID uint, o *ocm.AccessMethod_WebdavOptions) error {
 	accessMethod := &model.OcmShareProtocol{
 		OcmShareID:  uint(shareID),
-		Type:        model.WebDAVProtocol,
-		Permissions: utils.SharePermToInt(o.WebdavOptions.Permissions),
+		Type:        model.WebDAVAccessMethod,
```
glpatcern:
```suggestion
		Type:        model.WebDAVProtocol,
```

glpatcern (changes requested):

PR #5429, opened 2025-12-10
cephmount: add setfacl and getfacl support

labkode:
also adds some styling fixes (go fmt).


pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -331,37 +334,40 @@ func (fs *cephmountfs) fileAsResourceInfo(path string, info os.FileInfo, mdKeys
...
+			Move:                 true,
+			CreateContainer:      true,
+			Delete:               true,
+			PurgeRecycle:         true,
+
+			AddGrant:    true,
+			RemoveGrant: true,
+			UpdateGrant: true,
+			DenyGrant:   true,
 		},
```
glpatcern, line 371: I know this was already there, but how do you know that the user has all Grant-related permissions? (goes back to today's discussions!)

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,286 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+		perms.ListGrants = true
+	}
+
+	if strings.Contains(aclStr, "w") {
+		perms.CreateContainer = true
+		perms.Delete = true
+		perms.InitiateFileUpload = true
+		perms.Move = true
+		perms.PurgeRecycle = true
+		perms.RestoreFileVersion = true
```
glpatcern, line 1499: Given that we don't have versioning, shouldn't we exclude this?

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,286 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+		perms.AddGrant = true
+		perms.UpdateGrant = true
+		perms.RemoveGrant = true
+		perms.DenyGrant = true
+	}
+
+	if strings.Contains(aclStr, "x") {
+		perms.ListRecycle = true
+		perms.ListContainer = true
+		perms.ListFileVersions = true
```
glpatcern, line 1513: Same as above?

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,286 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+
+	if strings.Contains(aclStr, "w") {
+		perms.CreateContainer = true
+		perms.Delete = true
+		perms.InitiateFileUpload = true
+		perms.Move = true
+		perms.PurgeRecycle = true
+		perms.RestoreFileVersion = true
+		perms.RestoreRecycleItem = true
+		perms.AddGrant = true
```
glpatcern, line 1504:
```suggestion
		// TODO(lopresti) we should implement an additional xattr to store whether the user has
		// permissions to add/edit permissions. For now write access also grants that so this is readily
		// usable for personal spaces.
		perms.AddGrant = true
```

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,286 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+		perms.ListRecycle = true
+		perms.ListContainer = true
+		perms.ListFileVersions = true
+	}
+
+	return perms
+}
+
+// permissionsToACLString converts CS3 ResourcePermissions to ACL rwx string
+func (fs *cephmountfs) permissionsToACLString(perms *provider.ResourcePermissions) string {
```
glpatcern, line 1520: Similar comments for the versions apply here

glpatcern (changes requested):
I think it's all fine except a few minor things.

pkg/storage/fs/cephmount/cephmount.go
glpatcern: No, this is OK - we have to do that at the level of "storage capabilities"

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -866,78 +864,134 @@ func (fs *cephmountfs) RemoveGrant(ctx context.Context, ref *provider.Reference,
...
-	// List all grant-related extended attributes on user's thread
-	attrs, err := fs.listXattrsAsUser(ctx, path)
+	fullPath := filepath.Join(fs.chrootDir, path)
+
+	// Use getfacl to read ACLs
+	cmd := exec.CommandContext(ctx, "getfacl", "--omit-header", "--numeric", fullPath)
+	output, err := cmd.CombinedOutput()
 	if err != nil {
-		return nil, errors.Wrap(err, "cephmount: failed to list xattrs")
+		// No ACLs or error - return empty list
```
jessegeens, line 908: Shall we log something here in case there is an actual error? (not necessarily at the `err` level)

jessegeens: or can you check the exit code?

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,289 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+	fullPath := filepath.Join(fs.chrootDir, path)
+
+	// Check if it's a directory for recursive flag
+	info, err := os.Stat(fullPath)
+	if err != nil {
+		return errors.Wrap(err, "cephmount: failed to stat path")
+	}
+
+	// Build ACL removal entry string based on grantee type
+	var aclEntry string
```
jessegeens: Can you not factor out the acl building (since you do this for both the add and remove grant)?

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1258,6 +1323,289 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
...
+	} else {
+		result += "-"
+	}
+
+	return result
+}
+
+// permToInt converts ResourcePermissions to rwx bits
+func permToInt(rp *provider.ResourcePermissions) (result uint16) {
+	if rp == nil {
```
jessegeens, line 1551: why is the default full permissions?

pkg/storage/fs/cephmount/cephmount.go
labkode: I put all permission as the default, the permissions are then reduced through the spaces code. Ex: even I give all permissions as "resource permissions", if you have read-only access because you're only a reader, only read-related buttons are enabled in web.

Recycle, sharing and file versions permissions are also reduced through the interceptors.

The less magic in this driver the better.

pkg/storage/fs/cephmount/cephmount.go
labkode: the error is bubbled up where it will be logged, I don't think it adds anything to re-log here as the error is wrapped already containing all possible error messages. 

if err is not nil, the error code is non-0 (meaning error). During my tests I didn't find any error code that required different handling like we do in EOS for example. If some scenarios appear in the future that I haven't tested we can always come back, but so far I didn't have that need.

pkg/storage/fs/cephmount/cephmount.go
labkode: yep, I was too lazy and just copied paste the addACl one ;) I'll do for the grantee checking.

pkg/storage/fs/cephmount/cephmount.go
labkode: I had to chose, I left it as default to avoid potential denials if permissions are nil (not checked in upstream calls). This per se does not grant any permission on the storage as is the cephfs kernel mount that checks for permissions.

PR #5443, opened 2025-12-16
Make PROPFIND response proper for PROPFINDs to files

jessegeens:


glpatcern (approved):
(Reviewed before today's lunch!)
Not sure I entirely get the refactoring of `propfindResponse()` but the fix is very clear and contained.

PR #5451, opened 2025-12-19
fix: handling special characters

rodcoffani:
Some parts of this PR are a workaround while there isn't a fix in EOS for path encoding.
These can be reverted once there is one.
There is still a known issue with the character % (which should also be solved with the fix)

- [x] fix PROPFIND response when listing a folder named with special charactrers
- [x] enable the creation of files inside this folder
- [x] enable to open files located inside this folder (to be fixed in the web)

### DIFF
--- a/pkg/eosclient/eosgrpc/eoshttp.go
+++ b/pkg/eosclient/eosgrpc/eoshttp.go
@@ -65,7 +65,6 @@ type HTTPOptions struct {
 
 	// TTL for an idle conn per transport
 	IdleConnTimeout int
-
 	// If the URL is https, then we need to configure this client
 	// with the usual TLS stuff
 	// Defaults are /etc/grid-security/hostcert.pem and /etc/grid-security/hostkey.pem
@@ -229,9 +228,11 @@ func (c *EOSHTTPClient) buildFullURL(urlpath string, auth eosclient.Authorizatio
 		}
 	}
 
+	urlpathEncoded := strings.ReplaceAll(url.PathEscape(urlpath), "%2F", "/")
+
 	fullurl := strings.TrimRight(c.opt.BaseURL, "/")
 	fullurl += "/"
-	fullurl += strings.TrimLeft(urlpath, "/")
+	fullurl += strings.TrimLeft(urlpathEncoded, "/")
 
 	if pos < 0 {
 		fullurl += "?"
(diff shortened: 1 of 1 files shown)


pkg/eosclient/eosgrpc/eoshttp.go
```diff
@@ -446,6 +446,16 @@ func (c *EOSHTTPClient) PUTFile(ctx context.Context, remoteuser string, auth eos
 				log.Error().Str("func", "PUTFile").Str("url", finalurl).Str("err", err.Error()).Msg("can't get a new location for a redirection")
 				return err
 			}
+			// FIXME: remove once the URL from 'Location' is properly encoded (with EOS 5.4)
```
jessegeens: so can we remove this now, since the fix is in homedev?

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -511,24 +508,34 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
...
 
 	var href string
+	var err error
 	if parent != nil {
-		relativePath, err := filepath.Rel(parent.Path, md.Path)
+		// Encoding the paths to be able to handle special characters
+		encodedParentPath := url.PathEscape(parent.Path)
+		encodedMdPath := url.PathEscape(md.Path)
+		encodedRelativePath := strings.TrimPrefix(encodedMdPath, encodedParentPath)
+		_, err = url.PathUnescape(encodedRelativePath)
```
jessegeens: what's the point of doing this if you don't use the result?

internal/http/services/owncloud/ocdav/propfind.go
```diff
@@ -511,24 +508,34 @@ func (s *svc) mdToPropResponse(ctx context.Context, pf *propfindXML, md *provide
...
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to calculate path relative to parent: %v", md.Path)
 		}
-		href, err = url.JoinPath(hrefBase, relativePath)
+		href = url.PathEscape(hrefBase) + encodedRelativePath
```
jessegeens: take a look at https://pkg.go.dev/net/url#URL.ResolveReference for this

pkg/eosclient/eosgrpc/eoshttp.go
rodcoffani: it depends on how fast the new version of eos is going to production, my suggestion would be to merge this and add the removal to the `eos5.4` branch

internal/http/services/owncloud/ocdav/propfind.go
rodcoffani: to check if the URL is still valid after these operations, splitting the unescaping the `encodedRelativePath` and `href` in two separate lines.
this intends to have a more precise error message instead of doing all the operations in the same line.
but it shouldn't be a problem to remove this and deal only with the `href, err = url.PathUnescape(href)` later.

jessegeens: that might be a bit cleaner, in principle it works anyway so we should never see those messages, and if this ever needs to be debugged because something breaks they can just split the steps then

PR #5458, opened 2025-12-28
Add OCM Test Suite matrix workflow for CERNBox

MahdiBaghbani:
This PR is a part of https://github.com/cs3org/OCM-STA/issues/3
This PR is a part of https://github.com/cs3org/OCM-STA/issues/4

The document for this workflow can be found on [OCM Test Suite](https://github.com/cs3org/ocm-test-suite/blob/fabfce69fff106485331b1e288313d01fc304131/docs/vendor-ci/vendor-ci-reva.md)

A demo of the run can be found in the [GitHub Actions](https://github.com/MahdiBaghbani/reva/actions/runs/20552747067), you can also download and review the videos generated by Cypress from artifact section.

This pull request adds a dedicated GitHub Actions workflow that runs the OCM Test Suite against CERNBox and Reva using a small, fixed invite link matrix.

- The workflow builds and uploads a static revad binary from the current branch, then uses that binary inside the CERNBox containers.
- A single matrix actor job runs four invite link combinations:
  - Stub v1.1.0 <-> Stub v1.1.0
  - Stub v1.1.0 <-> CERNBox (current Reva build)
  - CERNBox (current Reva build) <-> Stub v1.1.0
  - CERNBox (current Reva build) <-> CERNBox (current Reva build)
- Each matrix entry prepares a small dev stock root, pulls the required Docker images, mounts the Docker socket, and runs the OCM Test Suite via the dev stock image with scenario invite link.
- When a combination needs the current Reva code, the workflow downloads the revad binary artifact into a host directory and passes it into the dev stock container via REVA_BINARY_DIR.
- A summary job reads one row JSON status artifacts for each combination, writes a compact Markdown table into the GitHub Actions summary, and then deletes the status artifacts while keeping the revad binary and all Cypress videos and screenshots.
- The matrix job is configured as non blocking with continue on error, so all four combinations are attempted and OCM regressions are reported without failing the entire CI run.

This gives CERNBox and Reva a clear, repeatable invite link compatibility check between Stub and the current Reva build.

The OCM-Stub is subject to be developed in the next milestone, and this workflow would also be updated, as of now it is expected that the results for the CERNBox to Stub and Stub to CERNBox to be failure.

### DIFF
--- a/.github/workflows/ocm-test-suite.yml
+++ b/.github/workflows/ocm-test-suite.yml
@@ -1,0 +1,323 @@
+# SPDX-FileCopyrightText: 2025 CERN and Reva contributors
+# SPDX-License-Identifier: Apache-2.0
+name: OCM Test Suite
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  # Orchestrator: compute version
+  ocm-prepare:
+    name: Prepare
+    runs-on: ubuntu-latest
+    outputs:
+      current_version: ${{ steps.version.outputs.version }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          submodules: recursive
+
+      - name: Derive current Reva version
+        id: version
+        shell: bash
+        run: |
+          if [[ -f VERSION ]]; then
+            ver=$(cat VERSION | tr -d '[:space:]')
+          else
+            ver="dev"
+          fi
+          echo "version=v${ver}" >> "$GITHUB_OUTPUT"
+          echo "Detected Reva version: v${ver}"
+
+  # Build: create static revad binary from current branch
+  ocm-build-revad:
+    name: Build Reva
+    needs: ocm-prepare
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          submodules: recursive
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.25"
+          cache: true
+
+      - name: Install musl tools for static build
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ocm-test-suite.yml
```diff
@@ -0,0 +1,323 @@
...
+          cache: true
+
+      - name: Install musl tools for static build
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y musl-tools
+
+      - name: Extract gaia from DockyPody image
+        run: |
+          docker pull ghcr.io/mahdibaghbani/containers/gaia:latest
```
glpatcern, line 69: Could you please use some "upstream" references here and in the following, instead of `/mahdibaghbani/...`?

.github/workflows/ocm-test-suite.yml
```diff
@@ -0,0 +1,323 @@
...
+            requires_reva_binary: true
+            images: |
+              pondersource/cypress:latest
+              pondersource/dev-stock:latest
+              pondersource/ocmstub:v1.1.0
+              ghcr.io/mahdibaghbani/containers/idp:latest
+              ghcr.io/mahdibaghbani/containers/cernbox-revad:mahdi_fix_localhome-development
+              ghcr.io/mahdibaghbani/containers/cernbox-web:testing
+          - combo_id: cernbox-stub
+            display_name: "CERNBox to Stub"
```
glpatcern, line 138: Do we have an option to mark a failure as "warning" as opposed to full red cross? If not, it's fine to have it as is, definitely better than hiding it

glpatcern (changes requested):
This is very good work, if you could also look into the comment below I think we can merge it right away (even if the test currently fails)

.github/workflows/ocm-test-suite.yml
glpatcern: As discussed today, for now we stay with those to ease Mahdi's life with poor bandwidth when re-downloading the containers. To be moved to the cs3org some time soon.

PR #5466, opened 2026-01-09
Clean up DAV PUT

jessegeens:
* Fixed a "bug" where a variable was overwritten (though this had no impact)
* Cleaned up a method
* Removed unused function

### DIFF
--- a/internal/http/services/owncloud/ocdav/put.go
+++ b/internal/http/services/owncloud/ocdav/put.go
@@ -113,15 +113,18 @@ func (s *svc) handlePathPut(w http.ResponseWriter, r *http.Request, ns string) {
 	ctx := r.Context()
 
 	fn := path.Join(ns, r.URL.Path)
-	ref := &provider.Reference{}
+	var ref *provider.Reference
 
 	// We check if the PUT was made to a resource ID instead of a path
-	if r, ok := requestWasMadeToResourceId(ctx, fn); ok {
-		ref = r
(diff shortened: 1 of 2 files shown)


internal/http/services/owncloud/ocdav/put.go
```diff
@@ -113,16 +113,17 @@ func (s *svc) handlePathPut(w http.ResponseWriter, r *http.Request, ns string) {
...
+	}
+	sublog := appctx.GetLogger(ctx).With().Any("ref", ref).Logger()
 
 	// We check if the PUT was made to a resource ID instead of a path
-	if r, ok := requestWasMadeToResourceId(ctx, fn); ok {
-		ref = r
-	} else {
-		ref.Path = fn
+	if req, ok := requestWasMadeToResourceId(ctx, fn); ok {
+		ref = req
```
glpatcern: I'm not sure this is any better, ref is still overwritten, which now makes the log in L119 incorrect. Shouldn't we do:
```
var ref ...
if ref, ok := requestWasMadeToResourceId(); !ok {
    ref = &provider.Reference{...}
}
sublog := ...
```

jessegeens: the point is to overwrite `ref`. But before, it also overwrote `r` (which is a method parameter)! and the logline is still correct, we want it to take the new value (which it will, because it's a pointer)

PR #5470, opened 2026-01-12
add indexes to share db for fields that may be queried

jessegeens:


glpatcern (commented):
Similarly, these are the indexes for `OcmShare`

pkg/share/manager/sql/model/model.go
glpatcern: (argh, GitHub didn't take my multi-line suggestion into account!)

jessegeens:
Looks good to me!

PR #5472, opened 2026-01-13
use uid instead of username for EOS remote-user

jessegeens:


glpatcern (commented):
This is good but it will only work with EOS 5.4. Do I understand it right that for file versions in personal spaces it currently works because the username matches, and we would actually break it on EOS 5.3? In that case, we may need to introduce some `if` - and a TODO to be removed once we are on 5.4...

PR #5474, opened 2026-01-16
Support updated eosgrpc bindings for trashbin

jessegeens:


glpatcern (approved):
This looks good, I guess you'll build on this to decide whether to query the "project shared" recycle bin vs the regular one.

Again let's merge only once we get close to releasing Reva for EOS 5.4.

PR #5475, opened 2026-01-16
Feat: ownerless trashbin for EOS

jessegeens:
Depends on https://github.com/cs3org/reva/pull/5474

### DIFF
--- a/pkg/storage/utils/eosfs/eosfs.go
+++ b/pkg/storage/utils/eosfs/eosfs.go
@@ -41,8 +41,8 @@ import (
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
-	"github.com/cs3org/reva/v3/pkg/permissions"
 	"github.com/cs3org/reva/v3/pkg/appctx"
+	"github.com/cs3org/reva/v3/pkg/permissions"
 	"github.com/cs3org/reva/v3/pkg/spaces"
 
 	"github.com/cs3org/reva/v3/pkg/eosclient"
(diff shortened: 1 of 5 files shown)


pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1876,6 +1898,11 @@ func (fs *Eosfs) convert(ctx context.Context, eosFileInfo *eosclient.FileInfo) (
 		if !strings.HasPrefix(k, "sys") {
 			filteredAttrs[k] = v
 		}
+		// we also want to expose the recycle id, if set
+		// see https://eos-docs.web.cern.ch/diopside/manual/interfaces.html#recycle-bin
+		if k == "sys.forced.recycleid" {
```
glpatcern: Could we make this a constant at the top? I understand the "exception" we want to apply (and we may have other ones too)

glpatcern (changes requested):
All good to me except for a mini change:

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1561,33 +1561,45 @@ func (fs *Eosfs) EmptyRecycle(ctx context.Context) error {
...
+		// project trashbin
+		// there are two options: ownerless or service-account-owned
 		md, err := fs.GetMD(ctx, &provider.Reference{Path: basePath}, nil)
 		if err != nil {
 			return nil, err
 		}
-		if md.PermissionSet.ListRecycle {
-			auth, err = fs.getUIDGateway(ctx, md.Owner)
+		if !md.PermissionSet.ListRecycle {
+			return nil, errtypes.PermissionDenied("eosfs: user doesn't have permissions to restore recycled items")
```
diocas, line 1586:
```suggestion
			return nil, errtypes.PermissionDenied("eosfs: user doesn't have permissions to list the recycle")
```

glpatcern (approved):

pkg/storage/utils/eosfs/eosfs.go
```diff
@@ -1561,33 +1562,45 @@ func (fs *Eosfs) EmptyRecycle(ctx context.Context) error {
...
+	if !ok {
+		return nil, errtypes.PermissionDenied("no user found in context for ListRecycle")
+	}
 
-	if fs.conf.AllowPathRecycleOperations && basePath != "/" {
-		// We need to access the recycle bin for a non-home reference.
-		// We'll get the owner of the particular resource and impersonate them
-		// if we have access to it.
+	// user's own trashbin
+	if basePath == "/" {
```
diocas, line 1574: Don't we have a separate way to detect projects? Not that this is a problem right now, but this prevents us from using folder navigation inside a trashbin in the future. Also, have the impression this was being handled differently in other parts of the code.

diocas: I'm also not sure why we need this special case. If we're talking about a home trashbin, wouldn't we do the same as bellow and request _as the owner_?

jessegeens: > Don't we have a separate way to detect projects?

Yes, sure, but that's not the only case: you can also be working in someone else's home when a folder has been shared with you. So that wouldn't work here

glpatcern: To me a proper way to detect project vs personal space going forward will be to look for this `recycleid`, as all projects will have it and none of the personal spaces will have it.

PR #5476, opened 2026-01-16
OCM: removed datatx protocol, now deprecated, and added accessTypes

glpatcern:
Depends on https://github.com/cs3org/cs3apis/pull/259

In this PR the Nextcloud storage driver tests have been disabled as we're not actively supporting Nextcloud as a storage provider.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/coreos/go-oidc/v3 v3.17.0
 	github.com/creasty/defaults v1.8.0
 	github.com/cs3org/cato v0.0.0-20200828125504-e418fc54dd5e
-	github.com/cs3org/go-cs3apis v0.0.0-20251211091140-ea69016c2473
+        github.com/cs3org/go-cs3apis v0.0.0-20260130145416-2dc593dc27e7
 	github.com/dgraph-io/ristretto v0.2.0
 	github.com/dolthub/go-mysql-server v0.14.0
 	github.com/glpatcern/go-mime v0.0.0-20221026162842-2a8d71ad17a9
--- a/internal/http/services/opencloudmesh/ocmd/protocols_test.go
+++ b/internal/http/services/opencloudmesh/ocmd/protocols_test.go
@@ -57,49 +57,53 @@ func TestUnmarshalProtocol(t *testing.T) {
 			},
 		},
 		{
-			raw: `{"name":"multi","options":{},"webdav":{"sharedSecret":"secret","permissions":["read","write"],"requirements":["req"],"uri":"http://example.org"}}`,
+			raw: `{"name":"multi","options":{},"webdav":{"sharedSecret":"secret","permissions":["read","write"],"requirements":["req"],"uri":"https://example.org"}}`,
 			expected: []Protocol{
 				&WebDAV{
 					SharedSecret: "secret",
 					Permissions:  []string{"read", "write"},
 					Requirements: []string{"req"},
-					URI:          "http://example.org",
+					URI:          "https://example.org",
 				},
 			},
 		},
 		{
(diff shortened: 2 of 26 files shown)


pkg/share/manager/sql/model/model.go
```diff
@@ -218,9 +233,8 @@ type OcmReceivedShareProtocol struct {
 	Uri                string           `gorm:"size:255"`
 	SharedSecret       string           `gorm:"type:text;not null"`
 	// WebDAV and WebApp Protocol fields
-	Permissions int `gorm:"default:null"`
-	// Transfer Protocol fields
-	Size uint64 `gorm:"default:null"`
```
rawe0, line 223: How come the size field is removed?

pkg/share/manager/sql/model/model.go
```diff
@@ -82,10 +82,24 @@ const (
 	WebDAVProtocol OcmProtocol = iota
 	// WebappProtocol is the OCM `webapp` protocol.
 	WebappProtocol
-	// TransferProtocol is the OCM `datatx` protocol.
-	TransferProtocol
 	// EmbeddedProtocol is the OCM `embedded` protocol.
 	EmbeddedProtocol
+	// SshProtocol is the OCM `ssh` protocol (not currently supported).
+	SshProtocol
```
rawe0, line 88: Is there a reason we are adding this now?

internal/http/services/opencloudmesh/ocmd/protocols_test.go
```diff
@@ -109,12 +113,16 @@ func TestUnmarshalProtocol(t *testing.T) {
...
+			t.Fatalf("unexpected error. Got=%+v expected=%+v", err, tt.err)
 		}
 
 		if tt.err == "" {
 			if !protocolsEqual(got, tt.expected) {
-				t.Fatalf("result does not match with expected. got=%#v expected=%#v", got, tt.expected)
+				got_json, _ := got.MarshalJSON()
+				expected_json, _ := tt.expected.MarshalJSON()
+				//spew.Dump(got)
+				//spew.Dump(tt.expected)
```
rawe0: Remove if commented?

go.mod
```diff
@@ -153,4 +153,5 @@ go 1.25.1
 replace (
 	github.com/eventials/go-tus => github.com/andrewmostello/go-tus v0.0.0-20200314041820-904a9904af9a
 	github.com/oleiade/reflections => github.com/oleiade/reflections v1.0.1
+	github.com/cs3org/go-cs3apis => /home/lopresti/dev/cs3/CS3API/build/go-cs3apis
```
rawe0: to be removed when cs3apis are merged I assume.

internal/http/services/opencloudmesh/ocmd/protocols_test.go
glpatcern: Right, left over from the debugging, now gone

pkg/share/manager/sql/model/model.go
glpatcern: I thought to add it as placeholder, as it's part of the OCM spec.

pkg/share/manager/sql/model/model.go
glpatcern: Very good question. We removed it from the spec and essentially delegated it to the actual wire transfer, also as in case of a folder the total size may not be that useful after all.

pkg/share/manager/sql/model/model.go
rawe0: Okay

pkg/share/manager/sql/model/model.go
rawe0: Okay, makes sense!

PR #5480, opened 2026-01-19
Implement CS3 Labels API

jessegeens:


pkg/labels/sql/sql.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return &mgr{
+		c:  &c,
+		db: db,
+	}, nil
+}
+
+func (m *mgr) ListLabels(ctx context.Context) ([]string, error) {
+	return nil, nil
```
glpatcern: Shouldn't we return `favorite` here? Or even running the unique query?

glpatcern (commented):
It all looks good, just a question below and also I wonder whether to put the "favorite" string in a constant rather than hard coding the literal in several places

pkg/labels/sql/sql.go
jessegeens: Yes, we should indeed run the query here. Though this function is not called yet because the HTTP layer has no support for it yet

glpatcern (approved):

PR #5482, opened 2026-01-20
Return 403 instead of 500 when EOS gives "no permissions"

jessegeens:


pkg/eosclient/eosgrpc/file_ops.go
```diff
@@ -371,14 +370,10 @@ func (c *Client) list(ctx context.Context, auth eosclient.Authorization, dpath s
...
-			// We got an error while reading items. We return the error to the user and break off the List operation
-			// We do not want to return a partial list, because then a sync client may delete local files that are missing on the server
-			log.Error().Err(err).Str("func", "List").Int("nitems", i).Str("path", dpath).Str("got err from EOS", err.Error()).Msg("")
-			if i > 0 {
-				log.Error().Str("path", dpath).Int("nitems", i).Msg("No more items, dirty exit")
-				return nil, errors.Wrap(err, "Error listing files")
+			} else {
+				// We got an error while reading items. We return the error to the user and break off the List operation
+				// We do not want to return a partial list, because then a sync client may delete local files that are missing on the server
+				return nil, err
```
glpatcern, line 377: Don't we want to keep the error log?

PR #5515, opened 2026-02-20
cephmount: added more fail-safe cases at init time

glpatcern:


rawe0 (commented):
Looks good to me! I wonder though what is `.gitkeep` in `changelog/unreleased/.gitkeep`? is this needed?

rawe0 (approved):
Approved!

glpatcern:
> Looks good to me! I wonder though what is `.gitkeep` in `changelog/unreleased/.gitkeep`? is this needed?

That's to hopefully force git to keep the `unreleased` folder around when we tag a release and move all changelogs out

PR #5552, opened 2026-03-25
Enhancement: add OCM code-flow token exchange

MahdiBaghbani:
This PR is a part of cs3org/OCM-STA#16

With inspirations from the https://github.com/cs3org/reva/pull/5373 😉 

## Summary

- add the OCM code-flow path end to end: discovery advertises `exchange-token` and `tokenEndPoint`, `/ocm/token` exchanges the authorization code, and the new `ocmsharecode` / `ocmexchangedtoken` auth managers keep code validation separate from exchanged-token validation
- keep the security boundary clear: shares that declare `must-exchange-token` reject legacy direct-secret access, and code-flow scopes carry share and resource identity without carrying the long-lived shared secret
- include the interop fixes that came out of running the real Nextcloud/CERNBox path: `client_id` handling, `/ocm` path-boundary checks, SQL/JSON persistence parity, root-mounted DAV share recovery, preserved single-file download paths, and early rejection of malformed protocol payloads

## Why this sits in one PR

This branch started as code-flow support, but once I ran the full path end to end it became clear that the failures were chained together: discovery -> `/ocm/token` -> exchanged-token DAV access -> storage routing -> download handoff.

That is why the touches in `static`, `localfs`, `storageprovider`, `wopi`, and the JSON/SQL share backends stay here. They are not opportunistic cleanup. Each one fixes a concrete break on that path.

## Interop context

This work was validated together with a matching Nextcloud branch.

On the Nextcloud side, the paired work is already open as [enriquepablo/server#2](https://github.com/enriquepablo/server/pull/2). That branch stores exchanged bearer tokens in the dedicated access-token fields, persists their expiry, refreshes them correctly, and tightens remote classification so Reva no longer has to carry a long-lived `status.php` compatibility workaround.

On the Reva side, this PR stays focused on code flow itself and on the fixes that were needed to make the real runtime path hold up. The main idea through the branch is that share identity becomes the stable handle across layers, while the exchanged-token flow stays cleanly separate from the legacy shared-secret path.

One detail worth calling out explicitly is the root-mounted DAV compatibility path. Nextcloud mounts the discovery WebDAV root (`/remote.php/dav/ocm/`) rather than always using a share-specific URI, so Reva has to recover the canonical share identity from the authenticated exchanged-token context in that case.

`/ocm/token` stays intentionally unprotected at the route level because the exchange step authenticates through the OCM share code itself. The handler keeps that contract explicit, and the tests around it cover the failure cases that mattered in practice: empty inputs, unsupported grants, and route/discovery coupling.

## Concrete references

If you want to review this by following the actual failure points rather than reading the branch file by file, these are the places I would start with:

- token exchange and grant handling on the Reva side: [`ocmd/token.go`](https://github.com/MahdiBaghbani/reva-opencloud/blob/b6d4f1d5a3db11dfc9dc27f2a5875e1f9d3e32a3/internal/http/services/opencloudmesh/ocmd/token.go#L65-L141)
- the point where Reva now rejects direct-secret access for shares that explicitly require token exchange: [`pkg/auth/manager/ocmshares/ocmshares.go`](https://github.com/MahdiBaghbani/reva-opencloud/blob/b6d4f1d5a3db11dfc9dc27f2a5875e1f9d3e32a3/pkg/auth/manager/ocmshares/ocmshares.go#L85-L123)
- the scope checks that had to stop matching lookalike `/ocm...` paths and instead stay tied to the actual share: [`pkg/auth/scope/ocmshare.go`](https://github.com/MahdiBaghbani/reva-opencloud/blob/b6d4f1d5a3db11dfc9dc27f2a5875e1f9d3e32a3/pkg/auth/scope/ocmshare.go#L104-L196)
- the root-mounted DAV recovery path, where Reva has to reconstruct the real share identity from the exchanged-token scope: [`pkg/ocm/storage/outcoming/ocm.go`](https://github.com/MahdiBaghbani/reva-opencloud/blob/b6d4f1d5a3db11dfc9dc27f2a5875e1f9d3e32a3/pkg/ocm/storage/outcoming/ocm.go#L157-L206)
- the single-file download fix for clients that hit the DAV root instead of a share-specific path: [`internal/grpc/services/storageprovider/storageprovider.go`](https://github.com/MahdiBaghbani/reva-opencloud/blob/b6d4f1d5a3db11dfc9dc27f2a5875e1f9d3e32a3/internal/grpc/services/storageprovider/storageprovider.go#L442-L496)
- the Nextcloud sender-side fix that made it use its own advertised DAV endpoint instead of rebuilding one from the remote cloud id: [`CloudFederationFactory.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/lib/private/Federation/CloudFederationFactory.php#L43-L159)
- the Nextcloud sender-side fix that now sends real OCM `read` / `write` / `share` permissions instead of the old DAV property token: [`CloudFederationShare.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/lib/private/Federation/CloudFederationShare.php#L74-L90) and [`CloudFederationShare.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/lib/private/Federation/CloudFederationShare.php#L381-L405)
- the Nextcloud receiver-side changes from the paired PR, where exchanged tokens are stored and reused properly and where `status.php` probing stops misclassifying Reva as an ownCloud-style peer: [`apps/files_sharing/lib/External/Storage.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/apps/files_sharing/lib/External/Storage.php#L97-L130), [`apps/files_sharing/lib/External/Storage.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/apps/files_sharing/lib/External/Storage.php#L351-L423), and [`lib/private/Files/Storage/DAV.php`](https://github.com/enriquepablo/server/blob/20255cf7f6310c6010a6b26d0ee827012eb14a1b/lib/private/Files/Storage/DAV.php#L147-L214)

## Scope note

This branch is meant to cover acceptance, visibility, and read/download for the validated code-flow pairs, while the root-mounted single-file save/edit edge is intentionally left for a follow-up PR rather than folded into this one. That is also why the OCM channel discussion between you and Jorn, with Rasmus mentioned, about CephFS and the deprecation of single-file share caught my attention. I have not managed to reply on Matrix yet because my messages seem to hit a wall, but I would be interested in your take on what that means in practice.


MahdiBaghbani:
This looks quite tempting for me right now 😄 Gianmaria made a good point 3 years ago #3691
https://github.com/cs3org/reva/blob/2660cddd6b38f76bb22a50d1b307077931690eed/pkg/auth/scope/ocmshare.go#L40-L42

But maybe later on another PR (I hope)

internal/http/services/opencloudmesh/ocmd/client.go
```diff
@@ -141,7 +143,8 @@ func (c *OCMClient) NewShare(ctx context.Context, endpoint string, r *NewShareRe
 	}
 
 	log := appctx.GetLogger(ctx)
-	log.Info().Str("url", url).Str("payload", string(body)).Msg("Sending OCM share")
+	// Log whitelist metadata only; OCM payloads carry shared secrets.
```
glpatcern, line 146: We know the OCM payload carries secrets, but we need the full payload for debugging purposes. I propose to redact the secret from the body for logging purposes, and log all the rest

internal/http/services/opencloudmesh/ocmd/client.go
```diff
@@ -195,7 +198,8 @@ func (c *OCMClient) InviteAccepted(ctx context.Context, endpoint string, r *Invi
 	}
 
 	log := appctx.GetLogger(ctx)
-	log.Info().Str("url", url).Str("payload", string(body)).Msg("Sending OCM invite-accepted")
+	// Log whitelist metadata only; OCM payloads carry invite tokens.
+	log.Info().Str("url", url).Int("payload_bytes", len(body)).Msg("Sending OCM invite-accepted")
```
glpatcern, line 202: Same as above

internal/http/services/opencloudmesh/ocmd/routes.go
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package ocmd
+
+const (
```
glpatcern, line 22:
```suggestion
// OCM official routes as defined by the spec
const (
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -170,14 +175,14 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
...
-	log.Info().Any("req", createShareReq).Msg("CreateOCMIncomingShare payload")
+	log.Info().Str("resource_id", req.ProviderID).Str("sender", req.Sender).Str("resource_type", req.ResourceType).Msg("CreateOCMIncomingShare payload")
 	createShareResp, err := h.gatewayClient.CreateOCMIncomingShare(ctx, createShareReq)
 	if err != nil {
 		reqres.WriteError(w, r, reqres.APIErrorServerError, "error creating ocm share", err)
 		return
 	}
 
-	if userRes.Status.Code != rpc.Code_CODE_OK {
+	if createShareResp.Status.Code != rpc.Code_CODE_OK {
```
glpatcern, line 185: Well spotted!

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -252,63 +254,76 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
+					authType = "ocmexchangedtoken"
+				} else {
+					authType = "ocmshares"
+				}
 				mode = "bearer"
 			} else {
 				username, _, ok := r.BasicAuth()
 				if ok {
-					// OCM v1.0 (OC10 and Nextcloud) uses basic auth for carrying the shared secret,
-					// and does not pass the shareId
```
glpatcern, line 271: Please keep those comments, also above (without the TODO of course!)

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -252,63 +254,76 @@ func (h *DavHandler) Handler(s *svc) http.Handler {
...
 					token = username
-					r.URL.Path = filepath.Join("/", token, r.URL.Path)
+					relPath = strings.TrimPrefix(r.URL.Path, "/")
+					authType = "ocmshares"
 					mode = "legacy"
 				} else {
-					log.Info().Any("url", r.URL.Path).Any("headers", r.Header).Msg("unauthenticated remote OCM access")
```
glpatcern, line 276: This helps in debugging when access was not granted, let's keep the headers even if they may contain a bearer token (that did not work)

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -452,10 +467,40 @@ func handleSignatureAuth(ctx context.Context, c gatewayv1beta1.GatewayAPIClient,
...
+			return false
+		}
+		if _, err := base64.RawURLEncoding.DecodeString(p); err != nil {
+			return false
+		}
+	}
+	return true
+}
+
+// Legacy direct-secret DAV access still routes internally by token. Exchanged-
```
glpatcern, line 497: Can't we make all internal routing via `shareID`?

internal/http/services/owncloud/ocdav/dav.go
```diff
@@ -452,10 +467,40 @@ func handleSignatureAuth(ctx context.Context, c gatewayv1beta1.GatewayAPIClient,
...
-		Type:         "ocmshares",
+		Type:         authType,
 		ClientId:     ocmshare,
 		ClientSecret: token,
 	})
 }
+
+// Three non-empty base64url segments (RFC 7515 JWS Compact Serialization).
+// Classifies the bearer as an exchanged JWT vs a legacy direct secret.
+// Not a cryptographic proof, @Mahdi: I can't think of anything better at this point.
```
glpatcern, line 480: I think this is good enough to tell apart the two cases. The legacy direct secret is typically just a hash, so it will fail the first check already.

pkg/auth/manager/ocmsharecode/ocmsharecode.go
```diff
@@ -0,0 +1,185 @@
...
+				return authpb.Role_ROLE_EDITOR, "editor"
+			}
+			if p.InitiateFileDownload {
+				return authpb.Role_ROLE_VIEWER, "viewer"
+			}
+		case *ocm.AccessMethod_WebappOptions:
+			viewMode := v.WebappOptions.ViewMode
+			if viewMode == provider.ViewMode_VIEW_MODE_VIEW_ONLY ||
+				viewMode == provider.ViewMode_VIEW_MODE_READ_ONLY ||
+				viewMode == provider.ViewMode_VIEW_MODE_PREVIEW {
```
glpatcern, line 176: This is actually read-write. `PREVIEW` means you're presented with the read-only view but can switch to full editor.

glpatcern (changes requested):
OK, this is a first pass (I didn't look at the tests yet), hopefully with not too many changes.
Rasmus will also have a look soon, and as I'll be offline in the coming days, maybe we could try and schedule some time for a chat for some parts. But overall it seems a nice good piece of work!

internal/http/services/opencloudmesh/ocmd/client.go
MahdiBaghbani: Yes! I also felt I should've not removed all the stuff from here and just remove the secret

pkg/auth/manager/ocmsharecode/ocmsharecode.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, nil, err
+	case shareRes.Status.Code == rpc.Code_CODE_NOT_FOUND:
+		return nil, nil, errtypes.NotFound(shareRes.Status.Message)
+	case shareRes.Status.Code == rpc.Code_CODE_PERMISSION_DENIED:
+		return nil, nil, errtypes.InvalidCredentials(shareRes.Status.Message)
+	case shareRes.Status.Code != rpc.Code_CODE_OK:
+		return nil, nil, errtypes.InternalError(shareRes.Status.Message)
+	}
+
+	// Resolve the accepted user (same pattern as ocmshares)
```
glpatcern, line 110: Another check that could be implemented here - and that IMHO would  improve the security stance of the flow - is to compare the shareRes' remote domain with the domain (not the host) the current client request is from. They must coincide, even in the case of reverse proxies. This would prevent another hijacking path a priori

glpatcern (commented):
A comment I forgot yesterday, at least for consideration

internal/http/services/owncloud/ocdav/dav.go
MahdiBaghbani: I think this is the right direction, but my read is that it is a bit bigger than what this PR is already trying to prove.

Right now, the split is still carrying two different models at once: 

The legacy direct-secret DAV path still leans on the token-shaped routing, while the exchanged-token path moved to share ID because the JWT no longer carries the long-lived secret. So, I did not want to quietly turn this PR into a broader routing refactor while I was still trying to keep the code-flow part reviewable.

Happy to do that next though. I think it deserves a focused follow-up PR where we can move the remaining token-shaped paths more deliberately and review the edge cases on their own.

And I have to mention that the code flow basically relies on the fact that we have http-signatures which Reva still doesn't have so the follow up is definitely necessary for this PR anyways

pkg/auth/manager/ocmsharecode/ocmsharecode.go
MahdiBaghbani: I think this is a good hardening idea, but I got uncertain about doing it inline here.

The part that made me stop is that "request domain" sounds simple at first, but in practice it quickly turns into questions about Host vs forwarded host, proxy setup, normalization, and what exactly we want to compare it against on the share side. 

There are many questions here that I cannot answer myself because I do not know how CERNBox is deployed and what do you have (load balancers, nginx etc)

pkg/auth/manager/ocmsharecode/ocmsharecode.go
MahdiBaghbani: I've fixed this along with adding tests in two locations here and in the `ocmshares.go`

Hope i didn't accidentally edit that file because it just looks the same!

internal/http/services/owncloud/ocdav/dav.go
MahdiBaghbani: I checked the old version in git history and restored that comment block first.
Then I only changed the parts that would read wrong with the current code. 
So, I tried to keep the original wroding, just without making the comment misleading.

internal/http/services/owncloud/ocdav/dav.go
glpatcern: Thanks, makes sense!

pkg/auth/manager/ocmsharecode/ocmsharecode.go
glpatcern: I agree this hardening deserves some further discussion, yet I wanted to bring the comment here so we could reference it (will add a TODO in the code)

pkg/auth/manager/ocmsharecode/ocmsharecode.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	m.c = &c
+	return nil
+}
+
+// Authenticate validates an exchange code (the existing long-lived OCM WebDAV shared secret)
+// and returns the accepted user with a shareId/resource-only scope for JWT minting.
+// The generic auth interface passes OAuth client_id as the first argument. In OCM code flow
+// that identifies the receiving server, so share lookup must come from the exchanged code.
+func (m *manager) Authenticate(ctx context.Context, clientID, code string) (*userpb.User, map[string]*authpb.Scope, error) {
```
glpatcern, line 93:
```suggestion
// that identifies the receiving server, so share lookup must come from the exchanged code.
// TODO(lopresti) here we could also enforce that the domain of the remote request matches
// the domain recorded in `shareRes`, taking into account forwarded host headers etc.
func (m *manager) Authenticate(ctx context.Context, clientID, code string) (*userpb.User, map[string]*authpb.Scope, error) {
```

glpatcern (dismissed):
I think this is OK for merge. Further tests and refinements to come next.

pkg/token/manager/jwt/jwt.go
```diff
@@ -132,3 +132,17 @@ func (m *manager) DismantleToken(ctx context.Context, tkn string) (*user.User, m
 
 	return nil, nil, errtypes.InvalidCredentials("invalid token")
 }
+
+// ValidatedExpiresAt parses the token, validates it, and returns the expiration time.
+func (m *manager) ValidatedExpiresAt(_ context.Context, tkn string) (time.Time, error) {
+	token, err := jwt.ParseWithClaims(tkn, &claims{}, func(token *jwt.Token) (any, error) {
```
rawe0, line 138: Why are we returning the internal secret here when parsing the token?

pkg/auth/scope/ocmshare.go
```diff
@@ -96,28 +97,145 @@ func ocmShareScope(_ context.Context, scope *authpb.Scope, resource any, _ *zero
...
+// checkStorageRefForOCMShare reports whether ref refers to a resource inside this OCM share.
+// Flow: ref is either by ResourceId (storage opaque id) or by path. We allow only when the
+// ref clearly belongs to this share. Used at gRPC (storage) scope check; for path refs at
+// space root we get ref.Path = "/<filename>", so pathUnderOCMPrefix avoids matching "/ocm-*.txt".
 func checkStorageRefForOCMShare(s *ocmv1beta1.Share, r *provider.Reference, ns string) bool {
+	shareID := s.Id.GetOpaqueId()
+
+	// Ref by ResourceId: allow only if ref equals share resource, or ref id is scoped to this share.
 	if r.ResourceId != nil {
-		return utils.ResourceIDEqual(s.ResourceId, r.GetResourceId()) || strings.HasPrefix(r.ResourceId.OpaqueId, s.Token)
```
rawe0, line 106: Are we not removing a case here that worked before? now this would not get evaluated

```
strings.HasPrefix(r.ResourceId.OpaqueId, s.Token)
```

pkg/storage/registry/static/static.go
```diff
@@ -170,6 +171,18 @@ func (b *reg) FindProviders(ctx context.Context, ref *provider.Reference) ([]*re
 				continue
 			}
 			if m := r.FindString(fn); m != "" {
+				if !matchesPathBoundary(fn, m) {
```
rawe0, line 174: I don't quite understand the purpose of this?

rawe0 (commented):
Looks good! I have a couple of questions and would also like to test this in our dev setup to make sure it doesn't break anything :)

pkg/token/manager/jwt/jwt.go
MahdiBaghbani: Good question! 
The value is not returned to HTTP clients, and it is not returned by `DismantleToken` or `ValidatedExpiresAt`.

That callback is the [`jwt.ParseWithClaims`][jwt-keyfunc] keyfunc.
The JWT library calls it so it can verify the HS256 signature using the same secret this token manager uses when [minting the token][jwt-mint].
After that, [`DismantleToken`][jwt-dismantle-return] returns the parsed user and scopes, while [`ValidatedExpiresAt`][jwt-exp-return] returns the validated expiry time.

The reason this came up in this PR is that M6 needed `expires_in` for `/ocm/token`, but we did not want the HTTP handler to parse JWTs itself or invent a second expiry path.

So the validation stays inside the configured JWT token manager, using the normal Reva/sharedconf secret wiring through [`ApplyDefaults`][jwt-sharedconf].

There is a second `ParseWithClaims` call in [`ValidatedExpiresAt`][jwt-exp-keyfunc] for the same reason: it validates the token before reading the expiry.

[jwt-keyfunc]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L120-L123
[jwt-exp-keyfunc]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L137-L140
[jwt-mint]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L94-L97
[jwt-dismantle-return]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L129-L130
[jwt-exp-return]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L144-L145
[jwt-sharedconf]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/token/manager/jwt/jwt.go#L59-L65

pkg/auth/scope/ocmshare.go
MahdiBaghbani: Yes, that prefix case was removed intentionally.

During debugging I hit cases where prefix-style matching was too loose for OCM paths and identifiers.

I also fall upon this by complete accident since I was sharing files from Nextcloud to CERNBox with names like `ocm-nc-to-cbox.txt`, I got completely lost in there about why it is not working at all :-)

For `ResourceId.OpaqueId`, the safer rule is that the reference must clearly identify this share.

So the current code checks exact equality against the share id, and exact equality against `s.Token` only when [`s.Token`][ocmshare-resourceid] is non-empty.

That last guard matters for code flow. Exchanged-token scopes deliberately do not carry the long-lived WebDAV share secret.

You can see that in [`AddCodeFlowOCMShareScope`][ocmshare-codeflow-scope], which copies the share id and resource id but does not copy `Token`.

If `s.Token` is empty, matching against it must be skipped, otherwise an empty token can accidentally widen what the scope accepts.

The same idea is used on the path side. [`pathUnderOCMPrefix`][ocmshare-prefix] allows `/ocm` and `/ocm/...`, but it does not let a root-level file like `/ocm-file.txt` or `/ocm-m6-proof.txt` count as part of the `/ocm` namespace.

I removed the broad prefix case.
The new code still supports the real compatibility cases:
- If the reference is exactly the share resource id, it works.
- If the opaque id is exactly the share id, it works.
- If the opaque id is exactly the legacy token, it works, but only when the token exists.
- If access is path-based, `/ocm/<share-id>/...` and `/ocm/<token>/...` still work.

So token-based access is not removed. What is removed is treating any opaque id that merely starts with the token as valid.

[ocmshare-resourceid]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/auth/scope/ocmshare.go#L123-L139
[ocmshare-codeflow-scope]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/auth/scope/ocmshare.go#L229-L238
[ocmshare-prefix]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/auth/scope/ocmshare.go#L108-L117

pkg/storage/registry/static/static.go
MahdiBaghbani: Quite same as the answer to previous question but a bit different.

The purpose is to make storage-provider routing follow the same segment-boundary rule as the OCM scope checks.

One of the failures I saw during debugging was that even after tightening the scope checks, a normal user file with a name like `/ocm-file.txt` could still be routed as if it belonged under the `/ocm` provider. That was because the static registry matched provider rules by prefix, so `/ocm` was too greedy.

This change keeps the existing provider selection shape, including longest-match behavior, but filters path matches through [`matchesPathBoundary`][static-boundary-call].

So `/ocm` still matches the mount itself, and `/ocm/share-id` still routes to the OCM provider, but `/ocm-file.txt` falls back to the normal root provider.

The helper makes that rule explicit: [`/ocm` matches only `/ocm` or `/ocm/...`][static-helper].

The added test covers exactly that distinction:
[`/ocm-file.txt` does not route to `/ocm`, while `/ocm/share-id` still does][static-test].

[static-boundary-call]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/storage/registry/static/static.go#L163-L194
[static-helper]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/storage/registry/static/static.go#L219-L231
[static-test]: https://github.com/cs3org/reva/blob/d780ccedf5494f90bc075e75bdc242668593cd18/pkg/storage/registry/static/static_test.go#L242-L273

MahdiBaghbani: - `matched == ""`: An empty match is meaningless. Do not route based on it.
- `matched == "/"`: The root provider is special. It is allowed to match everything, because everything is under /.
- `path == matched`: Exact mount path is valid. So /ocm matches /ocm.
- `strings.HasPrefix(path, matched+"/")`: A child path is valid only if the next character after the match is /.

So:
```text
path = /ocm/share-id
matched = /ocm
matched + "/" = /ocm/
/ocm/share-id starts with /ocm/, so it matches.
```

But:
```text
path = /ocm-file.txt
matched = /ocm
matched + "/" = /ocm/
/ocm-file.txt does not start with /ocm/, so it does not match.
```

pkg/token/manager/jwt/jwt.go
glpatcern: Thanks, will add a comment about the context where we're called back

glpatcern, line 138:
```suggestion
	token, err := jwt.ParseWithClaims(tkn, &claims{}, func(token *jwt.Token) (any, error) {
		// here we are called back by the JWT library to verify the signature of the token, which was encrypted
		// with the secret configured in Reva to mint tokens, therefore that's the one we have to return
```

glpatcern (approved):

rawe0 (approved):
I have tested these changes after rebasing on top of the current master, the current OCM functionality was tested and I've concluded that everything works as intended (it doesn't have any side effects). 


the tests included:

1. Deployment with current master <->  Deployment with this PR (Code flow off)
2. Deployment with current master <->  Deployment with this PR (Code flow on)
3. Deployment with current master <-> Deployment with current master (For reference)

I have not tested the final combination, but this is assumed to work as this was tested by @MahdiBaghbani:

Deployment with this PR (Code flow on) <-> Deployment with this PR (Code flow on)

PR #5556, opened 2026-03-25
allow using different field as identifier, configurable per idp

jessegeens:
For having CERN users keep sub, external users use email:

```
  [grpc.services.authprovider.auth_managers.oidc]                                                                                                                                                                                                                                                                           
  default_id_claim = "email"

  [grpc.services.authprovider.auth_managers.oidc.idp_to_id_claim]                                                                                                                                                                                                                                                           
  "f3700f7d-85b5-4b29-b6b4-987522ac9ea6" = "sub"                                                                                                                                                                     
```

This PR goes together with https://github.com/cernbox/reva-plugins/pull/67

### DIFF
--- a/changelog/unreleased/idp-config-identifier.md
+++ b/changelog/unreleased/idp-config-identifier.md
@@ -1,0 +1,3 @@
+Enhancement: allow using different field as identifier, configurable per idp
+
+https://github.com/cs3org/reva/pull/5556
--- a/internal/grpc/services/authprovider/authprovider.go
+++ b/internal/grpc/services/authprovider/authprovider.go
@@ -131,6 +131,10 @@ func (s *service) Authenticate(ctx context.Context, req *provider.AuthenticateRe
 		return &provider.AuthenticateResponse{
 			Status: status.NewNotFound(ctx, "unknown client id: "+err.Error()),
 		}, nil
+	case errtypes.Conflict:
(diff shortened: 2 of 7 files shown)


internal/grpc/services/gateway/storageprovider.go
```diff
@@ -726,6 +726,18 @@ func (s *svc) splitPath(_ context.Context, p string) []string {
 	return strings.SplitN(p, "/", 4) // ["home", "MyShares", "photos", "Ibiza/beach.png"]
 }
 
+func (s *svc) AddFavorite(ctx context.Context, req *provider.AddFavoriteRequest) (*provider.AddFavoriteResponse, error) {
```
diocas, line 729: All of this should be a separate commit.

diocas (changes requested):

diocas (dismissed):
From what I could see, the provider changes with linked accounts, right? So this would work as we expected and we don't need anything else, right? (even though users might be confused).

jessegeens:
> From what I could see, the provider changes with linked accounts, right? So this would work as we expected and we don't need anything else, right? (even though users might be confused).

Yes, so your opaque id will become your cern email address and not your username. So from this side this should be all that's needed.

glpatcern (dismissed):

glpatcern (approved):

PR #5562, opened 2026-04-01
Feat/share consistency

jessegeens:
This PR adds a hierarchical checking algorithm for shares to the gateway, as defined in ADR general/0005-sharing. Concrectely, the new algorithm does the following: 

* Before applying any ACL, the gateway checks for parent and child shares in the database.
* Based on their relationships and permission levels, the gateway decides whether to apply, reapply, or reject the operation.
* ACL updates will be applied orderd by path-length (where the shortest comes first) to maintain  consistent inheritance semantics (otherwise, you would overwrite child shares).
* The algorithm applies equally to create, update, and delete operations.

### DIFF
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -117,6 +117,8 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 	}
 
 	// If we commit to the storage, first we apply to the new resource, then we clean up
+	toDelete := sharehierarchy.Shares(result.ToDelete)
+	toReapply := sharehierarchy.Shares(result.ToReapply)
 	if s.c.CommitShareToStorageGrant {
 		if grantStatus, err := s.applyGrant(ctx, req.ResourceInfo.Id, req.Grant.Grantee, req.Grant.Permissions.Permissions); err != nil || grantStatus.Code != rpc.Code_CODE_OK {
 			if err != nil {
@@ -125,7 +127,7 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 			return &collaboration.CreateShareResponse{Status: grantStatus}, nil
 		}
 		// Remove grants for child shares made redundant by the new share.
-		if st, err := s.removeChildGrants(ctx, result.ToDelete); err != nil || st.Code != rpc.Code_CODE_OK {
+		if st, err := s.removeChildGrants(ctx, toDelete); err != nil || st.Code != rpc.Code_CODE_OK {
 			if err != nil {
 				return nil, err
 			}
@@ -133,7 +135,7 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 		}
 		// Re-apply grants for children that retain higher permissions (N=R, C=RW).
 		// ToReapply is already sorted shallowest-first by CheckGrantConsistency.
-		s.reapplyChildGrants(ctx, result.ToReapply)
+		s.reapplyChildGrants(ctx, toReapply)
 	}
 
 	// Finally, we write to the db
@@ -147,7 +149,7 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
 	}
 
 	// And we remove from the db the deleted shares made redundant by the new share.
-	s.removeChildShareRecords(ctx, shareClient, result.ToDelete)
+	s.removeChildShareRecords(ctx, shareClient, toDelete)
 
 	return res, nil
 
@@ -406,8 +408,8 @@ func (s *svc) UpdateShare(ctx context.Context, req *collaboration.UpdateShareReq
 				Status: status.NewConflict(ctx, conflictErr, conflictErr.MarshalToJSON()),
 			}, nil
 		}
-		toDelete = result.ToDelete
-		toReapply = result.ToReapply
+		toDelete = sharehierarchy.Shares(result.ToDelete)
+		toReapply = sharehierarchy.Shares(result.ToReapply)
 	}
 
 	if isPermUpdate && s.c.CommitShareToStorageGrant {
--- a/internal/grpc/services/gateway/usershareprovider.go
+++ b/internal/grpc/services/gateway/usershareprovider.go
@@ -31,6 +31,8 @@ import (
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	typesv1beta1 "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/cs3org/reva/v3/pkg/appctx"
+	revashare "github.com/cs3org/reva/v3/pkg/share"
+	"github.com/cs3org/reva/v3/pkg/sharehierarchy"
 	"github.com/cs3org/reva/v3/pkg/utils/resourceid"
 
 	"github.com/cs3org/reva/v3/pkg/errtypes"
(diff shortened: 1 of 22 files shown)


pkg/sharehierarchy/hierarchy.go
```diff
@@ -0,0 +1,210 @@
...
+		switch {
+		case isStrictAncestor(path, nodePath):
+			// Step 1: existing share S is a parent of the new node N.
+			// Allowed only when N strictly escalates beyond P in the permission ordering.
+			// Any other combination means N is redundant or conflicts with P.
+			if nodePermLevel > sharePerms {
+				continue
+			}
+			return nil, errtypes.ShareParentConflict(fmt.Sprintf(
+				"resource at %q is already accessible via a %s share on parent %q",
```
glpatcern:
```suggestion
				"resource at %q is already shared as a %s share on parent %q",
```
(the "higher" share might be a denial)

glpatcern (commented):
So far so good, I didn't go to the very end though

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -73,40 +75,74 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
+	// (https://gitlab.cern.ch/cernbox/adr/-/blob/master/decisions/general/0005-sharing.md)
+	// ---------------------------------
+	spaceId := req.ResourceInfo.Id.SpaceId
+	if spaceId == "" {
+		if stat, statErr := s.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{ResourceId: req.ResourceInfo.Id}}); statErr == nil && stat.Status.Code == rpc.Code_CODE_OK && stat.Info.Id.SpaceId != "" {
+			spaceId = stat.Info.Id.SpaceId
+			req.ResourceInfo.Id.SpaceId = spaceId
+			log.Debug().Str("keyword", "sharehierarchy").Str("spaceId", spaceId).Msg("sharehierarchy: populated missing spaceId via stat")
+		} else {
+			log.Warn().Str("keyword", "sharehierarchy").Msg("sharehierarchy: spaceId missing and stat could not resolve it")
```
diocas: Is it ok to continue? If so, explain why with a comment.

I see that existing shares are empty in this case, so we skip the check (fine, I guess we don't want to do a check of all shares). But at the same time, it looks like we allow creating a share without a space (don't see where we use it, because we haven't changed the sharing column name to reflect this, right?). It's just that it looks a bit weird that we allow creating a share without a space.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -73,40 +75,74 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
-			if err != nil {
-				log.Error().Err(err).Str("ResourceInfo", req.ResourceInfo.String()).Str("Grantee", req.Grant.Grantee.String()).Str("Message", addGrantStatus.Message).Msg("Failed to Create Share: error during addGrant")
-				return nil, errors.Wrap(err, "gateway: error adding grant to storage")
+			return &collaboration.CreateShareResponse{Status: grantStatus}, nil
+		}
+		// Remove grants for child shares made redundant by the new share.
+		for _, child := range result.ToDelete {
+			if st, err := s.removeGrant(ctx, child.ResourceId, child.Grantee, child.Permissions.Permissions); err != nil || st.Code != rpc.Code_CODE_OK {
+				if err != nil {
+					return nil, errors.Wrap(err, "gateway: error removing child grant")
```
diocas: Shall we return or skip to try to remove the most that we can (and reach the re-apply section as well)?

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -73,40 +75,74 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
+			if _, err := s.addGrant(ctx, child.ResourceId, child.Grantee, child.Permissions.Permissions); err != nil {
+				log.Error().Err(err).Str("shareId", child.Id.OpaqueId).Msg("error re-applying child grant after CreateShare")
 			}
 		}
 	}
 
-	// Then we commit to the db
-	// ------------------------
+	// Finally, we write to the db
 	res, err := shareClient.CreateShare(ctx, req)
```
diocas, line 142: Would it be more consistent if we brought this closer to the call to apply the new acl to storage? We do everything we need for the current share, _then_ we do the cleanups and consistencies?

Although, keeping it here would allow the user to re-issue the same share request again...

diocas: And we should think very well what happens if any action fails. Our consistency check should include this algorithm verification to cleanup unnecessary shares.

I'm just wondering, if we move the db creation upwards, and this call fails, how bad would it be? Shall we exit? In the meantime the ACLs in EOS are all broken, so we should still try to do the consistency cleanups...

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -483,6 +634,75 @@ func (s *svc) UpdateReceivedShare(ctx context.Context, req *collaboration.Update
 	return res, nil
 }
 
+// ---------------------------------------------------------------------------
+// Hierarchy helpers (ADR-0005-P01)
+// ---------------------------------------------------------------------------
+
+// parseForce reads the "force" flag from the request opaque map.
+func parseForce(opaque *typesv1beta1.Opaque) bool {
```
diocas, line 616: why not receiving a req? In the future, if we move from an opaque to some proper attr, then it would be a single line change.

diocas: Ah, because we have different types of request?

internal/grpc/services/gateway/usershareprovider.go
diocas: Also, all of this iteration logic (to delete and to reapply) is repeated in all methods. can we extract this to aux functions? it would make the flow of this also nicer if abstracted.

pkg/share/manager/sql/model/model.go
```diff
@@ -145,7 +147,8 @@ type Share struct {
 	DeletedAt         gorm.DeletedAt `gorm:"uniqueIndex:u_share"`
 	Inode             string         `gorm:"size:32;uniqueIndex:u_share;index"`
 	Instance          string         `gorm:"size:32;uniqueIndex:u_share;index"`
-	ShareWith         string         `gorm:"size:255;uniqueIndex:u_share;index"` // 255 because this can be an external account, which has a long representation
+	SpaceId           string         `gorm:"size:64;index:idx_space_sharewith"`
```
diocas: Why not protoshare? so that we don't update pl for now? Also, do we need UIDOwner if we have this one?

internal/grpc/services/gateway/usershareprovider.go
jessegeens: Yes, can be both Create and UpdateShareRequest

internal/grpc/services/gateway/usershareprovider.go
jessegeens: maybe indeed best to exit with an error there

internal/grpc/services/gateway/usershareprovider.go
jessegeens: > Would it be more consistent if we brought this closer to the call to apply the new acl to storage? We do everything we need for the current share, then we do the cleanups and consistencies?

I'm not sure. To me it's always best to keep the db last, because once we do this the share becomes user visible. If any of the previous steps failed then usually something is wrong.

jessegeens: > I'm just wondering, if we move the db creation upwards, and this call fails, how bad would it be? Shall we exit? In the meantime the ACLs in EOS are all broken, so we should still try to do the consistency cleanups...

Yeah but then it's really in a user-visible inconsistent state. Anyway the thing that happens before the db call is setting the ACLs. So if you got there, all the hierarchy logic went OK. And then either the ACLs fail: we abort or the ACLs succeed, then we do the db call, so all should be good.

DB call can also fail but this should be really rare. Esp since we do a liveness check in the beginning (and we did all the cleanup, which also calls the db). 

Then yes, we might still need the consistency tool for these cases.

internal/grpc/services/gateway/usershareprovider.go
jessegeens: Done

internal/grpc/services/gateway/usershareprovider.go
glpatcern: I'd tend to agree with Jesse here, we have pushed the DB update at the very end in the previous refactoring as we are also "pinging" it, so this seems the most consistent way forward.

internal/grpc/services/gateway/usershareprovider.go
```diff
@@ -73,40 +75,69 @@ func (s *svc) CreateShare(ctx context.Context, req *collaboration.CreateShareReq
...
-	// -------------------------------------
+	// Hierarchy check (ADR-GENERAL-005)
+	// (https://gitlab.cern.ch/cernbox/adr/-/blob/master/decisions/general/0005-sharing.md)
+	// ---------------------------------
+	spaceId := req.ResourceInfo.Id.SpaceId
+	if spaceId == "" {
+		if stat, statErr := s.Stat(ctx, &provider.StatRequest{Ref: &provider.Reference{ResourceId: req.ResourceInfo.Id}}); statErr == nil && stat.Status.Code == rpc.Code_CODE_OK && stat.Info.Id.SpaceId != "" {
+			spaceId = stat.Info.Id.SpaceId
+			req.ResourceInfo.Id.SpaceId = spaceId
+			log.Debug().Str("keyword", "sharehierarchy").Str("spaceId", spaceId).Msg("sharehierarchy: populated missing spaceId via stat")
```
glpatcern: Is this `Str("keyword", "sharehierarchy")` a leftover?

jessegeens: Was for debugging yes, I removed it from all the logs

diocas (dismissed):

glpatcern (approved):

PR #5563, opened 2026-04-01
Redis: implement sentinel model and make it configurable

rawe0:
This PR adds the configurable option of having Redis in sentinel mode

- If this is enabled we have to discover the redis master
- Added two configurable options RedisSentinelMode (enables the discovery) and RedisMasterName (name of the master).

### DIFF
--- a/pkg/share/cache/redis/redis.go
+++ b/pkg/share/cache/redis/redis.go
@@ -20,6 +20,8 @@ package redis
 
 import (
 	"encoding/json"
+	"fmt"
+	"strings"
 	"time"
 
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/share/cache/redis/redis.go
```diff
@@ -49,6 +53,12 @@ func (c *config) ApplyDefaults() {
 	if c.RedisAddress == "" {
 		c.RedisAddress = "localhost:6379"
 	}
+	if c.RedisMasterName == "" {
+		c.RedisMasterName = "cboxmaster"
```
glpatcern, line 57:
```suggestion
		c.RedisMasterName = "master"
```

glpatcern (commented):
Apart the minor change, I don't understand why we would need to connect to the master when in sentinel mode: are the sentinels/followers not capable of receiving updates?

IMHO we should have every Reva connecting to its own local Redis and do everything as before.

pkg/share/cache/redis/redis.go
rawe0: In this setup we always have a master and replicas, and the replicas are read-only, so at least writes we will have to do to the master, we could read from the replicas though.  

Sentinel is just a layer on top of this "master-replica" setup, where if a master is unreachable it will have an election and one of the replicas is promoted to master.

rawe0: You can disable this read-only mode, but it's not recommended since it can create inconsistencies, Redis has asynchronous replication so it's just the way it was designed.

glpatcern:
As discussed, we're not doing this for the main Reva and it's already done for the specific plugins where it matters

PR #5583, opened 2026-04-15
Update cs3apis (GetTenant)

jessegeens:
Follow-up on https://github.com/cs3org/reva/pull/5580

Since it is the same change, I didn't add a changelog


diocas (dismissed):
Add link in the changelog and merge

diocas (approved):

PR #5598, opened 2026-04-30
Fix/listshare perf improvements

jessegeens:


glpatcern (commented):
Nice!

Very good that you got rid of an opaque payload: but then, the consumer of that in https://github.com/cs3org/reva/blob/master/internal/grpc/services/storageprovider/storageprovider.go#L913 is to be removed as well?

jessegeens:
> Very good that you got rid of an opaque payload: but then, the consumer of that in https://github.com/cs3org/reva/blob/master/internal/grpc/services/storageprovider/storageprovider.go#L913 is to be removed as well?

We could, the question is whether we ever want to use that again

glpatcern:
> We could, the question is whether we ever want to use that again

In that case let's comment the code? It looks weird to have a StorageProvider attempting to parse an Opaque and do something with it, which is not set anywhere else

glpatcern (approved):
I think this is all good now, the refactoring goes in the right direction. I'm only afraid of test coverage, but we have time to exercise this.

PR #5610, opened 2026-05-12
bugfix: ocm shared folders are displayed correctly

rawe0:
Fixes an issue where the received ocm share wasn't displayed properly due to the ocm sql driver still using an outdated field.

### DIFF
--- a/pkg/share/manager/sql/conversions.go
+++ b/pkg/share/manager/sql/conversions.go
@@ -142,10 +142,10 @@ func convertToCS3OCMReceivedShare(s *model.OcmReceivedShare, p []*ocm.Protocol) 
 		Mtime: &types.Timestamp{
 			Seconds: uint64(s.Mtime),
 		},
-		ResourceType: convertToCS3ResourceType(s.ItemType),
-		ShareType:    convertToCS3OCMShareType(s.RecipientType),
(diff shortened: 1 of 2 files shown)


pkg/share/manager/sql/conversions.go
```diff
@@ -142,10 +142,11 @@ func convertToCS3OCMReceivedShare(s *model.OcmReceivedShare, p []*ocm.Protocol)
 		Mtime: &types.Timestamp{
 			Seconds: uint64(s.Mtime),
 		},
-		ResourceType: convertToCS3ResourceType(s.ItemType),
-		ShareType:    convertToCS3OCMShareType(s.RecipientType),
-		State:        convertToCS3OCMShareState(s.State),
-		Protocols:    p,
+		ResourceType:       convertToCS3ResourceType(s.ItemType), // depracated
```
glpatcern: Can we get rid of this?

glpatcern (dismissed):
Looks good, only one comment

pkg/share/manager/sql/conversions.go
rawe0: I believe so, but will test

rawe0: Can be deleted, and done.

glpatcern (approved):
Great!

PR #5611, opened 2026-05-14
OCM: reworked discovery logic and payload

glpatcern:
This PR modifies the OCM discovery endpoint following the recent discussions, and **enforces** that the sender server actually serves the protocol that was sent in a share.

The latter means that we attempt to discover the sender server in all cases, not only for the legacy shares. @rawe0 This would currently break the PoC for Data Commons but we need to go there sooner or later.


internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -243,13 +243,30 @@ func getOCMShareType(st string) ocm.RecipientType {
 	}
 }
 
-func getAndResolveProtocols(ctx context.Context, p Protocols, ownerServer string) (protos []*ocm.Protocol, legacy bool, err error) {
+func getAndResolveProtocols(ctx context.Context, p Protocols, resType string, ownerServer string) (protos []*ocm.Protocol, legacy bool, err error) {
 	protos = make([]*ocm.Protocol, 0, len(p))
 	legacy = false
+
+	// discover remote resource types
+	ocmRTs, ocmEndpoint, err := discoverOcmResourceTypes(ctx, ownerServer)
```
MahdiBaghbani, line 251: I think we need to fail closed here when discovery fails.

Right now `discoverOcmResourceTypes` can return a non-nil `err` together with
mocked resource types and `ocmEndpoint == ""`, but this loop never checks
`err`.

We still enter protocol resolution and can end up at `url.Parse("")`
below, which produces a path-only URI with no scheme/host.

That would bypass the nice `protoInfo == nil` guard you added and conflicts
with the PR description that the sender's advertised protocol is enforced.

```suggestion
	ocmRTs, ocmEndpoint, err := discoverOcmResourceTypes(ctx, ownerServer)
	if err != nil {
		return nil, false, fmt.Errorf("failed to discover remote OCM server %s: %w", ownerServer, err)
	}
	if ocmEndpoint == "" {
		return nil, false, fmt.Errorf("remote OCM server %s did not advertise an endpoint", ownerServer)
	}
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -271,61 +291,67 @@ func getAndResolveProtocols(ctx context.Context, p Protocols, ownerServer string
...
+		// (relative) path or missing
+		if protocolName != "webdav" {
+			return nil, false, fmt.Errorf("invalid protocol URI: missing host for protocol %s", protocolName)
+		}
+		protoRoot, ok := protoInfo.(string)
+		if !ok {
+			return nil, false, fmt.Errorf("missing host in URI %s and root webdav path not advertised by the remote OCM server", uri)
 		}
+
+		u, _ = url.Parse(ocmEndpoint)
```
MahdiBaghbani: `url.Parse("")` yields a URL with no scheme/host, so `u.String()` becomes a
path-only string. With the mocked fallback this is reachable today.

It is better to have a check even after fixing discovery:
```suggestion
		base, err := url.Parse(ocmEndpoint)
		if err != nil || base.Scheme == "" || base.Host == "" {
			return nil, false, fmt.Errorf("invalid OCM endpoint %q from sender discovery", ocmEndpoint)
		}
		u = base
```

MahdiBaghbani (dismissed):
I see this as WIP so I've stopped reviewing further.
I left inline suggestions with concrete code changes. Happy to re-check after WIP.

internal/http/services/opencloudmesh/ocmd/shares.go
glpatcern: Thanks Mahdi, indeed I'm not happy with this and I'm still considering ways to remove the mocked stuff by adapting the tests

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -310,13 +302,30 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
+	ocmEndpoint, err := ocmim.GetOCMEndpoint(req.RecipientMeshProvider)
+	if err != nil || ocmEndpoint == "" {
 		return &ocm.CreateOCMShareResponse{
 			Status: status.NewInvalidArg(ctx, "the selected provider does not have an OCM endpoint"),
 		}, nil
 	}
 
+	// and discover its resource types to understand how to send the share
+	ocmUrl, _ := url.Parse(ocmEndpoint)
+	//ocmDisco, err := s.client.Discover(ctx, ocmUrl.Scheme+"://"+ocmUrl.Host)
```
rawe0: Leftover?

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -310,13 +302,30 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
+	_, err = s.client.Discover(ctx, ocmUrl.Scheme+"://"+ocmUrl.Host)
+	if err != nil {
+		log.Error().Err(err).Msg("error discovering remote OCM provider capabilities")
+		return &ocm.CreateOCMShareResponse{
+			Status: status.NewInternal(ctx, err, "error discovering remote OCM provider capabilities"),
+		}, nil
+	}
+
+	// TODO(lopresti) does the remote OCM server support the protocols and requirements we're sending?
```
rawe0: Is there a reason we are not implementing this now?

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
rawe0: I guess this is from the TODO

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -243,13 +243,33 @@ func getOCMShareType(st string) ocm.RecipientType {
...
+	}
+
 	for _, data := range p {
 		var uri string
+		var protoInfo any
+		var ok bool
 		ocmProto := data.ToOCMProtocol()
 		protocolName := GetProtocolName(data)
+		for _, rt := range ocmRTs {
+			if rt.Name == resType {
```
rawe0, line 263: Does NextCloud still use "file" when it's a folder, would this cause issues?

rawe0: `shares.go:147` 
```
// in case of legacy OCM v1.0 shares, we have to PROPFIND the remote resource to check the type,
// because remote systems such as Nextcloud may send "file" even if the resource is a folder.
```

glpatcern: Yes, it does... so we need this logic for some time yet

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: No particular reason :) Initially I thought I'd only adjust the discovery endpoint, but now it's already quite a change. We could also implement this check, with the caveat that for `webdav` we have to be more permissive.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: Yes, it is

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: OK I introduced the minimal implementation we'll need once we wire in the `webapp` shares

rawe0 (dismissed):
Looks good for me now, let's give it a test and see!

internal/http/services/opencloudmesh/ocmd/client.go
```diff
@@ -71,35 +71,42 @@ func NewClient(timeout time.Duration, insecure bool) *OCMClient {
...
 	log := appctx.GetLogger(ctx)
 
+	if !strings.Contains(endpoint, "http") {
+		if strings.Contains(endpoint, "localhost") || strings.Contains(endpoint, "127.0.0.1") {
+			// for testing purposes we allow no TLS on localhost
+			endpoint = "http://" + endpoint
+		} else {
+			endpoint = "https://" + endpoint
+		}
+	}
```
MahdiBaghbani: This `strings.Contains(endpoint, "http")` check can skip scheme normalization for domains that just contain the substring `http`, for example `my-http-cloud.example`.

I think checking for an explicit HTTP(S) prefix keeps the intended localhost exception without changing unrelated domains.

```suggestion
	if !strings.HasPrefix(endpoint, "http://") && !strings.HasPrefix(endpoint, "https://") {
		if strings.HasPrefix(endpoint, "localhost") || strings.HasPrefix(endpoint, "127.0.0.1") {
			// for testing purposes we allow no TLS on localhost
			endpoint = "http://" + endpoint
		} else {
			endpoint = "https://" + endpoint
		}
	}
```

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -310,13 +304,53 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
+				// for now we ignore the targets as we only support "blank", which
+				// we assume the remote supports as well
+				webapp_supported = true
+			}
+			// if p == "webdav-receive" {
+			//   we cannot assume this is present as until OCM 1.2 it was given for granted,
+			//   therefore we optimistically assume it's supported
+			// }
+		}
+	}
```
MahdiBaghbani, line 373: My read is that this loop checks the map values, but OCM-API develop advertises receiver support as the `webapp-receive` key with a `targets` object. With the current shape Reva advertises, this will not find `webapp-receive`, so `webapp_supported` stays false.

Could we key the map directly and check that the receiver supports the target we actually use?

```suggestion
	for _, rt := range ocmDisco.ResourceTypes {
		if rt.Name != resType {
			continue
		}
		webappReceive, ok := rt.Protocols["webapp-receive"].(map[string]any)
		if !ok {
			continue
		}
		switch targets := webappReceive["targets"].(type) {
		case []any:
			for _, target := range targets {
				if target == "blank" {
					webapp_supported = true
					break
				}
			}
		case []string:
			for _, target := range targets {
				if target == "blank" {
					webapp_supported = true
					break
				}
			}
		}
	}
```

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -331,8 +365,8 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
 		}),
 		SenderDisplayName: user.DisplayName,
 		ShareType:         "user",
-		ResourceType:      getResourceType(info),
-		Protocols:         s.getProtocols(ctx, ocmshare),
+		ResourceType:      resType,
+		Protocols:         s.getProtocols(ctx, ocmshare, webapp_supported && token_exchange_supported),
```
MahdiBaghbani, line 397: I think the recipient's `exchange-token` capability should not be part of this gate. For the develop webapp flow, the sender hosts `tokenEndPoint` and puts `must-exchange-token` in the protocol payload. The receiver-side signal we need here is `webapp-receive` with a target we can use.

If webapp was requested but the receiver cannot accept it, my preference would be an explicit failure or making sure the locally persisted access methods match what was actually sent, rather than silently dropping webapp.

```suggestion
		Protocols:         s.getProtocols(ctx, ocmshare, webapp_supported),
```

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -298,6 +291,7 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
 		AccessMethods: req.AccessMethods,
 	}
 
+	// TOOD(lopresti) persist after sending to remote
```
MahdiBaghbani: This new preflight happens after `StoreShare`, so discovery or endpoint failures can leave a local outbound share that the remote never received.

I checked the JSON repository and `StoreShare` generates `ocmshare.Id`, so I do not think this can move all the way after building the `NewShareRequest`. But the endpoint lookup, URL validation, discovery call, and capability checks can happen before `StoreShare`. Then the existing rollback can stay for failures after the share id is generated and the request is sent.

Could you please move those preflight checks before storing the share, and keep rollback for the `NewShare` path?

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -237,15 +228,17 @@ func (s *service) walk(ctx context.Context, path string, fn walker.WalkFunc) err
...
+func (s *service) getProtocols(ctx context.Context, share *ocm.Share, include_webapp bool) ocmd.Protocols {
 	var p ocmd.Protocols
 	for _, m := range share.AccessMethods {
 		switch t := m.Term.(type) {
 		case *ocm.AccessMethod_WebdavOptions:
 			p = append(p, s.getWebdavProtocol(share, t))
 		case *ocm.AccessMethod_WebappOptions:
-			p = append(p, s.getWebappProtocol(share))
+			if include_webapp {
+				p = append(p, s.getWebappProtocol(share))
```
MahdiBaghbani: One more thing around webapp: I think this branch is probably behind current `master` here. `master` seems to have a richer webapp payload with `sharedSecret`, `permissions`, `requirements`, `targets`, and `appName`.

Could you please rebase on top of `master` first? I think it would be better to review this part again after the branch has that code, and then layer the discovery gate on top of it.

MahdiBaghbani (commented):
Hi, I was a bit late to this review.

My read is that the discovery payload direction is right if we target the
current OCM-API develop branch, especially `webapp: {}`,
`webapp-receive: { targets: [...] }`, and `must-invite`.

internal/http/services/opencloudmesh/ocmd/client.go
glpatcern: Yes, this is definitely better, I don't like this part anyhow but I couldn't find a decent alternative

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
rawe0: I will rebase!

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
rawe0: We have simplified this a bit but the outcome is the same.

glpatcern: Right, we've done this a bit differently

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: I actually disagree :) the idea here is that from a UI perspective, this is not worth a user-visible failure, as apps if available are "blended" in the file browsing UI. But anyway this is to be further expanded once we wire the webapp properly.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: This is trickier, I have left the TODO there: we really want to persist to the DB after the preflight checks and also after sending the share over, but we need to generate an ID. @rawe0 and I will do a separate PR for this.

rawe0 (dismissed):
Looks good for me after the fixes, @MahdiBaghbani feel free to review again.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -331,13 +328,57 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
+		if !ok {
+			continue
+		}
+		targets, ok := webappReceive["targets"].([]string)
+		if !ok {
+			continue
+		}
+		if slices.Contains(targets, "blank") {
+			webapp_supported = true
+		}
```
MahdiBaghbani: I may be missing a normalization step somewhere, but I think this still has the same shape problem with real discovery JSON.

Since `Protocols` is `map[string]any`, a remote JSON payload with `"targets": ["blank"]` should decode as `[]any`, not `[]string`. So this may still keep `webapp_supported` false even when the remote advertises `webapp-receive` with the target we need.

Could you please cross-check this?

```suggestion
		switch targets := webappReceive["targets"].(type) {
		case []string:
			if slices.Contains(targets, "blank") {
				webapp_supported = true
			}
		case []any:
			if slices.ContainsFunc(targets, func(target any) bool {
				targetString, ok := target.(string)
				return ok && targetString == "blank"
			}) {
				webapp_supported = true
			}
		}
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -271,61 +294,42 @@ func getAndResolveProtocols(ctx context.Context, p Protocols, ownerServer string
...
+		// (relative) path or missing
+		if protocolName != "webdav" {
+			return nil, false, fmt.Errorf("invalid protocol URI: missing host for protocol %s", protocolName)
+		}
+		protoRoot, ok := protoInfo.(string)
+		if !ok {
+			return nil, false, fmt.Errorf("missing host in URI %s and root webdav path not advertised by the remote OCM server", uri)
 		}
+
+		u, _ = url.Parse(ocmEndpoint)
```
MahdiBaghbani: One small thing I am still unsure about here.

The discovery call now fails closed on request errors, which is good, but the `endPoint` value from a successful JSON payload still seems to be trusted as-is. If that field is empty, relative, or malformed, this fallback can still build a path-only WebDAV URI.

Could we validate that the discovered endpoint is an absolute HTTP(S) URL before using it?

```suggestion
		u, err = url.Parse(ocmEndpoint)
		if err != nil {
			return nil, false, fmt.Errorf("invalid OCM endpoint %q: %w", ocmEndpoint, err)
		}
		if u.Scheme == "" || u.Host == "" || (u.Scheme != "http" && u.Scheme != "https") {
			return nil, false, fmt.Errorf("invalid OCM endpoint: %s", ocmEndpoint)
		}
```

MahdiBaghbani (commented):
Hi @glpatcern and @rawe0, thanks for the updates.

I think the stale local share ordering can be left for the follow-up PR you mentioned. My read is that this PR already got quite a bit bigger, so that split makes sense to me.

I still have two small doubts that I would like your eyes on before merge. They may just be me missing a conversion point somewhere, but both look like places where valid discovery JSON could still behave differently than the current code expects.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: I executed this snippet on a map defined as:
```
m := map[string]any{
		"targets": []string{"blank"},
	}
```
exactly like it is defined in the well-known endpoint, and it works. If the JSON decoder preserves the type to be a `map[string]any`, then we should be fine. If not, we will learn once we start to use it for real.

internal/http/services/opencloudmesh/ocmd/shares.go
glpatcern: Agreed we should add more input validation (though this is in the pending PR about hardening the OCM handling). I've added one here and one above for the protocol's URI.

But the `u.Scheme` check cannot be done as we may still receive payloads with incomplete URIs - it's only since very recently that we say _URIs MUST be absolute_ in the spec.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
MahdiBaghbani:
```go
package main

import (
	"encoding/json"
	"fmt"
	"slices"
)

// Mirrors wellknown.ResourceTypes: Protocols is map[string]any, JSON-decoded.
type ResourceTypes struct {
	Name      string         `json:"name"`
	Protocols map[string]any `json:"protocols"`
}

type OcmDiscoveryData struct {
	ResourceTypes []ResourceTypes `json:"resourceTypes"`
}

// The exact logic from ocmshareprovider.go lines 358-368.
func webappSupported(rt ResourceTypes) bool {
	webappReceive, ok := rt.Protocols["webapp-receive"].(map[string]any)
	if !ok {
		return false
	}
	targets, ok := webappReceive["targets"].([]string)
	if !ok {
		return false
	}
	return slices.Contains(targets, "blank")
}

func main() {
	mNative := map[string]any{
		"targets": []string{"blank"},
	}
	_, okA := mNative["targets"].([]string)
	fmt.Printf("CASE A (native Go map[string]any{targets: []string}): .([]string) ok = %v\n", okA)

	remoteJSON := `{
	  "resourceTypes": [
	    {
	      "name": "file",
	      "protocols": {
	        "webapp-receive": { "targets": ["blank"] }
	      }
	    }
	  ]
	}`
	var disco OcmDiscoveryData
	if err := json.Unmarshal([]byte(remoteJSON), &disco); err != nil {
		panic(err)
	}
	rt := disco.ResourceTypes[0]
	wr := rt.Protocols["webapp-receive"].(map[string]any)

	fmt.Printf("CASE B (json.Unmarshal): targets concrete type = %T\n", wr["targets"])
	_, okStr := wr["targets"].([]string)
	_, okAny := wr["targets"].([]any)
	fmt.Printf("CASE B: .([]string) ok = %v   .([]any) ok = %v\n", okStr, okAny)
	fmt.Printf("CASE B: webappSupported() with current code = %v   (expected true for a valid payload)\n", webappSupported(rt))
}

```

```
CASE A (native Go map[string]any{targets: []string}): .([]string) ok = true
CASE B (json.Unmarshal): targets concrete type = []interface {}
CASE B: .([]string) ok = false   .([]any) ok = true
CASE B: webappSupported() with current code = false   (expected true for a valid payload)
```

glpatcern: Long story short, the Go JSON parser decodes arrays of strings to `[]any`, so we have to cast to that on `targets`. Code fixed!

MahdiBaghbani (approved):
LGTM!

PR #5625, opened 2026-05-27
EOS: store attrs on version folders

jessegeens:


pkg/storage/fs/eos/client/grpc/acl.go
```diff
@@ -180,22 +173,40 @@ func (c *Client) getACLForPath(ctx context.Context, auth eosclient.Authorization
 }
 
 func (c *Client) fixupACLs(ctx context.Context, auth eosclient.Authorization, info *eosclient.FileInfo) *eosclient.FileInfo {
```
glpatcern: Can we rename this function as e.g. `fixupACLsAndAttrs` ?

glpatcern (commented):
I think it's all good, despite the added complexity and cost for each file operation (well, that's the unavoidable downside...). Only one comment (I had hard time to find where the attrs were reconciled)

glpatcern (approved):

PR #5627, opened 2026-05-27
Use destination field in received ocm share

rawe0:
PR for using the destination string instead of hijacking the opaque field in the UpdateOCMShareRequest

### DIFF
--- a/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
+++ b/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
@@ -475,21 +475,18 @@ func (s *service) ListReceivedOCMShares(ctx context.Context, req *ocm.ListReceiv
 
 func (s *service) UpdateReceivedOCMShare(ctx context.Context, req *ocm.UpdateReceivedOCMShareRequest) (*ocm.UpdateReceivedOCMShareResponse, error) {
 	user := appctx.ContextMustGetUser(ctx)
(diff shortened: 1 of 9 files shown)


internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -475,21 +475,19 @@ func (s *service) ListReceivedOCMShares(ctx context.Context, req *ocm.ListReceiv
...
-				if errors.Is(err, share.ErrShareNotFound) {
-					return &ocm.UpdateReceivedOCMShareResponse{
-						Status: status.NewNotFound(ctx, "share does not exist"),
-					}, nil
-				}
+	// This is the state we are setting when processing a share
+	// meaning it should be ACCEPTED in order to trigger the processing.
+	if req.Share.State == ocm.ShareState_SHARE_STATE_ACCEPTED {
+		destPath := req.Share.Destination
+		if _, err := s.repo.ProcessEmbeddedShare(ctx, user, req.Share, destPath); err != nil {
```
glpatcern: This is a bit nitpicking, but as `ProcessEmbeddedShare()` already has `req.Share`, you don't need to pass the path any longer

rawe0: Ahh yeah, I will update

glpatcern (approved):

PR #5628, opened 2026-05-27
feat: add ACL support for container-specific permissions

flash7777:
## Summary

Extend the grants ACL encoding/decoding to support the new `delete_container` and `move_container` permission fields proposed in cs3org/cs3apis#272.

### Changes

- `pkg/storage/utils/grants/grants.go`: Add ACL flags `+dc`/`!dc` (delete_container) and `+mc`/`!mc` (move_container) to the ACL string representation
- Parsing: Decode `+dc` and `+mc` back to the corresponding `ResourcePermissions` fields

## Motivation

This enables storage backends to distinguish between file and container (directory) delete/move operations, supporting DMS use cases like file plan (Aktenplan) structure protection.

With these ACL flags, a role can be configured that allows:
- Deleting files: yes
- Deleting directories: no
- Renaming files: yes  
- Renaming directories: no

## Depends on

- cs3org/cs3apis#272 (container-specific permissions and immutable flag)

## Test plan

- [ ] Verify `GetACLPerm` encodes `+dc`/`!dc` and `+mc`/`!mc` correctly
- [ ] Verify `GetGrantPermissionSet` decodes `+dc` and `+mc` correctly
- [ ] Verify backward compatibility: existing ACL strings without `dc`/`mc` default to `false`

### DIFF
--- a/pkg/storage/utils/grants/grants.go
+++ b/pkg/storage/utils/grants/grants.go
@@ -59,6 +59,18 @@ func GetACLPerm(set *provider.ResourcePermissions) (string, error) {
 		b.WriteString("!d")
 	}
 
+	if set.DeleteContainer {
+		b.WriteString("+dc")
+	} else {
+		b.WriteString("!dc")
+	}
+
+	if set.MoveContainer {
+		b.WriteString("+mc")
+	} else {
+		b.WriteString("!mc")
+	}
+
 	return b.String(), nil
 }
 
@@ -96,6 +108,16 @@ func GetGrantPermissionSet(perm string) *provider.ResourcePermissions {
 		rp.Delete = false
 		rp.PurgeRecycle = false
 	}
+
+	if strings.Contains(perm, "+dc") {
+		rp.DeleteContainer = true
+	}
+	// !dc is the default (false), no action needed
+
+	if strings.Contains(perm, "+mc") {
+		rp.MoveContainer = true
+	}
+	// !mc is the default (false), no action needed
 
 	if strings.Contains(perm, "m") && !strings.Contains(perm, "!m") {
 		rp.AddGrant = true


flash7777:
### Context

This PR is the Reva-side companion to the CS3 API proposal in cs3org/cs3apis#272.

We are developing an **EDMS layer for German municipalities** (Kommunalverwaltungen) on top of OpenCloud/Reva. A core requirement is protecting hierarchical file plan structures (Aktenpläne) — rigid directory trees that must remain stable while users freely work with documents inside them.

The current permission model does not distinguish between file and container operations for `delete` and `move`. These small ACL extensions (`+dc`/`!dc`, `+mc`/`!mc`) bridge that gap and enable real-world DMS and records management scenarios.

The changes here are intentionally minimal — just the ACL encoding/decoding layer. Storage driver implementations (e.g., decomposedfs) would add the corresponding `IsDir()` checks in their delete/move handlers once the CS3 API change is accepted.

We are happy to extend this with tests and additional driver support based on community feedback.

flash7777:
Quick note: The commits mention "Claude" as co-author — this just means we used Claude Code (AI coding assistant) as tooling support during development. The proposal and its rationale are entirely ours, based on real requirements from our EDMS project.

jessegeens:
Hi @flash7777 ,

Sorry for the late reply. I also saw your pull request on the CS3APIs. Since you want to build an EDMS on top of OpenCloud, you should probably make this PR in github.com/opencloud-eu/reva. They run a fork of our Reva which has completely diverged, so your commit would not get in there.

flash7777:
Indeed this was the plan. starting at cs3.

flash7777:
and yes, we're still in the process of founding a World Immutable File System Flag Association. Freedom for mutable flags, immutable flags for everyone, for every file system. We're still in the founding phase — it's currently held up by several current and upcoming PRs. 🙂

flash7777:
and speaking of which — this Reva from OpenCloud (opencloud.eu) of course relies on importing CS3's CS3 API, which actually has to be merged and shipped first before it can be included. And I really wanted to avoid making my own fork, merging that in as the import, and then saying 'no, let's just wait for CS3.' It's all not a big deal though — that's just how it is, one thing after another, apparently, obviously.

PR #5629, opened 2026-05-28
Parse zenodo crate

rawe0:
Adds support for OCM embedded shares whose payload is an RO-Crate wrapping the Zenodo JSON-LD format (the ScienceMesh-style RO-Crate is also supported).

The RO-Crate is stored by the `ocmshareprovider` (the SQL DB, in our case). At the provider level a new pluggable driver, the `embedded_driver`, parses the RO-Crate (Zenodo- or ScienceMesh-style) to extract the links to the publicly available data on Zenodo or GitHub, then transfers it. There is no authentication for the source downloads for now, so only publicly accessible data is supported.

### Configuration

The `ocmshareprovider` requires some additional configuration:

```
[grpc.services.ocmshareprovider]
driver = "sql"
embedded_driver = "webdav"   # optional, this is the default
...

[grpc.services.ocmshareprovider.embedded_drivers.webdav]
webdav_url = "https://cbox-ocisdev-rasmus.cern.ch/webdav"
# embedded_transfer_timeout = 3600       (default, seconds)
# embedded_transfer_idle_timeout = 120   (default, seconds)
# embedded_transfer_retries = 3          (default)
```
### Flow

1. The user calls the process-embedded-share endpoint to accept the share.
2. The `ocmshareprovider` retrieves the JSON payload from the DB.
3. The embedded driver parses the payload and extracts the links to the publicly available data.
4. For each link the driver does a GET and streams the response, via a PUT, to the configured WebDAV endpoint (usually the machine itself).

The transfer runs in the background: the endpoint returns immediately and the files are fetched asynchronously.

### Behaviour

- A transfer is aborted if no data flows for the idle timeout (e.g. due to rate limiting). Both the number of retries and the timeouts are configurable. If a file still fails after exhausting retries it is skipped (best-effort) and the remaining files continue.
- Each file also has an absolute timeout ceiling.
- Zenodo may return 429 Too Many Requests when there are too many attempts, along with a time at which requests can resume; this Retry-After is respected.
- Transfers are monitored by logging throughput every 15s (the first log is after 15s, so quick transfers are not logged).
- When all transfers finish, the number of successful and skipped transfers is logged.

### Tests
Tests have been implemented for the embedded driver where it is possible (non network related functionality). 

#### Further development

In a separate PR a share transfer state will be implemented:
- Current: `pending` -> `accepted` -> (transfer started)
- Future: `pending` -> `transfer` (transfer started) -> `accepted` (transfer finished) 

This will also be reflected in the WebUI and a seperate web PR for this will be made.

### DIFF
--- a/pkg/ocm/embedded/crate.go
+++ b/pkg/ocm/embedded/crate.go
@@ -1,0 +1,191 @@
+// Copyright 2018-2026 CERN
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// In applying this license, CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+package embedded
+
+import (
+	"encoding/json"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+
+	"github.com/rs/zerolog"
+)
+
+type crate struct {
+	Graph []crateEntity `json:"@graph"`
+}
+
+type crateEntity struct {
+	ID             string               `json:"@id"`
+	Type           json.RawMessage      `json:"@type"`
+	URL            json.RawMessage      `json:"url"`
+	Name           string               `json:"name"`
+	ContentSize    string               `json:"contentSize"`
+	EncodingFormat string               `json:"encodingFormat"`
+	Description    string               `json:"description"`
+	Distribution   []zenodoDistribution `json:"distribution"`
+}
+
+type zenodoDistribution struct {
+	Type           string `json:"@type"`
+	ContentURL     string `json:"contentUrl"`
+	EncodingFormat string `json:"encodingFormat"`
+}
+
+type transferEntry struct {
+	srcURL         string
+	name           string
+	sizeHint       int64
+	encodingFormat string
+}
+
+type idRef struct {
+	ID string `json:"@id"`
+}
+
+func (e crateEntity) URLString() string {
+	if len(e.URL) == 0 {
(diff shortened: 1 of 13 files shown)


pkg/ocm/share/share.go
```diff
@@ -58,8 +58,11 @@ type Repository interface {
 	// UpdateReceivedShare updates the received share with share state.
 	UpdateReceivedShare(ctx context.Context, user *userpb.User, share *ocm.ReceivedShare, fieldMask *field_mask.FieldMask) (*ocm.ReceivedShare, error)
 
-	// Process the embedded share with the given state.
-	ProcessEmbeddedShare(ctx context.Context, user *userpb.User, share *ocm.ReceivedShare) (*ocm.ReceivedShare, error)
+	// EmbeddedPayload returns the RO-Crate JSON payload carried by the embedded
+	// protocol of a received share, or an empty string if the share carries no
+	// embedded protocol. It is used to drive the background transfer of embedded
+	// share contents.
+	EmbeddedPayload(ctx context.Context, user *userpb.User, share *ocm.ReceivedShare) (string, error)
```
glpatcern: Let's rename this `GetEmbeddedPayload`, and return NotSupported for the repositories where we didn't implement it, as opposed to just return ""

pkg/ocm/embedded/crate.go
```diff
@@ -0,0 +1,191 @@
...
+		for _, d := range e.Distribution {
+			if entry, ok := zenodoEntry(log, d); ok {
+				entries = append(entries, entry)
+			}
+		}
+	}
+	return entries
+}
+
+func scienceMeshEntry(log *zerolog.Logger, e crateEntity) (transferEntry, bool) {
```
glpatcern: is this really a 'ScienceMesh' format? :)

rawe0: I guess it's a `eosc-dc` style ro-crate with added ocm fields, I don't really know what to name it... suggestions are welcome!

pkg/ocm/share/share.go
rawe0: Yep I agree

pkg/ocm/embedded/crate.go
glpatcern: let's call it `plain`, a priori it's any vanilla RO-Crate without any assumption (and we should make no assumptions on the payload)

rawe0: Done!

glpatcern (approved):

PR #5639, opened 2026-06-03
fix: bound Prometheus HTTP handler label cardinality (#4509)

### LINKED ISSUE #4509: Prom interceptor uses URL as label, meaning infinity cardinality
The prom http interceptor is using the URL as label, as the HTTP endpoints are user-bound, means that the cardinality of this label is unbound and that has triggered to store more than 50GB of metrics per unique URL under the `/metrics` endpoint.

The fix should include the name of the HTTP service only.

mvanhorn:
## Summary

The HTTP metrics interceptor in `internal/http/interceptors/metrics/metrics.go` used the full per-request URL path (`r.URL.Path`) as the value of the `handler` label on the `http_request_duration_seconds` histogram. This change derives the label from the leading static path segment instead (e.g. `/remote.php/dav/files/einstein/foo` becomes `/remote.php`), so the `handler` label cardinality is bounded by the number of route prefixes rather than the number of distinct user URLs.

A doc comment now records that the `handler` label must stay bounded, referencing #4509, so the regression is not silently reintroduced. A unit test verifies that multiple distinct user-bound URLs sharing a prefix collapse to a single `handler` label value.

## Why this matters

Reva's HTTP endpoints are user-bound (e.g. `/remote.php/dav/files/<userid>/<deep/path>`), so every distinct request path produced a new `handler` label value. Per #4509 this made the label cardinality effectively unbounded and caused Reva to store more than 50GB of metrics behind the `/metrics` endpoint. Bounding the label to the route prefix keeps the metrics series count proportional to the number of services.

## Testing

- `gofmt -l internal/http/interceptors/metrics/` is clean
- `go vet ./internal/http/interceptors/metrics/` passes
- `go build ./internal/http/...` succeeds
- `go test ./internal/http/interceptors/metrics/...` passes (8 tests), including the new cardinality regression test

## Note

This is the minimal first-segment normalization described in the issue. It does not yet coalesce unmatched/non-existent first segments (a request to `/foo123` is still recorded as `/foo123` before routing returns 404). Mapping unmatched prefixes to a single bucket would need the interceptor to know the registered service prefixes, which are config-driven; happy to follow up with that hardening if you prefer it in this PR.

Fixes #4509

AI was used for assistance.

### DIFF
--- a/changelog/unreleased/fix-prom-http-url-label-cardinality.md
+++ b/changelog/unreleased/fix-prom-http-url-label-cardinality.md
@@ -7,4 +7,4 @@ caused excessive metrics storage. The label is now derived from the leading
 static path segment, keeping cardinality bounded to the number of route
 prefixes.
 
-https://github.com/cs3org/reva/issues/4509
+https://github.com/cs3org/reva/pull/5639```
--- a/changelog/unreleased/fix-prom-http-url-label-cardinality.md
+++ b/changelog/unreleased/fix-prom-http-url-label-cardinality.md
@@ -1,0 +1,10 @@
+Bugfix: Bound the Prometheus HTTP handler label cardinality
+
+The HTTP metrics interceptor used the full request URL path as the value of the
+`handler` label on the `http_request_duration_seconds` histogram. Because Reva
+HTTP endpoints are user-bound, this produced unbounded label cardinality and
+caused excessive metrics storage. The label is now derived from the leading
+static path segment, keeping cardinality bounded to the number of route
+prefixes.
+
+https://github.com/cs3org/reva/issues/4509
--- a/go.mod
+++ b/go.mod
@@ -85,6 +85,7 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.29.0 // indirect
 	github.com/jinzhu/inflection v1.0.0 // indirect
 	github.com/jinzhu/now v1.1.5 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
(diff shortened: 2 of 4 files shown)


changelog/unreleased/fix-prom-http-url-label-cardinality.md
```diff
@@ -0,0 +1,10 @@
+Bugfix: Bound the Prometheus HTTP handler label cardinality
+
+The HTTP metrics interceptor used the full request URL path as the value of the
+`handler` label on the `http_request_duration_seconds` histogram. Because Reva
+HTTP endpoints are user-bound, this produced unbounded label cardinality and
+caused excessive metrics storage. The label is now derived from the leading
+static path segment, keeping cardinality bounded to the number of route
+prefixes.
+
+https://github.com/cs3org/reva/issues/4509
```
glpatcern:
```suggestion
https://github.com/cs3org/reva/pull/5639```

glpatcern (commented):
Thanks for this contribution @mvanhorn !

May I ask you how you're developing/deploying Reva?

I'll apply a mini change to pass the CI.

glpatcern (approved):

mvanhorn:
Glad to get this in, thanks @glpatcern. Bounding the Prometheus handler label cardinality should keep the metrics interceptor from blowing up series count.

mvanhorn:
Thanks @glpatcern. Bounding the Prometheus handler label cardinality should keep that metric from blowing up on per-request URL paths.

PR #5642, opened 2026-06-04
Feat/4348 Return linked-primary auth conflict contract

Gerry-Cern:
## Summary

- Return a machine-readable `409 Conflict` response from protected HTTP auth middleware when gateway `Authenticate` surfaces `CODE_ABORTED`.
- Add the linked-primary Web contract: `X-Oc-Linked-Primary-Account: true`, `Cache-Control: no-store`, and Libre Graph/OData JSON with `error.code = linkedPrimaryAccount`.
- Document the identity auth HTTP error contract and add a changelog entry.

## Test plan

- `go test ./internal/http/interceptors/auth/...`

## Notes

CERNBox deployments using `reva-plugins` already emit `errtypes.Conflict` for linked external accounts via the REST userprovider, which maps to `CODE_ABORTED` and triggers this HTTP response on protected bootstrap requests such as `GET /graph/v1.0/me`.


glpatcern (changes requested):
As I understand, this is possibly not needed and we get the minimal required functionality with the web PR only, correct?

PR #5643, opened 2026-06-05
Add transfer state

rawe0:
Utilizes the new `transferring` state in the cs3api for received ocm shares.

- Puts the ocm share in a `transferring` state when doing the processing of an embedded share is ongoing.
- A callback is implemented that puts the share in the `accepted` state when the transfer is finished.


glpatcern (approved):
Looks good!

PR #5644, opened 2026-06-05
Sync Client Login Flow

jessegeens:
Config is available in the puppet `dev` environment

  How to test

  1. Point the NextCloud Desktop client at `https://<box>:9998`.
  2. The client will POST /index.php/login/v2 → receives poll-token + login URL.
  3. It opens the login URL in a browser: GET /index.php/login/v2/flow/<lt> → the server auto-approves and shows "Access granted".
  4. The client's next poll returns `{loginName: "jgeens", appPassword: "..."}`.
  5. From then on, WebDAV/OCS calls use Basic-Auth with those credentials, verified via the existing appauth auth manager chai


glpatcern (approved):
Eager to see also the ADR, but it looks good on a first pass. A possible addition is a cache of `Authenticate()` results (positive and negative), so that we don't hammer the DB, but that's for later. Let's go ahead and exercise it!

PR #5650, opened 2026-06-09
Fix attribute errors

rawe0:
Two bugs are fixed here: 

- **ReadArbitraryMetadata** - Returns stored user.* xattrs in GetMD instead of nothing.
- **Paths** - Unlock/SetLock open the full chroot path instead of the incorrect relative path. 
 
Quality of life improvements

- **[ENOATTR](https://pkg.go.dev/github.com/pkg/xattr#pkg-constants)**  Use errors.Is(err, xattr.ENOATTR) instead of matching error strings.
- **Debug logs** are added for the locking functions.


glpatcern (approved):
Looks good, just one comment: I think there was a concept of `sys` attributes, how are they handled?

glpatcern:
Answering my comment, the only place where we need access to system attributes is for quota management, which takes place in a thread run as root.

PR #5651, opened 2026-06-10
Background jobs framework

gmgigi96:
Adds a framework for running background work: tasks that run periodically (e.g. warming a cache, cleaning up expired state) and tasks that run once on demand (e.g. on a user request). This is the framework only, no production jobs are shipped, just a ping-pong example.

It provides:
- A jobs serverless service hosting a per-process runner that owns scheduling, a worker pool, and the job lifecycle.
- Periodic jobs with an explicit scope: ScopeAllNodes (runs on every replica as a local ticker, no infra) or ScopeLeader (runs once across the cluster, durable).
- On-demand jobs triggered in-process via Enqueue, returning a RunID.
- Durability via NATS JetStream (work queue + schedule state) and per-run status in SQL, queryable by RunID.


pkg/rjobs/store/sql/sql.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (s *store) Close(ctx context.Context) error {
+	db, err := s.db.DB()
+	if err != nil {
+		return err
+	}
+	return db.Close()
+}
+
+func toModel(st rjobs.Status) (*model.Run, error) {
```
diocas, line 101: It's not possible to just have Status and make that the model? If so, not sure why we need to convert between 2 different but equal structures.

pkg/rjobs/store/nats/nats.go
```diff
@@ -0,0 +1,506 @@
...
+	// when the run finishes, and ignored once older than runningHold so a
+	// crashed worker cannot block the schedule forever.
+	RunningSince *time.Time `json:"running_since,omitempty"`
+}
+
+// runningHold bounds how long a RunningSince mark is trusted. A legitimate long
+// run keeps its claim alive with heartbeats, but the schedule mark has no
+// heartbeat of its own, so we cap it: after this long the job is allowed to be
+// scheduled again even if the mark is still set (e.g. the marker crashed).
+const runningHold = 24 * time.Hour
```
diocas, line 94: with the heartbeats, maybe 1h would be enough

diocas (approved):
Would benefit from more detail in the readme related to the nats queues (with names, how jobs are tracked, etc).

Otherwise, seems good to merge!

glpatcern (approved):

pkg/rjobs/store/sql/sql.go
glpatcern: Just discussing, the idea is to keep the model "private", whereas `rjobs.Status` may also contain methods that have nothing to do with the persistence layer

pkg/rjobs/store/nats/nats.go
gmgigi96: This is an extra protection in case the job "fails" but the thread sending heartbeats is still alive, so for the scheduler is still "running". It's on purpose very big for long running jobs (think about the creation of an archive of a very big space). For now it's hardcoded, in future we can make it configurable if needed.

PR #5654, opened 2026-06-14
Enable external accounts to list received OCM shares

glpatcern:


changelog/unreleased/ocm-for-externals.md
```diff
@@ -0,0 +1,6 @@
+Enhancement: enable external accounts to receive OCM shares
```
glpatcern:
```suggestion
Enhancement: enable external accounts to list received OCM shares
```

rawe0 (approved):
Looks good from my end, I remember there was some issue with enabling ocm shares for external users, but lets see what the issue is and then we can fix.

PR #5664, opened 2026-06-16
New webapp protocol

rawe0:
This PR introduces the new webapp protocol for OCM

### DIFF
--- a/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
+++ b/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
@@ -26,8 +26,6 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
-	"strings"
-	"text/template"
 	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
@@ -80,7 +78,7 @@ type config struct {
 	GatewaySVC      string                    `mapstructure:"gatewaysvc"                                    validate:"required"`
 	ProviderDomain  string                    `docs:"The same domain registered in the provider authorizer" mapstructure:"provider_domain" validate:"required"`
 	WebDAVEndpoint  string                    `mapstructure:"webdav_endpoint"                               validate:"required"`
-	WebappTemplate  string                    `mapstructure:"webapp_template"                               validate:"required"`
+	WebAppEndpoint  string                    `mapstructure:"webapp_endpoint"                               validate:"required"`
 }
 
 type service struct {
@@ -88,7 +86,6 @@ type service struct {
 	repo        share.Repository
 	client      *ocmd.OCMClient
 	gateway     gateway.GatewayAPIClient
-	webappTmpl  *template.Template
 	walker      walker.Walker
 	transferrer embedded.Transferrer
 }
@@ -147,10 +144,6 @@ func New(ctx context.Context, m map[string]any) (rgrpc.Service, error) {
 		return nil, err
 	}
 
-	tpl, err := template.New("webapp_template").Parse(c.WebappTemplate)
-	if err != nil {
-		return nil, err
-	}
 	walker := walker.NewWalker(gateway)
 
 	ocmcl := ocmd.NewClient(time.Duration(c.ClientTimeout)*time.Second, c.ClientInsecure)
@@ -159,7 +152,6 @@ func New(ctx context.Context, m map[string]any) (rgrpc.Service, error) {
 		repo:        repo,
 		client:      ocmcl,
 		gateway:     gateway,
-		webappTmpl:  tpl,
 		walker:      walker,
 		transferrer: transferrer,
 	}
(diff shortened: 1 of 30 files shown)


internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -222,13 +222,42 @@ func (s *service) getWebdavProtocol(share *ocm.Share, m *ocm.AccessMethod_Webdav
...
+
+	// requirements are required by the OCM specifications: fall back to the
+	// defaults for shares created before these were stored.
+	reqs := m.WebappOptions.Requirements
+	if len(reqs) == 0 {
+		reqs = share.DefaultWebappRequirements
+	}
+	// targets are no longer carried by the webapp access method, so always
+	// advertise the defaults required by the OCM specifications.
+	targets := share.DefaultWebappTargets
```
glpatcern: Targets were never part of webapp in the past, this comment seems a remnant of the intermediate CS3 API

internal/http/services/opencloudmesh/ocmd/specs.go
```diff
@@ -299,6 +338,38 @@ func (p Protocols) Validate() error {
...
+				return errors.New("protocol webapp missing permissions")
+			}
+			for _, permission := range data.Permissions {
+				if _, ok := validWebappPermissions[permission]; !ok {
+					return fmt.Errorf("protocol webapp has unsupported permission %q", permission)
+				}
+			}
+			if len(data.Targets) == 0 {
+				return errors.New("protocol webapp missing targets")
+			}
```
glpatcern: I guess this can be factored out with the webdav protocol handling above

pkg/auth/manager/ocmshares/ocmshares.go
```diff
@@ -190,15 +189,13 @@ func getRole(s *ocm.Share) (authpb.Role, string) {
...
-				viewMode == provider.ViewMode_VIEW_MODE_READ_ONLY {
-				return authpb.Role_ROLE_VIEWER, "viewer"
-			}
-			if viewMode == provider.ViewMode_VIEW_MODE_READ_WRITE ||
-				viewMode == provider.ViewMode_VIEW_MODE_PREVIEW {
+			p := v.WebappOptions.GetPermissions()
+			if p.GetInitiateFileUpload() {
 				return authpb.Role_ROLE_EDITOR, "editor"
 			}
+			if p != nil {
```
glpatcern: What about `"reader"` ? Also this seems a duplicate of the function(s) in `ocmsharecode.go`

pkg/ocm/share/repository/nextcloud/nextcloud.go
```diff
@@ -175,6 +175,16 @@ func (sm *Manager) StoreShare(ctx context.Context, share *ocm.Share) (*ocm.Share
 	return share, nil
 }
 
+// webappViewModeToResourcePermissions maps the legacy EFSS view mode strings
+// to CS3 resource permissions.
```
glpatcern: Why do we need this?

pkg/share/manager/sql/conversions.go
```diff
@@ -161,22 +160,6 @@ func convertToCS3OCMReceivedShare(s *model.OcmReceivedShare, p []*ocm.Protocol)
...
-		return 1
-	case appprovider.ViewMode_VIEW_MODE_READ_ONLY:
-		return 2
-	case appprovider.ViewMode_VIEW_MODE_READ_WRITE:
-		return 3
-	case appprovider.ViewMode_VIEW_MODE_PREVIEW:
-		return 4
-	}
-	return -1
-}
```
glpatcern, line 178: This function is still used by the app provider I believe, no?

glpatcern (changes requested):
Some relatively minor comments

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -222,13 +222,42 @@ func (s *service) getWebdavProtocol(share *ocm.Share, m *ocm.AccessMethod_Webdav
 	}
 }
 
-func (s *service) getWebappProtocol(share *ocm.Share) *ocmd.Webapp {
+func (s *service) getWebappProtocol(ocmShare *ocm.Share, m *ocm.AccessMethod_WebappOptions) *ocmd.Webapp {
 	var b strings.Builder
-	if err := s.webappTmpl.Execute(&b, share); err != nil {
+	if err := s.webappTmpl.Execute(&b, ocmShare); err != nil {
```
glpatcern: Also this is to be removed, we don't have templates any longer

pkg/share/manager/sql/conversions.go
rawe0: I don't think so, the only reference I could find was in `pkg/share/manager/sql/ocm_shares.go`

glpatcern: Ah ok! then sure we can drop it

pkg/ocm/share/repository/nextcloud/nextcloud.go
rawe0: The comment is a bit misleading I will change it, it's because we changed the `NewWebAppAccessMethod` to taking ResourcePermissions as input (it's a trickle down effect from the cs3apis changes).

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
rawe0: To be removed

pkg/auth/manager/ocmshares/ocmshares.go
rawe0: reader and viewer is essentially the same here, that's why, downstream in the eosfs.go we check the following:

eosfs.go:412
```
if role, ok := utils.HasOCMShareRole(u); ok {
	if role == "editor" {
		return permissions.NewEditorRole().CS3ResourcePermissions()
	}
	return permissions.NewViewerRole().CS3ResourcePermissions()
}
```
Basically the viewer role gets download permissions currently, we could change this ofc.

glpatcern (approved):

PR #5676, opened 2026-06-26
OCM: fixed parsing of discovery payloads

glpatcern:
Following tests with EOSC Data Commons.


changelog/unreleased/fix-ocm-discovery-parsing.md
```diff
@@ -0,0 +1,3 @@
+Bugfix: fixed parsing of OCM discovery payload
+
```
glpatcern, line 2:
```suggestion

The principle is that we should not fail the JSON parsing on unsupported capabilities, but only on out of spec payloads. Therefore, `ResourceType` must be a generic string, to be validated afterwards.
```

jessegeens (approved):

PR #5685, opened 2026-07-01
Add sciencemesh paths for external accounts

glpatcern:
This still relies on an old mechanism that should be refactored.


rawe0 (dismissed):
Looks good!

jessegeens (approved):

PR #5689, opened 2026-07-01
misc: modernize go code with go fix

labkode:
Run `go fix` on the codebase.


glpatcern (approved):
Nice, we could have this in the CI!

PR #5690, opened 2026-07-01
Add remote users automatically for OCM shares

rawe0:
This PR Introduces:

- Functionality to add remote users even without the invitation flow when an OCM share is received.
- This configuration option is a whitelist of hosts of which this functionality is enabled for.

In order to do this a `machine secret` also has to be introduced since we have to impersonate the user receiving the share (since the ocm share endpoint is unauthenticated).

### DIFF
--- a/internal/http/services/opencloudmesh/ocmd/shares.go
+++ b/internal/http/services/opencloudmesh/ocmd/shares.go
@@ -182,12 +182,12 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// By default all incoming shares are accepted. Configuring a non-empty
-	// auto-accept whitelist restricts this: a share is then stored only if the
-	// sender's provider is on the whitelist, or if the sender is already an
+	// By default all incoming shares are accepted only if the sender is already an
 	// accepted user of the recipient (i.e. they went through a prior invitation
-	// flow). Shares matching neither are rejected outright.
-	if len(h.autoAcceptProviders) > 0 && !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
+	// flow).  Configuring a non-empty auto-accept whitelist (that may include `.*`)
+	// allows any sender whose provider is on the whitelist to be accepted.
+	// Shares matching neither are rejected outright.
+	if !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
 		reqres.WriteError(w, r, reqres.APIErrorUnauthenticated, "sender provider not in the auto-accept whitelist", nil)
 		return
 	}
--- a/internal/http/services/opencloudmesh/ocmd/shares.go
+++ b/internal/http/services/opencloudmesh/ocmd/shares.go
@@ -187,7 +187,8 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 	// flow).  Configuring a non-empty auto-accept whitelist (that may include `.*`)
 	// allows any sender whose provider is on the whitelist to be accepted.
 	// Shares matching neither are rejected outright.
-	if !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
+	wasAccepted := h.isAcceptedUser(ctx, userRes.User, sender)
+	if !h.matchesAutoAccept(sender.Idp) && !wasAccepted {
 		reqres.WriteError(w, r, reqres.APIErrorUnauthenticated, "sender provider not in the auto-accept whitelist", nil)
 		return
 	}
@@ -247,7 +248,7 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 	// resolution of them would fail. Register them locally now, while the payload still
 	// carries their display names. This is auxiliary: any failure is logged but does
 	// not fail the share creation.
-	if h.matchesAutoAccept(sender.Idp) {
+	if !wasAccepted {
 		// We leave the email blank since it's not available in the ocm share payload.
 		h.registerAcceptedUser(ctx, userRes.User, &userpb.User{
 			Id:          sender,
@@ -270,12 +271,6 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 // and simply expires. All failures are logged and swallowed.
 func (h *sharesHandler) registerAcceptedUser(ctx context.Context, recipient *userpb.User, remoteUser *userpb.User) {
 	log := appctx.GetLogger(ctx)
-
-	// skip if the remote user is already an accepted user of the recipient
-	if h.isAcceptedUser(ctx, recipient, remoteUser.Id) {
-		log.Debug().Str("remote_user", remoteUser.Id.OpaqueId).Msg("auto-register: remote user already accepted, skipping")
-		return
-	}
 
 	// impersonate the recipient so the minted invite token is owned by them
 	recipientCtx, err := h.impersonate(ctx, recipient)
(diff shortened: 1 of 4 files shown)


internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -93,6 +117,14 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// When an auto-accept whitelist is configured it acts as a mandatory allowlist:
+	// shares from a provider that does not match are rejected outright and never stored.
+	// An empty whitelist disables this gate, preserving the default "accept all" behavior.
+	if len(h.autoAcceptProviders) > 0 && !h.matchesAutoAccept(sender.Idp) {
```
glpatcern: Claude got it wrong here. The logic is reverted:
* Check if user was invited and accepted via GetAcceptedUser().
  * if yes, go ahead
  * if not, i.e. an unknown user, check the sender.Idp as here:
    *  if match, call autoRegisterRemoteUser (only for the `sender` user).
    *  if no match, fail

rawe0: Fixed!

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -138,6 +182,16 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// When an auto-accept whitelist is configured it acts as a mandatory allowlist:
+	// shares from a provider that does not match are rejected outright and never stored.
```
glpatcern: the logic is now correct but this comment is outdated

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -188,6 +242,15 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Shares from a whitelisted provider arrive without a prior invitation flow, so
```
glpatcern:
```suggestion
	// Shares from a whitelisted provider may arrive without a prior invitation flow, so
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -196,6 +259,95 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
...
+	})
+
+	if owner.OpaqueId == sender.OpaqueId && owner.Idp == sender.Idp {
+		return
+	}
+	if owner.OpaqueId == "" {
+		log.Debug().Msg("auto-register: skipping empty owner")
+		return
+	}
+	h.registerAcceptedUser(ctx, recipient, &userpb.User{
```
glpatcern: As a matter of fact, we only need to register the `sender`, not the `owner` (and Data Commons right now is putting a random ID for the owner). We'd need to also inject the email address, which we don't have, so fine we leave it blank but maybe add a comment that we're not populating it.

glpatcern (commented):
Some relatively minor comments

internal/http/services/opencloudmesh/ocmd/shares.go
rawe0: I think for the case of data commons the owner will always be the same as the sender in this case - after the changes made to the dispatcher (this is only in the ocm share request, not in the ro-crate) but yes we don't need to add the sender. 

I'll add a comment regarding the email.

internal/http/services/opencloudmesh/ocmd/shares.go
rawe0: To me this seems correct, if you read the full comment?

rawe0: I will rephrase it though, I can see how it is confusing.

internal/http/services/opencloudmesh/ocmd/shares.go
rawe0: Done, I removed this entire function in fact.

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -138,6 +182,16 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// By default all incoming shares are accepted. Configuring a non-empty
+	// auto-accept whitelist restricts this: a share is then stored only if the
+	// sender's provider is on the whitelist, or if the sender is already an
+	// accepted user of the recipient (i.e. they went through a prior invitation
+	// flow). Shares matching neither are rejected outright.
+	if len(h.autoAcceptProviders) > 0 && !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
```
glpatcern:
```suggestion
	// By default all incoming shares are accepted only if the sender is already an
	// accepted user of the recipient (i.e. they went through a prior invitation
	// flow).  Configuring a non-empty auto-accept whitelist (that may include `.*`)
	// allows any sender whose provider is on the whitelist to be accepted.
	// Shares matching neither are rejected outright.
	if !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -138,6 +182,16 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// By default all incoming shares are accepted only if the sender is already an
+	// accepted user of the recipient (i.e. they went through a prior invitation
+	// flow).  Configuring a non-empty auto-accept whitelist (that may include `.*`)
+	// allows any sender whose provider is on the whitelist to be accepted.
+	// Shares matching neither are rejected outright.
+	if !h.matchesAutoAccept(sender.Idp) && !h.isAcceptedUser(ctx, userRes.User, sender) {
+		reqres.WriteError(w, r, reqres.APIErrorUnauthenticated, "sender provider not in the auto-accept whitelist", nil)
```
glpatcern:
```suggestion
	wasAccepted := h.isAcceptedUser(ctx, userRes.User, sender)
	if !h.matchesAutoAccept(sender.Idp) && !wasAccepted {
		reqres.WriteError(w, r, reqres.APIErrorUnauthenticated, "sender provider not in the auto-accept whitelist", nil)
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -188,6 +242,19 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Shares from a whitelisted provider may arrive without a prior invitation flow, so
+	// their remote sender/owner are not yet known as accepted users and a later
+	// resolution of them would fail. Register them locally now, while the payload still
+	// carries their display names. This is auxiliary: any failure is logged but does
+	// not fail the share creation.
+	if h.matchesAutoAccept(sender.Idp) {
```
glpatcern:
```suggestion
	if !wasAccepted {
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -196,6 +263,71 @@ func (h *sharesHandler) CreateShare(w http.ResponseWriter, r *http.Request) {
...
+// and simply expires. All failures are logged and swallowed.
+func (h *sharesHandler) registerAcceptedUser(ctx context.Context, recipient *userpb.User, remoteUser *userpb.User) {
+	log := appctx.GetLogger(ctx)
+
+	// skip if the remote user is already an accepted user of the recipient
+	if h.isAcceptedUser(ctx, recipient, remoteUser.Id) {
+		log.Debug().Str("remote_user", remoteUser.Id.OpaqueId).Msg("auto-register: remote user already accepted, skipping")
+		return
+	}
+
```
glpatcern:
```suggestion
```

glpatcern (dismissed):

glpatcern (approved):

PR #5691, opened 2026-07-02
Gormify the invite repository

rawe0:


glpatcern (dismissed):

pkg/ocm/invite/repository/sql/sql.go
```diff
@@ -66,6 +66,29 @@ func (c *Config) ApplyDefaults() {
...
+// OcmRemoteUser is the gorm model for the ocm_remote_users table.
+//
+// The i_unique index enforces a single row per (initiator, opaque_user_id, idp).
+// A soft-deleted row keeps occupying that index, so AddRemoteUser revives it
+// rather than inserting a duplicate.
+type OcmRemoteUser struct {
+	gorm.Model
+	Initiator    string `gorm:"size:255;index:i_initiator;uniqueIndex:i_unique"`
+	OpaqueUserID string `gorm:"size:255;uniqueIndex:i_unique"`
+	Idp          string `gorm:"size:255;uniqueIndex:i_unique"`
```
glpatcern: Sorry that I didn't spot it in the first pass:
- the index should be `i_ocmremoteuser_initiator` to avoid collisions
- the `uniqueIndex` must be named differently, let's follow the convention `i_ocmremoteuser_xyz`.
For the latter, I know we had in the past this constraint that the uniqueness is for pairs of (local_initiator, remote_users), but it's always been weird: what we want is uniqueness of remote users (so on `OpaqueUserID` and `Idp`), regardless who invited them.

Can you check if we display all users regardless the initiator when sharing? If that's the case, let's go for the stricter uniqueness. If not, this is a fix to be done in a later PR.

pkg/ocm/invite/repository/sql/sql.go
```diff
@@ -66,6 +66,29 @@ func (c *Config) ApplyDefaults() {
 	c.Database = sharedconf.GetDBInfo(c.Database)
 }
 
+// OcmToken is the gorm model for the ocm_tokens table.
+type OcmToken struct {
+	gorm.Model
+	Token       string `gorm:"size:255;uniqueIndex:i_token"`
+	Initiator   string `gorm:"size:255;index:i_initiator"`
```
glpatcern: also here, we want `i_ocmtoken_initiator` etc.

glpatcern (changes requested):
In fact there's a fix and a question:

pkg/ocm/invite/repository/sql/sql.go
glpatcern: So the current behavior is that invites are personal and a user can share only to recipients he/she invited. Let's keep the schema as is for now then.

glpatcern (approved):

PR #5692, opened 2026-07-02
Admin APIs

gmgigi96:
This PR adds an admin API to reva: an authenticated, audited way to look at and operate a running fleet from a single `reva admin` command. It is **off by default** and only enabled where a process is given an admin group.

The building block is a small **control channel** that every reva process runs beside its services. The admin reaches into a live instance through it to read state or run an operation (an *invocation*), resolving targets through the registry, so a command can hit one instance, all instances of a service, or the whole fleet, with the results merged.

On top of it, an operator can:

- **inspect the fleet**: list services and their health, read and diff configuration, read or follow logs, trace a single request or user across services, dump goroutines, and watch an instance's in-flight request activity;
- **act on it**: set a process's log level at runtime, take instances out of and back into rotation for maintenance, drive the background jobs runner (list and enqueue jobs, trigger or cancel runs), and impersonate a user.

On the same host it needs no login, it is reached over a Unix socket and authorized by the caller's OS identity; remotely, an operator first steps up to a short-lived admin token.

Jobs is the first example of the intended pattern: typed admin API methods whose implementation reuses the invoke channel underneath, fanning out to every instance for live state, or targeting one when the state is shared.


jessegeens:
https://github.com/cs3org/reva/pull/5773

PR #5695, opened 2026-07-05
Bugfix: normalize remote user ids in OCM

MahdiBaghbani:
## Summary

Cross-vendor OCM shares fail when Reva is the **sender** and the peer is a
non-conformant OCM server. The OCM spec requires the invite `userID` to be the
**bare** identifier of the user at their OCM Server, with the host carried
separately in `recipientProvider`. Some peers instead send a fully-qualified
`userID`:

- oCIS sends `id@host`
- OpenCloud sends `id@https://host`

Reva stored that qualified string verbatim as the federated user's opaque id,
then re-appended the provider domain when building `shareWith`, producing
malformed recipients:

- `id@host@host`
- `id@https://host@host`

Receivers cannot resolve those to a local user, so the share silently never
materializes even though the `/shares` POST returns `200 OK`.

Observed on the OCM Test Suite:

- CERNBox -> oCIS (FAIL): `shareWith: "f7fbf8c8-...@ocis2.docker@ocis2.docker"`
- CERNBox -> OpenCloud (FAIL): `shareWith: "60708dda-...@https://opencloud2.docker@opencloud2.docker"`
- CERNBox -> CERNBox (PASS): `shareWith: "f7fbf8c8-...@cernbox2.docker"`

### DIFF
--- a/internal/http/services/opencloudmesh/ocmd/specs.go
+++ b/internal/http/services/opencloudmesh/ocmd/specs.go
@@ -424,3 +424,50 @@ func GetUserIdFromOCMAddress(user string) (*userpb.UserId, error) {
 		Type: userpb.UserType_USER_TYPE_FEDERATED,
 	}, nil
 }
+
+// TrimOCMScheme removes a leading http(s):// scheme from an OCM host string.
+// OCM Addresses are not URIs and MUST NOT carry a scheme, but some servers
+// (Nextcloud, oCIS, OpenCloud) include one; we strip it defensively.
+func TrimOCMScheme(host string) string {
+	host = strings.TrimPrefix(host, "https://")
+	host = strings.TrimPrefix(host, "http://")
+	return host
+}
+
+// NormalizeRemoteUserID returns the bare OCM identifier for a remote user.
+//
+// Per the OCM spec the invite `userID` MUST be the bare identifier of the user
+// at their OCM Server, and the host travels separately in `recipientProvider`.
+// Some non-conformant servers append the host to `userID` anyway (oCIS sends
+// "id@host", OpenCloud sends "id@https://host"). If stored verbatim, CERNBox
+// keeps that qualified string as the OpaqueId and later re-appends the provider
+// domain when building `shareWith`, producing "id@host@host" (or with a scheme),
+// which the receiver cannot resolve to a local user.
+//
+// We strip a trailing "@<provider>" suffix ONLY when it matches the already-known
+// provider domain, repeating to collapse accidental double-qualification. A
+// spec-conformant identifier that legitimately contains '@' (e.g. an email local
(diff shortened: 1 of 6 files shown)


internal/http/services/opencloudmesh/ocmd/specs.go
```diff
@@ -424,3 +424,50 @@ func GetUserIdFromOCMAddress(user string) (*userpb.UserId, error) {
...
+// "id@host", OpenCloud sends "id@https://host"). If stored verbatim, CERNBox
+// keeps that qualified string as the OpaqueId and later re-appends the provider
+// domain when building `shareWith`, producing "id@host@host" (or with a scheme),
+// which the receiver cannot resolve to a local user.
+//
+// We strip a trailing "@<provider>" suffix ONLY when it matches the already-known
+// provider domain, repeating to collapse accidental double-qualification. A
+// spec-conformant identifier that legitimately contains '@' (e.g. an email local
+// part such as "a@b.org" belonging to a different provider) is left untouched.
+func NormalizeRemoteUserID(userID, providerDomain string) string {
```
glpatcern, line 451: How does this and the other new functions play with the existing `GetUserIdFromOCMAddress()`? Seems we have some duplication now...

glpatcern (changes requested):
Thanks for finding this out! I have a comment, can we consolidate?

internal/http/services/opencloudmesh/ocmd/specs.go
MahdiBaghbani: My read is that `GetUserIdFromOCMAddress()` should stay as the small parser for a full OCM address (`id@provider`), while `NormalizeRemoteUserID()` is the repair step for peers that put the provider back into the `userID`.

I kept them separate because in the invite flow we do not always have a full address, we have `userID` and `recipientProvider` as two fields.

That said, I agree the current shape can be cleaner. I will make the parser use the same scheme trimming helper, and then use the normalizer at the persistence boundary so direct `AcceptInvite` calls do not store a qualified opaque id again.

MahdiBaghbani: One thing I am not sure about: do we want to handle backwards compatibility for already stored rows in `ocm_remote_users`? Older rows may have `opaque_user_id = id@host`, while the new code will look up the bare `id`.

I can add a migration/backfill for those rows, or keep that as a focused follow-up if you prefer.

glpatcern: No don't worry about that, we can manage the small set of rows ourselves - this so far did not get exposed to users outside our own team

glpatcern (approved):
Cool, that's good enough. Will merge right after we tag a release

PR #5696, opened 2026-07-05
Bugfix: Resolve single-file OCM shares over WebDAV

MahdiBaghbani:
## Summary
This is a cascade PR, it relies on the previous #5695, I hope this ends up being more review friendly than a giant PR.

Cross-vendor OCM shares of a **single file** never became visible on the
receiver when Reva was the sender. The share was created and accepted, but the
receiver's "shared with me" listing silently dropped it. The root cause is a
path-doubling bug in the `ocmoutcoming` storage driver that made the receiver's
WebDAV `PROPFIND` return HTTP 500.

## Root cause

When a single file is shared over OCM, the shared resource is the file itself.
Remote receivers (oCIS, OpenCloud) mount the share as a directory and address
the file as `<share>/<name>`, so the incoming WebDAV request carries the file's
own name as a relative path.

`translateOCMShareResourceToCS3Ref` resolved the share token to the shared
resource (the file) and then unconditionally joined the relative path onto it:

    filepath.Join("/home/einstein/report.txt", "report.txt")
      => "/home/einstein/report.txt/report.txt"

That path does not exist, so the stat failed (CS3 code 15) and the `PROPFIND`
returned HTTP 500. Receivers treat a failed stat as a missing share and drop it
from the listing, so the file never appeared even though the share existed and
had been accepted.

Observed on the receiver:

    graph: "could not stat received ocm share, skipping"
           error="stat:PROPFIND /...report.txt: 500"

and on the sender:

    ERR error statting: path:"/home/einstein/report.txt/report.txt"
    PROPFIND /remote.php/dav/ocm/<share>/report.txt -> 500

### DIFF
--- a/changelog/unreleased/fix-ocm-single-file-webdav-path.md
+++ b/changelog/unreleased/fix-ocm-single-file-webdav-path.md
@@ -1,0 +1,25 @@
+Bugfix: Resolve single-file OCM shares correctly over WebDAV
+
+This fixes cross-vendor OCM shares failing to appear on the receiver when Reva
+is the sender and the shared resource is a single file. Remote receivers
+(oCIS, OpenCloud) mount an OCM share as a directory and address the shared file
+as `<share>/<name>`, so the incoming WebDAV request carries the file's own name
+as a relative path.
+
+The `ocmoutcoming` storage driver resolved the share token to the shared
+resource (the file itself) and then unconditionally joined the relative path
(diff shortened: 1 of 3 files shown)


glpatcern (dismissed):
Great, I'm formally approving but will have to re-approve once the underlying PR is merged. And yes stacked PRs would be awesome here...

glpatcern (approved):

PR #5698, opened 2026-07-06
Bugfix: Normalize user ids on incoming OCM shares

MahdiBaghbani:
## Summary

On the receiving side of an OCM share (`POST /ocm/shares`), CERNBox failed to
create incoming shares from oCIS and OpenCloud senders. Both peers send OCM
address fields where the host is glued onto the user id instead of kept separate,
and the incoming-share handler could not resolve those ids to users.

This is the receiver-side counterpart to #5695, which fixed the same class of
malformed address on the sending side (`FormatOCMUser`).

## Root cause

The OCM spec expects an address as `<bare-id>@<provider>`. Two peers deviate, and
by the time the payload reaches us the id is doubled:

- oCIS sends `id@host@host` (no scheme)
- OpenCloud sends `id@https://host@host` (with scheme)

`GetUserIdFromOCMAddress` only strips the final `@host`, so the parsed opaque id
keeps a leftover host suffix. That leftover breaks two different lookups, so the
two vendors fail at two different points:

- `shareWith` (the local recipient): the residual host means the local user
  lookup misses and the request returns `404 user not found`. This is what oCIS
  hit first, because its `shareWith` is doubled.
- `sender` / `owner` (the remote sharer): the residual host means the accepted-user
  lookup misses (we store the accepted remote user with a bare id during the
  invitation flow), and the request returns `401 unauthenticated`. This is what
  OpenCloud hit, because its `shareWith` happened to be clean but its
  `sender`/`owner` were doubled.

Because oCIS also sends doubled `sender`/`owner`, fixing only `shareWith` would
have moved oCIS from the 404 straight to the same 401. All three fields need the
same treatment.

### DIFF
--- a/changelog/unreleased/fix-ocm-incoming-share-user-id-normalization.md
+++ b/changelog/unreleased/fix-ocm-incoming-share-user-id-normalization.md
@@ -1,0 +1,27 @@
+Bugfix: Normalize user ids on incoming OCM shares
+
+This fixes cross-vendor OCM shares failing on the receiving side when Reva is
+the receiver and the peer is a non-conformant OCM server. It is the
+receiver-side counterpart to the sender-side normalization in #5695.
+
+The OCM spec expects an address as `<bare-id>@<provider>`, with the host kept
+separate from the user id. Some peers glue the host onto the id: oCIS sends
+`id@host` and OpenCloud sends `id@https://host`, and by the time it reaches the
+`POST /ocm/shares` handler the id is doubled up as `id@host@host` (with or
+without a scheme).
+
+`GetUserIdFromOCMAddress` only strips the final `@host`, so the parsed opaque id
+kept a leftover host suffix that broke user resolution, and the two vendors
+failed at two different points. For `shareWith` (the local recipient) the
+leftover host made the local user lookup miss and the request returned 404 user
+not found. For `sender` and `owner` (the remote sharer) it made the accepted-user
+lookup miss, because the accepted remote user is stored with a bare id during the
+invitation flow, and the request returned 401 unauthenticated.
+
+The incoming-share handler now parses these three addresses through a helper that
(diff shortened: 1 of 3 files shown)


MahdiBaghbani:
## What this fixes

I was getting CERNBox to share with oCIS and OpenCloud, and incoming shares
kept failing. The cause is how they write the remote user id: oCIS sends
`id@host` and OpenCloud sends `id@https://host`, so by the time it reaches our
`/ocm/shares` handler it is doubled up as `id@host@host`. We only strip the
last `@`, so a stray host stayed stuck to the id and the user lookup missed
(404 for the recipient, 401 for the sender/owner).

## Why I did it this way

I am not trying to be strict about the spec here. The id is opaque, so a peer
is technically allowed to shape it like this. My goal is just to make CERNBox side
tolerant so cross-vendor sharing works today, and let the others tidy up their
addresses whenever they get to it.

For servers that already send a bare id with the host kept separate, we still
do the right thing: store it as-is, send `id@host`, resolve it cleanly.
CERNBox-to-CERNBox and Nextcloud are unaffected.

## One honest caveat

We only trim a trailing host when it matches the provider we already know. So
if a fully conformant server ever uses an id that genuinely ends in its own
host (`id@idp@host` with `idp == host`) and expects it back untouched, we would
over-trim it. I think that is not the case today, but it is real. If we hit such a peer,
the right fix is to match on the invite `userID` verbatim instead of reparsing,
and I am happy to do that follow-up then.

If anything here does not match what you see, say so and we can dig in :)

MahdiBaghbani:
I cannot wait for [Stacked PR](https://github.github.com/gh-stack/introduction/overview) to become mainstream (and out of private beta)

glpatcern (commented):
This also looks good, to be reevaluated once we resolve the underlying PRs

glpatcern (approved):

PR #5699, opened 2026-07-06
OCM: Store share after remote operations succeed

rawe0:
This commit introduces a behaviour where the OCM shares are only stored locally if the remote system accepts them first, meaning we don't have any orphaned OCM shares locally.

This is done by exposing the GenerateID function via the repository, so that the `ocmshareprovider` can generate an ID before storing the share, which was previously done in the ocm sql driver when the storing of the share happened.

### DIFF
--- a/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
+++ b/internal/grpc/services/ocmshareprovider/ocmshareprovider.go
@@ -313,20 +313,38 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
 		AccessMethods: req.AccessMethods,
 	}
 
-	// TODO(lopresti) this is to be persisted after sending to remote,
-	// which requires the repo to expose a GenerateID() method to create
-	// the share here as it's sent in the payload
-	ocmshare, err = s.repo.StoreShare(ctx, ocmshare)
-	if err != nil {
-		if errors.Is(err, share.ErrShareAlreadyExisting) {
-			return &ocm.CreateOCMShareResponse{
-				Status: status.NewAlreadyExists(ctx, err, "share already exists"),
-			}, nil
-		}
+	// Since the share is persisted only after the remote server has accepted it,
+	// check upfront whether it already exists: this avoids sending it to the
+	// remote (and leaving a dangling share there) only to fail on the local
+	// duplicate afterwards.
+	switch _, err := s.repo.GetShare(ctx, user, &ocm.ShareReference{
+		Spec: &ocm.ShareReference_Key{
+			Key: &ocm.ShareKey{
+				Owner:      info.Owner,
+				ResourceId: req.ResourceId,
+				Grantee:    req.Grantee,
+			},
+		},
+	}); {
+	case err == nil:
+		return &ocm.CreateOCMShareResponse{
+			Status: status.NewAlreadyExists(ctx, share.ErrShareAlreadyExisting, "share already exists"),
+		}, nil
+	case !errors.Is(err, share.ErrShareNotFound):
 		return &ocm.CreateOCMShareResponse{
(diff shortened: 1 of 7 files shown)


pkg/ocm/share/share.go
```diff
@@ -30,7 +30,13 @@ import (
 
 // Repository is the interface that manipulates the OCM shares repository.
 type Repository interface {
-	// StoreShare stores a share.
+	// GenerateID generates a unique ID for a new share, so that the share can
+	// be referenced (e.g. in the payload sent to the remote OCM server) before
+	// it is persisted with StoreShare.
+	GenerateID(ctx context.Context) (*ocm.ShareId, error)
```
glpatcern, line 36: The interface is good, but we need to implement it also for the JSON driver

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -428,9 +414,24 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
 		}
 	}
 
+	stored, err := s.repo.StoreShare(ctx, ocmshare)
+	if err != nil {
+		// the remote server has already accepted the share, so a failure here
+		// leaves a dangling share on the remote end
+		log.Error().Err(err).Str("shareid", ocmshare.Id.OpaqueId).Msg("remote share created but persisting it locally failed")
+		if errors.Is(err, share.ErrShareAlreadyExisting) {
```
glpatcern: Shouldn't we prevent this from happening in the first place? How do we do this for normal shares?
The dangling share on the remote end is a bit of a burden...

glpatcern (commented):
Some comments for now, happy to discuss

pkg/ocm/share/share.go
rawe0: It is implemented for the json driver as well

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
rawe0: I've updated it now so it checks if the share already exists locally before creating the remote, if so it returns already exists to the remote if that's the case. I think this is the only check we can do before creating it remotely. 

If the remote succeeds and the DB/Shareprovider is down for some reason we will have a remote dangling share still until we implement the notifications so we can remove it remotely if it fails locally.

glpatcern: Very good, that is acceptable and indeed once we implement notifications we can improve this and all other areas where we may have dangling references.

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -313,20 +313,38 @@ func (s *service) CreateOCMShare(ctx context.Context, req *ocm.CreateOCMShareReq
...
 	}
 
-	// TODO(lopresti) this is to be persisted after sending to remote,
-	// which requires the repo to expose a GenerateID() method to create
-	// the share here as it's sent in the payload
-	ocmshare, err = s.repo.StoreShare(ctx, ocmshare)
+	// Since the share is persisted only after the remote server has accepted it,
+	// check upfront whether it already exists: this avoids sending it to the
+	// remote (and leaving a dangling share there) only to fail on the local
+	// duplicate afterwards.
```
glpatcern:
```suggestion
	// duplicate afterwards. This also allows to check that the repo backend is online,
	// and fail early if it is not.
```

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
glpatcern: ...but then this check can be dropped from here

glpatcern (approved):

PR #5701, opened 2026-07-08
Enhancement: Add webapp launch support to sciencemesh OpenInApp

MahdiBaghbani:
This PR is a part of https://github.com/cs3org/OCM-STA/issues/20


glpatcern:
Great to see this coming up! You can drop the `resolveTemplate()` function,  it's a leftover that is not needed any longer - the URI is expected to point to the root of the Web app "space" and you append the relative path

glpatcern (approved):

glpatcern:
Converting to draft as discussed with Mahdi: this works but it's based on assumptions on the frontend, whereas we want to properly populate the libregraph payload and make web expose it only for `webapp` shares, not for all remote shares

PR #5704, opened 2026-07-10
add support for making PROPFIND support responses relative to a basepath

jessegeens:


jessegeens:
Not needed in the end

PR #5707, opened 2026-07-10
Takeout

Tristan-C789:
# CERNBox Takeout

Users currently have no convenient, self-service way to obtain a full, offline copy of everything they store in their CERNBox personal space. This is needed for personal backups, offboarding (e.g. leaving CERN), and data-access requests. "Takeout" lets a user trigger an export from the web interface; the system packages their space into one or more archives asynchronously, stores them in a pre-configured EOS location, and notifies the user by email with download links when the export is ready. Exports expire automatically after a configurable retention period.

## Features

This PR mplements the core takeout logic as a Reva background job. Given a user, a max archive size, and a format, the job collects the contents of the user's personal space, packages them into one or more archives of the requested format, splitting output so that no single archive exceeds the configured max size, and uploads the artifact(s) to a pre-configured EOS folder.

It exposes two HTTP endpoints: one to create a new takeout job (taking max size and format), and one to get the status of the current/last takeout for the authenticated user. The create endpoint enforces one active job per user, if a takeout is already running, it returns the existing job rather than starting a new one. The status endpoint returns the job state and, when completed, the download link(s) and expiry. In particular:

- POST endpoint creates a takeout for the authenticated user with body {max_archive_size, format}; unsupported size/format is rejected with a clear error.
- If the user already has an active takeout, the create endpoint returns the existing job's id/status instead of starting a second one.
- GET status endpoint returns the current state (none / queued / running / completed / failed) and progress if available.
- When completed, the status response includes the download link(s) and expiry time.
- When failed, the status response includes an error/cause indication.


pkg/takeout/cleanup/cleanup.go
```diff
@@ -0,0 +1,134 @@
...
+	CleanupSchedule string
+	// The minimal delay in hours before a directory is removed by the cleanup job
+	CleanupDelay int64
+}
+
+func (cj *CleanupJob) Run(ctx context.Context) error {
+	appctx.GetLogger(ctx).Info().Msg("running cleanup job")
+
+	// Setup gateway
+	gtw, err := pool.GetGatewayServiceClient(pool.Endpoint("localhost:9142"))
```
diocas: Why hardcoded? Temporary while doing dev?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+type params struct {
+	MaxArchiveSize int64  `mapstructure:"maxArchiveSize"` // < 10GB
+	ArchiveFormat  string `mapstructure:"archiveFormat"`  // One of ["zip", "tgz"]
+	Username       string `mapstructure:"username" validate:"required"`
+}
+
+// Run walks the user's archives the userspace
+func (j *job) Run(ctx context.Context, p rjobs.Params) (rjobs.Params, error) {
+	// Decode run parameters
+	pp := params{
```
diocas: We are setting defaults in the post request and here. I would do it in a single place.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+		// Default values in case they're not provided
+		MaxArchiveSize: 2 << 30, // 2 GiB
+		ArchiveFormat:  "zip",
+	}
+	if err := mapstructure.Decode(map[string]any(p), &pp); err != nil {
+		return nil, errors.Wrap(err, "takeout: decoding params failed")
+	}
+	if pp.MaxArchiveSize > 10<<30 {
```
diocas: Should this be a service configuration?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+		return nil, errors.Errorf("takeout: MaxArchiveSize cannot be larger than 10GB")
+	}
+	j.log.Info().Msgf("takeout: using parameters %+v", pp)
+
+	// Setup gateway
+	gtw, err := pool.GetGatewayServiceClient(pool.Endpoint("localhost:9142"))
```
diocas: as above, hardcoded gw

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "takeout: user authentication failed")
+	}
+	adminCtx, err := j.authenticate(ctx, gtw, j.conf.TakeoutAdminUsername)
+	if err != nil {
+		return nil, errors.Wrap(err, "takeout: admin authentication failed")
+	}
+
+	// Setup downloader
+	tr := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
```
diocas: Config?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	dl := downloader.NewDownloader(gtw, hc)
+
+	// Setup upload client
+	upHC := httpclient.New(httpclient.RoundTripper(tr), httpclient.Timeout(0))
+
+	// Setup walker
+	wk := walker.NewWalker(gtw)
+
+	// Set the source & destination directories
+	root := "/eos/user/" + pp.Username[0:1] + "/" + pp.Username
```
diocas: This should be configured somewhere else. I would put this as a param of the job, potentially set at the http layer (and full freedom via any other means, e.g. so that in the future we allow other spaces to be taken out).

And I would use functions we already have to derive a user home directory, instead of hardcoding it.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+		err = j.createTgzArchives(userCtx, adminCtx, root, archPath, wk, dl, gtw, upHC, pp.MaxArchiveSize)
+		if err != nil {
+			return nil, errors.Wrap(err, "takeout: tgz archive could not be created")
+		}
+	default:
+		return nil, errors.Errorf("takeout: %s is not a supported archive format", pp.ArchiveFormat)
+	}
+
+	// Share the folder containing the archives through a public link
+	token, err := j.createPublicShare(adminCtx, gtw, archPath)
```
diocas: Shall we protect it? This might contain personal data, so it shouldn't be exposed not even for a few hours.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+		if copyErr != nil {
+			return copyErr
+		}
+		if closeErr != nil {
+			return closeErr
+		}
+
+		return gw.Flush()
+	})
+	if err != nil {
```
diocas: Shall we ignore/skip errors with files and write them (the error) to a file (that the user can check)?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+
+		// Get relative path of the current file
+		fileName, err := filepath.Rel(filepath.Dir(rootPath), currentPath)
+		if err != nil {
+			return err
+		}
+		fileName = filepath.ToSlash(fileName)
+
+		// Cut the archive if adding the current file could exceed maxArchiveSize
+		if !isDir && cw.n > 0 && cw.n+int64(info.Size) > maxArchiveSize {
```
diocas: I think we discussed and this was a pending issue: what happens if a single file is bigger than max archive size? I would at least create an archive with that file, even if bigger, unless you can chunk it.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	// Finalize last archive
+	if err := flush(); err != nil {
+		j.log.Err(err).Msg("takeout: archive upload failed")
+		return err
+	}
+
+	return nil
+}
+
+func (j *job) createZipArchives(userCtx, adminCtx context.Context, rootPath, archPath string, wk walker.Walker, dl downloader.Downloader, gtw gateway.GatewayAPIClient, hc *httpclient.Client, maxArchiveSize int64) error {
```
diocas: I'm seeing many similarities between this and the tar method. There are things that are different, like headers, but we can split this, others are not so clear, like deferring close or closing and checking the error. Given that the logic seems quite similar otherwise, wouldn't there be a way to make this a single function?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	// Finalize last archive
+	if err := flush(); err != nil {
+		j.log.Err(err).Msg("takeout: archive upload failed")
+		return err
+	}
+
+	return nil
+}
+
+func (*job) createTakeoutContainer(adminCtx context.Context, gtw gateway.GatewayAPIClient, arch_path string) error {
```
diocas: We should add comments explaining why we are deleting. Kinda obvious, but might be overlooked. Just explain that we keep a takeout _per day_ (but we allow more than one takeout per user, should we?)

diocas (changes requested):
I didn't check the archiver code, but I have the impression that this is a copy of it. Didn't we discuss that we would integrate the current archive code in the takeout, so that we only set the logic in a single place? The advantage of the takeout would be the possibility to keep track of error instead of just killing the connection. And the fact that we can chunk, but maybe we can achieve a single logic for both (infinite chunk size would have the same output as the archiver, no?)

internal/http/services/takeout/takeout.go
```diff
@@ -0,0 +1,244 @@
...
+func (s *svc) Prefix() string {
+	return s.conf.Prefix
+}
+
+// Unprotected sets the unprotected paths
+func (s *svc) Unprotected() []string {
+	return nil
+}
+
+// Handler propagates the request dependanding on the suffix
```
jessegeens: Typo :)

internal/http/services/takeout/takeout.go
```diff
@@ -0,0 +1,244 @@
...
+	s.log.Info().Msgf("takeout: takeout job %s enqueued", runId)
+	w.WriteHeader(http.StatusOK)
+}
+
+func (s *svc) handleGet(w http.ResponseWriter, r *http.Request) {
+	// Get job runner
+	runner := rjobs.Default()
+	if runner == nil {
+		s.log.Error().Msg("takeout: could not find runner")
+		w.WriteHeader(http.StatusFailedDependency)
```
jessegeens, line 208: You can make this one `InternalServerError`. Taking a 400-type code implies that this is something the client can solve, but this error is really on the server side

internal/http/services/takeout/takeout.go
```diff
@@ -0,0 +1,244 @@
...
+	user := appctx.ContextMustGetUser(r.Context())
+	jobs, err := runner.ListByOwner(r.Context(), user.Username, rjobs.ListFilter{Job: "takeout"})
+	if err != nil {
+		s.log.Err(err).Msg("takeout: could not list user's jobs")
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if len(jobs) == 0 {
+		s.log.Debug().Msgf("takeout: user %s has no takeout job listed", user.Username)
+		w.WriteHeader(http.StatusBadRequest)
```
jessegeens: 404 instead maybe? since "no jobs found"?

internal/http/services/takeout/takeout.go
```diff
@@ -0,0 +1,244 @@
...
+		return
+	}
+
+	// Handle latest job
+	st := jobs[0]
```
jessegeens, line 227: is the latest one always in front?

pkg/takeout/cleanup/cleanup.go
```diff
@@ -0,0 +1,134 @@
...
+	// Get container list
+	containerRes, err := gtw.ListContainer(ctx, &provider.ListContainerRequest{
+		Ref: &provider.Reference{
+			Path: cj.conf.TakeoutPath,
+		},
+	})
+	if err != nil {
+		return errors.Wrap(err, "cleanup: could not get takeout containers")
+	}
+	if containerRes.Status.Code != rpc.Code_CODE_OK {
```
jessegeens, line 96: Should never be the case, but make sure `ContainerRes.Status != nil`, otherwise you get a panic.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	})
+	switch {
+	case err != nil:
+		return "", err
+	case statRes.Status.Code != rpc.Code_CODE_OK:
+		return "", errtypes.InternalError(statRes.Status.Message)
+	}
+
+	// Create the read-only public link
+	shareRes, err := gtw.CreatePublicShare(ctx, &link.CreatePublicShareRequest{
```
jessegeens: Should we set a password on the link? Or / and an expiry date?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,595 @@
...
+	)
+
+	// Initiate the file upload request
+	req := &provider.InitiateFileUploadRequest{
+		Ref: &provider.Reference{Path: archFile},
+		Opaque: &types.Opaque{
+			Map: map[string]*types.OpaqueEntry{
+				"Upload-Length": {
+					Decoder: "plain",
+					Value:   []byte("-1"),
```
jessegeens: Why `Upload-Length` -1?

jessegeens (changes requested):

internal/http/services/takeout/takeout.go
Tristan-C789: Specification says so : "ListByOwner returns the runs created for the given user, most recently enqueued first."

pkg/takeout/takeout.go
Tristan-C789: It's a leftover from when we didn't use streams and had to specify the size of each request. It's been removed now.

pkg/takeout/takeout.go
Tristan-C789: Currently, if a file is a larger than MaxArchiveSize, it's put in its own archive by itself, no matter what size that archive ends up being. There's no easy way to chunk without delegating some dechunking work to the user, which we want to avoid as Takeout is supposed to be as easy to use as possible for anyone.

pkg/takeout/takeout.go
Tristan-C789: I will try to merge the Tar, Zip and archiver logic into a single new package on a new branch ; if it's not too different, it will solve this issue naturally.

PR #5708, opened 2026-07-10
Refactor notifications system

jessegeens:
This PR refactors the notification system into an event-driven architecture, where accumulated events are guaranteed to be handled on the same instance using a lease system. It implements ADR `general/0007-notifications-refactor.md`.


diocas (approved):
Conditional approval. Proper review will be made after and fixes committed separately.

internal/grpc/services/gateway/events.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	scopes, ok := appctx.ContextGetScopes(ctx)
+	if _, isMachine := scopes[scope.MachineScope]; !ok || !isMachine {
+		return &gateway.PublishEventResponse{
+			Status: status.NewPermissionDenied(ctx, errtypes.PermissionDenied("machine auth required"), "gateway: PublishEvent requires machine authentication"),
+		}, nil
+	}
+
+	u, ok := appctx.ContextGetUser(ctx)
```
diocas, line 59: why do we need to have a context user? wondering if we can have some internal "system" events generated... (and that we might not want to rare limit?)

diocas: or in that case we use an internal user?

internal/http/services/appprovider/mentions.go
```diff
@@ -147,35 +137,64 @@ func (s *svc) handleMentions(w http.ResponseWriter, r *http.Request) {
...
-				"path":                 statRes.Info.Path,
-				"fileID":               documentID,
-				"appName":              req.AppName,
-				"commentText":          req.CommentText,
-				"anchorText":           req.AnchorText,
-				"documentURL":          documentURL,
-				"eventID":              req.EventID,
-			},
-		})
+		recipients = append(recipients, recipient.Mail)
```
diocas, line 151: why not the user instead of the mail? the event dispatcher should decide if it wants to send by email or not.

internal/http/services/owncloud/ocdav/notifications.go
```diff
@@ -0,0 +1,180 @@
...
+
+	templateData := map[string]any{
+		"resource_id":   spaces.EncodeToStringifiedResourceID(info.GetId()),
+		"resource_name": info.GetName(),
+		"resource_path": info.GetPath(),
+		"resource_type": info.GetType().String(),
+		"mime_type":     info.GetMimeType(),
+		"size":          info.GetSize(),
+	}
+	if publicShare != nil {
```
diocas, line 61: why haven't we returned above if publicShare is nil?

diocas: It would avoid a lot of if's in this method...

internal/http/services/owncloud/ocdav/notifications.go
```diff
@@ -0,0 +1,180 @@
...
+
+	res, err := client.GetUser(ctx, &userpb.GetUserRequest{
+		UserId:                 owner,
+		SkipFetchingUserGroups: true,
+	})
+	if err != nil || res.GetStatus().GetCode() != rpc.Code_CODE_OK || res.GetUser() == nil {
+		log.Debug().Err(err).Msg("failed to resolve public share owner for upload notification")
+		return ""
+	}
+	return res.GetUser().GetMail()
```
diocas, line 126: here it makes more sense that we use emails instead of users (as the extra field asks for emails, at least _for now_...). but maybe for the owner itself, this would be a user.

pkg/notifications/accumulation/sql.go
```diff
@@ -0,0 +1,283 @@
...
+	defaultMaxItems = 100
+)
+
+// PendingEvent is one accepted envelope awaiting accumulation. One event is one
+// row. It is grouped with the other events sharing its dedup key by the Group
+// that owns them.
+type PendingEvent struct {
+	gorm.Model
+	EventID  string `gorm:"uniqueIndex;size:64"`
+	DedupKey string `gorm:"index;size:512"`
```
diocas, line 47: I guess this cannot be a fk because we add the group afterwards?

pkg/notifications/accumulation/sql.go
```diff
@@ -0,0 +1,283 @@
...
+		if res.RowsAffected == 0 {
+			// Redelivered event: leave the group untouched.
+			return tx.First(&group, "dedup_key = ?", envelope.DedupKey).Error
+		}
+
+		created := Group{
+			DedupKey:      envelope.DedupKey,
+			FirstSeen:     now,
+			LatestSeen:    now,
+			FlushAfter:    now.Add(time.Duration(envelope.Accumulation.WindowSeconds) * time.Second),
```
diocas, line 122: what if we extended the flush time on updates as well? we still have max items to prevent infinite accumulation.

pkg/notifications/accumulation/sql.go
```diff
@@ -0,0 +1,283 @@
...
+	return res.RowsAffected == 1, nil
+}
+
+// LockDueForFlush transitions a due group owned by this box to flushing.
+func (s *SQLStore) LockDueForFlush(ctx context.Context, dedupKey, owner string, now time.Time) (bool, error) {
+	res := s.db.WithContext(ctx).Model(&Group{}).
+		Where("dedup_key = ?", dedupKey).
+		Where("status = ?", statusOpen).
+		Where("lease_owner = ?", owner).
+		Where(s.db.Where("flush_after <= ?", now).Or("item_count >= max_items")).
```
diocas, line 166: I wonder if max items should not be a reason to flush, but only to ignore new messages. otherwise, if we have a huge flux of messages, we will accumulate but still send all of them (when maybe discarding would be a better option).

pkg/notifications/accumulation/sql.go
```diff
@@ -0,0 +1,283 @@
...
+		var pending []PendingEvent
+		if err := tx.Where("dedup_key = ?", dedupKey).
+			Where("status = ?", statusPending).
+			Order("created_at ASC, id ASC").
+			Find(&pending).Error; err != nil {
+			return err
+		}
+
+		if len(pending) == 0 {
+			if err := tx.Unscoped().Where("dedup_key = ?", dedupKey).Delete(&Group{}).Error; err != nil {
```
diocas, line 228: there's a chance an event arrives in between, deleting and selecting, no? would a fk resolve the issue?

pkg/notifications/handlers/email.go
```diff
@@ -0,0 +1,337 @@
...
+
+const validTemplateNameRegex = "[a-zA-Z0-9-]"
+
+// EmailConfig configures the email notification handler.
+type EmailConfig struct {
+	SMTPAddress    string         `docs:";The hostname and port of the SMTP server." mapstructure:"smtp_server"`
+	SenderLogin    string         `docs:";The email to be used to send mails."       mapstructure:"sender_login"`
+	SenderPassword string         `docs:";The sender's password."                    mapstructure:"sender_password"`
+	DisableAuth    bool           `docs:"false;Whether to disable SMTP auth."        mapstructure:"disable_auth"`
+	DefaultSender  string         `docs:"no-reply@cernbox.cern.ch;Default sender when not specified in the request." mapstructure:"default_sender"`
```
diocas, line 57: I would not let requests set the sender

pkg/notifications/handlers/email.go
```diff
@@ -0,0 +1,337 @@
...
+	host, portStr, ok := strings.Cut(h.conf.SMTPAddress, ":")
+	if !ok {
+		return errors.New("invalid SMTP address format: expected host:port")
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return fmt.Errorf("invalid SMTP port %s: %w", portStr, err)
+	}
+
+	dialer := &mail.Dialer{
```
diocas, line 255: do we need to create a new dialer every time we send an email?

internal/grpc/services/gateway/events.go
jessegeens: Yes indeed; this is a gateway method so it must come from "someone". System events can come from a system account

internal/http/services/appprovider/mentions.go
jessegeens: true, and then we can also resolve their preferences

pkg/notifications/accumulation/sql.go
jessegeens: Maybe. But `max_items` is already quite large. If you get 1000 files uploaded at once, this would only result in 10 emails. On the other hand if we choose the ignore option, we would not be able to to extend the flush time on new uploads

pkg/notifications/accumulation/sql.go
jessegeens: This could be a good idea. I would reduce the flush time to maybe 2 minutes then though. This would lead to more immediate feedback which is good, and possibly less emails if many things are being slowly uploaded.

pkg/notifications/handlers/email.go
jessegeens: So one sender for all emails? Or configurable per-template?

diocas: I would say for all (otherwise you need login/pwd for each as well).

PR #5727, opened 2026-07-23
Enhancement: Add new resource_admins column to projects table

pmedinar01:


diocas (approved):

glpatcern (commented):
How is that different from the admins? Can this be answered in the corresponding ADR record we already have for project spaces?

PR #5730, opened 2026-07-27
Enhancement: Add deleted status to projects table

pmedinar01:


glpatcern (commented):
As this is a GORM object, an entry is deleted if `deleted_at` is not NULL. Why would we need an additional status? and why would we override the GORM one?

pmedinar01:
> e?

As we might receive status changes to deleted from the resource portal ... we thought it was a way to keep track of precisely that.

@diocas

glpatcern:
> As we might receive status changes to deleted from the resource portal ... we thought it was a way to keep track of precisely that.

But then what is the workflow we're trying to address? We always agreed that for us, a "delete" action = we move to `archived`, we never delete from a user action. We have a delete in case of mistakes as you know.

diocas:
@glpatcern precisely because we do not delete (and archive instead), we need to keep track of the state that FIM expects... If it tells us to "delete", we say we did, but don't. And if it asks us again, we need to be able to say that it's deleted, even if it's not, to be consistent.

What I think is that this is CERN specific, so I'm not a fan of having this in Reva...

glpatcern:
(Sorry forgot to answer here) I see the need to follow FIM (and I actually agree with Diogo that this does not really belong to Reva). But still why can't we use the GORM `deleted_at` for this state? That is we soft-delete and effectively keep the record on the table, and on the back we archive the data. I still see an extra field redundant and in conflict with GORM.

PR #5733, opened 2026-07-28
Fix nilpointer in ListUsers

jessegeens:
Specifically, when the OrderBy parameter was omitted


diocas (approved):
We could just have a default in case this is not given, but fine...

PR #5735, opened 2026-07-28
fix nil map assignment error in appprovider

jessegeens:


diocas (approved):

glpatcern:
Good catch! but if that's targeting CodiMD I'd rather ignore the theme altogether, because CodiMD does not support form parameters - and actually works with GET, possibly not POST

PR #5738, opened 2026-07-30
Respond file name on PUT to upload-only public link

jessegeens:


diocas (approved):
Approved, but.. should we `sendCreatedResponse` only for specific links?

jessegeens:
I'm not sure. I mean right now it will only send on upload-only links. Adding an option for this makes the interface more complicated, and I don't see a downside to always doing that. On top of that, this would also be needed if we want to e.g. integrate with forms in WordPress (so that these can do unauthenticated uploads, and have the path in the form itself). 

I'll merge for now since you approved, but if you have a different opinion we can always change it later.

PR #5740, opened 2026-07-31
cephmount driver: sharing for external accounts. 

rawe0:
Adds missing functionality for shares to external accounts on ceph

* Grants to external accounts are now stored in a `user.reva.lwshare.<account>` xattr and are read back by ListGrants. 
* Operations on their behalf run as the service account configured with `external_accounts_user_name`/`_uid`/`_gid`, `cboxexternal` by default, the same account the EOS driver uses.
* What they are allowed to do is decided by Reva from the stored grant rather than by the permissions of that account.

* The ResourceInfo of a shared resource now also carries a name and an recursive etag instead of mtime, meaning that it will show changes in the subtree at the top level as well.

### DIFF
--- a/pkg/storage/fs/cephmount/cephmount.go
+++ b/pkg/storage/fs/cephmount/cephmount.go
@@ -437,14 +437,20 @@ func (fs *cephmountfs) fileAsResourceInfo(path string, info os.FileInfo, mdKeys 
 
 	owner, _ := user.LookupId(fmt.Sprint(stat.Uid))
 
+	externalPath := fs.fromChroot(path)
+
 	ri := &provider.ResourceInfo{
 		Type:     resourceType,
 		Id:       resourceId,
 		Checksum: &provider.ResourceChecksum{},
 		Size:     size,
 		Mtime:    &typepb.Timestamp{Seconds: uint64(info.ModTime().Unix())},
-		Path:     fs.fromChroot(path), // Convert chroot path back to external path
-		Owner:    &userv1beta1.UserId{OpaqueId: owner.Username},
+		Path:     externalPath,
+		// Clients list a received share by its name and etag, so both must be set
+		// even though the resource itself is addressed by path or id.
+		Name:  filepath.Base(externalPath),
+		Etag:  calcEtag(info, stat),
+		Owner: &userv1beta1.UserId{OpaqueId: owner.Username},
 		PermissionSet: &provider.ResourcePermissions{
 			GetPath:              true,
 			GetQuota:             true,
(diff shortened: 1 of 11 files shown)


pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -989,18 +1095,26 @@ func (fs *cephmountfs) ListGrants(ctx context.Context, ref *provider.Reference)
...
+
 	fullPath := filepath.Join(fs.chrootDir, path)
 
+	log := appctx.GetLogger(ctx)
+
+	// Grants to external accounts live in xattrs, not in the POSIX ACL.
+	glist = append(glist, fs.listExternalAccountGrants(ctx, fullPath)...)
+
 	// Use getfacl to read ACLs
 	cmd := exec.CommandContext(ctx, "getfacl", "--omit-header", "--numeric", fullPath)
```
jessegeens, line 1178: This fails if no ACLs are set? Or also on other occasions?

Because this looks like it might cause silent errors

pkg/storage/fs/cephmount/options.go
```diff
@@ -54,6 +68,17 @@ func (c *Options) ApplyDefaults() {
...
 	}
 
+	// Same service account the eos driver uses for external accounts.
+	if c.ExternalAccountsUserName == "" || c.ExternalAccountsUserUID == 0 || c.ExternalAccountsUserGID == 0 {
+		c.ExternalAccountsUserName = "cboxexternal"
+		c.ExternalAccountsUserUID = 193339
+		c.ExternalAccountsUserGID = 2766
+	}
+
+	if c.SpaceDepth == 0 {
```
jessegeens: It would be best to take this immediately from the storageprovider, instead of having to re-define this. I also did that for the EOS one recently, I think that commit has already been merged so you can take a look. This is in the registry code I think, because that's where the driver is loaded

jessegeens (changes requested):

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -98,6 +104,15 @@ func New(ctx context.Context, m map[string]any) (storage.FS, error) {
 	}
 
 	log := appctx.GetLogger(ctx)
+
+	if o.SpaceDepth <= 0 {
```
jessegeens: I would say this is invalid and should result in an error. That or you reset it to a default value

pkg/storage/fs/cephmount/cephmount.go
```diff
@@ -1515,6 +1741,135 @@ func (fs *cephmountfs) Unlock(ctx context.Context, ref *provider.Reference, lock
 	return nil
 }
 
+// externalAccountQualifier returns the xattr qualifier for a lightweight or
+// federated grantee. Those accounts only exist inside reva and must never be
+// resolved as local users.
+func externalAccountQualifier(g *provider.Grantee) (string, bool) {
```
jessegeens: maybe put this in `pkg/storage/fs/cephmount/external_accounts.go` ? Seems to me like it belongs there more. But feel free to ignore if you disagree

jessegeens (approved):

PR #5741, opened 2026-07-31
Orphan job

jessegeens:
This PR adds support for a periodic job that checks the shares and public links database for orphaned shares / links and marks them as such.


pkg/reconciliation/config.go
```diff
@@ -0,0 +1,82 @@
...
+	DryRun bool `mapstructure:"dry_run"`
+	// RunOnStart fires the job once as soon as the runner starts, instead of
+	// waiting a full interval for the first run.
+	RunOnStart bool `mapstructure:"run_on_start"`
+	// LogFile is the path the job writes its own log to. It takes "stdout" or
+	// "stderr" to write to the standard streams instead.
+	LogFile string `mapstructure:"log_file"`
+}
+
+// OpenLog opens a job's own log. A job keeps its own log rather than writing to
```
diocas, line 57: Should this (logging) be part of the job framework instead?

pkg/reconciliation/README.md
```diff
@@ -0,0 +1,65 @@
+# Deploying the reconciliation jobs
+
+The jobs span the share and the public link tables, so they are wired as their
+own serverless service rather than by one of the grpc share services. They read
+and write those tables directly through gorm, so they always use the `sql`
```
diocas: Why don't we instantiate a share provider, of wtv type we need? This way, we wouldn't assume the sql driver, which might change.

pkg/reconciliation/README.md
```diff
@@ -0,0 +1,65 @@
...
+The jobs runner hands a run a bare context, but we need a valid auth
+so we can do stat's etc. Therefore, each run mints itself a token for
+`service_user_name` and sends it with every call. The three `service_user_*`
+keys are thus required and it needs to have the proper permissions on EOS
+(for CERNBox, `cbox` does the trick).
+
+## Jobs
+
+`reconciliation.orphans` marks the shares and public links whose resource or
+recipient no longer exists.
```
diocas, line 51:
```suggestion
recipient no longer exists as orphans.
```

diocas: Side comment: should our orphan column be a timestamp? So that we can (hard) delete orphans that are older than X (like we might do for soft deletions older than X)

internal/serverless/services/reconciliation/reconciliation.go
```diff
@@ -0,0 +1,258 @@
...
+func New(ctx context.Context, m map[string]any) (_ rserverless.Service, err error) {
+	var c config
+	if err := cfg.Decode(m, &c); err != nil {
+		return nil, err
+	}
+	if len(c.Jobs) == 0 {
+		return nil, errors.Errorf("reconciliation: no jobs enabled, set jobs to %q", jobOrphan)
+	}
+
+	sm, err := sql.NewShareManager(ctx, c.DB)
```
diocas: Can this be configurable so that we allow different types of providers/managers? And dynamic? (maybe dynamic not, as it needs to be casted...)

pkg/reconciliation/orphan.go
```diff
@@ -0,0 +1,418 @@
...
+	KindPublicLink Kind = "publiclink"
+)
+
+// OrphanReason says why an item was marked orphaned.
+type OrphanReason string
+
+const (
+	// ReasonResourceMissing means the shared resource no longer exists. Its
+	// storage space being gone shows up here too, since the resource Stat then
+	// fails.
```
diocas: What about moving to trash bin? Should we include that or we allow restored files to keep their shares, just hiding them from the shares listing?

diocas: Either decision should be recorded as a comment.

pkg/reconciliation/orphan.go
```diff
@@ -0,0 +1,418 @@
...
+	// KindPublicLink is a public link.
+	KindPublicLink Kind = "publiclink"
+)
+
+// OrphanReason says why an item was marked orphaned.
+type OrphanReason string
+
+const (
+	// ReasonResourceMissing means the shared resource no longer exists. Its
+	// storage space being gone shows up here too, since the resource Stat then
```
diocas: By storage space, you also include the owner no longer exists? (a space should no longer exist if the owner leaves, but this is handled independently so maybe we should care about it here as well as a safeguard)

diocas (changes requested):

pkg/reconciliation/config.go
jessegeens: Perhaps, but usually these can go anywhere I think. Here it is more like a "journal" so that it's easy to revert in case we have any issue

pkg/reconciliation/README.md
jessegeens: Yeah but these methods are not exposed via the CS3API...

diocas: Fine, but maybe this should be put as "only sql driver supported for now".  And that "any other provider needs to implement the struct ShareStore" or smth.

internal/serverless/services/reconciliation/reconciliation.go
jessegeens: well we can cast and just check that it satisfies the right interface

pkg/reconciliation/README.md
jessegeens: I've fixed it by having the job load a configurable driver and then casting it to an interface with the methods it needs, so at least its swappable

pkg/reconciliation/orphan.go
jessegeens: no no, there is no check for the owner: only the resource and the recipient. Ive updated the comment, it was not very clear

pkg/reconciliation/README.md
jessegeens: Yes, but let's tackle that in another PR

diocas (approved):

PR #5748, opened 2026-08-05
Return fileid and etag in MKCOL response

### LINKED ISSUE #4769: Port #4767 to master
We may need to add something like:
```
w.Header().Set(HeaderOCFileID, resourceid.OwnCloudResourceIDWrap(newInfo.Id))
w.Header().Set(HeaderOCETag, newInfo.Etag)
```
To internal/http/services/owncloud/ocdav/mkcol.go in order to "comply" to Reva edge

btwshivam:
Fixes #4769

MKCOL now returns the `OC-FileId` and `OC-ETag` headers of the collection it just created, the same way PUT already does. Clients can skip the PROPFIND they currently have to send right after creating a folder.

to get there.. `storage.FS.CreateDir` now returns the new directory's `ResourceInfo` alongside the error. The storage provider puts the fileid and etag in the `CreateContainerResponse` opaque and ocdav turns them into headers. Drivers get the info from a local stat, which is cheaper than the gateway round trip a caller would otherwise make, and it keeps the fileid encoding identical to what PROPFIND reports so clients can match them up.

This changes the storage.FS interface, so every driver in the tree is updated and any out-of-tree driver will need the new signature. It is a one line change per driver: return nil, err instead of err, and return the info if a stat after creation succeeds.

It is best effort throughout. If the stat fails the directory still exists, MKCOL still returns 201, and the headers are simply absent. Public links get the fileid rewritten into the public namespace so they never leak the id of the resource behind the link.

### DIFF
--- a/changelog/unreleased/mkcol-fileid-etag.md
+++ b/changelog/unreleased/mkcol-fileid-etag.md
@@ -3,4 +3,10 @@ Enhancement: return fileid and etag in MKCOL response
 A successful MKCOL now returns the OC-FileId and OC-ETag headers of the
 newly created collection, so clients do not need a follow-up PROPFIND.
 
+The storage.FS CreateDir method was changed to return the created
+directory's ResourceInfo, which the storage provider passes along in the
+CreateContainerResponse opaque. This way no extra stat call through the
+gateway is needed to obtain the fileid and etag.
+
+https://github.com/cs3org/reva/pull/5748
 https://github.com/cs3org/reva/issues/4769
--- a/changelog/unreleased/mkcol-fileid-etag.md
+++ b/changelog/unreleased/mkcol-fileid-etag.md
@@ -1,0 +1,6 @@
+Enhancement: return fileid and etag in MKCOL response
+
+A successful MKCOL now returns the OC-FileId and OC-ETag headers of the
+newly created collection, so clients do not need a follow-up PROPFIND.
+
+https://github.com/cs3org/reva/issues/4769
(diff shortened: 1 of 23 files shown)


jessegeens:
Hi @btwshivam ,

Thanks for opening this PR. Unfortunately your implementation suffers from the same behaviour as https://github.com/cs3org/reva/pull/4767 : it still requires an extra stat, it's just that now the server does this instead of the client. Many clients don't need the FileID though, but the `MKCOL` would be slower for all of them like this. 

A better approach to this would be if the `CreateDir` method returned the created folder's id and etag directly.

btwshivam:
> Hi @btwshivam ,
> 
> Thanks for opening this PR. Unfortunately your implementation suffers from the same behaviour as #4767 : it still requires an extra stat, it's just that now the server does this instead of the client. Many clients don't need the FileID though, but the `MKCOL` would be slower for all of them like this.
> 
> A better approach to this would be if the `CreateDir` method returned the created folder's id and etag directly.

Hi @jessegeens thanks for the feedback, that makes sense.. i reworked the implementation:

* `storage.FS.CreateDir` now returns the created directory's `ResourceInfo` (`nil` if the driver can't provide it), and I updated all drivers.
* Since `CreateContainerResponse` doesn't expose an info field, the storage provider passes the fileid/etag via the response opaque, and `ocdav` sets the headers directly.. so there's no extra `Stat` through the gateway anymore.
* Clients that don't need the FileID don't pay any extra cost now.

For EOS, `mkdir` doesn't return metadata, so the driver still needs a single `fileinfo` call after creation. I also updated the public storage provider to translate the fileid into the public namespace, consistent with `Stat`
Let me know if you'd like anything adjusted further :)

btwshivam:
Screenshot from 2026-08-07 05-40-03

btwshivam:
@jessegeens Hi.. please take a look when you have some free time .. Thank you!

PR #5749, opened 2026-08-06
feat: implement CommitUpload and MarkProcessing for OCM received shares

f-firas:


jessegeens:
Hi @f-firas , 

Thanks for your PR. I see that you are merging from `owncloud:`, so you probably want to merge into their fork instead: https://github.com/owncloud/reva

PR #5389, opened 2025-11-01
feat(ocm): Strengthen OCM client security

MahdiBaghbani:
This PR is a part of https://github.com/cs3org/OCM-STA/issues/1
This PR depends on the https://github.com/cs3org/reva/pull/5385 to be merged first.

## Description
This PR introduces security measures to the OCM client to prevent common attack vectors such as SSRF, DNS rebinding, and resource exhaustion attacks. The implementation provides configurable security controls while maintaining backward compatibility with existing code.

PR #5752, opened 2026-08-08
Harden OCM federation against SSRF, spoofed IPs and TLS downgrade

btwshivam:
Three small unrelated things I hit while reading the OCM / ScienceMesh code, one
commit each.

**SSRF via `/sciencemesh/discover`** endpoint is unauthenticated
and fetches whatever host the body names, so it could be pointed at
`169.254.169.254` or anything internal. Discovery of caller-supplied domains now
goes through a client that refuses loopback, private, link-local, CGNAT and
metadata addresses.. check is in the dialer, so redirects, DNS rebinding and
the NAT64 prefix are covered too. Operator-configured targets keep the plain
client.

**Allow list trusts `X-Forwarded-For`**  sender IP was read from
the header whenever present, so a peer could pick which address it was checked as.
Now it's only read behind the new `trust_forwarded_for` option, taking the last
hop. Defense in depth, not a bypass: the invite token and accepted-user checks
still gate the real actions.

**Discovery skips TLS verification**  Two call sites hardcoded it off
while every other one reads it from config. A MITM on discovery could redirect
where content is fetched from. Both now verify certificates by default

## Testing

Table-driven tests for the IP checks, the header parsing and TLS-by-default
Confirmed the SSRF fix against a running revad: the old build connects to an
attacker-named internal port, the fixed one doesn't

### DIFF
--- a/internal/http/services/opencloudmesh/ocmd/client.go
+++ b/internal/http/services/opencloudmesh/ocmd/client.go
@@ -90,6 +90,8 @@ func NewClient(timeout time.Duration, insecure bool) *OCMClient {
 // it also covers redirects and DNS rebinding.
 func NewPublicOnlyClient(timeout time.Duration, insecure bool) *OCMClient {
 	tr := newOCMTransport(insecure)
+	// with a proxy the dial goes to the proxy, so Control never sees the target
+	tr.Proxy = nil
 	tr.DialContext = (&net.Dialer{
 		Timeout:   timeout,
 		KeepAlive: 30 * time.Second,
@@ -116,7 +118,7 @@ func refuseNonPublicAddr(_, address string, _ syscall.RawConn) error {
 }
 
 // 64:ff9b::/96 embeds an IPv4 address in its low 32 bits (RFC 6052), so on a
-// NAT64 network it can still reach internal hosts. Check the embedded address.
+// NAT64 network it can still reach internal hosts.
 var nat64WellKnownPrefix = net.IPNet{IP: net.ParseIP("64:ff9b::"), Mask: net.CIDRMask(96, 128)}
 
 func isPublicIP(ip net.IP) bool {
--- a/internal/http/services/opencloudmesh/ocmd/client.go
+++ b/internal/http/services/opencloudmesh/ocmd/client.go
@@ -25,9 +25,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/cs3org/reva/v3/internal/http/services/wellknown"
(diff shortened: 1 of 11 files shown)


glpatcern:
Thanks for this contribution! We are aware those capabilities are not yet production-grade, and some hardening was in the making (see e.g. #5389), so this is very timely. @MahdiBaghbani how does that fit with your work?

btwshivam:
Thanks @glpatcern! Just looked at #5389, and the SSRF piece here clearly overlaps with Mahdi's work, so I'll drop that commit and defer to it.  two ranges I block that #5389 doesn't yet, CGNAT (100.64.0.0/10) and the NAT64 well-known prefix (64:ff9b::/96), we can keep these..
other two changes don't overlap: the X-Forwarded-For handling in the provider allow list, and a couple of discovery call sites that had TLS verification hardcoded off

MahdiBaghbani:
Hi! Will do the review soon, the overlap does exist I think I can rebase my changes on top of this and do the review and add the stuff that belongs to me just on my PR and remove the SSRF ones

MahdiBaghbani:
@btwshivam please do not change 😄 

The two PRs are complementary, not duplicates, but they overlap on the SSRF surface where yours is technically better and mine is broader

internal/http/services/opencloudmesh/ocmd/client.go
```diff
@@ -83,6 +85,57 @@ func NewClient(timeout time.Duration, insecure bool) *OCMClient {
 	}
 }
 
+// NewPublicOnlyClient returns an OCMClient that only dials public addresses, for
+// discovery of hosts named by an untrusted caller. The check is in the dialer so
+// it also covers redirects and DNS rebinding.
+func NewPublicOnlyClient(timeout time.Duration, insecure bool) *OCMClient {
+	tr := newOCMTransport(insecure)
```
MahdiBaghbani, line 92: `NewPublicOnlyClient` builds its transport with `newOCMTransport`, which sets `Proxy: http.ProxyFromEnvironment`. The dial `Control` (`refuseNonPublicAddr`) only runs on the raw TCP dial, so when `HTTP_PROXY`/`HTTPS_PROXY` is set the client CONNECTs to the proxy and the guard never sees the target IP.

My read is that the "only dials public addresses" guarantee is bypassable wherever an outbound proxy is configured.

I think clearing `Proxy` on this transport is enough, right after `newOCMTransport`:

```go
func NewPublicOnlyClient(timeout time.Duration, insecure bool) *OCMClient {
	tr := newOCMTransport(insecure)
	tr.Proxy = nil // never proxy untrusted discovery; the dial Control must see the target
	tr.DialContext = (&net.Dialer{
		Timeout:   timeout,
		KeepAlive: 30 * time.Second,
		Control:   refuseNonPublicAddr,
	}).DialContext
	return &OCMClient{
		client: &http.Client{
			Transport: tr,
			Timeout:   timeout,
		},
	}
}
```

I'd keep `newOCMTransport` and the trusted `NewClient` path as-is, those talk to operator-configured targets and may legitimately need a corporate proxy, and as example I definitely do need the proxy setup in my environment since I have a legitimate need in OCM Test Suite to run a TLS decrypting MITM, so this only touches the public-only transport.

`TestNewOCMTransportUsesProxyFromEnvironment` can stay exactly as it is.

A small test would lock the public-only side in:

```go
func TestNewPublicOnlyClientTransportProxyNil(t *testing.T) {
	c := NewPublicOnlyClient(5*time.Second, true)
	tr, ok := c.client.Transport.(*http.Transport)
	if !ok || tr.Proxy != nil {
		t.Fatal("public-only client must not use a proxy")
	}
}
```

internal/http/services/opencloudmesh/ocmd/shares.go
```diff
@@ -487,8 +491,8 @@ func getAndResolveProtocols(ctx context.Context, p Protocols, resType string, ow
 	return protos, legacy, nil
 }
 
-func discoverOcmResourceTypes(ctx context.Context, ownerServer string) ([]wellknown.ResourceTypes, string, error) {
-	ocmClient := NewClient(time.Duration(10)*time.Second, true)
+func (h *sharesHandler) discoverOcmResourceTypes(ctx context.Context, ownerServer string) ([]wellknown.ResourceTypes, string, error) {
+	ocmClient := NewClient(time.Duration(10)*time.Second, h.ocmClientInsecure)
```
MahdiBaghbani, line 495: Not blocking this PR, just flagging it since it's the same surface.

`discoverOcmResourceTypes` dials the share sender's `sender.Idp` (the caller-supplied `ownerServer`) with the plain `NewClient`, so the inbound share path can still be pointed at an internal host.

This one is pre-existing, the SSRF exposure isn't introduced here, and this PR actually improves things by reading `insecure` from config instead of hardcoding it.

Since you already said you'd defer the broader inbound SSRF to #5389, I'll switch this call site to `NewPublicOnlyClient` there and reuse the dial guard this PR adds.

pkg/ocm/provider/authorizer/open/open.go
```diff
@@ -75,7 +79,7 @@ func (a *authorizer) GetInfoByDomain(ctx context.Context, domain string) (*ocmpr
 	}
 
 	// not yet known: try to discover the remote OCM endpoint
-	ocmClient := client.NewClient(time.Duration(10)*time.Second, true)
+	ocmClient := client.NewClient(time.Duration(10)*time.Second, a.insecure)
```
MahdiBaghbani, line 82: Same story as the inbound discover site and not blocking.
`GetInfoByDomain` probes a user-supplied domain with the plain `NewClient`, so it can be aimed at an internal host too.

Pre-existing not regressed here. I'll move it to the public-only client in #5389.

The `insecure` config wiring you added here is the right call regardless, thanks for that.

MahdiBaghbani (commented):
Thanks for this @btwshivam - hardening the unauthenticated `/discover` path is the right place to start, and I like that the check sits in the dialer, so it covers redirects and DNS rebinding, not just the first resolve.

The NAT64 well-known prefix unwrap, and the CGNAT range are nice catches, and the test coverage (the `isPublicIP` table, the public-only refusal, the XFF parsing, the TLS-by-default cases) is good and CI is happy about it.

One thing I'd like to sort out before these merges, plus two follow-ups that I'm happy to carry in #5389.

The one blocker: `NewPublicOnlyClient` still keeps the environment proxy.
The dial `Control` only fires on the raw TCP dial, but `newOCMTransport` sets `Proxy: http.ProxyFromEnvironment`, so when `HTTP_PROXY`/`HTTPS_PROXY` is set the transport CONNECTs to the proxy first and `refuseNonPublicAddr` never sees the target IP.

My read is that the "only dials public addresses" promise is bypassable in any environment with an outbound proxy. It's a one-liner - clear `Proxy` on the public-only transport right after `newOCMTransport`.

I left an inline suggestion on `client.go` with a small test.

The two follow-ups aren't blocking, and they line up with what you already said about deferring the broader inbound SSRF to #5389:

- `discoverOcmResourceTypes` in `shares.go` still dials the share sender's `sender.Idp` with the plain `NewClient`.
- `GetInfoByDomain` in `open/open.go` still probes a user-supplied domain with the plain `NewClient`.

Both are pre-existing and not regressed here, so I'd rather move them to the public-only client in #5389 than widen this PR.

One optional nit while we're here: `isPublicIP` could also reject the documentation/benchmark ranges (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`, `2001:db8::/32`) and `240.0.0.0/4`.
Not reachable in practice, so really just a note - happy to fold it into #5389 too.

With the `Proxy` one-liner in, this is good to go from my side. I'll rebase #5389 on top once it lands.

btwshivam:
@MahdiBaghbani Hey.. good catch on the proxy..
fixed with `tr.Proxy = nil` on the public-only transport plus your test. left NewClient alone so your MITM setup still works

agreed on the other two and the doc ranges they fit better in #5389 Thanks!!

MahdiBaghbani (approved):
Thanks @btwshivam 
LGTM, I'll rebase #5389 on top once this lands.

glpatcern (approved):
Thanks both! I'm happy to merge this right away without further tests as this capability is not in production yet.

PR #5761, opened 2026-08-10
Ceph: fix dir size

rawe0:
Directories were reported with a size of 0, so a folder's size did not match the contents seen inside it. The size of the subtree is now taken from the CephFS `ceph.dir.rbytes` extended attribute.


glpatcern (commented):
All good, I'd actually remove some unnecessary comments (!) as Claude gets them a bit too verbose (I wonder if this can be tuned)

rawe0:
> All good, I'd actually remove some unnecessary comments (!) as Claude gets them a bit too verbose (I wonder if this can be tuned)

Will tune them down a bit :)

glpatcern (approved):
Great, all good

PR #5762, opened 2026-08-10
add permission checks for (public)share methods to the gateway 

jessegeens:
instead of relying only on plugins


glpatcern (approved):
Looks good!

PR #5770, opened 2026-08-13
Resolve `/dav/files/<username>/<path>` in the right namespace

### LINKED ISSUE #5767: `/dav/files/<user>/` returns 500 on OPTIONS and the paths below it resolve into a different namespace
Two things are wrong with this route

**OPTIONS returns 500**, so WebDAV clients can't mount it:

Image

Image

`files` branch in `dav.go` calls `GetHome` unconditionally but the auth
middleware lets OPTIONS through before authenticating.. so there is no user in the
context and GetHome comes back Unauthenticated.

**root and the paths below it end up in different providers.** `GetHome` is only
joined for the root.. everything else returns early into `FilesHandler`, which uses
`files_namespace` and no longer has the username since `applyLayout` returns the
namespace unchanged.

Image

soo the folder is created and then it is not in the listing of the home it was
created in

Since the username is dropped every user lands on the same path. marie has never
created anything here:

Image

last part is on localfs which is permissive. On EOS the driver impersonates the
end user so the ACLs would likely deny the read, but the routing is wrong either way :)

setup is a single revad with `files_namespace = "/users"` and
`webdav_namespace = "/home"`, same as tests/revad/frontend.toml

btwshivam:
Fixes #5767

OPTIONS returned 500 because the files branch looks up the home unconditionally and OPTIONS isn't authenticated.. so there's no user to look one up for.

home was also only joined for the root. Everything below it went to the files namespace with the username already stripped off, soo `/dav/files/einstein/Documents` ended up at `/users/Documents` and every user collided on the same path

This restores the two routes: your own files resolve against the home, someone else's stay addressed by name under f`iles_namespace`. home goes in as the handler namespace and not as a prefix on the path, because MOVE and COPY take their destination from the Destination header, which is resolved through the namespace

Screenshot from 2026-08-14 02-40-33

Screenshot from 2026-08-14 02-39-34

### DIFF
--- a/changelog/unreleased/dav-files-namespace.md
+++ b/changelog/unreleased/dav-files-namespace.md
@@ -1,0 +1,8 @@
+Bugfix: resolve /dav/files/<username>/<path> in the right namespace
+
+The home was only applied to the collection root, so everything below it fell
+back to the files namespace without the username and every user ended up on the
+same path. OPTIONS also 500 because it is not authenticated and had no user to
+look a home up for, which stopped clients from mounting the endpoint.
+
+https://github.com/cs3org/reva/pull/5770
--- a/internal/http/services/owncloud/ocdav/dav.go
+++ b/internal/http/services/owncloud/ocdav/dav.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	gatewayv1beta1 "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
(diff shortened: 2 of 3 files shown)


btwshivam:
@glpatcern @jessegeens PATL .. let me know if any improvement needed

btwshivam:
Closing this..  #5644 got merged since I opened it and reworks /dav/files/{user}/ into a root-based view (home, eos,...) which already covers the OPTIONS 500 and the namespace split I was fixing here.. So this is redundant now. Thanks!

PR #5772, opened 2026-08-14
Service registry (#5665)

jessegeens:
* registry: add discovery interface, BaseRegistry, memory and NATS backends with liveness

* config: add [shared.registry] block and registry driver factory

* service: add registry-backed peer resolver with global accessor and HTTP endpoint lookup

* runtime: self-register services and install the process-wide resolver

* gateway: discover the data gateway and data provider through the registry

* gateway: resolve peers through the registry instead of fixed addresses

* auth: resolve the gateway and userprovider through the registry in the interceptors

* ocs, ocdav, ocgraph: resolve the gateway through the registry

* http services: resolve the gateway through the registry in sciencemesh, ocmd and others

* grpc services: resolve the gateway lazily through the registry

* auth managers, storage drivers and wopi: resolve the gateway through the registry

* rgrpc: remove the pool package now that discovery goes through the registry

* changelog: add service registry entry

* tests: add registry discovery and self-registration integration tests

* tests: co-locate gateway with its callers in the OCM integration fixtures

* service: expose package-level resolver accessors and drop the embedded Base

* ocs: drop the hardcoded gateway and storage registry addresses, resolve through the registry


glpatcern (dismissed):
This is good to go a priori, further improvements after we give it a run for real

glpatcern (approved):

PR #5774, opened 2026-08-16
Report failures on COPY and third-party copy instead of success

btwshivam:
COPY handler and the third-party copy push/pull handlers reported success even when the upload to the data server (or the remote PUT) returned a non-2xx status. For overwrite COPY the destination is deleted before the upload, so a false success silently lost the file, for a third-party copy a transfer manager could delete the source after a copy that never actually happened.

While in there, a third-party copy push of a source that doesn't exist dereferenced a nil stat result and panicked.. it now returns 404
**Now its:**
- report an error when a COPY upload fails
- report an error when a third-party copy upload fails
- return 404 for a third-party copy push of a missing source


btwshivam:
@glpatcern @jessegeens PTAL

PR #5780, opened 2026-08-17
Takeout

Tristan-C789:
# CERNBox Takeout

Users currently have no convenient, self-service way to obtain a full, offline copy of everything they store in their CERNBox personal space. This is needed for personal backups, offboarding (e.g. leaving CERN), and data-access requests. "Takeout" lets a user trigger an export from the web interface; the system packages their space into one or more archives asynchronously, stores them in a pre-configured EOS location, and notifies the user by email with download links when the export is ready. Exports expire automatically after a configurable retention period.

## Features

This PR mplements the core takeout logic as a Reva background job. Given a user, a max archive size, and a format, the job collects the contents of the user's personal space, packages them into one or more archives of the requested format, splitting output so that no single archive exceeds the configured max size, and uploads the artifact(s) to a pre-configured EOS folder.

It exposes two HTTP endpoints: one to create a new takeout job (taking max size and format), and one to get the status of the current/last takeout for the authenticated user. The create endpoint enforces one active job per user, if a takeout is already running, it returns the existing job rather than starting a new one. The status endpoint returns the job state and, when completed, the download link(s) and expiry. In particular:

- POST endpoint creates a takeout for the authenticated user with body {max_archive_size, format}; unsupported size/format is rejected with a clear error.
- GET status endpoint returns the current state (none / queued / running / completed / failed / cancelling / cancelled) and progress if available.
- DELETE endpoint cancels the ongoing takeout for the authenticated user, if any.
- If the user already has an active takeout, the create endpoint returns the existing job's id/status instead of starting a second one.
- When completed, the email notification is sent, which includes the download link(s) and expiry time.
- When failed, the status response includes an error/cause indication.


internal/http/services/archiver/manager/archiver.go
```diff
@@ -53,209 +46,44 @@ func NewArchiver(files []string, w walker.Walker, d downloader.Downloader, confi
...
+func getDeepestCommonDir(files []string) string {
+	return bundler.DeepestCommonDir(files)
+}
 
-			dst, err := w.CreateHeader(&header)
-			if err != nil {
-				return err
-			}
+// ErrMaxFileCount is the error returned when the max files count specified in the config is reached
+type ErrMaxFileCount = bundler.ErrMaxFileCount
```
jessegeens: Why do you need to re-export these?

Tristan-C789: It allows to not modify the archiver API at all when introducing the bundler. It's useful for the tests for example, but it's quite conservative and could be adapted quite easily.

internal/http/services/takeout/takeout.go
```diff
@@ -0,0 +1,314 @@
...
+	if len(jobs) == 0 {
+		s.log.Debug().Msgf("takeout: user %s has no takeout job attached", user.Username)
+		w.WriteHeader(http.StatusNotFound)
+		return rjobs.Status{}, false
+	}
+	return jobs[0], true
+}
+
+// replyStatus sends the JSON reply describing the given job
+func (s *svc) replyStatus(w http.ResponseWriter, st rjobs.Status) {
```
jessegeens: I would call this `respondWithStatus` or something similar (since response is the HTTP terminology)

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,487 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "takeout: user authentication failed")
+	}
+	adminCtx, err := j.authenticate(ctx, j.conf.TakeoutAdminUsername)
+	if err != nil {
+		return nil, errors.Wrap(err, "takeout: admin authentication failed")
+	}
+
+	// Set the destination directory and create it
+	archPath := fmt.Sprintf("%s/%s_%s/", j.conf.TakeoutPath, pp.Username, time.Now().Format("2006-01-02"))
```
jessegeens, line 177: Will the runner block multiple takeouts in one day? Otherwise there will be a collision if you only use the day for the foldername

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,487 @@
...
+	// Setup the bundler with a walker and a downloader
+	b := bundler.New(walker.NewWalker(gtw), downloader.NewDownloader(gtw, hc))
+
+	return &job{conf: &c, log: l, gtw: gtw, hc: hc, bundler: b}, nil
+}
+
+// The per-run takeout parameters
+type params struct {
+	MaxArchiveSize uint64 `mapstructure:"max_archive_size"`
+	ArchiveFormat  string `mapstructure:"archive_format"` // One of ["zip", "tgz"]
```
jessegeens: That is what you say, but there is no check here that `tar` is not passed :)
I think it's perfectly okay to do that check in the HTTP layer, but then the comment is a bit redundant. Especially if you would change the type to be a `bundler.Format`

internal/http/services/takeout/takeout.go
Tristan-C789: Fixed

pkg/takeout/takeout.go
Tristan-C789: The job deletes the folder if it already exists ; in theory, the frontend won't allow for more than one takeout per day.

jessegeens (changes requested):
Very, very good job! I have a few small comments, after that we can merge 🚀

internal/http/services/archiver/manager/archiver.go
jessegeens: in that case, I think we can do a `s/ErrMaxFileCount/bundler.ErrMaxFileCount/g` in the tests to make it pass, re-exporting it seems a bit pointless

pkg/bundler/bundler.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	switch opts.Format {
+	case FormatTar, FormatZip, FormatTgz:
+	default:
+		return ErrUnsupportedFormat{Format: opts.Format}
+	}
+
+	// Entry names are relative to the deepest common parent of the roots
+	baseDir := DeepestCommonDir(opts.Roots)
+	if pathIn(opts.Roots, baseDir) {
```
diocas, line 157: Why is this check necessary? I would add a comment explaining.

pkg/bundler/bundler.go
```diff
@@ -0,0 +1,452 @@
...
+			r, err := b.downloader.Download(ctx, currentPath, "")
+			if err != nil {
+				return err
+			}
+			writeErr := aw.WriteFile(fileName, int64(info.Size), mtime, r)
+			_ = r.Close()
+			return writeErr
+		})
+		if walkErr != nil {
+			break
```
diocas, line 264: Why do we bail out? can't we continue?

internal/http/services/archiver/manager/archiver.go
```diff
@@ -53,209 +46,44 @@ func NewArchiver(files []string, w walker.Walker, d downloader.Downloader, confi
...
+	return a.bundler.Create(ctx, a.opts(bundler.FormatZip), bundler.SinglePart(dst))
+}
 
-			if isDir {
-				header.Name += "/"
-			} else {
-				header.UncompressedSize64 = info.Size
-			}
+// getDeepestCommonDir returns the deepest common parent directory
+func getDeepestCommonDir(files []string) string {
```
diocas: Why do we do these things? Just to avoid changing the tests? Same for the errors bellow.

pkg/takeout/takeout.go
Tristan-C789: Changed the use of strings to a subset of bundler.Format in the job, thus the job can only accept correct formats.

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,481 @@
...
+
+	// Create the archives
+	err = j.bundler.Create(userCtx, j.bundleOpts(pp), j.newPartFunc(adminCtx, archPath, string(pp.ArchiveFormat)))
+	if err != nil {
+		return nil, errors.Wrapf(err, "takeout: %s archive could not be created", pp.ArchiveFormat)
+	}
+
+	// Share the folder containing the archives through a public link
+	expiration := time.Now().Add(time.Duration(j.conf.ExpirationDelay) * time.Hour)
+	token, err := j.createPublicShare(adminCtx, archPath, expiration)
```
diocas: In the previous PR I asked whether we should protect the pl. Was there a conclusion?

pkg/takeout/takeout.go
```diff
@@ -0,0 +1,481 @@
...
+	// Pass the archives and the expiration to the email template
+	templateData := map[string]any{
+		"user_display_name": user.DisplayName,
+		"username":          user.Username,
+		"archives":          archives,
+		"expiration_date":   expiration.Format("Monday, 2 January 2006 at 15:04 MST"),
+	}
+
+	// Publish the completion event
+	event := notifications.EncodeEvent(model.EventTakeout, []string{user.Mail}, templateData)
```
diocas: This might've changed and now receives users instead of emails, right @jessegeens ?

diocas (commented):
Looks very good. But still wonder if we should keep going on failure and just upload a txt file to the same share with the errors. This way users get the most the takeout can take, and can try to figure out the rest. Thoughts?

Also, from the web ui you can only query the status, right? To get the url's to download, this means opening the email, only? Maybe we could give the pl link there as well, so that users could open, but it's fine to stay as is for now.

What I wonder, is if we should protect the pl (this would not allow users to download from the urls in the email, instead would need to open the pl in the web ui, but I think that is fine as well). At least their private data would stay private. Thoughts? cc @jessegeens @gmgigi96

internal/http/services/archiver/manager/archiver.go
Tristan-C789: Fixed

pkg/bundler/bundler.go
Tristan-C789: Fixed

pkg/bundler/bundler.go
Tristan-C789: With the new error handling, this error has now been through the entryError function ; we will bail out only if entryError deems it necessary, depending on usecase.

pkg/takeout/takeout.go
Tristan-C789: Per @gmgigi96's request, the only way to access the takeout is through the direct download links sent by email. This is incompatible with password protected public links.

internal/http/services/archiver/manager/archiver.go
Tristan-C789: Correct, this way the archiver exposes exactly the same interface as before. `getDeepestCommonDir` is indeed only in tests, it will be removed.

pkg/takeout/takeout.go
Tristan-C789: Every occurence of `notifications.EncodeEvent` uses the mail field ; and this works with no issue.

jessegeens: Because I changed that in a PR that was not merged yet :) But that's OK. If we merge this one before mine I'll change it

jessegeens (approved):

PR #5781, opened 2026-08-17
Return proper error codes instead of NewInternal

### LINKED ISSUE #4915: Review all internal errors
The current code base uses `status.NewInternal()` in far too many places as a "shortcut" to report an error. In many/most occurrences, a more tailored error code should be used for the caller to display proper error messages, and logging should be adapted as well.

Here follows the count per GRPC service:

```
applicationauth: 4
appregistry: 6
authregistry: 2
datatx: 4
gateway: 104
groupprovider: 5
ocmcore: 1
ocminvitemanager: 10
ocmproviderauthorizer: 3
ocmshareprovider: 12
preferences: 2
publicshareprovider: 6
publicstorageprovider: 9
storageprovider: 64
storageregistry: 3
userprovider: 4
usershareprovider: 8
```

btwshivam:
Fixes #4915.
`NewInternal()` was used as a catch-all across the grpc services and flattened errors that already carried a code (notfound, permission denied, ...) into INTERNAL. routes those through `NewStatusFromErrType` so the real code reaches the client, adds the inverse `NewErrtypeFromStatus` for the gateway helpers that receive a downstream status and fixes a few miscoded BadRequests

also i see there are genuinely internal sites (transport failures, marshaling, client setup) i left them as-it-is

### DIFF
--- a/changelog/unreleased/review-internal-errors.md
+++ b/changelog/unreleased/review-internal-errors.md
@@ -1,0 +1,10 @@
+Change: return tailored grpc error codes instead of always INTERNAL
+
+Many grpc handlers wrapped every error in status.NewInternal, hiding the real
+code from callers. They now map known errtypes (not found, permission denied,
+already exists, invalid argument, ...) to the matching grpc status code through
+a fixed status.NewStatusFromErrType, and the gateway share handlers propagate
+the downstream status instead of flattening it.
+
+https://github.com/cs3org/reva/pull/5781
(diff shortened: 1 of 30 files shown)


btwshivam:
@glpatcern @jessegeens done! please take a look!.. let me knw if i am missing anything..

btwshivam:
@glpatcern Please take a look when you get a chance :)

btwshivam:
Fixed the faling test

PR #5801, opened 2026-08-31
implement token cache

jessegeens:


glpatcern (approved):
Good to go for the sync clients tests, the JWT part is to be activated at a later time

PR #5804, opened 2026-09-04
fix rjobs timeout and skip-overlap

jessegeens:
* Jobs that are killed because of an overlap are no longer seen as a failure and are not retried
* Daemons that cannot reach the gateway die
* Failed resolutions are retried for a few times
* The scheduler reads only its own schedule keys instead of listing the whole bucket, which timed out and stopped all periodic jobs


pkg/service/clients.go
```diff
@@ -163,6 +204,95 @@ func (c *clients) resolve(name string) (*grpc.ClientConn, string, error) {
...
+	// the services it hosts.
+	log := appctx.GetLogger(ctx)
+	if name != NameGateway || calls < unresolvableCalls || since < unresolvableFor {
+		log.Error().Err(cause).Str("service", name).Int("failed_lookups", calls).
+			Str("registry", c.registrySnapshot(name)).Msg("cannot resolve peer, retrying")
+		return
+	}
+	reason := fmt.Sprintf("reva: %q has been unresolvable for %s over %d lookups, exiting: %v",
+		name, since.Truncate(time.Second), calls, cause)
+	log.Error().Err(cause).Str("service", name).Int("failed_lookups", calls).
```
glpatcern:
```suggestion
	log.Fatal().Err(cause).Str("service", name).Int("failed_lookups", calls).
```

glpatcern (changes requested):
A small change, noting that we have a `if name != NameGateway` condition but ok that's fair for now.

glpatcern (dismissed):

glpatcern (approved):

PR #5811, opened 2026-09-07
report folder size in public links

jessegeens:


rawe0 (dismissed):
Looks good to me 👍

rawe0 (approved):
Still good!

PR #5815, opened 2026-09-09
appregistry: discover app providers through the service registry

gmgigi96:
## The problem

App providers registered by *pushing*: on startup each one slept three seconds and called `AddAppProvider` on the gateway, once. Four things followed from that.

**Startup order mattered.** The sleep existed only to let the app registry come up first, with a `TODO` next to it admitting as much. On a slow start the provider registers into nothing and the app is silently missing.

**Restarting the app registry lost every app.** Nothing ever pushed a second time, so restarting the gateway left it empty until every app provider was restarted too. 

**Providers were keyed by address, and the address was configuration.** It came from `app_provider_url`, which fell back to the *gateway* address when unset, so all providers collapsed into one entry and every app but the last to register became unreachable by name. `/app/list` still showed them all, since that index is keyed by name, so the only symptom was `app '<name>' not found` on open.

**No liveness, no redundancy.** A dead provider stayed on offer, and one address per app made a second node serving it unusable.

## What this does

App providers are ordinary reva services, and the runtime already registers them in the service registry with an address, a transport, a liveness state and reaping. They now describe their app in that node's metadata and the app registry reads it: take the `appprovider` nodes, keep the live gRPC ones, group them by app name, pick a node per app. The mime type catalogue stays in configuration, since it is deployment policy rather than something a provider knows about itself.

## What it buys

- **Order independence.** Nothing has to start before anything else; the state is read at request time.
- **Self-healing.** Restarting the registry costs nothing, and a provider that dies comes back on its own.
- **Liveness.** Offline and draining nodes drop out through the registry.
- **Redundancy.** An app served by several nodes gets a live one picked per lookup.
- **Nothing to misconfigure.** The address comes from the node.

## Breaking changes

```toml
# before
[grpc.services.appregistry]
driver = "static"

[grpc.services.appregistry.drivers.static]
mime_types = [ ... ]

# after
[grpc.services.appregistry]
mime_types = [ ... ]
```

- `app_provider_url` is removed, drop it if you set it
- `AddAppProvider` returns `UNIMPLEMENTED` and is no longer an unprotected
  endpoint
- the `static` driver and the `providers` list it took are gone


internal/grpc/services/appregistry/registry.go
```diff
@@ -0,0 +1,385 @@
...
+		info := infoByApp[name]
+		info.Address = node.Address()
+		out[name] = info
+	}
+	return out, nil
+}
+
+// priorityOf reads the priority an app provider advertises, defaulting to the
+// lowest when it advertises none.
+func priorityOf(p *registrypb.ProviderInfo) uint64 {
```
glpatcern, line 183: Just a comment about this priority: what is it for? So far there was no "prioritization" of applications

gmgigi96: There was already a prioritization of the applications in the static driver: https://github.com/cs3org/reva/blob/819b45c4579c69fc8bc0084e4af4777078507cca/pkg/app/registry/static/static.go#L334

glpatcern (approved):

PR #5826, opened 2026-09-16
Add failure mode for processing embedded shares

rawe0:
Embedded transfers may fail due to a number of reasons, an internal service could be down or the remote repository could be unreachable. Instead of being stuck in the "Transferring" state, shares are now put in the "Rejected" state so that this can be surfaced to the user in a reasonable way (accompanying [web PR](https://github.com/cernbox/web-extensions/pull/65)).


glpatcern (commented):
Just one question: was the transfer really panicking in case of errors? I don't think we were observing that...

rawe0:
> Just one question: was the transfer really panicking in case of errors? I don't think we were observing that...

It was not, we could remove this part, but I added it for extra redundancy... which may be a bit overkill, especially now that we can revert a "stuck" transfer (if it happens somehow, which it shouldn't).

internal/grpc/services/ocmshareprovider/ocmshareprovider.go
```diff
@@ -673,17 +674,27 @@ func (s *service) processEmbeddedShare(ctx context.Context, user *userpb.User, r
 	detached := context.WithoutCancel(ctx)
 	onComplete := func(transferErr error) {
 		log := appctx.GetLogger(detached)
+		defer func() {
+			if r := recover(); r != nil {
+				log.Error().Interface("panic", r).Msg("panic while ending an embedded transfer")
```
glpatcern:
```suggestion
				log.Error().Interface("error", r).Msg("failure while ending an embedded transfer")
```

pkg/ocm/embedded/embedded.go
```diff
@@ -146,10 +151,20 @@ func (d *driver) Process(ctx context.Context, payload, destination string, onCom
 	return nil
 }
 
+func (d *driver) runTransfer(log *zerolog.Logger, token, destination string, entries []transferEntry, timeout time.Duration) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Error().Interface("panic", r).Bytes("stack", debug.Stack()).Msg("embedded transfer failed")
```
glpatcern:
```suggestion
			log.Error().Interface("error", r).Bytes("stack", debug.Stack()).Msg("embedded transfer failed")
```

glpatcern (approved):

PR #5833, opened 2026-09-21
add(ocm): shared OCM HTTP transport package

MahdiBaghbani:

## Summary

- Add `pkg/ocm/client`, a shared OCM HTTP transport package with `TransportConfig` and trusted/public-only HTTP client constructors, centralizing transport and address-policy logic currently inlined in `ocmd`. Existing `ocmd.OCMClient` callers are unchanged; public-only clients still bypass environment proxies and use the guarded dialer.
- Supersedes #5389.


MahdiBaghbani:
This pulls the OCM transport into one `TransportConfig` in `pkg/ocm/client` so the rest of the stack shares a single dial policy. Existing `ocmd` constructor signatures don't change.

```mermaid
flowchart TD
    CFG["TransportConfig"] --> T["trusted client"]
    CFG --> P["public-only client"]
    T --> TP["always ProxyFromEnvironment"]
    P --> G["Dialer.Control"]
    G --> Z{"zoned IPv6?"}
    Z -->|yes| R["reject"]
    Z -->|no| IP["isAllowedIP"]
    P --> PX{"UseEnvProxy"}
    PX -->|false, default| D["Proxy = nil"]
    PX -->|true| E["ProxyFromEnvironment"]
```

The zone check is the part you might ask about. `netip.Prefix.Contains` returns false for any zoned address, so something like `64:ff9b::a9fe:a9fe%0` used to slip past the NAT64 and denied-prefix checks. 

I reject zones outright instead of stripping them, since a public-only client has no reason to honor a link-local zone.

The other thing is `UseEnvProxy`. It defaults to false. If you turn it on, `Control` checks the proxy hop and not the OCM target, so it's an operator opt-in and doesn't add a second guarded mode.

The proxy change is its own commit, so it can be reverted without touching the zone fix.

And the reason is that you might need to have all your OCM outbound routed through MITM, like I do when I'm in the Test Suite.

MahdiBaghbani:
Hi Giuseppe, good question in Matrix. I went to check all the places and the distinction between trusted and untrusted.


A party counts as trusted once the operator has admitted it through the provider authorizer (mentix or the json file), and that's what `IsProviderAllowed` checks at `/ocm/invite-accepted` and `/ocm/shares`.

```mermaid
flowchart LR
    Auth["provider authorizer (mentix / json)"] --> Gate{"IsProviderAllowed"}
    Gate -->|admitted| GI["GetInfoByDomain"]
    GI --> T["NewTrustedHTTPClient"]
    Dir["WAYF configured directory URLs"] --> T
    T --> T1["invite manager: InviteAccepted"]
    T --> T2["share provider: NewShare"]
    Gate -->|unknown or caller-named host| P["NewPublicOnlyHTTPClient"]
    P --> P1["inbound /shares: sender IdP + WebDAV"]
    P --> P2["received-share storage: WebDAV"]
    P --> P3["open authorizer probe"]
    P --> P4["WAYF untrustedClient: /discover + listed servers"]
```

The trusted client keeps the env proxy and skips the dial guard, and it's used in exactly three places: 
1. WAYF dialing the directory URLs from config.
2. The invite manager calling `InviteAccepted`.
3. The share provider calling `NewShare`.

In the two gRPC cases the target comes from `GetInfoByDomain`, so it's the authorizer picking the host and not the caller.

Anywhere a caller gets to name the host, it goes through the public-only client.

One thing I want to be sure about: 
those two gRPC call sites take the `ProviderInfo` they're handed and don't run `IsProviderAllowed` again before dialing.
Right now, that's safe because the HTTP handlers always resolve through `GetInfoByDomain` first, but the check lives upstream and not at the dial site. I can add a re-check there if you think it's worth it.

PR #5834, opened 2026-09-21
fix(ocm): bound peer-controlled OCM JSON

MahdiBaghbani:

## Summary

- Bound every OCM control-plane response body and peer-controlled error snippet to one 64 KiB default limit via `readOCMBody` and `decodeOCMJSON` helpers, which read the body once and return `ErrResponseTooLarge` for oversized input. This prevents decode-after-drain regressions in `NewShare` and `InviteAccepted` and bounds the existing unsigned directory JSON at the same limit.


internal/http/services/opencloudmesh/ocmd/client.go
```diff
@@ -36,6 +36,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// DefaultResponseLimit is the maximum OCM control-plane response body size.
+const DefaultResponseLimit int64 = 64 << 10
```
MahdiBaghbani, line 41: I do understand that this seems like a magic number and should be configurable but at the same time I think that too much configuration defeats the purpose of the security and simply adds unnecessary overheads (config wiring and examples etc) for unrealistic scenarios.

for this instance, I really cannot imagine a payload would exceed the 64KiB limit I set here.

Happy to discuss this and do adjustments.

MahdiBaghbani: is 64 KiB enough for directories? to be seen 😄

MahdiBaghbani:
```mermaid
flowchart TD
    WK["GET /.well-known/ocm"] --> CAP["readOCMBody, 64 KiB cap"]
    CAP -->|fails| LEG["GET /ocm-provider, same cap"]
    CAP -->|ok| JSON["decode"]
    LEG -->|ok| JSON
    LEG -->|fails| ERR["ErrResponseTooLarge / ErrPolicyViolation kept"]
```

If the well-known response is oversized, discovery still falls back to the legacy endpoint.

When both fail, you get the real sentinel back through `errors.Is` instead of a generic `InternalError`.

The cap counts bytes actually read, not Content-Length, and it doesn't limit JSON nesting depth.

MahdiBaghbani:
Thanks Giuseppe for the chat on the Matrix, agreed on hardcoding it.

I've already bumped `DefaultResponseLimit` from 64 KiB to 1 MiB on the branch.

That's the same cap the appprovider mentions and feedback handlers use with `http.MaxBytesReader`, and they don't have a config knob either, so no new TOML key and the config file stays clean.

One thing I noticed while looking at this: the embedded-shares payload actually comes in on the incoming `/shares` POST, which is decoded straight from `r.Body` with no cap. `readOCMBody` only limits the responses we get back from peers (discovery, the NewShare reply, invite-accepted, token, and directory-service), and none of those carry an embedded payload.

So the response cap and the embedded-shares case are really two separate things.

Right now the incoming bodies on `/shares`, `/invites`, and `/notifications` have no size limit at all, and `/notifications` just does an `io.ReadAll`.

Would you like me to open a few more small PRs to cap those incoming requests as well?

Notification can wait and be part of the (hopefully and fingers crossed) next* milestones 😄

PR #5836, opened 2026-09-21
fix(ocm): guard inbound share discovery

MahdiBaghbani:

## Summary

- Make unauthenticated inbound `/shares` discovery public-only: build one
  public-only HTTP client at handler init and reuse it per request, closing
  the SSRF where a sender-derived server could steer the inbound service at
  private networks. Loopback is allowed only behind an explicit
  `allow_loopback_federation` flag for the local two-provider integration
  topology; both local integration fixtures declare the exception.
- Make the discovery timeout configurable via `ocm_client_timeout` (seconds,
  default 10, preserving the existing inbound-discovery timeout).


MahdiBaghbani:
`POST /ocm/shares` is unauthenticated, and it used to run discovery against the sender-supplied server with a trusted client.
`IsProviderAllowed` checks the caller's IP, not that URL.
Now the handler builds one public-only client at init and reuses it for every request.

```mermaid
flowchart TD
    POST["unauthenticated POST /shares"] --> S["sender server URL"]
    S --> C["public-only client, built once at init"]
    C --> G["Dialer.Control"]
    G -->|private / loopback / zoned| D["reject"]
    G -->|public| OK["discover"]
```

Two config details can trip people up.

`ocm_client_timeout` is an integer number of seconds (`10`, not `"10s"`), because reva's decoder has no duration hook.

`ocm_client_use_env_proxy` is the inbound proxy opt-in. It's a different key from the one received storage gets in the next PR, so please don't mix them up.

PR #5837, opened 2026-09-21
fix(ocm): guard receiver OCM and WebDAV

MahdiBaghbani:

## Summary

- Apply the public-only transport to receiver-side connections: inbound
  legacy WebDAV type probes, received-share OCM discovery and token exchange,
  and received-share WebDAV operations. One stored OCM client guards both
  discovery and token exchange; one driver helper builds every gowebdav client
  and always applies the guarded transport, including the retry and one-off
  upload paths.
- Add `allow_loopback_federation` to the received-storage config; loopback is
  allowed only behind the explicit flag, private targets stay denied.
  Enforcement is at dial time, not in URI syntax validation.


MahdiBaghbani:
Received shares carry a token endpoint and a WebDAV URI in stored protocol data, and the peer controls both.
This puts both of them on the public-only transport.

```mermaid
flowchart TD
    Share["received-share protocol data"] --> OCM["stored public-only OCM client"]
    Share --> DAV["newWebDAVClient"]
    OCM --> D1["Discover + ExchangeToken"]
    DAV --> D2["every gowebdav client, incl. 401 retry and upload"]
    D1 --> G["Dialer.Control"]
    D2 --> G
```

Discover and ExchangeToken now share one stored client, so an advertised token URL gets the same dial guard.

`newWebDAVClient` is the only place in the driver that calls `gowebdav.NewClient`, and it always sets the guarded transport.

That's the invariant to keep an eye on in future changes.

The proxy opt-in here is `ocm_use_env_proxy`, which is separate from the inbound key in the previous PR.

The private-token test checks that the error says `non-public address` and does not say `refusing scheme`.

That shows the dial guard blocked the connection, so we don't need a hit counter.

PR #5838, opened 2026-09-21
fix(ocm): guard OCM discovery inputs

MahdiBaghbani:

## Summary

- Use the shared public-only OCM client for the two untrusted discovery
  sources: the open provider authorizer's `GetInfoByDomain`, and the
  directory-listed provider discovery in ScienceMesh WAYF `FetchInfo`. The
  operator-configured directory fetch stays on the trusted client; only
  provider URLs taken from the unsigned directory response (and request-supplied
  WAYF discovery) go through the public-only client.
- Loopback is disabled for open-authorizer discovery; both untrusted paths
  bypass environment proxies. Blocked targets are rejected before any request
  reaches the server.


MahdiBaghbani:
```mermaid
flowchart TD
    Dir["operator directory URL"] --> Trust["trusted fetch"]
    Trust --> Listed["listed provider URLs"]
    Open["open authorizer GetInfoByDomain"] --> Pub["public-only Discover"]
    Listed --> Pub
    Pub --> G["Dialer.Control"]
    Pub -.->|endpoint copied into ProviderInfo| Later["share / invite / gateway, still trusted"]
```

This PR only guards the first hop. The peer's endpoint still gets copied into `ProviderInfo`, and share, invite, and gateway still dial it with a trusted client.

PR #5839, opened 2026-09-21
fix(ocm): enforce transport TLS and scheme policy

MahdiBaghbani:

## Summary

- Enforce a TLS 1.2 floor on every shared OCM HTTP transport (including
  trusted and insecure modes), make public-only clients HTTPS-only with
  redirect-downgrade rejection (10-redirect limit), and close the bounded
  Q6 address-classification gaps (IANA special-purpose ranges, PCP/TURN
  anycast exceptions, NAT64 unwrap).


MahdiBaghbani:
`Dialer.Control` only checks IPs, so it misses `ftp`/`gopher` and
https-to-http downgrades. This wraps the public-only transport in a
RoundTripper that requires https, sets a TLS 1.2 minimum, and refuses
downgrade redirects.

```mermaid
flowchart TD
    URL["peer URL"] --> W["publicOnlyTransport.RoundTrip"]
    W -->|http / ftp / gopher| R["ErrPolicyViolation"]
    W -->|redirect to http| R
    W -->|https| B["base transport, then Dialer.Control"]
```

I wrapped `RoundTrip` instead of using `Transport.RegisterProtocol`, because
that can't override `http`/`https` and would actually enable `gopher` if we
registered it. The wrapper is also the only redirect gate for WebDAV, since
gowebdav ignores `CheckRedirect`. The gowebdav test shows a 302 to plain
HTTP failing with no HTTP request sent. `HTTPTransport` is exported only so
tests can inspect the base transport. Production code shouldn't use it to
get around the wrapper.

